PR #1, opened 2021-08-08
Journeys API

tataihono:
- add eslint standard
- add github actions workflow
- default project node v15
- add graphql codegen
- add prisma with journey migration
- add graphql-modules


murphomatic (approved):
Looks good Tatai -

PR #4, opened 2021-08-18
Implemented candidate way to add apollo federation / subgraph support…

morleycb:
…, and added the basic federation directives to the ‘journey’ api


morleycb:
test failed due to federation directives

PR #5, opened 2021-08-18
add apollo federation

morleycb:
Added apollo federation support


apps/api-journeys/src/modules/application/buildModuleSubgraphSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  const modules = typeDefs.map((typeDefs, i) => {
+    const module: GraphQLSchemaModule = { typeDefs: typeDefs, resolvers: resolvers[i] }
+    return module
+  })
+
+  const schema = buildSubgraphSchema(modules)
+  return schema
+}
```
tataihono, line 26: seems to me this is going to be needed across services. Do you think we should move this into a lib?

apps/gateway/index.js
```diff
@@ -0,0 +1,20 @@
...
+  // provides metrics, schema management and trace reporting.
+  engine: false,
+
+  // Subscriptions are unsupported but planned for a future Gateway version.
+  subscriptions: false,
+});
+
+server.listen({ port: gw-port }).then(({ url }) => {
+  console.log(`🚀 Server ready at ${url}`);
+});
```
tataihono, line 20: can you redefine this as a typescript file?

tataihono (changes requested):
some file movements needed. Otherwise looking good. How do we connect various services into the apollo gateway?

PR #6, opened 2021-08-19
add first-draft journey graph design

soberstadt:
Based on https://www.figma.com/file/HksZCdBmh2NRdV94eViu7Q/Block-content-variations?node-id=0%3A1

I don't expect this code to get merged in, but I figured a PR was as good of places as any to discuss the design.


soberstadt:
Seems like this is going a different direction.

PR #11, opened 2021-08-31
add block schema

soberstadt:


apps/api-journeys/src/modules/block/index.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    data: {
+      title: 'not "that" journey',
+      published: true
+    }
+  })
+  await db.block.create({ data: { journeyId: otherJourney.id, blockType: 'StepBlock' } })
+
+  const { data } = await testkit.execute(app, {
+    document: gql`
+      query($id: ID!) {
```
soberstadt, line 55: @tataihono well that makes me feel dumb...

soberstadt: But it makes me think that the graphql engine must have been raising an exception here. how did you find this? and can we do something so that next time we make a dumb mistake it doesn't take this long?

tataihono: The `testkit.execute` function puts an object out with `data`, but also `errors`. So line 53 I just put `const { data, errors }` and then console logged that out.

soberstadt: :( I tried that, but I guess I tried it wrong... sigh.

tataihono (approved):

PR #14, opened 2021-09-01
journeys

tataihono:
[Basecamp Reference - A note about blocks.](https://3.basecamp.com/3105655/buckets/23405694/messages/4071254773)

This PR implements the pattern in the document above. Particularly:
- conductor
- transformer
- block render

It also provides some configuration for deployment on NextJS.

Some basic components have been implemented including:
- RadioOption
- RadioQuestion
- Stub for Video


gdub01 (approved):
I think it's good. It hurts a little dropping vime because it's pretty well-written but I could not figure out how to have jest/jsdom use its webcomponents in tests. 

We'll have to probably adjust the data models of radio questions and radio options at a later date as we try to handle things like multi vs single select and advancing or not to next step on click. But I think that will be a future project when we know the interactions we want.

gdub01 (dismissed):

gdub01 (approved):

PR #15, opened 2021-09-01
refactor conductor

tataihono:
- conductor should use redux over provider for more robust state management


gdub01 (approved):
Looks great

PR #18, opened 2021-09-06
Storybook controls

GeronimoJohn:
- Add stories to radio option component
- Add disable button functionality once an option is chosen

Question: 
1. Do you have any feedback on the disable button functionality?
2. Do you have any feedback on the current stories (Radio question story to come)?
2. Setting a type on a redux action


apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.tsx
```diff
@@ -28,29 +34,42 @@ export function RadioQuestion ({
...
+            fullWidth={true}
+          >
             {children?.map(
               (option) =>
                 option.__typename === 'RadioOption' && (
-                  <RadioOption {...option} key={option.id} />
+                  <RadioOption
+                    {...option}
+                    key={option.id}
+                    selected={selectedId !== ''}
```
storyworks, line 69: Could use `selected={ selectedId === option.id }` ?

and then in RadioOption instead of `startIcon={selected && !disabled ? <CheckCircle...` do `selected ? <CheckCircle...`

apps/journeys/src/components/blocks/RadioOption/RadioOption.spec.tsx
```diff
@@ -11,17 +11,37 @@ const block: RadioOptionType = {
 describe('RadioOption', () => {
   it('should render option props', () => {
```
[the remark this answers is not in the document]
GeronimoJohn: I think the test for rendering out label was written before we implemented visual testing. But that's a good point that you made I can remove this test now since visual testing should be able to cover this. 

Will start working on those other tests!

apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.spec.tsx
```diff
@@ -27,14 +27,28 @@ describe('RadioQuestion', () => {
...
   })
 
-  it('should render dark them', () => {
-    const { getByTestId } = renderWithStore(<RadioQuestion {...block} variant="dark" />)
-    expect(getByTestId('RadioQuestionCard')).toHaveClass('MuiRadioQuestionComponent-dark')
+  it('should select an option OnClick', () => {
+    const { getByTestId, getAllByRole } = renderWithStore(<RadioQuestion {...block}/>)
+    const buttons = getAllByRole('button')
+    fireEvent.click(buttons[0])
+    expect(getByTestId('RadioOptionCheckCircleIcon')).toBeInTheDocument()
```
storyworks, line 40: Maybe instead of checking that the checked icon appears anywhere in the document, we could check:
`expect(buttons[0]).not.toBeDisabled()
    expect(buttons[0]).toContainElement(getByTestId('RadioOptionCheckCircleIcon'))`

apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.spec.tsx
```diff
@@ -27,14 +27,28 @@ describe('RadioQuestion', () => {
...
+  })
+
+  it('should disable unselected options', () => {
+    const { getByTestId, getAllByRole } = renderWithStore(
+      <RadioQuestion {...block} />
+    )
+    const buttons = getAllByRole('button')
+    fireEvent.click(buttons[0])
+    expect(getByTestId('RadioOptionRadioButtonUncheckedIcon')).toBeInTheDocument()
```
storyworks, line 49: Similar here:
`expect(buttons[1]).toContainElement(getByTestId('RadioOptionRadioButtonUncheckedIcon'))`
and you already have the check for button[1] is disabled, so just the 1 line here :)

storyworks (dismissed):
Other than those 2 comments, looks ready to merge :) I'll approve tomorrow morning?

I'm also still not sure why the title looks messed up in storybook even though it looks good in deployment. Maybe a cooldown task..

storyworks (approved):

PR #19, opened 2021-09-08
Connect Gateway to Journeys

tataihono:
- allow gateway to run locally 
- support pulling journey data from gateway
- add actions to data model


morleycb:
Tatai, just an FYI from the Apollo docs:

> We hope to deprecate the serviceList option in the coming months, but we'd love to hear from you if it enables an important use case that either supergraphSdl or managed federation currently doesn't.

To get the gateway to run against locally running subgraphs, there are several steps that we may want Sean to script into nx at some point.  E.g:  

Introspect and generate the subgraph schema:

`nx run api-journeys:serve`
`rover subgraph introspect http://127.0.0.1:3000/graphql > journeys-schema.graphql`


Then, if there was a configuration such as:  
```
subgraphs:
  journeys:
    routing_url: http://127.0.0.1:3000/graphql
    schema: 
      file: ./journeys-schema.graphql

```

the following would generate the supergraph schema:
`rover supergraph compose --config ./supergraph.yaml > ark-weave-schema.graphql`

the result of which could be read from the filesystem into a string and used as the 'supergraphSdl' option when starting the gateway.  Of course, for the gateway to work, both it and the 'journeys' endpoints would have to be running...

morleycb (dismissed):

storyworks (approved):

PR #25, opened 2021-09-09
add block-response module

soberstadt:


apps/api-journeys/src/modules/block-response/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    FINISHED
+  }
+
+  extend type Mutation {
+    journeySessionCreate(journeyId: ID!): ID!
+
+    signupBlockResponseCreate(journeySessionId: ID!, blockId: ID!, name: String!, email: String!): ID!
+    videoBlockResponseCreate(journeySessionId: ID!, blockId: ID!, position: Float!, state: VideoBlockResponseStateEnum!): ID!
+    radioQuestionBlockResponseCreate(journeySessionId: ID!, blockId: ID!, selectedResponseBlockId: ID!): ID!
+  }
```
tataihono, line 18:
```suggestion
  input SignupBlockResponseInput {
  	journeySessionId: ID!
  	blockId: ID!
  	name: String!
  	email: String!
  }
  
  input VideoBlockResponseInput {
  	journeySessionId: ID!
  	blockId: ID!
  	position: Float!
  	state: VideoBlockResponseStateEnum!
  }
  
  input RadioQuestionBlockResponseInput {
  	journeySessionId: ID!
  	blockId: ID!
  	selectedResponseBlockId: ID!
  }

  extend type Mutation {
    journeySessionCreate(journeyId: ID!): ID!

    signupBlockResponseCreate(input: SignupBlockResponseInput!): ID!
    videoBlockResponseCreate(input: VideoBlockResponseInput!): ID!
    radioQuestionBlockResponseCreate(input: RadioQuestionBlockResponseInput!): ID!
  }
```

This pattern is used by Relay and arguably has become a [best practice](https://atheros.ai/blog/input-object-type-as-an-argument-for-graphql-mutations-and-queries) for defining mutation inputs.

It also provides better type safety for frontend clients validating mutation inputs in files separate from where the mutation query is being executed i.e types are generated for SignupBlockResponseInput etc.

If you accept this suggestion you'll need to update your resolvers to destructure the input object too.

apps/api-journeys/src/modules/block-response/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const typeDefs = gql`
+  enum VideoBlockResponseStateEnum {
+    PLAYING
+    PAUSED
+    FINISHED
+  }
+
+  extend type Mutation {
+    journeySessionCreate(journeyId: ID!): ID!
```
tataihono, line 13: I think this should be moved into journey-session module?

apps/api-journeys/src/modules/block-response/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    },
+    async signupBlockResponseCreate (_parent, { journeySessionId, blockId, name, email }, { db }) {
+      const response = await db.blockResponse.create({
+        data: {
+          journeySessionId,
+          blockId,
+          responseData: { name, email }
+        }
+      })
+      return response.id
```
tataihono, line 39: I like to leverage destructuring to be explicit about what is being used in the response. Not needed though if you don't like it.

```suggestion
      const { id } = await db.blockResponse.create({
        data: {
          journeySessionId,
          blockId,
          responseData: { name, email }
        }
      })
      return id
```

apps/api-journeys/src/modules/block-response/index.ts
```diff
@@ -0,0 +1,69 @@
...
+          responseData: { selectedResponseBlockId }
+        }
+      })
+      return response.id
+    }
+  }
+}
+
+export default createModule({
+  id: 'block-response',
```
tataihono, line 65:
```suggestion
  id: 'blockResponse',
```

can we change all instances where we are using dashes here and with file names to camelCase? This file directory should be renamed blockResponse. This is to match the other project's (FE) extensive use of this kind labeling.

tataihono (changes requested):
Great start! Mostly just changes to match styles and best practices now.

Wondering why the JourneySession didn't get split out into its own module?

apps/api-journeys/src/modules/block-response/index.ts
soberstadt: Awesome, thanks for the reminder on this, I forget about destructuring in TS

apps/api-journeys/src/modules/block-response/index.ts
soberstadt: I didn't know how we were doing module naming, I'm good with renaming to `blockResponse`

apps/api-journeys/src/modules/block-response/index.ts
soberstadt: Done.

apps/api-journeys/src/modules/block-response/index.ts
soberstadt: I updated this on the blockResponse module, I didn't think it should be done for `userSessionCreate` since that is a single input value at this point.

soberstadt:
@tataihono I believe this is ready for re-review.

tataihono (dismissed):

soberstadt:
@tataihono which of the many generate functions did you use to update the graphql schema?

apps/api-journeys/src/lib/mockDb.ts
```diff
@@ -0,0 +1,16 @@
+import { PrismaClient } from '@prisma/client'
+import { mockDeep, mockReset } from 'jest-mock-extended'
+import { DeepMockProxy } from 'jest-mock-extended/lib/cjs/Mock'
+
+import prisma from './db'
```
tataihono, line 5: this should match the exported name which is db

```suggestion
import db from './db'
```

apps/api-journeys/src/lib/mockDb.ts
```diff
@@ -0,0 +1,16 @@
...
+jest.mock('./db', () => ({
+  __esModule: true,
+  default: mockDeep<PrismaClient>()
+}))
+
+beforeEach(() => {
+  mockReset(prismaMock)
+})
+
+export const prismaMock = prisma as unknown as DeepMockProxy<PrismaClient>
```
tataihono, line 16:
```suggestion
export const dbMock = db as unknown as DeepMockProxy<PrismaClient>
```

apps/api-journeys/src/lib/mockDb.ts
```diff
@@ -0,0 +1,16 @@
+import { PrismaClient } from '@prisma/client'
```
tataihono, line 1:
```suggestion
import { PrismaClient } from '.prisma/api-journeys-client'
```

this will give you strong types.

apps/api-journeys/src/lib/mockDb.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import prisma from './db'
+
+jest.mock('./db', () => ({
+  __esModule: true,
+  default: mockDeep<PrismaClient>()
+}))
+
+beforeEach(() => {
+  mockReset(prismaMock)
```
tataihono, line 13:
```suggestion
  mockReset(dbMock)
```

apps/api-journeys/src/modules/userSession/index.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    mutation ($journeyId: ID!) {
+      userSessionCreate(journeyId: $journeyId) {
+        id
+      }
+    }
+  `
+
+  const { data } = await testkit.execute(app, {
+    document: query,
+    variableValues: { journeyId: journeyId },
```
tataihono, line 28:
```suggestion
    variableValues: { journeyId },
```

apps/api-journeys/src/modules/userSession/index.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  })
+  const sessionId = uuidv4()
+  const journeyId = uuidv4()
+  prismaMock.userSession.create.mockResolvedValue({ id: sessionId })
+
+  const query = gql`
+    mutation ($journeyId: ID!) {
+      userSessionCreate(journeyId: $journeyId) {
+        id
+      }
```
tataihono, line 22: I am surprised that this query doesn't fail. This mutation returns id not { id }. When I try this command against the served graph (`nx serve`) it fails.

```suggestion
      userSessionCreate(journeyId: $journeyId)
```

apps/api-journeys/src/modules/userSession/index.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { testkit, gql } from 'graphql-modules'
+import { schemaBuilder } from '@core/shared/util-graphql'
+import module from '.'
+import { v4 as uuidv4 } from 'uuid'
+import { prismaMock } from '../../lib/mockDb'
+import Journey from '../journey'
+import Block from '../block'
+
+it('creates a journey session', async () => {
```
tataihono, line 9: I think a spec needs to be added here to capture what happens when the journeyId does not exist.

tataihono (changes requested):

tataihono:
Looks like your changes to the package-lock.json file are run against an old version of npm. Try `npm i -g npm@7.24.0`

PR #26, opened 2021-09-09
Setup shared-ui library for common ui components

storyworks:
I'm thinking we should house Text / Icon / Button in a common lib?

Keeping this basic so it can be quickly merged / John and I can use as a base for Text / Icon

This is literally just using the NX generator to create a new lib + giving it a basic component to test it works with storybook

### DIFF
--- a/workspace.json
+++ b/workspace.json
@@ -35,7 +35,7 @@
           }
         },
         "serve-all": {
-          "builder": "@nrwl/workspace:run-commands",
+          "executor": "@nrwl/workspace:run-commands",
           "options": {
             "commands": [
               {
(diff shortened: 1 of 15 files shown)


workspace.json
```diff
@@ -35,7 +35,7 @@
           }
         },
         "serve-all": {
-          "builder": "@nrwl/workspace:run-commands",
+          "executor": "@nrwl/workspace:run-commands",
```
storyworks, line 38: I'm not sure why this changed - it occurred after I used nx generate to create shared-lib

storyworks: Spoke to Tatai, executor is the one we want.

storyworks:
Bump? @tataihono , @gdub01 , @GeronimoJohn

GeronimoJohn (approved):
Looks good!

PR #27, opened 2021-09-10
Typography

storyworks:
https://3.basecamp.com/3105655/buckets/23904268/todolists/4127141243


tataihono (changes requested):

libs/shared/ui/src/components/Typography/Typography.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Typography from './Typography'
+
+describe('Typography', () => {
+  it('should render successfully', () => {
+    const { getByRole } = render(
+      <Typography
+        __typename="TypographyBlock"
+        id="heading3"
+        parentBlockId="question"
+        content={'Hello World!'}
```
tataihono, line 18:
```suggestion
        content="Hello World!"
```

libs/shared/ui/src/components/Typography/Typography.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+        variant={TypographyVariant.h3}
+        color={TypographyColor.primary}
+        align={TypographyAlign.left}
+      />
+    )
+    expect(
+      getByRole('heading', { name: 'Hello World!', level: 3 })
+    ).toBeTruthy()
```
tataihono, line 26:
```suggestion
    ).toBeInTheDocument()
```

libs/shared/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,32 @@
+import { ReactElement } from 'react'
+import {
+  Typography as MuiTypography,
+  TypographyProps as MuiTypographyProps
+} from '@mui/material'
+import { TypographyFields } from './__generated__/TypographyFields'
+
+export interface TypographyProps
```
tataihono, line 8: These props should just be from the API not from Mui Typography.

tataihono (changes requested):
please move typography block component back into the journeys project

tataihono (changes requested):
I would have expected to see pages/[journeyId].tsx to be updated to include the new fragment. Can you add that in then this should be done :)

tataihono:
can you generate types for your query adjustment `nx codegen journeys`

storyworks:
> can you generate types for your query adjustment `nx codegen journeys`

I have run that? This was the change
image

tataihono (approved):

PR #29, opened 2021-09-12
remove redux store and use client

tataihono:
- remove redux dependencies
- migrate conductor slice reducers to libs/client/cache/blocks.ts
- move action handling from RadioQuestion to RadioOption
- change handleClick prop to onClick

The useBlock hook is the main thing I'd love feedback on. Perhaps there is a better way to handle this. Apollo doesn't really prescribe how to manage state transformations. Instead it allowing anyone to change it.

I think this is a better solution than having frontend state managed by redux and backend state by apollo. One state manager to rule them all. Great suggestion @gdub01.

This work is needed to progress with resolvers for response storage.


gdub01 (approved):
This looks great!! 💯

PR #30, opened 2021-09-13
Video-Component

GeronimoJohn:
- Create Video Component
- Migrated VideoJS to Typescript
- Fix Firefox autoplay
- Add stories for Video Component
- Merge main branch to video-component branch

**Video Component**
Remaining TODO: 
Update VideoBlock src/url to be an m3u8/hls file instead of a youtube link (as youtube was out of our scope before). Currently no video will be played if you run journeys until src is changed to an m3u8/hls file. Though a working example can be seen in the storybook.


apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.tsx
```diff
@@ -51,7 +51,7 @@ export function RadioQuestion({
           <Typography variant="h1" gutterBottom>
             {label}
           </Typography>
-          <Typography variant="h6">{description}</Typography>
+          <Typography variant="h2">{description}</Typography>
```
tataihono, line 54: this change is unrelated. can you revert this and address in a separate PR @GeronimoJohn

apps/journeys/src/components/blocks/Video/Video.spec.tsx
```diff
@@ -1,22 +1,22 @@
...
-    id: 'main',
-    src: 'https://www.youtube.com',
-    title: 'title',
-    parentBlockId: null,
-    provider: VideoProviderEnum.YOUTUBE,
-    children: []
-  }
-  it('should render successfully', () => {
-    const { getByText } = render(<Video {...block} />)
+const block: TreeBlock<VideoBlock> = {
```
tataihono, line 6: You don't need to specify VideoBlock in here if you don't want to. Merely specifying the `__typename` field does this automatically.

apps/journeys/src/components/blocks/Video/Video.tsx
```diff
@@ -1,8 +1,31 @@
...
-import { TreeBlock } from '../../../libs/transformer/transformer'
 import React, { ReactElement } from 'react'
+import { Container } from '@mui/material'
+import { VideoPlayer } from './VideoPlayer'
+import { TreeBlock } from '../../../libs/transformer/transformer'
 import { GetJourney_journey_blocks_VideoBlock as VideoBlock } from '../../../../__generated__/GetJourney'
 
-export const Video = ({ title }: TreeBlock<VideoBlock>): ReactElement => {
-  return <Box data-testid="Video">Render {title} Here</Box>
+export const Video = ({
```
tataihono, line 7: This component seems like a pointless wrapper component. VideoPlayer should just replace this video component.

apps/journeys/src/components/blocks/Video/Video.tsx
```diff
@@ -1,8 +1,31 @@
...
+  src,
+  volume = 0,
+  autoplay = false
+}: TreeBlock<VideoBlock>): ReactElement => {
+  const videoJsOptions = {
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
tataihono, line 13: You should use the videojs.PlayerOptions type here. I think that should remove this eslint failure.

apps/journeys/src/components/blocks/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }, [player, isReady, autoPlaySuccess])
+
+  return (
+    <Container className={classes.container}>
+      <video ref={videoNode} className="video-js" />
+      {children != null ? (
+        <Container className={classes.overlayHolder}>
+          {children as unknown as JSX.Element}
+        </Container>
+      ) : null}
```
tataihono, line 109: can you drop this for now. Deal with overlays in the other branch.

apps/journeys/src/components/blocks/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,112 @@
...
+    overlayHolder: {
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      width: '100%',
+      height: 'calc(100% - 30px)',
+      justifyContent: 'center',
+      alignItems: 'center',
+      overflow: 'hidden'
+    }
```
tataihono, line 23: can you drop this for now. Deal with overlays in the other branch.

apps/journeys/src/components/blocks/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,112 @@
...
+  })
+)
+
+interface IVideoPlayerProps {
```
tataihono, line 27:
```suggestion
interface VideoPlayerProps {
```

apps/journeys/src/components/blocks/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useEffect(() => {
+    if (videoNode.current != null) {
+      player.current = videojs(videoNode.current, {
+        ...initialOptions,
+        ...options
+      })
+      player.current.on('ready', () => {
+        onReady?.(player.current)
+        setIsReady(true)
+      })
```
tataihono, line 84: do you have to remove these listeners as a function of useEffects return value?

apps/journeys/src/types.ts
```diff
@@ -0,0 +1,93 @@
+export type TypeName =
```
tataihono, line 1: I don't think this file is needed.

tataihono (dismissed):

apps/journeys/src/components/blocks/Video/Video.spec.tsx
GeronimoJohn: @tataihono Our other tests have the same pattern, I don't mind changing it, but I think for consistency it might be a good idea to stick with it and follow the pattern we currently have. Unless we're thinking of changing those.

apps/journeys/src/components/blocks/Video/Video.tsx
GeronimoJohn: Made some changes - applied your suggestion, sadly error is still there

tataihono (approved):

PR #32, opened 2021-09-14
add devcontainers

tataihono:
- add devcontainers configuration
- run services from 0.0.0.0 instead of localhost
- add contributing readme


.nvmrc
```diff
@@ -1,2 +1 @@
-16.6.2
-
+14
```
tataihono, line 1: 16 has a bunch of memory leak issues when running in containers. this falls back to the current LTS.

storyworks (approved):
Walked through Developing inside a Container + Starting gateway/back-end/front-end projects - worked as expected.

PR #34, opened 2021-09-16
add prettier linting

tataihono:


storyworks (approved):
Probably a cooldown job but I'd love to create alias' for the different projects so we don't need to `import "../../../../../libs/shared/ui..."` when using components in another lib and set some standard ordering for imports :)

PR #36, opened 2021-09-16
SignUp

storyworks:
**Ready for Review**


apps/journeys/src/components/Conductor/Conductor.stories.tsx
```diff
@@ -27,47 +27,58 @@ export const Default = (): ReactElement => (
           nextBlockId: 'step2.id',
```
storyworks, line 27: This file should be overwritten when John puts his changes in..

apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.tsx
```diff
@@ -50,7 +50,7 @@ export function RadioQuestion({
   const selectedId = data?.radioQuestionResponseCreate?.radioOptionBlockId
 
   return (
-    <Card data-testid="RadioQuestionCard">
+    <div data-testid="RadioQuestionCard">
```
storyworks, line 53: This will also be overwritten by Johns changes - needed to update in my branch so Conductor story wouldn't break. I'm also using Conductor in SignUp story.

apps/journeys/src/components/blocks/SignUp/SignUp.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <Formik
+      initialValues={initialValues}
+      validationSchema={signUpSchema}
+      onSubmit={(values) => {
+        // TODO: Handle server error responses when available
+        void onSubmitHandler(values)
+        // TODO: Replace with generic action handler
+        switch (action?.__typename) {
```
storyworks, line 72: onSubmit doesn't fire if there are validation errors, since we're assuming all valid input will return a correct response - we can fire actions right after submitting form values. Will move this to triggering after signup response is returned once we handle server errors / return successful responses.

If I add it there right now, we're not able to see it working in Storybook (see comments in that file)

apps/journeys/src/components/blocks/SignUp/SignUp.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { ReactElement } from 'react'
+
+const Demo = {
+  ...journeysConfig,
+  component: SignUp,
+  title: 'Journeys/Blocks/SignUp'
+}
+
+export const Default = (): ReactElement => (
+  <MockedProvider>
```
storyworks, line 16: Adding a custom mock here like in the test file doesn't work because we can't mock `uuid` :(

apps/journeys/src/components/Conductor/Conductor.stories.tsx
storyworks: Conductor container also needs to be fixed so we can fully see the shadows on cards
image

apps/journeys/src/libs/action/handleAction.ts
```diff
@@ -9,5 +9,11 @@ export function handleAction(action?: ActionFields | null): void {
     case 'NavigateAction':
       nextActiveBlock()
       break
+    case 'NavigateToJourneyAction':
+      nextActiveBlock({ id: action.journeyId })
+      break
+    case 'LinkAction':
+      nextActiveBlock({ id: action.url })
+      break
```
tataihono, line 17: remove this.

tataihono (changes requested):
just one removal here

libs/shared/ui/src/libs/themes/base/tokens/components.ts
```diff
@@ -1,6 +1,5 @@
-// import { lightThemeToken } from './colors'
 import { ThemeOptions } from '@mui/material'
-import { spacingThemeToken } from './spacing'
+import { spacingThemeToken } from '../tokens/spacing'
```
tataihono, line 2: revert

tataihono (changes requested):

tataihono (approved):

tataihono:
I approved the design for this as they match pixel-perfect to the designs. We can revisit modifications to this in future cycles if I was wrong to approve the design.

PR #37, opened 2021-09-16
set node version to specific version of 14 and update package-lock.json

soberstadt:
In #32 the node version was changed to v14. [Node 14 ships with npm 7](https://github.com/nodejs/node/discussions/37689), which uses lockfile format 1 instead of 2.

I set node to the latest 14 version and ran `npm i` which re-wrote the lock file.


tataihono:
@soberstadt you might want to try running the project in a [container](https://github.com/JesusFilm/core/blob/main/CONTRIBUTING.md#developing-inside-a-container). The container build process runs `npm install -g npm@latest` which today installs `7.24.0` which supports the new lock file format.

Since Microsoft doesn't ship container variants down to the patch version (e.g 14.17.6) I've intentionally let the .nvmrc version install the latest version of 14.

tataihono:
@soberstadt you might want to try running the project in a [container](https://github.com/JesusFilm/core/blob/main/CONTRIBUTING.md#developing-inside-a-container). The container build process runs npm install -g npm@latest which today installs 7.24.0 which supports the new lock file format.

Since Microsoft doesn't ship container variants down to the patch version (e.g 14.17.6) I've intentionally let the .nvmrc version install the latest version of 14.

Closing this PR. Let me know if you really think we should add in that change to the .nvmrc file but this should become less relevant as people just depend on the container for the correct node version anyhow.

PR #38, opened 2021-09-16
stub remaining prisma calls in tests

soberstadt:
Separate PR since #25 is already big! But now we don't even setup a DB to run api-journey specs.

The main problem is that I was not able to figure out [`calledWith`](stub remaining prisma calls in tests), so our stubs don't do any argument matching. I would have liked done more stubbing like:
```js
prismaMock.journey.findUnique.calledWith({ data: { id: journeyId } }).mockResolvedValue({ id: journeyId })
```


tataihono:
pulled this work into https://github.com/JesusFilm/core/pull/44.

PR #41, opened 2021-09-20
add firebase authentication

tataihono:
- api-gateway adds a user-id header to subgraph request
- api-gateway simplifies server implementation and removes fastify
- api-gateway accepts an Authorization header with value JWT from Firebase
- api-journeys adds userId to the context object


morleycb:
Hey Tatai, is the nullish coalesce part of this for local development environments?  
`admin.initializeApp({
   credential: admin.credential.cert(
     JSON.parse(process.env.GOOGLE_APPLICATION_JSON ?? '{}')
   )
 })`

It looks like for either this case, or an invalid token, we are just logging to the console and leaving out the "userId" for now.

apps/api-gateway/src/main.ts
```diff
@@ -1,13 +1,49 @@
...
+  buildService({ url }) {
+    return new AuthenticatedDataSource({ url })
+  }
+})
+
+const server = new ApolloServer({
+  gateway,
+  context: async ({ req }) => {
+    const token = req.headers.authorization ?? ''
+    if (token == null) return {}
```
morleycb, line 33: Is it possible for token to be null after the nullish coalesce operator?

tataihono: no. So I removed the coalese. Good spot!

tataihono:
https://github.com/JesusFilm/core/pull/41#issuecomment-923146517

@morleycb The nullish coalesce is because typescript says that the JSON.parse method cannot accept an undefined value. `process.env.GOOGLE_APPLICATION_JSON` has a type of `string | undefined` so I just provide a default string if the environment variable is not set.

morleycb (dismissed):

tataihono:
@morleycb I've added in a check for GOOGLE_APPLICATION_JSON. Basically, you should still be able to run the gateway for the read-only portions that don't rely on userId.

apps/api-gateway/.env
```diff
@@ -0,0 +1,9 @@
+# Firebase Admin SDK
+# https://firebase.google.com/docs/admin/setup
+# 1. Set up a Firebase project and service account
+# 2. Generate a private key file for your service account
+# 3. Copy contents of file to clipboard and open up node in your terminal
+# 4. Run `JSON.stringify({ "type":... })` and copy response
+# 5. Paste response in quotes below e.g GOOGLE_APPLICATION_JSON='{"type":...'
+# 6. Find and replace `\\n` with `\n`
+GOOGLE_APPLICATION_JSON=
```
morleycb, line 9: Tatai, for the `if (process.env.GOOGLE_APPLICATION_JSON != null) {` logic in main.ts to avoid trying to parse the non-existent JSON, I think this needs to have a comment in front of it.  E.g.:

`# GOOGLE_APPLICATION_JSON=`

morleycb (approved):

apps/api-gateway/.env
tataihono: Thanks for this. I just added a code fix to address null | undefined | ''.

PR #42, opened 2021-09-20
Image

gdub01:
Takes width, height, src and alt. Width and height are currently mandatory but wasn't sure if that should always be true. I think it makes sense for the db to know the _stored_ image sizes so I'm imagining when we request the image block, we get the original image sizes back. The client can then determine how/if it wants to transform the image... either by modifying the source url or by changing some properties it adds to the next/image component. But I'm thinking we'll deal with some of those things later.


storyworks (approved):
Just need to merge in main before you merge - there's some updates to BlockRenderer

tataihono:
I think the backend is going to fail. Did you manage to test this before approval @storyworks?

apps/journeys/src/components/blocks/Image/Image.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const Demo = {
+  ...journeysConfig,
+  component: Image,
+  title: 'Journeys/Blocks/Image'
+}
+
+const DefaultTemplate: Story<TreeBlock<ImageBlock>> = ({
+  ...props
+  // eslint-disable-next-line jsx-a11y/alt-text
```
tataihono, line 15: I don't think we should be adding in eslint disables here.

tataihono (dismissed):

apps/journeys/src/components/blocks/Image/Image.tsx
```diff
@@ -0,0 +1,34 @@
...
+        borderRadius: (theme) => theme.spacing(3),
+        overflow: 'hidden',
+        mb: 4,
+        ...sx,
+        '> div': {
+          display: 'block !important'
+        }
+      }}
+    >
+      <NextImage src={src} alt={alt} height={height} width={width} />
```
storyworks, line 31: Add `objectFit='cover'` ?

storyworks (changes requested):
There's also the issue of the conductor container cutting off the shadows on Cards, but I'm not sure if you want to fix that in another PR. https://github.com/JesusFilm/core/pull/36#discussion_r719867474

apps/journeys/src/components/blocks/Image/Image.tsx
storyworks: Actually, with that - maybe use a picture that makes it more obvious if it's been squashed in some way? It's really easy to miss with this one..

storyworks (approved):

PR #47, opened 2021-09-23
Responses frontend

tataihono:
Depends on https://github.com/JesusFilm/core/pull/46

- when option is selected send RadioQuestionResponseCreate mutation

@storyworks only need to look at changes to apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.spec.tsx everything else is server side or auto generated.

If you try and run this in your local environment you'll need to setup firebase. check the .env file for instructions on how to do that in apps/journeys folder.


storyworks (approved):
I'd expect this test file to be a bit simpler / not need a MockedProvider if we're making RadioQuestion a shared-ui component later on.

PR #51, opened 2021-09-24
JourneyProgress

tataihono:
calculates progress based on current state (uses treeBlocks and activeBlock for calculation)


lumberman:
Hey @tataihono, 0% state should have no button and be colored in solid light gray color as in Figma:
Screen Shot 2021-09-27 at 4 45 02 PM

storyworks (approved):

PR #58, opened 2021-09-29
add CardBlock to Block module

tataihono:
- add CardBlock to api-journeys block module
- add CardBlock to BlockType enum on api-journeys database
- update codegen and regenerate backend types
- regenerate api-journeys and api-gateway


morleycb (approved):

apps/api-journeys/prisma/seed.ts
```diff
@@ -32,11 +32,23 @@ async function main(): Promise<void> {
       parentOrder: 0
     }
   })
+  const card = await prisma.block.create({
+    data: {
+      journeyId: journey.id,
+      blockType: 'CardBlock',
+      parentBlockId: step.id,
+      extraAttrs: {
+        ThemeMode: ThemeMode.dark,
```
morleycb, line 41: I think this needs to be themeMode (starting with lower case t)

apps/api-journeys/prisma/seed.ts
```diff
@@ -32,11 +32,23 @@ async function main(): Promise<void> {
...
+        ThemeName: ThemeName.base
```
morleycb, line 42: I think this needs to be themeMode (starting with lower case t)

morleycb (approved):

PR #60, opened 2021-09-29
Card

tataihono:
Depends on #58 & #61 

- [x] update storybook/preview to use CommonJS instead of ESM (causing failing tests)
- [x] add CardBlock fields to pages/[journeyId].tsx
- [x] add TypographyBlock to BlockRenderer
- [x] add CardBlock component
- [x] update ThemeProvider to prevent it from loading CssBaseline when nested

To do in a future PR once ImageBlock is added. 

- [ ] use coverBlockId correlated child block as background


libs/shared/ui/src/libs/themes/base/tokens/colors.ts
```diff
@@ -4,7 +4,7 @@ export const baseColorsLight: Required<Pick<ThemeOptions, 'palette'>> = {
   palette: {
     mode: 'light',
     // DEFAULT CONTAINER COLORS
-    background: { default: '#FEFEFE' },
+    background: { default: '#FEFEFE', paper: '#FEFEFE' },
```
storyworks, line 7: Shouldn't paper be the same as primary.main?

tataihono: Screen Shot 2021-09-30 at 1 08 39 PM

I don't think that's right. Otherwise light card looks like this.

storyworks: That's what I meant yesterday when I said we'd have to update the typography tokens because using the primary palette colors, we had a background color and default text color defined in a way so they wouldn't clash. 

Currently we're using palette.text.primary for the default text color on `background.default`. We could use palette.text.secondary for default text color on `background.paper`. 

I don't know if that gets confusing though since we have primary / secondary colors in the palette too. So there are trade offs with both approaches..

storyworks: image
Another thing to note, input label color inherits from `text.secondary` and input color inherits from `text.primary` currently in the designs, Vlad intends them both to be white

storyworks (approved):

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
tataihono: we resolved this in a conversation. How it is is "correct".

PR #66, opened 2021-10-03
provide uuid as a di prop for RadioQuestion

tataihono:


apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.tsx
```diff
@@ -18,28 +18,34 @@ export const RADIO_QUESTION_RESPONSE_CREATE = gql`
   }
 `
 
+interface RadioQuestionProps extends TreeBlock<RadioQuestionBlock> {
+  uuid?: () => string
```
storyworks, line 22: Context for others - we know it's not good practice to create props solely for the purpose of testing. In this case, we can mock uuid in jest, but not for Storybook - and we're needing to mock the uuid so the stories correctly respond when we interact with it. If anyone else has thoughts on how to get around this, that would be good

apps/journeys/src/components/blocks/RadioQuestion/RadioQuestion.tsx
```diff
@@ -18,28 +18,34 @@ export const RADIO_QUESTION_RESPONSE_CREATE = gql`
...
-}: TreeBlock<RadioQuestionBlock>): ReactElement {
+  children,
+  uuid = uuidv4
+}: RadioQuestionProps): ReactElement {
   const [radioQuestionResponseCreate, { data }] =
     useMutation<RadioQuestionResponseCreate>(RADIO_QUESTION_RESPONSE_CREATE)
 
   const handleClick = async (radioOptionBlockId: string): Promise<void> => {
-    const uuid = uuidv4()
+    console.log('HELLO')
```
storyworks, line 36: remove console log?

storyworks (approved):

PR #69, opened 2021-10-04
Simplify Resolvers

tataihono:
- no longer need to define resolvers when part of the extraAttrs json object
- split large block test into independent tests for each block type


morleycb (approved):
Took me a bit to follow the new resolver pattern, but I think I get it now.  And it works!

PR #70, opened 2021-10-04
Shorten storybook script from "storybook" to "demo"

storyworks:


tataihono (commented):
can we go with storybooks own shortened version `sb`?

storyworks:
> can we go with storybooks own shortened version `sb`?

I liked `demo` over `sb` since it was shorter but still verbose. I also like how script commands aren't linked to any particular technology (lint is not `eslint`, test is not `jest`).

tataihono (approved):

PR #71, opened 2021-10-04
Add JesusFilm logo to storybook

storyworks:
Just for fun :)

image


tataihono (changes requested):
any chance you can copy the logo into the repo? If the site ever moves away from AEM the image could move.

storyworks:
> any chance you can copy the logo into the repo? If the site ever moves away from AEM the image could move.

Would it be ok if I removed the "A Cru Ministry" section while I'm at it? It's a bit too small to read easily in this case.

tataihono (approved):

PR #72, opened 2021-10-05
Fix Radio Question/Option Design

GeronimoJohn:
- Update Radio Question/Option to follow the design on figma

### DIFF
--- a/apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
+++ b/apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
@@ -13,12 +13,12 @@ const useStyles = makeStyles(() => ({
     color: '#54A055'
   },
   buttonLabels: {
-    fontSize: 16,
-    fontWeight: 700,
+    fontWeight: 600,
     lineHeight: 1.4,
     textAlign: 'start',
     justifyContent: 'flex-start',
-    borderRadius: '8px'
+    borderRadius: '8px',
+    padding: '14px 10px 14px 14px'
   }
 }))
 
--- a/apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
+++ b/apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
@@ -1,4 +1,4 @@
-import { Button } from '@mui/material'
+import { Button, useTheme } from '@mui/material'
 import RadioButtonUncheckedIcon from '@mui/icons-material/RadioButtonUnchecked'
 import CheckCircleIcon from '@mui/icons-material/CheckCircle'
 import { ReactElement } from 'react'
@@ -22,6 +22,8 @@ export function RadioOption({
   selected = false,
   onClick
 }: RadioOptionProps): ReactElement {
+  const theme = useTheme()
+
   const handleClick = (): void => {
     handleAction(action)
     onClick?.(id)
@@ -41,9 +43,10 @@ export function RadioOption({
         )
       }
       sx={{
-        fontSize: (theme) => theme.typography.body2.fontSize,
+        fontFamily: theme.typography.body2.fontFamily,
+        fontSize: theme.typography.body2.fontSize,
         fontWeight: 600,
-        lineHeight: 1.4,
+        lineHeight: theme.typography.body2.lineHeight,
         textAlign: 'start',
         justifyContent: 'flex-start',
         borderRadius: '8px',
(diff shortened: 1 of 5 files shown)


apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
```diff
@@ -13,12 +13,12 @@ const useStyles = makeStyles(() => ({
     color: '#54A055'
```
storyworks, line 13: makeStyles is now deprecated https://mui.com/styles/basics/

use `sx` instead? This way you can also reference the theme and apply a typography variant directly to this:

` sx={{
        fontSize: (theme) => theme.typography.body1.fontSize
      }}`

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
```diff
@@ -48,6 +48,15 @@ export const baseColorsLight: Required<
           }
         }
       }
+    },
+    MuiButtonGroup: {
+      styleOverrides: {
+        groupedContainedVertical: {
+          '&:not(:last-of-type)': {
+            borderColor: '#AAACBB'
```
storyworks, line 56: image
Still no border when the option is dark.. it only appears once it is clicked - do you know why? Also instead of hardcoding colors, use a color token?

storyworks: Sorry to add onto this ticket but are you also able to fix this corner? 
It looks like a div is set to have a larger `borderRadius` to it's parent div. (Both div's need same borderRadius)
image

storyworks (changes requested):

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
GeronimoJohn: That border thing is very odd. It's currently working for me at the moment. But will look into that more.

apps/journeys/src/components/blocks/RadioOption/RadioOption.tsx
GeronimoJohn: Used body2 instead of body1 - followed figma design

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
```diff
@@ -45,6 +44,24 @@ export const baseColorsLight: Required<
...
+            color: '#FFFFFF80'
+          }
+        }
+      }
+    },
+    MuiButtonGroup: {
+      styleOverrides: {
+        groupedContainedVertical: {
+          '&:not(:last-of-type)': {
+            borderColor: secondary.dark
```
storyworks, line 58: I like that you found a color token that fit for both light / dark themes here. 

To avoid having to add `color: '#FFFFFF80'` in disabled mode you could replace these 2 borderColor and override "borderBottom: `1px solid ${secondary.main}`" secondary dark throws an error :(

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
```diff
@@ -67,8 +84,7 @@ export const baseColorsDark: Required<
     },
     action: {
       // DISABLED BUTTON COLORS
-      disabled: '#26262E',
```
storyworks, line 70: Use `disabled: secondary.dark` and remove the `&.disabled : {color: _____ }` references? It's also hard because the designs use primary but with opacity changes.. Although it looks like the opacity has just been randomly picked based on what looks nice..

Let's use the tokens for now and if we need the change secondary tokens to be an rgb value (includes opacity), then it should already work for these disabled buttons.

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
```diff
@@ -67,8 +84,7 @@ export const baseColorsDark: Required<
...
-      disabledBackground: '#BDBFCF'
+      disabledBackground: secondary.light
     }
   },
   components: {
```
storyworks, line 90: After this line, add `...baseColorsLight.components,`
Then you should be able to remove `MuiButtonGroup` styling in the dark mode and inherit the same one that's used in the light theme.

storyworks (dismissed):

libs/shared/ui/src/libs/themes/base/tokens/colors.ts
GeronimoJohn: Have also applied this logic to the light modes disabled.

PR #75, opened 2021-10-07
Autogenerate dark mode components

storyworks:
Allows us to create stories for light mode and dark mode stories will be auto generated. View the stories side by side with the ThemeMode toolbar icon 
image
 

Any color issues found in the stories will be resolved in another branch


apps/journeys/src/components/blocks/Card/Card.tsx
```diff
@@ -20,57 +19,73 @@ export function Card({
   )
 
   return (
-    <CardWrapper themeMode={themeMode} themeName={themeName}>
-      <Paper
```
storyworks, line 24: Moved Paper to CardWrapper so we could reuse it in StoryCard

apps/journeys/src/libs/storybook/containers.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { CardWrapper } from '../../components/blocks/Card'
+
+interface StoryCardProps {
+  id?: string
+  backgroundColor?: string
+  themeMode?: ThemeMode
+  children: ReactNode
+}
+
+export const StoryCard = ({
```
storyworks, line 13: Allows us to wrap block components that would always render within a Card without the limitation of only passing in TreeBlock children and having to pass in null for theme props.

libs/shared/ui/src/libs/storybook/decorators.tsx
```diff
@@ -1,19 +1,71 @@
...
+const ThemeContainer = ({
+  mode,
+  layout = 'padded',
+  children
+}: ThemeDecoratorProps): ReactElement => {
+  return (
+    <div
+      style={{
+        position: 'absolute',
+        inset: mode === 'light' ? '0px 50vw 0px 0px' : '0px 0px 0px 50vw',
```
storyworks, line 24: Styling derived from Storybooks own https://5a375b97f4b14f0020b0cda3-wbeulgbetj.chromatic.com/?path=/story/core-layout--padded-inline

tataihono (approved):

PR #81, opened 2021-10-11
API Deployment

murphomatic:
Tatai - At long last, here are the workflows and requisite assets to support deployments in both stage and production (main) environments. This should merge fairly easily, I kept up on merging in main to stay abreast of work other developers were doing. One small issue - package.json includes a package called subscriptions-transport-ws, which is 0.10.0 in main .. however that creates some sort of dependency conflict, so I have it at 0.9.19 in this branch. Unless there's a pressing reason to push this to 0.10.0 - we might want to prefer 0.9.19 so that all dependency trees are happy.


apps/api-journeys/.eslintrc.json
```diff
@@ -8,9 +8,6 @@
     },
     {
       "files": ["*.ts", "*.tsx"],
-      "parserOptions": {
-        "project": ["apps/api-journeys/tsconfig.*?.json"]
-      },
```
tataihono, line 13: revert. This is required for linting api-journeys with type safety.

apps/api-journeys/src/main.ts
```diff
@@ -1,21 +1,33 @@
...
+    })
+    .catch((err) => {
+      console.error(err)
+    })
+} else {
+  const lambdaserver = new ApolloServerLambda({
+    schema,
+    context: {
+      db
+    }
```
tataihono, line 30: the userId context must be passed here.

apps/api-journeys/tests/setupTests.ts
```diff
@@ -0,0 +1,30 @@
...
+  })
+}
+
+beforeAll(async () => {
+  await truncate()
+})
+
+afterEach(async () => {
+  await truncate()
+})
```
tataihono, line 30: delete.

apps/api-journeys/types.d.ts
```diff
@@ -1,4 +1,4 @@
-import { PrismaClient } from '.prisma/api-journeys-client'
+import { PrismaClient } from '.prisma/api-journeys-1-client'
```
tataihono, line 1: this is not a file?

output.js
```diff
@@ -0,0 +1,5 @@
+function handler(data, serverless, options) {
+    console.log('Output received', data);
+}
+
+module.exports = { handler };
```
tataihono, line 5: can this file get a reason why its not in an app? this is not related to any nx related config right?

package.json
```diff
@@ -85,6 +86,7 @@
     "@nrwl/tao": "^12.6.3",
     "@nrwl/web": "^12.7.2",
     "@nrwl/workspace": "12.6.3",
+    "@ns3/nx-serverless": "*",
```
tataihono, line 89: should probably be versioned

workspace.json
```diff
@@ -125,7 +151,7 @@
             "passWithNoTests": true
           }
         },
-        "generate": {
+        "prismagenerate": {
```
tataihono, line 154:
```suggestion
        "generate-prisma": {
```

.github/workflows/prod-deploy.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Update Apollo Gateway schema to proper API Gateway URL
+        run: ssh gateway "sudo sed -i 's|http://127.0.0.1:4001/graphql|$JOURNEYS_API_ENDPOINT|g' /opt/apollo-gateway/apps/api-gateway/schema.graphql"
+
+      - name: Restart Apollo Gateway
+        run: ssh gateway "sudo -- sh -c '. /.nvm/nvm.sh; \cp -r /etc/.pm2 ~; pkill pm2; pkill node; pm2 resurrect; pm2 start 0'"
+
+      # Cleanup / remove Github access to postgres.
+      - name: Remove Github Actions Access to Postgres
+        if: always()
+        run: aws ec2 revoke-security-group-ingress --group-id $LAMBDA_SECURITY_GROUP_ID --protocol tcp --port 5432 --cidr ${{ steps.ip.outputs.ipv4 }}/32
```
tataihono, line 123: this all needs to move out to a deploy script for api-gateway and api-journey respectively. Otherwise the deploy is going to run every time any project in the monorepo is pushed.

workspace.json
tataihono: just to align with generate-graphql

PR #88, opened 2021-10-18
Responsive Card

storyworks:
Designs shown here https://www.figma.com/file/Rt5hcy8zJIUwz4SmnfniSG/NextSteps---Design-System?node-id=525%3A11525


gdub01:
The designs look great!

What do you think about updating seeds so that it uses this card/conductor? Seeds currently makes the journey run without these changes. 

Also currently swiper loads all blocks and swiping left/right for me is not the fastest with these fully-loaded cards. The performance will likely be worst on low-end devices. 

It may be outside the scope of this, but what do you think about on loading card content once it's focused.. and have the card just look black while it's moving?

storyworks:
> What do you think about updating seeds so that it uses this card/conductor? Seeds currently makes the journey run without these changes.

This is part of the NUA project pitch also running in this cycle so I'll leave them to merge my changes and update seed.

> Also currently swiper loads all blocks and swiping left/right for me is not the fastest with these fully-loaded cards. The performance will likely be worst on low-end devices.

> It may be outside the scope of this, but what do you think about on loading card content once it's focused.. and have the card just look black while it's moving?

Right now I haven't enabled the swipe, will do that in a seperate PR since this one only focuses on making the Conductor responsive. But yes totally agreed we should look into improving performance - and that's a good suggestion. Feel free to add it to the list to investigate during cooldown :)

gdub01 (approved):
Looks great!

PR #89, opened 2021-10-18
Columns

gdub01:
This branch probably should have been named grid in hindsight... but it supports grids like: 

```
type: Enum('container' | 'item' | 'container item')
bpMd: Enum(_6 | _12)
```

and it would be used like this:
step:
  grid(container):
    grid(item):
      radioQuestions
    grid(item):
      image


apps/journeys/src/components/blocks/Grid/Grid.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children
+}: TreeBlock<GridBlock>): ReactElement {
+  const columnSize = md === null ? 12 : (parseInt(md.split('_')[1]) as GridSize)
+  const gridType = type === null ? GridType.container : type
+
+  return (
+    <MaterialGrid
+      container={gridType.includes('container')}
+      item={gridType.includes('item')}
+      md={gridType.includes('item') ? columnSize : undefined}
```
storyworks, line 20: `md` is Tablet (P),  so though there are no designs for it now, I'm assuming it should be 12. `sm / lg` are Mobile /tablet landscape - so should take custom column size.

Are you also able to fix the broken stories/tests?

apps/api-journeys/schema.graphql
```diff
@@ -64,6 +64,34 @@ type CardBlock implements Block {
...
+  _9
+  _10
+  _11
+  _12
+}
+
+type GridBlock implements Block {
+  id: ID!
+  parentBlockId: ID
+  md: ColumnSize
```
storyworks, line 85: Maybe we could rename md to `gridSize: GridSize` ? or something not md, since md is Tablet portrait..

apps/api-journeys/schema.graphql
```diff
@@ -64,6 +64,34 @@ type CardBlock implements Block {
   themeName: ThemeName
 }
 
+enum ColumnSize {
```
storyworks, line 67: What do you think of adding 'auto' to this list? like mui's GridSize

apps/api-journeys/schema.graphql
```diff
@@ -64,6 +64,34 @@ type CardBlock implements Block {
...
+  _10
+  _11
+  _12
+}
+
+type GridBlock implements Block {
+  id: ID!
+  parentBlockId: ID
+  md: ColumnSize
+  type: GridType
```
storyworks, line 86: We also need a prop that lets each column know the alignment of the content:

image

storyworks (changes requested):

apps/api-journeys/schema.graphql
gdub01: renamed it to lg

apps/journeys/src/components/blocks/Grid/Grid.tsx
storyworks: You're probably already making the change but adding here to document our convo. It seems like from the designs the landscape viewports can be both single or double column. So this check would need to be on `sm, lg, xl`

image

I think Tatai's PR is determining if the content is offset or full screen so the check here should be enough.

apps/journeys/src/components/blocks/Grid/Grid.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children
+}: TreeBlock<GridBlock>): ReactElement {
+  const isContainer = container !== null && container !== undefined
+  const isItem = item !== null && item !== undefined
+
+  return (
+    <MaterialGrid
+      container={isContainer}
+      item={isItem}
+      lg={item?.lg.replace('_', '') as GridSize}
```
storyworks, line 19: Could have just kept that `columnSize` check you had before and applied it to all breakpoints but make the direction `row` for landscape viewports or `column` for portrait? No need for `item.lg` prop?

apps/journeys/src/components/blocks/GridContainer/GridContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+  justifyContent,
+  alignItems,
+  children
+}: TreeBlock<GridContainerBlock>): ReactElement {
+
+  return (
+    <MaterialGrid 
+      container
+      spacing={spacing}
+      direction={hyphenate(direction) as GridDirection}
```
storyworks, line 19: Possibly don't need this now and linting needs to be fixed, but other than that - looks good :)
```suggestion
      direction={hyphenate(direction)}
```

storyworks (changes requested):

apps/journeys/src/components/blocks/GridContainer/GridContainer.tsx
storyworks: From Slack:

image

storyworks (approved):

PR #90, opened 2021-10-19
Video Player Extensions

GeronimoJohn:
- Create trigger block schema
- Implement trigger block to the front end
- Add media support for arclight
- Add VideoBlockResponses
- Add test for trigger block
- Update Video tests and storybook


apps/api-journeys/src/modules/video/video.spec.ts
```diff
@@ -112,7 +193,10 @@ describe('VideoModule', () => {
         blockId: block1.id,
         userId,
         extraAttrs: {
-          state: 'PLAYING'
+          mediaComponentId: '2_0-FallingPlates',
+          languageId: '529',
+          state: 'PLAYING',
+          position: 0.3
```
tataihono, line 199:
```suggestion
          state: 'PLAYING',
          position: 30
```

apps/api-journeys/prisma/migrations/20211014003622_add_trigger_block_to_block_type/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterEnum
+ALTER TYPE "BlockType" ADD VALUE 'TriggerBlock';
```
tataihono, line 2:
```suggestion
ALTER TYPE "BlockType" ADD VALUE 'VideoTriggerBlock';
```

since this block is pretty much only going to be used by video.

apps/api-journeys/prisma/seed.ts
```diff
@@ -47,19 +47,45 @@ async function main(): Promise<void> {
...
   })
-  await prisma.block.create({
+  const video = await prisma.block.create({
     data: {
       journeyId: journey.id,
       blockType: 'VideoBlock',
       parentBlockId: card.id,
       extraAttrs: {
-        src: 'https://playertest.longtailvideo.com/adaptive/elephants_dream_v4/index.m3u8',
+        video: {
```
tataihono, line 56:
```suggestion
        content: {
```

maybe content would be more appropriate here. Otherwise its a video prop inside a video.

apps/api-journeys/src/modules/trigger/trigger.ts
```diff
@@ -0,0 +1,16 @@
+import { createModule, gql } from 'graphql-modules'
+
+const typeDefs = gql`
+  type TriggerBlock implements Block {
```
tataihono, line 4: can you add a comment above for clients to understand what this block is for? Also try VideoTriggerBlock for this.

apps/api-journeys/src/modules/trigger/trigger.ts
```diff
@@ -0,0 +1,16 @@
...
+    id: ID!
+    parentBlockId: ID
+    triggerStart: Int!
```
tataihono, line 7: can you add a comment above this for clients to understand that this is the number of seconds since the start of the video

tataihono (changes requested):
this PR needs to be split up into frontend and backend PRs.

apps/api-journeys/codegen.yml
```diff
@@ -21,6 +21,7 @@ generates:
         RadioOptionBlock: .prisma/api-journeys-client#Block as BlockType
         SignUpBlock: .prisma/api-journeys-client#Block as BlockType
         StepBlock: .prisma/api-journeys-client#Block as BlockType
+        TriggerBlock: .prisma/api-journeys-client#Block as BlockType
```
tataihono, line 24:
```suggestion
        VideoTriggerBlock: .prisma/api-journeys-client#Block as BlockType
```

apps/api-journeys/src/modules/trigger/index.ts
```diff
@@ -0,0 +1 @@
+export { videoTriggerModule } from './trigger'
```
tataihono, line 1: this should be moved from the trigger folder to the videoTrigger folder.

apps/api-journeys/src/modules/trigger/trigger.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  journeyModule,
+  blockModule,
+  actionModule
+} from '..'
+import dbMock from '../../../tests/dbMock'
+import { v4 as uuidv4 } from 'uuid'
+import { Block, ThemeName, ThemeMode } from '.prisma/api-journeys-client'
+import { DocumentNode, ExecutionResult } from 'graphql'
+
+describe('TriggerModule', () => {
```
tataihono, line 14:
```suggestion
describe('VideoTriggerModule', () => {
```

Can you make sure to update all instances of trigger to videoTrigger?

apps/api-journeys/src/modules/video/video.ts
```diff
@@ -19,22 +21,40 @@ const typeDefs = gql`
...
+    muted: Boolean
     autoplay: Boolean
+    content: Video
   }
 
   type VideoResponse implements Response {
     id: ID!
     userId: ID!
     state: VideoResponseStateEnum!
+    position: Float
```
tataihono, line 57: this should just be an int. we don't need to store more than seconds.

apps/api-journeys/src/modules/video/video.ts
```diff
@@ -19,22 +21,40 @@ const typeDefs = gql`
     id: ID
     blockId: ID!
     state: VideoResponseStateEnum!
+    position: Float
+  }
+
+  interface Video {
```
tataihono, line 27:
```suggestion
  interface VideoContent {
```

can we go with VIdeoContent instead of just video.

apps/api-journeys/src/modules/video/video.ts
```diff
@@ -19,22 +21,40 @@ const typeDefs = gql`
...
     parentBlockId: ID
-    src: String!
     title: String!
+    startAt: Int
+    endAt: Int
     description: String
-    volume: Int
+    muted: Boolean
     autoplay: Boolean
+    content: Video
```
tataihono, line 50: this should be required.

apps/api-journeys/src/modules/video/video.ts
```diff
@@ -19,22 +21,40 @@ const typeDefs = gql`
...
+  type VideoGeneric implements Video {
+    src: String!
   }
 
   type VideoBlock implements Block {
     id: ID!
     parentBlockId: ID
-    src: String!
     title: String!
+    startAt: Int
```
tataihono, line 45: can you add comments to many of these props so we know what they are for. The name doesn't really show us what they are needed for.

apps/journeys/pages/[journeyId].tsx
```diff
@@ -20,6 +20,7 @@ import {
   STEP_FIELDS,
   RADIO_OPTION_FIELDS,
   RADIO_QUESTION_FIELDS,
+  TRIGGER_FIELDS,
```
Comment (tataihono) [inline apps/journeys/pages/[journeyId].tsx:23]:
```suggestion
  VIDEO_TRIGGER_FIELDS,
```

same changes on backend should be made here (rename trigger to video trigger).

apps/journeys/src/components/BlockRenderer/BlockRenderer.spec.tsx
```diff
@@ -152,10 +152,15 @@ describe('BlockRenderer', () => {
     const block: TreeBlock = {
       __typename: 'VideoBlock',
       id: 'main',
-      src: 'https://www.youtube.com',
+      videoContent: {
```
tataihono, line 155: I thought on the backend this field was just called content

apps/journeys/src/components/blocks/Trigger/Trigger.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface TriggerProps extends TreeBlock<TriggerBlock> {
+  player?: videojs.Player
+}
+
+export function Trigger({
+  player,
+  triggerAction,
+  triggerStart
+}: TriggerProps): ReactElement {
+  const handleTrigger = (player: videojs.Player): void => {
```
tataihono, line 16: I actually think this should just be implemented as a useEffect.

```Typescript
useEffect(() => {
	if(player != null) {
		player.on('timeupdate', () => {
      if (player.currentTime() >= triggerStart) {
        if (player.isFullscreen()) player.exitFullscreen()
        player.pause()
        handleAction(triggerAction)
      }
    })
  }
}, [player])
```

apps/journeys/src/components/blocks/Video/Video.spec.tsx
```diff
@@ -1,22 +1,92 @@
...
+      <MockedProvider
+        mocks={[
+          {
+            request: {
+              query: VIDEO_RESPONSE_CREATE,
+              variables: {
+                id: 'uuid',
+                blockId: 'Video1',
+                state: VideoResponseStateEnum.PLAYING,
+                position: 0.3
```
tataihono, line 36: should provide whole seconds

tataihono (changes requested):

apps/journeys/src/components/BlockRenderer/BlockRenderer.spec.tsx
GeronimoJohn: I think there's another prop from a different block that's also called content and graphql at the client-side didn't like that. Hence I had to give it a different name on the front-end (videoContent)

GeronimoJohn: Should we leave the backend to be named content or just change it to videoContent? But if we do videoContent will have a prop called videoContent

apps/api-journeys/src/modules/video/video.ts
GeronimoJohn: I just tried setting the position to have the type of INT. But VideoJS gives us back the time in float. When I tested this to record time I just get a type error in the front end. Can convert the value but not sure if this is how we should go about it.

tataihono: yup just turn it itno a seconds int.

apps/journeys/src/components/BlockRenderer/BlockRenderer.spec.tsx
storyworks: Seems like you've changed it to `content` in the backend. 

Now you just need to update videoFields.ts to 
```
content {
      src
    }
```
Run `nx codegen journeys` and update the stories / test files

apps/journeys/pages/[journeySlug].tsx
```diff
@@ -110,6 +112,9 @@ export const getServerSideProps: GetServerSideProps<JourneyPageProps> = async (
             ... on StepBlock {
               ...StepFields
             }
+            ... on VideoTriggerBlock {
```
Comment (storyworks) [inline apps/journeys/pages/[journeySlug].tsx:115]: These should be referenced in alphabetical order, move under VideoBlock :)

apps/journeys/src/components/BlockRenderer/BlockRenderer.spec.tsx
storyworks: And also rename Trigger.tsx etc VideoTrigger.tsx?

I know you've moved the folder within Video which helps us find it if we're looking through the folder tree, but naming it VideoTrigger.tsx will also help us find it when doing a `cmd+P` search.

apps/journeys/src/components/blocks/Video/Trigger/Trigger.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { ReactElement, useEffect } from 'react'
+import { GetJourney_journey_blocks_VideoTriggerBlock as TriggerBlock } from '../../../../../__generated__/GetJourney'
+import { handleAction } from '../../../../libs/action'
+import { useRouter } from 'next/router'
+
+export interface TriggerProps extends TreeBlock<TriggerBlock> {
+  player?: videojs.Player
+}
+
+export function Trigger({
```
storyworks: Seems really strange to have this as a seperate component that just triggers a useEffect (seemed strange when it was just triggering a function as well)

Could VideoTriggerBlock not have just been a VideoTriggerAction? And then VideoBlock has a prop `triggers` that takes an array of VideoTriggerActions?

And then in Video.tsx, on first render we loop through the triggers and start the `player.on('timeupdate'....` listeners?

storyworks: After talking to @tataihono , VideoTrigger was made a block so its children could be the overlays that appear at different timestamps. Right now we're not implementing overlays so have held off from passing children into the render function. 

If we decide not to use video overlays in the end, this should be refactored, removing the VideoTrigger.tsx component and treating the Video `children` as `triggers`, looping through them to add the listeners in `Video.tsx`

storyworks (approved):

storyworks (approved):

storyworks (approved):

storyworks (approved):

PR #104, opened 2021-11-04
updated slug tests

csiyang:
Changed mocked error object to include the error code on tests


tataihono:
handled in https://github.com/JesusFilm/core/pull/107

PR #117, opened 2021-11-09
Storybook - stacked "all" mode

storyworks:
Make "all" mode stacked so it displays correctly for the 5 breakpoints in Chromatic

**Reminder that when manually testing components in the viewports, use either "light" or "dark".** 
"all" mode is for Chromatic testing or with default viewport (where height is variable).

https://3.basecamp.com/3105655/buckets/24865634/todos/4327677656

I've also tidied up the now redundant tests.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -13,8 +13,8 @@ const customViewports = {
   mobileMin: {
     name: 'Small Mobile',
     styles: {
-      width: '320px',
-      height: '568px'
+      width: '360px',
+      height: '640px'
     },
     type: 'mobile'
   },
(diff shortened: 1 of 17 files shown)


.storybook/preview.js
```diff
@@ -44,9 +44,7 @@ module.exports = {
         disable: true
       }
     },
-    // 2x viewport breakpoints (XS, SM, MD) Largest out of bounds for Chromatic.
-    // TODO: Split out light / dark mode tests during Cooldown
-    chromatic: { viewports: [640, 1136, 1200] },
+    chromatic: { viewports: [360, 568, 600, 961, 1200] },
```
storyworks, line 47: Analysis into global and our real data device sizes shows 360x640 as the minimum sized screen we should support.

GeronimoJohn (approved):
This is looking really good. The only thing I had in mind is that majority of our stories now stretch to their parent's width. I think these stories are wrapped around `StoryCard`. In terms of visuals, it can just be a little bit off.

Here's an example
Screen Shot 2021-11-10 at 11 31 10 AM

storyworks:
> This is looking really good. The only thing I had in mind is that majority of our stories now stretch to their parent's width. I think these stories are wrapped around `StoryCard`. In terms of visuals, it can just be a little bit off.

Agreed, it would be nice it if had a maxWidth the same as Card. Although how we display it may change once we have docs created for each component. Will look into it in another PR.

PR #118, opened 2021-11-10
Render by step

csiyang:
- Renders card content for only the active card
- Place holder card used for cards that are not active (for the swiper slide)
- added e2e tests for journeys


lumberman:
Hey, @csiyang I appreciate very much you working on this fix!
I have an issue with how it feels now. Initially, I thought that the slider somehow skips one step (grey coloured card) and goes from 1st to 3rd card directly. Then I realized it's due to simultaneous colour changes. I'm not sure what solution will be to that problem, but can you try adding a bit of colour transition or delay to remove this confusion?

csiyang:
Hi @lumberman, we updated how the step is rendered, please let me know if it's what you had in mind?

tataihono (approved):
approving everything but the video blocks.

apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -133,7 +139,19 @@ const Conductor = ({ blocks }: ConductorProps): ReactElement => {
...
                 }}
               >
-                <BlockRenderer {...block} />
+                {activeBlock?.id === block.id ||
+                previousBlocks[previousBlocks.length - 1]?.id === block.id ? (
+                  <BlockRenderer {...block} />
+                ) : (
+                  <CardWrapper
+                    id={block.id}
+                    backgroundColor={theme.palette.background.paper}
```
storyworks, line 148:
```suggestion
                    backgroundColor={theme.palette.primary.light}
```
Reverting back to this - I've just remembered we did this so the next Card was more visible

storyworks (changes requested):

storyworks (approved):

PR #121, opened 2021-11-16
Add publishedAt and createdAt props on Journey model

jianwei1:
Added publishedAt and createdAt props for the journey schema(removed the "published" prop). Updated the tests to reflect this change


apps/api-gateway/schema.graphql
```diff
@@ -210,11 +210,12 @@ type Journey
   @join__type(graph: JOURNEYS, key: "id")
 {
   blocks: [Block!] @join__field(graph: JOURNEYS)
+  createdOn: String! @join__field(graph: JOURNEYS)
   description: String @join__field(graph: JOURNEYS)
   id: ID! @join__field(graph: JOURNEYS)
   locale: String! @join__field(graph: JOURNEYS)
   primaryImageBlock: ImageBlock @join__field(graph: JOURNEYS)
-  published: Boolean! @join__field(graph: JOURNEYS)
+  publshedOn: String @join__field(graph: JOURNEYS)
```
storyworks, line 218:
```suggestion
  publishedAt: String @join__field(graph: JOURNEYS)
```

There are a couple places where it's still called `On` rather than `At`, you can do a global search in VS Code by doing `CMD + SHIFT + F` 
image

Same for createdAt

apps/api-journeys/schema.graphql
```diff
@@ -183,13 +183,14 @@ input ImageBlockCreateInput {
...
 type Journey @key(fields: "id") {
   id: ID!
-  published: Boolean!
   title: String!
   locale: String!
   themeMode: ThemeMode!
   themeName: ThemeName!
   description: String
   slug: String!
+  publshedOn: String
```
storyworks, line 192:
```suggestion
  publishedOn: String
```
Spelling

apps/api-journeys/src/modules/button/button.spec.ts
```diff
@@ -24,13 +24,14 @@ describe('ButtonModule', () => {
...
       title: 'published',
-      published: true,
       locale: 'en-US',
       themeMode: ThemeMode.light,
       themeName: ThemeName.base,
       description: null,
       primaryImageBlockId: null,
-      slug: 'published-slug'
+      slug: 'published-slug',
+      publishedAt: new Date(),
```
storyworks, line 33:
```suggestion
      publishedAt: '2021-11-15T03:42:22.322Z',
```
Use a hard coded value in these all the component tests that aren't Journey, they are here just as a placeholder - we don't need to properly test `publishedAt` and `createdAt` here.

apps/api-journeys/prisma/factories/nua1.ts
```diff
@@ -12,7 +12,7 @@ export async function nua1(prisma: PrismaClient): Promise<void> {
     journey = await prisma.journey.create({
       data: {
         title: 'Fact or Fiction',
-        published: true,
+        publishedAt: '2021-11-15T03:42:22.322Z',
```
storyworks, line 15: These nua seeds would need `createdAt` values too? You've said it was a required prop everywhere else

apps/api-journeys/src/__generated__/types.ts
```diff
@@ -14,6 +14,7 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
+  Date: any;
```
storyworks, line 17: Why have you added both a Date Scalar and a DateTime type?

apps/api-journeys/src/modules/journey/journey.spec.ts
```diff
@@ -85,13 +91,64 @@ describe('JourneyModule', () => {
...
+            'themeName',
+            'themeMode',
+            'slug',
+            'publishedAt',
+            'createdAt'
+          ])
+        )
+      })
+
+      it('publishedAt prop is null if journey is not yet published', async () => {
```
storyworks, line 140:
```suggestion
      it('should not be published by default', async () => {
```

Also add tests:
`should set createdAt with current date time` - just check the date and time to the hour is correct.
`should set publishedAt when _____` - trigger the published condition and check the date/time like above.

apps/api-journeys/src/modules/button/button.spec.ts
csiyang: Can't use a hardcoded string as publishedAt needs to be a date object. As discussed we will hard code a date object in.

apps/api-journeys/src/__generated__/types.ts
csiyang: Just using scalar also works!

apps/api-journeys/src/__generated__/types.ts
```diff
@@ -956,3 +967,5 @@ export type Resolvers<ContextType = GraphQLModules.Context> = {
   VideoTriggerBlock?: VideoTriggerBlockResolvers<ContextType>;
 };
 
+
```
storyworks, line 970: Remove extra line

apps/api-journeys/src/__generated__/types.ts
```diff
@@ -83,6 +84,7 @@ export type CardBlock = Block & {
   themeName?: Maybe<ThemeName>;
 };
 
+
```
storyworks, line 87: Remove extra line

apps/api-journeys/schema.graphql
```diff
@@ -71,6 +71,12 @@ type CardBlock implements Block {
   themeName: ThemeName
 }
 
+scalar Date
+
+type dateTime {
```
storyworks, line 76: Where is this ever used? Remove this as well?

apps/api-journeys/src/modules/journey/journey.ts
```diff
@@ -18,20 +18,27 @@ const typeDefs = gql`
...
   }
 
   enum IdType {
     databaseId
     slug
   }
 
+  scalar Date
+
+  # type dateTime {
```
storyworks, line 38: Remove commented out?

apps/journeys/pages/[journeySlug].tsx
```diff
@@ -79,6 +79,8 @@ export const getServerSideProps: GetServerSideProps<JourneyPageProps> = async (
           themeMode
           title
           description
+          publishedAt
```
Comment (storyworks) [inline apps/journeys/pages/[journeySlug].tsx:82]: Don't need to retrieve these for journeys but nice that you tested that it works.

storyworks (approved):

PR #123, opened 2021-11-17
basic authentication schema

gdub01:
user & userJourney schema and types in api-journeys


tataihono (changes requested):
user should be a stand-alone microservice as many services will depend on the user object. should be able to just reference the userId which is already passed to every microservice instead of creating your own internal userId.

apps/api-journeys/schema.graphql
```diff
@@ -389,6 +394,54 @@ enum TypographyVariant {
   overline
 }
 
+type User @key(fields: "id") {
+  id: ID!
+  firebaseId: ID
+  firstName: String
```
storyworks, line 400: Would firstName be required? So we have a at least a name to display with all the change access modals / avatars.

apps/api-journeys/src/modules/user/user.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        ...contextValue,
+        db: dbMock
+      }
+    })
+  }
+
+  describe('Query', () => {
+    describe('get user by id', () => {
+      it('returns user', async () => {
+        const user: User = {
```
storyworks, line 37: You could extract this, define it outside of the `describe` and reuse it in all the tests?

Also 'lo' at the examples xD

apps/api-journeys/schema.graphql
gdub01: That's fair. The issue is email signup doesn't have a first name field. There's a way to maybe split the email at the @ sign and get a first name from there, though. Think we should do that?

gdub01:
> user should be a stand-alone microservice as many services will depend on the user object. should be able to just reference the userId which is already passed to every microservice instead of creating your own internal userId.

I'm not sure what this means. 

The user model has an id that's used for the user_journey join table that needs to be unique. 

A service can pass any value they'd like for the id, but it defaults to a UUID if that helps?

You mentioned a userId is already passed to every micro service... but I don't know anything about that. Can you elaborate?  ie. where does that id come from?

apps/api-journeys/schema.graphql
storyworks: Hmm if we can't add first name to the email signup, possibly we can just display the email as the fallback.

Although.. if we can't get first name from email sign up and there's no userUpdate mutation - how will we ever get firstName and lastName?

tataihono:
So if you call the api with an authorization header containing a user's firebase id token. This is decoded by the api gateway here: https://github.com/JesusFilm/core/blob/main/apps/api-gateway/src/main.ts#L45

The uid is then passed to each graphql api as userId in the context object. You can see an example of this here: https://github.com/JesusFilm/core/blob/main/apps/api-journeys/src/modules/image/image.ts#L70

This uid is set by firebase and is guaranteed unique by firebase. It is a string not a uuid.

A user service could be created to take in any additional information about a user in a user service (e.g. api-users).  You can define a new User table where the id is just the uid from firebase.

gdub01:
Okay ty for reply. I'm afraid I'm only kind of understanding. 

The task we have is for 'basic authentication' which in the strictest sense is Authorization: Basic <credentials> where credentials is the Base64 encoding of ID and password joined by a single colon :.

However we're using oauth with multiple external providers using a 3P of firebase which is not basic auth, and it seems to have already been at least partially implemented which is... a bit surprising. 

It also seems we're assuming auth is always going to be done by firebase. ie. if you don't have a session from firebase, you can't access our graphql setup? Last I heard there was not a small chance we were using aws' cognito in the long term and I hadn't heard differently?

By "user's firebase id token" do you mean the "access token" firebase issues, or the user's "id" firebase issues? I'm storing firebase's id on the user's table which I believe makes sense... unless you're wanting a firebase_user's table? I'm not really understanding the need for that, though. I'm not storing the access token as they are meant to be temporary. 

The general plan is to have access to the user's firebase id so that we can find journeys relating to the logged in user. Looking at the code you've written in the past, it seems you're getting the firebase id already, but calling firebase's user id a generic userId instead of an explicit firebaseUserId. Using any alternative auth provider could make that confusing?

If we make the firebaseUserId co-opt our user table's 'id' that's fine, but risk of collision increases as we add more external providers, each with their own id. We could make the index unique from both the id column and provider column, though. ie. we set the id as firebase's user id and provider as 'firebase' for example. But the user_journey join table will be a little harder.  @tataihono

apps/api-journeys/schema.graphql
gdub01: Yes that's true - we can't ever get that information. There was an idea to provide a screen to the user to allow them to add/update info, but that got vetoed at this time.

tataihono:
Okay ty for reply. I'm afraid I'm only kind of understanding.

**The task we have is for 'basic authentication' which in the strictest sense is Authorization: Basic where credentials is the Base64 encoding of ID and password joined by a single colon :.**

Sorry. The pitch name Basic Authentication here just means the first version of Auth. There was a much more complex Authentication pitch so this was called Basic Authentication to note that it was a stripped-down version of the original pitch.

**However we're using oauth with multiple external providers using a 3P of firebase which is not basic auth, and it seems to have already been at least partially implemented which is... a bit surprising.**

Right, it was implemented to support anonymous authentication on the visitor client.

**It also seems we're assuming auth is always going to be done by firebase. ie. if you don't have a session from firebase, you can't access our graphql setup? Last I heard there was not a small chance we were using aws' cognito in the long term and I hadn't heard differently?**

We are using firebase moving forward. AWS Cognito does not support anonymous authentication nor transitioning those anonymous accounts to a properly authenticated account.

**By "user's firebase id token" do you mean the "access token" firebase issues, or the user's "id" firebase issues? I'm storing firebase's id on the user's table which I believe makes sense... unless you're wanting a firebase_user's table? I'm not really understanding the need for that, though. I'm not storing the access token as they are meant to be temporary.**

The thing you are referring to as an access token is more formally called an ID Token by [firebase](https://firebase.google.com/docs/auth/admin/verify-id-tokens). This token contains encoded data that allows us to verify it without needing to connect to a third-party service or connect to a local database which makes it a perfect fit for the api-gateway as it needs to do its verification without a datastore.

**The general plan is to have access to the user's firebase id so that we can find journeys relating to the logged in user. Looking at the code you've written in the past, it seems you're getting the firebase id already, but calling firebase's user id a generic userId instead of an explicit firebaseUserId. Using any alternative auth provider could make that confusing?**

We are not going to use an alternative auth provider. Firebase is our auth provider going forward for this repo.

**If we make the firebaseUserId co-opt our user table's 'id' that's fine, but risk of collision increases as we add more external providers, each with their own id. We could make the index unique from both the id column and provider column, though. ie. we set the id as firebase's user id and provider as 'firebase' for example. But the user_journey join table will be a little harder.**

As above Firebase is our auth provider. There is no need to manage collisions here. Just replace the ID primary key with the one from Firebase.

Hope that helps.

gdub01:
Okay thanks @tataihono I understand.

I removed the firebase id and it allows for sending the firebase id when we create the user.

CONTRIBUTING.md
```diff
@@ -71,7 +71,7 @@ We recommend using [Visual Studio Code](https://code.visualstudio.com/) with the
 1. Install [Visual Studio Code](https://code.visualstudio.com/)
 1. Install [Docker Desktop](https://www.docker.com/get-started)
 1. Start Docker Desktop
-1. Docker set up:
+1. Docker set up:nx generate-graphql api-gateway
```
tataihono, line 74:
```suggestion
1. Docker set up:
```

apps/journeys-admin/pages/_app.tsx
```diff
@@ -11,6 +11,7 @@ function CustomApp({ Component, pageProps }: AppProps): ReactElement {
   const auth = getAuth(firebaseClient)
   const [user] = useAuthState(auth)
   const client = createApolloClient(user?.accessToken)
+
```
tataihono, line 14:
```suggestion
```

tataihono (changes requested):
user needs to be in a separate service not api-journeys using federation.

gdub01:
@tataihono By separate service, are you meaning you want another app created? Would it have its own db?

gdub01:
Also, are you imagining apps would hit api-gateway which would then either send requests to api-users or api-journeys?

tataihono:
@gdub01 yes. please connect with @mikeallisonJS about what that might look like. You'll need to read into federation to see how data is shared between services.

gdub01:
Hello @mikeallisonJS! I'm not sure what timezone you're in but would you be free to discuss this today or tomorrow? I'm in EST.

gdub01:
Hello @mikeallisonJS ... just wanted to follow up again. Do you have another contact I could reach out to? Can you describe what it means to create a new service and whether a new db is required?

mikeallisonJS:
Sorry @gdub01 , you're comments were lost amongst the initial github noise. I have fixed the notifications and they should work going forward. The JFP digital slack is the best place to catch me (I'm in pacific time, but schedule work just before the the NZ team hours)

Please reach out and I'll walk you through the new microservice architecture we've been working on this cycle

apps/api-journeys/prisma/factories/nua1.ts
```diff
@@ -40,6 +40,25 @@ export async function nua1(prisma: PrismaClient): Promise<void> {
...
+  })
+
+  await prisma.userJourney.create({
+    data: {
+      userId: user.id,
+      journeyId: journey.id,
+      role: 'inviteRequested'
+    }
+  })
+
```
tataihono, line 61: no seed necessary.

apps/api-journeys/prisma/schema.prisma
```diff
@@ -29,10 +45,21 @@ model Journey {
...
-  primaryImageBlock Block?    @relation("PrimaryImageJourney", fields: [primaryImageBlockId], references: [id])
+  primaryImageBlock Block?    @relation("PrimaryImageJourney", fields: [primaryImageBlockId], references: [id], onDelete: SetNull)
   slug      String    @unique
   createdAt DateTime  @default(now())   
   publishedAt DateTime?
+  UserJourney         UserJourney[] @relation("Journey")
+}
+
+model UserJourney {
+  userId    String
```
tataihono, line 56: needed for cache management

```suggestion
  id String @id
  userId    String
```

apps/api-journeys/src/modules/user/user.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  extend type UserJourney {
+    user: User
+  }
+
+  extend type Query {
+    me: User
+    users: [User!]!
+    user(id: ID!): User
```
tataihono, line 29: these endpoints are not required

apps/api-journeys/src/modules/user/user.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  extend type Mutation {
+    userCreate(input: UserCreateInput!): User!
```
tataihono, line 33: it looks like i can call this endpoint and set it to any email i want it to. I think we should rely on firebase to set this up.

apps/api-journeys/src/modules/user/user.ts
```diff
@@ -0,0 +1,100 @@
...
+      })
+      return user
+    }
+  },
+  Query: {
+    async users(_, __, { db }) {
+      return await db.user.findMany({})
+    },
+    async user(_parent, { id }, { db }) {
+      return await db.user.findUnique({ where: { id } })
```
tataihono, line 53: looks like i can get user data for any user in the system. This seems like a security hole.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,245 @@
...
+        throw new AuthenticationError('No user token provided')
+
+      return await db.userJourney.create({
+        data: {
+          userId: input.userId,
+          journeyId: input.journeyId,
+          role: input.role as UserJourneyModule.UserJourneyRole
+        }
+      })
+    },
```
tataihono, line 108: remove. this isn't being used at the moment right? Looks like user calls the userJourneyRequest.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,245 @@
...
+      return await db.userJourney.create({
+        data: {
+          userId: input.userId,
+          journeyId: input.journeyId,
+          role: 'inviteRequested'
+        }
+      })
+    },
+    // update naming
+    async userJourneyUpdate(_parent, { input }, { db, userId }) {
```
tataihono, line 122: this should be userJourneyApprove or something like that.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,245 @@
...
+
+      return await db.userJourney.create({
+        data: {
+          userId: input.userId,
+          journeyId: input.journeyId,
+          role: 'inviteRequested'
+        }
+      })
+    },
+    // update naming
```
tataihono, line 121: remove.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,245 @@
...
+      })
+
+      if (actor === null || actor?.role !== 'owner')
+        throw new AuthenticationError(
+          'You do not own this journey so you cannot remove users'
+        )
+
+      if (!includes(['inviteRequested', 'editor', null], input.role))
+        throw new AuthenticationError(
+          'Owners cannot remove themselves from their journey'
```
tataihono, line 174:
```suggestion
          ' you cannot remove yourself from this journey when you are an owner'
```

apps/journeys-admin/src/libs/firebaseClient/UseFirebase.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { useAuthState } from 'react-firebase-hooks/auth'
+
+interface UseFirebaseProps {
+  logOut: () => void
+  user
+  loading
+  error
+}
+
+export const UseFirebase = (): UseFirebaseProps => {
```
tataihono, line 13: only components should start with a capital letter. Also Props refers to the input props provided to a component. This should be renamed to something like UseFirebaseResponse.

```suggestion
export const useFirebase = (): UseFirebaseResponse => {
```

apps/journeys-admin/test/testingLibrary.tsx
```diff
@@ -0,0 +1,17 @@
+import { ReactElement } from 'react'
```
tataihono, line 1: drop. If you are using MockedProvider you don't need to wrap an apollo provider.

apps/journeys-admin/src/components/InviteUserModal/ApproveUser/ApproveUser.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+    email: 'drew@drew.com',
+    firstName: 'drew',
+    lastName: 'drew',
+    imageUrl: ''
+  }
+}
+
+describe('Approve button', () => {
+  it('should render approve button', () => {
+    const { getByText } = renderWithApolloClient(
```
tataihono, line 24: renderWithApolloClient should be just render

apps/journeys-admin/src/components/InviteUserModal/PromoteUser/PromoteUser.spec.tsx
```diff
@@ -0,0 +1,54 @@
...
+    email: 'drew@drew.com',
+    firstName: 'drew',
+    lastName: 'drew',
+    imageUrl: ''
+  }
+}
+
+describe('Promote button', () => {
+  it('should render promote button', () => {
+    const { getByText } = renderWithApolloClient(
```
tataihono, line 24: should just be render

tataihono (dismissed):

apps/api-journeys/src/modules/user/user.ts
GeronimoJohn: I've made some changes - is this what you meant

apps/api-journeys/src/modules/user/user.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      // are all of these properties required?
+      // it's making us do a type check
+      return await db.user.create({
+        data: {
+          id: userId,
+          firstName: firstName as string,
+          lastName: lastName as string,
+          email: email as string,
+          imageUrl: imageUrl as string
```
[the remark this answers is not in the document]
GeronimoJohn: I've made some changes - is this what you meant

apps/api-journeys/src/modules/user/user.ts
GeronimoJohn: I'm not too sure what to do for this one

gdub01: it now uses the email belonging to the logged-in user in firebase

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+    userJourneyApprove(input: UserJourneyUpdateInput!): UserJourney!
+    userJourneyPromote(input: UserJourneyUpdateInput!): UserJourney!
+    userJourneyRemove(input: UserJourneyRemoveInput!): UserJourney!
+    userJourneyRequest(input: UserJourneyRequestInput!): UserJourney!
+  }
+`
+
+const resolvers: UserJourneyModule.Resolvers = {
+  Journey: {
+    async usersJourneys(journey, __, { db }) {
```
tataihono, line 65:
```suggestion
    async userJourneys(journey, __, { db }) {
```

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+    journeyId: ID!
+  }
+
+  extend type Journey {
+    usersJourneys: [UserJourney!]
+  }
+
+  extend type User {
+    usersJourneys: [UserJourney!]
+  }
```
tataihono, line 46: i don't think this is necessary. This will allow you to navigate through other users to find out what journeys they have access to.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+    role: UserJourneyRoleForUpdates!
+  }
+
+  input UserJourneyRequestInput {
+    userId: ID!
+    journeyId: ID!
+  }
+
+  extend type Journey {
+    usersJourneys: [UserJourney!]
```
tataihono, line 41:
```suggestion
    userJourneys: [UserJourney!]
```

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+  input UserJourneyUpdateInput {
+    userId: ID!
+    journeyId: ID!
+  }
+
+  input UserJourneyRemoveInput {
+    userId: ID!
+    journeyId: ID!
+    role: UserJourneyRoleForUpdates!
+  }
```
tataihono, line 33: drop

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+    id: ID!
+    userId: ID!
+    journeyId: ID!
+    role: UserJourneyRole!
+  }
+
+  extend type Mutation {
+    userJourneyApprove(input: UserJourneyUpdateInput!): UserJourney!
+    userJourneyPromote(input: UserJourneyUpdateInput!): UserJourney!
+    userJourneyRemove(input: UserJourneyRemoveInput!): UserJourney!
```
tataihono, line 58:
```suggestion
    userJourneyApprove(id: ID!): UserJourney!
    userJourneyPromote(id: ID!): UserJourney!
    userJourneyRemove(id: ID!): UserJourney!
```

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  input UserJourneyRemoveInput {
+    userId: ID!
+    journeyId: ID!
+    role: UserJourneyRoleForUpdates!
+  }
+
+  input UserJourneyRequestInput {
+    userId: ID!
```
tataihono, line 36: don't need userId. You should use the id of the authenticated user.

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+  },
+  Mutation: {
+    async userJourneyRequest(_parent, { input }, { db, userId }) {
+      if (userId == null)
+        throw new AuthenticationError('No user token provided')
+
+      return await db.userJourney.create({
+        data: {
+          userId: input.userId,
```
tataihono, line 96:
```suggestion
          userId,
```

tataihono (dismissed):

apps/api-journeys/src/modules/userJourney/userJourney.ts
```diff
@@ -0,0 +1,190 @@
...
+    userId: ID!
+    journeyId: ID!
+    role: UserJourneyRole!
+  }
+
+  extend type Mutation {
+    userJourneyApprove(id: ID!): UserJourney!
+    userJourneyPromote(id: ID!): UserJourney!
+    userJourneyRemove(id: ID!): UserJourney!
+    userJourneyRequest(input: UserJourneyRequestInput!): UserJourney!
```
mikeallisonJS, line 32: Since the input is not being used to modify data, can this be a parameter journeyId: ID! ?

mikeallisonJS (changes requested):

tataihono (approved):

PR #131, opened 2021-11-22
use block fields

tataihono:
queries shouldn't exist in libs. Rather fragments should. This fixes that.


storyworks (approved):

storyworks (commented):
Oh oops, approved once tests are fixed.

storyworks (approved):

PR #132, opened 2021-11-22
C4 - Fetch all journeys, add status prop to Journey

storyworks:


apps/journeys/pages/index.tsx
```diff
@@ -37,14 +37,17 @@ export const getServerSideProps: GetServerSideProps<JourneysPageProps> =
   async () => {
     const { data } = await client.query<GetJourneys>({
       query: gql`
-        query GetJourneys {
-          journeys {
+        query GetJourneys($status: JourneyStatus) {
+          journeys(status: $status) {
```
tataihono, line 41:
```suggestion
          journeys(status: "published") {
```

apps/journeys/pages/index.tsx
```diff
@@ -37,14 +37,17 @@ export const getServerSideProps: GetServerSideProps<JourneysPageProps> =
   async () => {
     const { data } = await client.query<GetJourneys>({
       query: gql`
-        query GetJourneys {
-          journeys {
+        query GetJourneys($status: JourneyStatus) {
```
tataihono, line 40:
```suggestion
        query GetJourneys {
```

tataihono (changes requested):

apps/journeys/pages/index.tsx
storyworks: Suggested format didn't work when I tried it in tests so I didn't write it like that here either.

csiyang (approved):

tataihono (approved):

PR #135, opened 2021-11-24
JourneyList - Remove blocks folder, fix storybook naming

storyworks:
Requested by @tataihono


csiyang (commented):
What do you think about changing the file structure in Storybook to match what's in the project?

eg:  JourneyCard not nested inside JourneyList

I feel like it would make it easier to find components esp when we have more, and it would match the structure from the other projects.

storyworks:
@csiyang Agreed the project structure should match storybook. I was playing around wondering if we could reuse JourneyCard in the SingleJourneyPage but decided against it because designs aren't finalised. I forgot to move it back into JourneyList - have updated now :)

csiyang (approved):

PR #137, opened 2021-11-24
Change useBreakpoint to use .up method over .only

storyworks:
.up() is considerably more useful than .only().

Consider the case where we want a component to render only on tablet and above - this can be done easily if useBreakpoint uses the .up() method. And this would be the most common use case.

Any checks already styling a component based on the .only() method can also be easily rewritten using only .up() checks. See the Conductor file for an example of refactoring.

### DIFF
--- a/libs/shared/ui/src/libs/themes/base/theme.stories.tsx
+++ b/libs/shared/ui/src/libs/themes/base/theme.stories.tsx
@@ -286,15 +286,28 @@ const ViewportTemplate: Story<ThemeStoryProps> = (args) => {
         display: 'flex',
         flexDirection: 'column',
         justifyContent: 'center',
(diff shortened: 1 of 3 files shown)


libs/shared/ui/src/libs/themes/base/theme.stories.tsx
```diff
@@ -286,41 +286,66 @@ const ViewportTemplate: Story<ThemeStoryProps> = (args) => {
         display: 'flex',
```
storyworks, line 286: Updated the story to still test the breakpoint helpers (theme.breakpoints / useBreakpoints) but is less wordy onscreen

image

tataihono (approved):

PR #138, opened 2021-11-25
C4/ Make JourneyStatus required on Journeys

csiyang:
Updated status to be non-nullable


storyworks (approved):

storyworks (dismissed):
Don't add status to prisma.schema

apps/api-journeys/src/modules/journey/journey.ts
```diff
@@ -85,7 +85,6 @@ const typeDefs = gql`
     journey: Journey
   }
 `
-
```
tataihono, line 88: readd to keep spacing

tataihono (changes requested):

apps/api-journeys/src/modules/journey/journey.ts
csiyang: added spacing

storyworks (approved):

PR #140, opened 2021-11-25
Sort Journeys

storyworks:


apps/journeys-admin/src/components/JourneyList/JourneyList.tsx
```diff
@@ -1,68 +1,80 @@
-import { ReactElement } from 'react'
+import { ReactElement, useState } from 'react'
 import { Box, Card, Typography, Button } from '@mui/material'
 import AddIcon from '@mui/icons-material/Add'
 import { useBreakpoints } from '@core/shared/ui'
 import { GetJourneys_journeys as Journey } from '../../../__generated__/GetJourneys'
+import { JourneySort, SortBy } from './JourneySort'
 import { JourneyCard } from './JourneyCard'
+import JourneysAppBar from '../JourneysAppBar'
```
tataihono, line 8: didn't the import change on this?

apps/journeys-admin/src/components/JourneyList/JourneyList.tsx
```diff
@@ -1,68 +1,80 @@
...
+  const [sortBy, setSortBy] = useState(SortBy.UNDEFINED)
+
+  if (sortBy === SortBy.TITLE) {
+    journeys.sort((a, b) => {
+      return a.title.localeCompare(b.title)
+    })
+  } else {
+    journeys.sort((a, b) => {
+      return new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+    })
```
tataihono, line 25: this logic could be moved out into a function and tested?

apps/journeys-admin/src/components/JourneyList/JourneySort/JourneySort.tsx
```diff
@@ -0,0 +1,144 @@
...
+            flexDirection: 'row',
+            justifyContent: 'flex-end'
+          }}
+        >
+          <Button
+            sx={{ mt: 1, mr: 1 }}
+            onClick={() => setShowSortBy(false)}
+            variant="text"
+          >
+            Cancel
```
tataihono, line 105: this button bottom margin on mine looks off

apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCard.tsx
```diff
@@ -21,10 +22,30 @@ export function JourneyCard({ journey }: JourneyCardProps): ReactElement {
...
       variant="outlined"
-      sx={{ borderRadius: 0, px: 6, py: 4, border: 'none' }}
+      sx={{
+        px: 6,
+        py: 4,
+        borderRadius: 0,
+        borderColor: 'divider',
+        borderBottom: 'none',
+        '&:first-child': {
+          borderTopLeftRadius: border,
```
tataihono, line 39: I'm not a fan of using breakpoints for styling. We should reserve useBreakpoints where codeflow is affected by responsiveness. You should use the expressive syntax inline that mui provides for this.

```suggestion
          borderTopLeftRadius: { xs: 0, sm: 12 },
```

apps/journeys-admin/src/components/JourneyList/JourneyList.tsx
tataihono, line 8: didn't i see somewhere where this changed from a default export?

tataihono (changes requested):

apps/journeys-admin/src/components/JourneyList/JourneySort/JourneySort.tsx
storyworks: It is off because it uses Button medium and that's something that comes from theming which I don't want to touch yet because it'll affect your buttons too..

apps/journeys-admin/src/components/JourneyList/JourneyList.tsx
storyworks: This is unit tested in JourneyList at the moment, didn't extract it out yet since it was still small..

tataihono (approved):

PR #141, opened 2021-11-25
added LinkAction to handleAction

csiyang:
Added link action to shared lib handleAction for navigating to external URL's


libs/journeys/ui/src/libs/action/handleAction.ts
```diff
@@ -19,5 +19,8 @@ export function handleAction(
     case 'NavigateAction':
       nextActiveBlock()
       break
+    case 'LinkAction':
+      void router.push(action.url)
```
tataihono, line 23:
```suggestion
      window.location.href = action.url
```

csiyang:
PR outside of scope for current cycle

PR #145, opened 2021-11-29
Locale helpers

storyworks:
So we can created the following:
image
image
image

getLocaleLabel formatting (used in Autocomplete) keeps changing in designs - will keep current format for now and update when final design is out..


libs/journeys/ui/src/libs/locales/locales.ts
```diff
@@ -0,0 +1,36 @@
...
+    locale: 'en-NZ',
+    language: 'English',
+    region: 'New Zealand'
+  }
+}
+
+export const getLocaleOptions = (): SelectOption[] =>
+  Object.keys(locales).map((locale) => getLocaleOption(locale))
+
+export const getLocaleOption = (locale: string): SelectOption => {
```
storyworks, line 28: Used when getting default value of autocomplete from locale string

gdub01:
Awesome =) 

How much do we want to integrate with arclight here or cldr? 

Arclight supports I think like 1800 bcp-47 codes which is more than most formats. It has language names and alternative language names updated by a couple sources I believe.

https://cldr.unicode.org supports localization.

Or are we thinking we're going to hand-code just a select few languages ourselves and not worry about localization?

gdub01 (approved):

csiyang:
Locale data from [here](https://github.com/mozilla/language-mapping-list/blob/29385e7627cfa8fb7a9cd19c1405a292b7d177e6/language-mapping-list.js)

jianwei1 (approved):

PR #156, opened 2021-12-06
card preview

tataihono:
- add CardPreview
- allow parent to control selected
- when selection changes trigger event


storyworks (approved):
Possibly might be worth having a preview-only story (possible spacing differences). But it's minor - we'll add this in and update if need be.

PR #163, opened 2021-12-08
c4/ admin theme typography update

csiyang:
- updated styling for existing variants
- added new variants (label, buttom-sm, button-md, button-lg)


apps/journeys-admin/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,14 +1,15 @@
 import { CssBaseline, ThemeProvider as MuiThemeProvider } from '@mui/material'
 import { ReactElement, ReactNode } from 'react'
-import { theme } from './theme'
+import { adminLight } from './admin/theme'
 
 interface ThemeProviderProps {
   children?: ReactNode
 }
 
+// TODO: Cooldown Use shared/ui ThemeProvider?
```
tataihono, line 9: I am not a fan of committing these kinds of comments in code. We should trake them in basecamp.

tataihono (changes requested):
remove comment. Looks good otherwise.

tataihono (approved):

PR #172, opened 2021-12-17
updated some styling for the app bar

jianwei1:


csiyang (approved):

csiyang:
@storyworks changes look good, but I think we need these in both the single journey and journey list

apps/journeys-admin/pages/journeys/[journeySlug].tsx
```diff
@@ -43,17 +52,18 @@ export const getServerSideProps: GetServerSideProps<SingleJourneyPageProps> =
   async (context) => {
     const { data } = await client.query<GetJourney>({
       query: gql`
+        ${BLOCK_FIELDS}
```
Comment (storyworks) [inline apps/journeys-admin/pages/journeys/[journeySlug].tsx:55]: Remove these updates here as well as anything associated with CardOverview  / will cause more conflict errors when we merge feature branches...

Comment (storyworks) [reply apps/journeys-admin/pages/journeys/[journeySlug].tsx:55]: Same on the JourneyList page

apps/journeys-admin/src/components/JourneyList/journeyListData.ts
```diff
@@ -9,5 +13,8 @@ export const defaultJourney: Journey = {
   themeName: ThemeName.base,
   themeMode: ThemeMode.light,
   slug: 'default',
-  locale: 'en_US'
+  locale: 'en_US',
```
storyworks, line 16: This will all be updated when merging feature branches, let's not make that harder for ourselves..

apps/journeys-admin/src/components/SingleJourney/SingleJourneyData.ts
```diff
@@ -0,0 +1,14 @@
+import { GetJourney_journey as Journey } from '../../../__generated__/GetJourney'
```
storyworks, line 1: Same here, don't need this file just yet..

storyworks (changes requested):

storyworks (approved):

PR #175, opened 2021-12-21
Fix flakey UI tests

GeronimoJohn:
- fix the chromatic tests problems that we encounter with video


GeronimoJohn:
_**Leaving this comment here for documentation purposes and to remember what happened in this PR**_

After running different configurations to see what will help fix the flakey UI tests in chromatic. This is what I've found. 

The affected UI tests most of the time, at least from this PR, are Video, Card, Editor, and Control Panel components. 

Setting a delay for the Video component works wonders in terms of fixing the problem with the UI tests. But for the rest of the components, I had to play around with different thresholds to see which one would help solve our problem. 

The default threshold for chromatic is at 0.63 where 0 is the most accurate and 1 is the least accurate. Since the changes that get checked in the UI tests for our specific components are minor. I've decided to put up the threshold to 0.75 except for the Video in the Card component, which is set to 0.85. Doing so seems to fix the flakey UI tests.

apps/api-journeys/prisma/migrations/20211223205429_nx_prisma/migration.sql
```diff
@@ -0,0 +1,17 @@
+-- DropForeignKey
```
GeronimoJohn, line 1: Did a migration in this branch and this file was created. Can drop if needed.

storyworks: Drop this migration for now? And merge in latest main (I've merged in the JourneyViewPage). 

Also nice work with the PR write up! Was really helpful in remembering the context for this PR :) 

If we find that more and more components become flakey, that could be an indication of us building things kinda funky or we could think about updating the chromatic settings globally..

storyworks (changes requested):

storyworks (approved):

PR #177, opened 2021-12-23
Remove duplicated components

GeronimoJohn:
- [Remove duplicated components](https://3.basecamp.com/3105655/buckets/24946810/todos/4379121193)


GeronimoJohn:
Due to some recent findings - we won't be needing to remove the duplicated components.

PR #179, opened 2022-01-09
Run VR test only if lint, build, test passing

storyworks:
To reduce snapshots taken so we don't exceed our capacity like last time

Downside is that it does take a bit longer now for all test to pass if we are running VR

image


apps/journeys-admin/src/components/JourneyView/Properties/Properties.tsx
```diff
@@ -16,7 +16,7 @@ export function Properties(): ReactElement {
           Properties
         </Typography>
       </Toolbar>
-      <Box sx={{ px: 6, py: 9 }}>
```
tataihono, line 19: err. these change should be removed

tataihono (changes requested):

tataihono (approved):

PR #184, opened 2022-01-11
speed up journey load time

tataihono:
- index should load props at build time
- journeySlug should load props at build time


GeronimoJohn (approved):
Looks good!

PR #186, opened 2022-01-11
fix UserJourney user object

mikeallisonJS:
Fix the graphql api schema to connect the services


apps/api-users/src/app/modules/user/user.service.spec.ts
```diff
@@ -61,6 +61,18 @@ describe('UserService', () => {
     })
   })
 
+  describe('getByUserId', () => {
+    beforeEach(() => {
+      ;(service.db as DeepMockProxy<Database>).query.mockReturnValue(
```
tataihono, line 66: does it work without the ;?
```suggestion
      (service.db as DeepMockProxy<Database>).query.mockReturnValue(
```

tataihono (changes requested):

apps/api-users/src/app/modules/user/user.service.spec.ts
mikeallisonJS: The semi is created by prettier, won't pass without it

tataihono (approved):

PR #193, opened 2022-01-13
Attributes for typography block

csiyang:


apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/Attributes.tsx
```diff
@@ -50,7 +53,7 @@ export function Attributes({ selected }: AttributesProps): ReactElement {
           borderTop: (theme) => `1px solid ${theme.palette.divider}`
         }}
       >
-        <Typography align="center">Editing Card Properties</Typography>
+        <MuiTypography align="center">Editing Card Properties</MuiTypography>
```
storyworks, line 56: Add ability to switch this label based on block selected

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { render } from '@testing-library/react'
+import { GetJourneyForEdit_journey_blocks_TypographyBlock as TypographyBlock } from '../../../../../../../__generated__/GetJourneyForEdit'
+import {
+  TypographyAlign,
+  TypographyColor,
+  TypographyVariant
+} from '../../../../../../../__generated__/globalTypes'
+import { Typography } from '.'
+
+describe('Typography', () => {
```
storyworks, line 11: I know this label is just copying what's there but maybe in the future we can rename it "Typography Properties" or something like that so test analysis is clearer..

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
```diff
@@ -0,0 +1,62 @@
...
+        value={_.capitalize(color?.toString()) ?? 'None'}
+        description="Text Color"
+        // onClick open drawer
+      />
+      <Attribute
+        id={`${id}-font-variant`}
+        icon={<TextFieldsRoundedIcon />}
+        name="Font Variant"
+        value={
+          _.upperFirst(
```
storyworks, line 45: Named imports for lodash

storyworks (changes requested):

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
```diff
@@ -0,0 +1,62 @@
...
+        description="Text Color"
+        // onClick open drawer
+      />
+      <Attribute
+        id={`${id}-font-variant`}
+        icon={<TextFieldsRoundedIcon />}
+        name="Font Variant"
+        value={
+          captialize(lowerCase(variant?.toString()).replace('h', 'header')) ??
+          'None'
```
storyworks, line 47:
```suggestion
          'None'
```
Replace this with the actual default variant - find it in the add blocks pitch

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
```diff
@@ -0,0 +1,62 @@
...
+        }
+        description="Font Variant"
+        // onClick open drawer
+      />
+      <Attribute
+        id={`${id}-text-alignment`}
+        icon={<FormatAlignLeftRoundedIcon />}
+        name="Text Alignment"
+        value={captialize(align?.toString()) ?? 'None'}
```
storyworks, line 56:
```suggestion
        value={captialize(align?.toString()) ?? 'None'}
```
Same here and for color - replace None with actual default values.

storyworks (changes requested):

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
storyworks: https://3.basecamp.com/3105655/buckets/23266566/messages/4514250301

csiyang: Doesn't that mean it can technically display a value but its actually null? If attributes are set correctly onCreate, the 'None' should never appear.

csiyang: Screen Shot 2022-01-14 at 9 28 29 AM

The text alignment is what it looks like when I remove the `?? 'None'` part of it. Thoughts on just removing the null check?

storyworks: Not too sure what you mean but if the property is returning null, it means that the journey builder hasn't set it themselves and we're using the default property that's based on the theme / hard coded on the component. 

I thought onCreate would just set the required props leaving the optional ones to default. So if we get that null, we have to match it to what it displays by default..

csiyang: As discussed will change these to display default values

storyworks (dismissed):

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
```diff
@@ -20,12 +20,12 @@ export function Typography({
...
         id={`${id}-text-color`}
         icon={
-          <Paper sx={{ borderRadius: 1000, overflow: 'hidden' }}>
+          <Paper sx={{ borderRadius: 1000 }}>
             <Box
               data-testid="backgroundColorIcon"
               sx={{
-                width: 25,
-                height: 25,
+                width: 20,
```
storyworks, line 27: If you decrease the icon size here you'd need to do it for Card as well? Are you able to find out from the designs how they want to handle what to do when the text overflows? 
And stick to one consistent size for the attribute icons (I think Vlad mentioned it in a design comment for Ariel)

apps/journeys-admin/src/components/Editor/ControlPanel/Button/Button.tsx
```diff
@@ -31,7 +31,7 @@ export function Button({
   return (
     <Box
       sx={{
-        maxWidth: 160
+        maxWidth: 150
```
storyworks, line 34: Nice that this was reverted back to match the rest

storyworks (dismissed):

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Typography/Typography.tsx
csiyang: Wasn’t sure if I should do it in this PR?

storyworks: I thought your fix for overflowing text on the attributes was to revert the width of the box back to before (which you've done), keeping the icon consistent size with the others, but updating the css that affects text display so it uses ellipsis ?

storyworks: I thought you said you'd fix the overflowing text here and that the was to revert the width of the box back to before (which you've done), keeping the icon consistent size with the others, but updating the css that affects text display so it uses ellipsis (assuming that's what they want) ?

csiyang: I add back the overflow hidden in the button component. This fixes both the color icon clipping as well as preserving the right margin.

storyworks (approved):

PR #195, opened 2022-01-13
Blocks tab

jianwei1:


tataihono (changes requested):
apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/NewBlocks

should be moved to apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab

NewBlocks should be renamed to BlocksTab also. In future we will change Attributes to PropertiesTab to make it clearer to the developer.

apps/journeys-admin/src/components/Editor/Add/Add.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactElement, useContext } from 'react'
```
tataihono, line 1: remove this file

tataihono (changes requested):
just one change.

tataihono (approved):

PR #202, opened 2022-01-14
Fix Journey data id references

storyworks:
Would be nice to keep the mock data for these card previews in a central place...

Also this just fixes the parentBlockId references for block selection to work. It'd be nice to have the consolidated format for id's / id naming fixes done once in a shared data file..

### DIFF
--- a/apps/journeys-admin/src/components/CardPreview/CardPreview.stories.tsx
+++ b/apps/journeys-admin/src/components/CardPreview/CardPreview.stories.tsx
@@ -106,10 +106,10 @@ const steps: Array<TreeBlock<StepBlock>> = [
     nextBlockId: 'step2.id',
     children: [
       {
-        id: 'card0.id',
+        id: 'card1.id',
         __typename: 'CardBlock',
         parentBlockId: 'step1.id',
-        coverBlockId: 'image0.id',
+        coverBlockId: 'image1.id',
         backgroundColor: null,
         themeMode: null,
         themeName: null,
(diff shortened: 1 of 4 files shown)


apps/journeys-admin/src/components/CardPreview/CardPreview.stories.tsx
```diff
@@ -191,28 +191,18 @@ const steps: Array<TreeBlock<StepBlock>> = [
...
-            variant: TypographyVariant.h3,
-            children: []
-          },
           {
             id: 'radioQuestion1.id',
             __typename: 'RadioQuestionBlock',
-            parentBlockId: 'step2.id',
-            label: '',
+            parentBlockId: 'card2.id',
+            label: 'What is Christianity to you?',
```
storyworks, line 205: Is it too late to rename this to heading? Seems like it's not even clear to devs that this should be used to create the heading rather than adding a seperate typography block

csiyang (commented):
Couldn't add the comments in line (I think it's because you didnt change those lines), but saw a few things :
- for CardPreview, Canvas and CardView, the radioOption skips id 2
- button id and typography id naming for all files is of a different format to all the other id's

Otherwise, looks algood.

storyworks:
> Couldn't add the comments in line (I think it's because you didnt change those lines), but saw a few things :
> 
> * for CardPreview, Canvas and CardView, the radioOption skips id 2
> * button id and typography id naming for all files is of a different format to all the other id's
> 
> Otherwise, looks algood.

Ok, can those other ones be a cooldown thing? This one I intended just to fix the references (the each block points to the correct parent since it affects selection)

csiyang:
> > Couldn't add the comments in line (I think it's because you didnt change those lines), but saw a few things :
> > 
> > * for CardPreview, Canvas and CardView, the radioOption skips id 2
> > * button id and typography id naming for all files is of a different format to all the other id's
> > 
> > Otherwise, looks algood.
> 
> Ok, can those other ones be a cooldown thing? This one I intended just to fix the references (the each block points to the correct parent since it affects selection)

Sounds good

csiyang (approved):

PR #203, opened 2022-01-17
Select Step block

storyworks:
Change back to selecting step, we shouldn't ever set Card as activeBlock

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
@@ -61,5 +61,9 @@ describe('Canvas', () => {
     expect(getByTestId('step-step0.id')).toHaveStyle(
       'border: 2px solid #c52d3a'
     )
+    fireEvent.click(getByTestId('step-step1.id'))
+    expect(getByTestId('step-step0.id')).toHaveStyle(
+      'border: 2px solid #efefef'
+    )
   })
 })
--- a/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
@@ -12,19 +12,7 @@ describe('Canvas', () => {
       parentBlockId: null,
       locked: false,
       nextBlockId: null,
-      children: [
-        {
-          id: 'card0.id',
-          __typename: 'CardBlock',
-          parentBlockId: 'step0.id',
-          coverBlockId: null,
-          backgroundColor: null,
-          themeMode: null,
-          themeName: null,
-          fullscreen: false,
-          children: []
-        }
-      ]
+      children: []
     }
     const step1: TreeBlock<StepBlock> = {
       id: 'step1.id',
@@ -32,19 +20,7 @@ describe('Canvas', () => {
       parentBlockId: null,
       locked: false,
       nextBlockId: null,
-      children: [
-        {
-          id: 'card1.id',
-          __typename: 'CardBlock',
-          parentBlockId: 'step1.id',
-          coverBlockId: null,
-          backgroundColor: null,
-          themeMode: null,
-          themeName: null,
-          fullscreen: false,
-          children: []
-        }
-      ]
+      children: []
     }
     const { getByTestId } = render(
       <ThemeProvider>
--- a/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
@@ -12,7 +12,19 @@ describe('Canvas', () => {
       parentBlockId: null,
       locked: false,
       nextBlockId: null,
-      children: []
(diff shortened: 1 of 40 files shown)


apps/journeys-admin/src/components/Editor/Canvas/Canvas.tsx
```diff
@@ -105,7 +112,8 @@ export function Canvas(): ReactElement {
                   zIndex: 1,
                   transition: '0.2s opacity ease-out 0.1s',
                   backgroundColor: (theme) => theme.palette.background.default,
-                  opacity: step.id === selectedStep?.id ? 0 : 1
+                  opacity: step.id === selectedStep?.id ? 0 : 1,
+                  pointerEvents: 'none'
```
storyworks, line 116: This allows us to click into each card instead of having the drag card interaction.

apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
```diff
@@ -37,9 +61,5 @@ describe('Canvas', () => {
     expect(getByTestId('step-step0.id')).toHaveStyle(
       'border: 2px solid #c52d3a'
     )
-    fireEvent.click(getByTestId('step-step1.id'))
```
storyworks, line 40: When I removed the 'shortSwipes' prop on the Swiper, we could no longer click onto the next card to switch to it. I tested using the `slideToClickedSlide` but the event doesn't trigger..

apps/journeys-admin/src/components/Editor/Canvas/Canvas.tsx
```diff
@@ -86,14 +89,19 @@ export function Canvas(): ReactElement {
                 position: 'relative',
                 overflow: 'hidden',
                 border: (theme) =>
-                  step.id === selectedStep?.id
+                  step.children[0].id === selectedBlock?.id
```
storyworks, line 92: Design removes selection around card when selectedBlock changes

apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.tsx
```diff
@@ -101,7 +101,15 @@ export function ControlPanel(): ReactElement {
         />
       </TabPanel>
       <TabPanel value={activeTab} index={1}>
-        {selectedBlock != null && <Attributes selected={selectedBlock} />}
+        {selectedBlock !== undefined && selectedStep !== undefined && (
+          <Attributes
+            selected={
+              selectedBlock.__typename === 'CardBlock'
```
storyworks, line 107: Use the Step attributes on Card as well

apps/journeys-admin/src/components/Editor/Canvas/Canvas.spec.tsx
```diff
@@ -61,9 +61,5 @@ describe('Canvas', () => {
...
```
storyworks, line 40: Even waitFor doesn't produce a success, works in storybook.

apps/journeys-admin/src/components/Editor/Canvas/Canvas.tsx
```diff
@@ -105,7 +112,8 @@ export function Canvas(): ReactElement {
                   zIndex: 1,
                   transition: '0.2s opacity ease-out 0.1s',
                   backgroundColor: (theme) => theme.palette.background.default,
-                  opacity: step.id === selectedStep?.id ? 0 : 1
+                  opacity: step.id === selectedStep?.id ? 0 : 1,
+                  pointerEvents: step.id === selectedStep?.id ? 'none' : 'auto'
```
storyworks, line 116: Allows us to click on elements in active card instead of dragging/swiping

libs/journeys/ui/src/libs/context/EditorContext.tsx
```diff
@@ -113,7 +117,7 @@ export function EditorProvider({
   const [state, dispatch] = useReducer(reducer, {
     steps: [],
     selectedStep: initialState?.steps?.[0],
-    selectedBlock: initialState?.steps?.[0],
+    selectedBlock: initialState?.steps?.[0].children[0],
```
mikeallisonJS, line 120: @tataihono this all looks good to me. I just want to be sure selecting the card block initially is what we are expecting

tataihono: no. should select step.

apps/journeys-admin/src/components/Editor/Canvas/Canvas.tsx
```diff
@@ -63,11 +63,19 @@ export function Canvas(): ReactElement {
           justifyContent: 'center'
         }
       }}
+      onClick={() => {
+        dispatch({
+          type: 'SetSelectedBlockAction',
+          block: selectedStep
```
storyworks, line 69: Moved this here so clicking outside of the Card would also reset the selectedBlock to the step.

selectedStep is set onSlideChange.

mikeallisonJS (approved):

PR #204, opened 2022-01-17
Select Typography, Button and RadioQuestion

storyworks:


libs/journeys/ui/src/components/Button/Button.tsx
```diff
@@ -44,7 +67,11 @@ export function Button({
           />
         )
       }
-      onClick={handleClick}
+      sx={{
+        outline: selectedBlock?.id === props.id ? '3px solid #C52D3A' : 'none',
```
storyworks, line 71: Hard-coding color for now since admin theme isn't accessible in journey-ui library. Can move over theme in seperate PR and fix..

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.stories.tsx
```diff
@@ -112,8 +112,6 @@ export const Default: Story<TreeBlock<RadioQuestionFields>> =
 Default.args = {
   id: 'RadioQuestion1',
   label: 'How can we help you know more about Jesus?',
-  description:
```
storyworks, line 115: default stories should pass only data that is required

mikeallisonJS (approved):

mikeallisonJS (commented):
Failing tests now

mikeallisonJS (approved):
Looks good

PR #205, opened 2022-01-17
fix card heights

tataihono:
- add Div100vh (this uses JS to determine the height. This is a `feature` of safari)
- use new height calculations in the conductor
- fix border-radius glitch on safari


GeronimoJohn (dismissed):
Looks good!

GeronimoJohn (approved):

PR #206, opened 2022-01-17
Editor attributes (Button, RadioQuestion, RadioOption, SignUp)

csiyang:
Editor attributes for 
- Button
- RadioOption
- SignUp


apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+  buttonColor,
+  size,
+  startIcon,
+  endIcon,
+  action
+}: TreeBlock<ButtonBlock>): ReactElement {
+  return (
+    <>
+      <Attribute
+        id={`${id}-link-action`}
```
storyworks, line 24:
```suggestion
        id={`${id}-button-action`}
```
Add `button` to all these id's and change id-font-variant in Typography.ts

storyworks (changes requested):

storyworks (approved):

PR #211, opened 2022-01-18
Switch to properties tab on select

storyworks:
And don't persist last selected attribute.

Manually tested but we really need e2e tests set up.. Unit tests don't render items within the Canvas iframe in `Editor.spec.ts`

Confirming with Vlad if we still switch to properties if the same block is clicked [He said he wanted us to switch to properties each time the block is click, even if it's already selected.]

 Reproduce by clicking a block > Navigating to Cards tab > click on already selected block > takes you to properties


apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
```diff
@@ -28,7 +28,7 @@ export function CardPreview({
         <Box
           id={step.id}
           key={step.id}
-          data-testid={`step-${step.id}`}
+          data-testid={`preview-${step.id}`}
```
storyworks, line 31: fix duplicate id from Canvas

libs/journeys/ui/src/components/Video/Video.tsx
```diff
@@ -150,7 +150,7 @@ export function Video({
   const handleSelectBlock = (e: MouseEvent<HTMLElement>): void => {
     e.stopPropagation()
 
-    if (selectedBlock?.id !== blockId) {
+    if (props.parentBlockId !== selectedStep?.id) {
```
storyworks, line 153: Only select VideoBlock if it's not a video only step.

mikeallisonJS (approved):

PR #219, opened 2022-01-21
add typography component

tataihono:


apps/api-journeys/src/app/modules/block/typography/typography.resolvers.ts
```diff
@@ -0,0 +1,43 @@
...
+      UserJourneyRole.owner,
+      UserJourneyRole.editor
+    ])
+  )
+  @IdAsKey()
+  async typographyBlockCreate(
+    @Args('input') input: TypographyBlockCreateInput & { __typename }
+  ): Promise<TypographyBlock> {
+    input.__typename = 'TypographyBlock'
+    return await this.blockService.save(input)
```
mikeallisonJS, line 28: This will need the parent order code from PR #225 to work properly once that goes in

apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab/Typography/Typography.tsx
```diff
@@ -0,0 +1,90 @@
...
+    state: { selectedStep },
+    dispatch
+  } = useEditor()
+
+  const handleClick = async (): Promise<void> => {
+    const card = selectedStep?.children.find(
+      (block) => block.__typename === 'CardBlock'
+    ) as TreeBlock<CardBlock> | undefined
+    const checkTypography = card?.children.map((block) =>
+      block.children.find((child) => child.__typename === 'TypographyBlock')
```
mikeallisonJS, line 41: Does this find any typography block? Can there be multiples?

mikeallisonJS (changes requested):

apps/api-journeys/src/app/modules/block/typography/typography.resolvers.ts
GeronimoJohn: Agree, just waiting for it

apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab/Typography/Typography.tsx
GeronimoJohn: Yes, there can be multiple Typography blocks, and had the logic to be looking for all of them. But we later found that all we really need is to find at least one Typography block. And even if one exists we set the variant to be body2 

[Requirement:](https://docs.google.com/document/d/15rY7PsOPmZU6wMcii-tP4bt8bOrQYxdsvulPF5-MuFE/edit#)
_If this is a first text block on the card: H1 in all other 6cases Body 2_

apps/api-journeys/src/app/modules/block/typography/typography.resolvers.ts
storyworks: I can update that on my PR since I'm still waiting on Vlad to review the PR #225 is dependant on.

storyworks (approved):
I'm approving this assuming Mike was ok with everything else and it was just #225 holding it up.

PR #229, opened 2022-01-26
Fix missing syntax in getSiblings

GeronimoJohn:
- Fix the missing syntax in the getSiblings function

### DIFF
--- a/apps/api-journeys/src/app/modules/block/block.service.ts
+++ b/apps/api-journeys/src/app/modules/block/block.service.ts
@@ -9,14 +9,10 @@ import { Block, Journey } from '../../__generated__/graphql'
 export class BlockService extends BaseService {
   async forJourney(journey: Journey): Promise<Block[]> {
     const primaryImageBlockId = journey.primaryImageBlock?.id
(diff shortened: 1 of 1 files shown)


tataihono:
@storyworks can you review this?

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -33,7 +33,7 @@ export class BlockService extends BaseService {
     const res = await this.db.query(aql`
       FOR block in ${this.collection}
         FILTER block.journeyId == ${journeyId}
-          ${filterParentBlockId}
+          AND ${filterParentBlockId}
```
storyworks, line 36: Line 32 in this file already includes the AND - does that break if there is a parentBlockId?

GeronimoJohn: I think line 32 is fine? AQL gave a syntax error when I tried to add a new typography block through our UI. It first gave a err 500 message. Then the AQL syntax error/warning on the terminal.

storyworks (approved):

PR #247, opened 2022-02-01
Add Move buttons

storyworks:


apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  const firstBlockIndex = coverBlock != null ? 1 : 0
+  const lastBlockIndex =
+    parentBlock != null ? parentBlock.children.length - 1 : 0
+
+  return (
+    <Box>
+      <ButtonGroup
+        data-testid="move-block-buttons"
```
tataihono, line 112: id not used anywhere

```suggestion
```

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+            parentOrder: 0
+          }
+        })
+      }
+    }
+
+    void fixCoverOrder()
+  }, [coverBlock, journey, blockOrderUpdate])
+
+  const handleMove = async (move: 'up' | 'down'): Promise<void> => {
```
tataihono, line 91: could you just provide a function that returns a function?
```suggestion
  const handleMove = async (move: 'up' | 'down'): () => Promise<void> => {
```

would simplify button logic below
```
          onClick={handleMove('up')}
```

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+        variables: {
+          id: selectedBlock.id,
+          journeyId: journey.id,
+          parentOrder: selectedBlock.parentOrder + moveBy
+        }
+      })
+    }
+  }
+
+  const firstBlockIndex = coverBlock != null ? 1 : 0
```
tataihono, line 105: i think one of the options in this ternary is not tested.

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const parentBlock =
+    selectedBlock.parentBlockId != null
+      ? searchBlocks(selectedStep.children, selectedBlock.parentBlockId)
+      : selectedStep
+
+  const coverBlock =
+    parentBlock != null &&
+    parentBlock.__typename === 'CardBlock' &&
+    parentBlock.coverBlockId != null
+      ? searchBlocks(parentBlock.children, parentBlock.coverBlockId)
```
tataihono, line 72: this line is not tested.

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+      : undefined
+
+  useEffect(() => {
+    const fixCoverOrder = async (): Promise<void> => {
+      if (coverBlock != null && coverBlock.parentOrder !== 0) {
+        await blockOrderUpdate({
```
tataihono, line 78: this line is also not tested.

libs/journeys/ui/src/libs/context/EditorContext.tsx
```diff
@@ -79,7 +79,10 @@ type EditorAction =
   | SetActiveTabAction
   | SetStepsAction
 
-function search(tree: TreeBlock[], id: string): TreeBlock | undefined {
+export function searchBlocks(
```
tataihono, line 82: this function should be extracted out as a lib function and tested separately since it's no longer tightly coupled with EditorContext.

tataihono (changes requested):
I wonder if we should allow for parentOrder to be null in the case where it is a cover block. That way we would know that it shouldn't be displayed without needing to know anything about the parent.

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
storyworks: This is used in Attributes.spec.ts

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
storyworks: Every async function must return a Promise typescript error - it suggests `Promise<() => Promise<void>>` which isn't right either..

storyworks: Every async function must return a Promise typescript error - it suggests `Promise<() => Promise<void>>` which isn't right either..

storyworks: Every async function must return a Promise typescript error - it suggests `Promise<() => Promise<void>>` which isn't right either..

storyworks: Every async function must return a Promise typescript error - it suggests `Promise<() => Promise<void>>` which isn't right either.. but noted to simplify onClick

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/MoveBlockButtons/MoveBlockButtons.tsx
storyworks: Possibly lastBlockIndex  - I don't have a test showing that when parent == null we use parentOrder 0 since parent should always exist (Move Buttons don't get applied to StepBlock). 

However, I've added a test showing both move buttons are disabled when there's a single child now.

tataihono (approved):

PR #248, opened 2022-02-01
Editor SignUp Update Mutation

GeronimoJohn:
- Add SignUp update mutation in the backend


tataihono:
@GeronimoJohn this needs to be refactored to meet new icon block configuration.

PR #251, opened 2022-02-02
Optional parent order

storyworks:
Doing this so cover image / videos won't need parentOrder - so can ignore their ordering when sorting / moving blocks around.


apps/api-journeys/src/app/modules/block/block.resolvers.ts
```diff
@@ -49,7 +49,10 @@ export class BlockResolvers {
   ): Promise<Array<Promise<Block>>> {
     const selectedBlock: Block = await this.block(_key)
 
-    if (selectedBlock.journeyId === journeyId) {
+    if (
```
tataihono, line 52: this needs a test

tataihono (changes requested):

tataihono (approved):

PR #267, opened 2022-02-08
add wrappers to block renderer

tataihono:


storyworks (approved):

storyworks (commented):
Found some issues using this with SelectWrapper:

- When a button/option is click, don't trigger handleAction, just the select. 
- RadioOption gets rendered in RadioQuestion without going through BlockRenderer (since it needs to pass in selected / disabled props). Thus it won't have the wrappers on it..

Also, I'm getting a lot of errors in console since we're passing `block` to a fragment which makes debugging difficult.. Clicking 1 element generates a massive list of that same error:

image


To solve the console errors issue maybe you could do like:
```
  const withWrapper = (children: ReactNode): ReactElement => {
    if (wrappers != null) {
       const Wrapper = wrappers.Wrapper
      return <Wrapper block={block}>{children}</Wrapper>
    } else {
      return <>{children}</>
    }
  }

  const withComponentWrapper = (
    children: ReactNode,
    ComponentWrapper?: WrapperFn
  ): ReactElement => {
    if (wrappers != null && ComponentWrapper != null) {
      return <ComponentWrapper block={block}>{children}</ComponentWrapper>
    } else {
      return <>{children}</>
    }
  }

  switch (block.__typename) {
    case 'ButtonBlock':
      return withWrapper(withComponentWrapper(<Button {...block} />, wrappers?.ButtonWrapper))

```

storyworks (approved):

PR #271, opened 2022-02-10
fix: add IconBlock, remove Icon type

storyworks:
Icon's weren't being cached correctly since they didn't have an id. 

Needed to cache icon data for optimistic response when updating icon attributes.

This is a breaking API change so build check will never pass - need to force changes through (Tatai)

Linting failing due to previous commit messages not fitting new format.

Tested linting & type-check locally for journeys-admin, journeys-ui, journeys & api-journeys.

[Basecamp](https://3.basecamp.com/3105655/buckets/26244169/todos/4632616338)


apps/api-journeys/src/app/modules/block/icon/icon.resolver.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  })
+
+  describe('IconBlock', () => {
+    it('returns IconBlock', async () => {
+      expect(await resolver.block('1')).toEqual(blockResponse)
+      expect(await resolver.blocks()).toEqual([blockResponse, blockResponse])
+    })
+  })
+
+  describe('iconBlockCreate', () => {
```
tataihono, line 116:
```suggestion
  describe('IconBlockCreate', () => {
```

apps/api-journeys/src/app/modules/block/icon/icon.resolver.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    journeyId: '2',
+    parentBlockId: '0',
+    parentOrder: 0,
+    name: 'ArrowForwardRounded',
+    color: 'secondary',
+    size: 'lg'
+  }
+
+  const input: IconBlockCreateInput & { __typename: string } = {
+    __typename: '',
```
tataihono, line 43: this should get a correct typename

tataihono (changes requested):
I think Icons need to be deletable? Also do icons need parentOrder set?

storyworks:
> I think Icons need to be deletable? Also do icons need parentOrder set?

Updated parentOrder, delete won't be implemented in this PR - will do with the delete blocks tasks in upcoming pitch

tataihono (dismissed):

tataihono (dismissed):

PR #293, opened 2022-02-15
feat: delete block api

mikeallisonJS:
# Description

These are the necessary API calls for deleting blocks. 

This code is currently only unit tested (although was used in Background Media prior to adding the parentOrder code)

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/block/block.service.ts
+++ b/apps/api-journeys/src/app/modules/block/block.service.ts
@@ -1,8 +1,8 @@
 import { Injectable } from '@nestjs/common'
 import { aql } from 'arangojs'
-
 import { BaseService } from '@core/nest/database'
 import { DocumentCollection } from 'arangojs/collection'
+
 import { Block, Journey } from '../../__generated__/graphql'
 
 @Injectable()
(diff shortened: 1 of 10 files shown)


apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -42,5 +42,47 @@ export class BlockService extends BaseService {
     return await res.all()
   }
 
+  async updateChildrenParentOrder(
```
storyworks, line 45: This is similar to line 60 in `block.resolver.ts` do we want reuse this logic there too?

mikeallisonJS: I have refactored it to use the new methods using updateAll

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -42,5 +42,53 @@ export class BlockService extends BaseService {
...
+    ])
+  }
+
+  async removeBlockAndChildren(
+    blockId: string,
+    parentBlockId: string,
+    journeyId: string
+  ): Promise<Block[]> {
+    const res: Block = await this.remove(blockId)
+    await this.updateChildrenParentOrder(journeyId, parentBlockId)
```
storyworks, line 89: Does parentOrder also need to be refreshed after removing a block?

mikeallisonJS: It needs to be refreshed for the parent it was deleted from

storyworks (approved):

apps/api-journeys/src/app/modules/block/block.service.ts
storyworks: Oh oops, I totally glanced over line 88 and thought line 90 was deleting both the original block it's children. Approved.

PR #300, opened 2022-02-17
fix: Add parentBlockId to action in api-journeys

storyworks:
# Description

Add parentOrderId to action so we can used optimised response in editing actions. This unblocks Siyang

[Basecamp](https://3.basecamp.com/3105655/buckets/26244169/todos/4632617491)

- [x] Updated Unit tests
- [x] Viewed data returned by apollo server

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/action/linkAction/linkAction.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/action/linkAction/linkAction.resolver.spec.ts
@@ -19,6 +19,7 @@ describe('LinkActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       url: 'https://google.com'
     }
--- a/apps/api-journeys/src/app/modules/action/navigateAction/navigateAction.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/action/navigateAction/navigateAction.resolver.spec.ts
@@ -18,12 +18,13 @@ describe('NavigateActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName'
     }
   }
 
   const navigateActionInput = {
-    gtmEventName: 'gtmEventName',
+    gtmEventName: 'gtmEventNameUpdated',
     blockId: null,
     journeyId: null,
     url: null,
--- a/apps/api-journeys/src/app/modules/action/navigateToBlockAction/navigateToBlockAction.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/action/navigateToBlockAction/navigateToBlockAction.resolver.spec.ts
@@ -18,6 +18,7 @@ describe('NavigateToBlockActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       blockId: '4'
     }
--- a/apps/api-journeys/src/app/modules/action/navigateToJourney/navigateToJourney.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/action/navigateToJourney/navigateToJourney.resolver.spec.ts
@@ -23,6 +23,7 @@ describe('NavigateToJourneyActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       journeyId: '4'
     }
@@ -37,6 +38,7 @@ describe('NavigateToJourneyActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       journeyId: '4'
     }
--- a/apps/api-journeys/src/app/modules/block/button/button.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/block/button/button.resolver.spec.ts
@@ -6,7 +6,8 @@ import { BlockService } from '../block.service'
 import {
   ButtonVariant,
   ButtonColor,
-  ButtonSize
+  ButtonSize,
+  ButtonBlock
 } from '../../../__generated__/graphql'
 import { UserJourneyService } from '../../userJourney/userJourney.service'
 import { ButtonBlockResolver } from './button.resolver'
@@ -29,6 +30,7 @@ describe('Button', () => {
     startIconId: 'start1',
     endIconId: 'end1',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       url: 'https://jesusfilm.org',
       target: 'target'
@@ -48,6 +50,7 @@ describe('Button', () => {
     startIconId: 'start1',
     endIconId: 'end1',
     action: {
+      parentBlockId: '1',
       gtmEventName: 'gtmEventName',
       url: 'https://jesusfilm.org',
       target: 'target'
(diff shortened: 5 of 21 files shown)


apps/api-journeys/src/app/modules/block/button/button.resolver.spec.ts
```diff
@@ -6,14 +6,15 @@ import { BlockService } from '../block.service'
...
+  ButtonSize,
+  ButtonBlock
 } from '../../../__generated__/graphql'
 import { UserJourneyService } from '../../userJourney/userJourney.service'
 import { ButtonBlockResolver } from './button.resolver'
 
 describe('Button', () => {
-  let resolver: BlockResolver,
-    buttonBlockResolver: ButtonBlockResolver,
+  let resolver: ButtonBlockResolver,
```
storyworks, line 16: Just cleaning up, most tests use resolver to be the relevant class resolver and `blockResolver: BlockResolver`

apps/api-journeys/src/app/modules/block/videoTrigger/videoTrigger.resolver.spec.ts
```diff
@@ -1,22 +1,23 @@
...
     __typename: 'VideoTriggerBlock',
     parentBlockId: '3',
     parentOrder: 0,
-    extraAttrs: {
-      triggerStart: 5,
-      action: {
-        gtmEventName: 'gtmEventName',
-        journeyId: '4'
-      }
+    triggerStart: 5,
```
storyworks, line 16: Not sure why we had the `extraAttrs` prop, updated to be in step with `.graphl` file.

apps/api-journeys/src/app/modules/block/button/button.resolver.ts
```diff
@@ -12,6 +13,17 @@ import { RoleGuard } from '../../../lib/roleGuard/roleGuard'
 @Resolver('ButtonBlock')
 export class ButtonBlockResolver {
   constructor(private readonly blockService: BlockService) {}
+
+  @ResolveField()
+  action(@Parent() block: ButtonBlock): Action | null {
```
storyworks, line 18: We decided to update `parentBlockId` for actions on each parent of action since we couldn't reference the parent block in `action.resolver.ts` itself. 
This would need to be remembered for each future component. Doing this meant we didn't need to touch current nua seed data (less room for error while updating seed data)

It also prompted Tatai to reconsider making Icon a block and revert it back to a type.. Possibly that task will be raised early next week. Please let us know if you have any thoughts @mikeallisonJS

apps/api-journeys/src/app/modules/block/button/button.resolver.ts
```diff
@@ -13,6 +14,17 @@ import { RoleGuard } from '../../../lib/roleGuard/roleGuard'
...
+    return block.action != null
+      ? {
+          ...block.action,
+          parentBlockId: block.id
+        }
+      : null
```
tataihono, line 25: multi-line ternaries are really ugly. Can we do something a bit simpler like:
```suggestion
    if (block.action == null) return null
    
    return {
      ...block.action,
      parentBlockId: block.id
    }
```

apps/api-journeys/src/app/modules/block/radioQuestion/radioQuestion.resolver.spec.ts
```diff
@@ -24,6 +25,7 @@ describe('RadioQuestionBlockResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
```
tataihono, line 28: this doesn't need to be added here. The underlying resolver should handle adding this.

apps/api-journeys/src/app/modules/action/linkAction/linkAction.resolver.spec.ts
```diff
@@ -19,6 +19,7 @@ describe('LinkActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
```
tataihono, line 22: not needed. The underlying resolver should do this.

apps/api-journeys/src/app/modules/action/navigateAction/navigateAction.resolver.spec.ts
```diff
@@ -18,12 +18,13 @@ describe('NavigateActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
```
tataihono, line 21: not needed. The underlying resolver should do this.

apps/api-journeys/src/app/modules/action/navigateToBlockAction/navigateToBlockAction.resolver.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('NavigateToBlockActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
```
tataihono, line 21: not needed. The underlying resolver should do this.

apps/api-journeys/src/app/modules/action/navigateToJourney/navigateToJourney.resolver.spec.ts
```diff
@@ -23,6 +23,7 @@ describe('NavigateToJourneyActionResolver', () => {
     label: 'label',
     description: 'description',
     action: {
+      parentBlockId: '1',
```
tataihono, line 26: not needed. The underlying resolver should do this.

apps/api-journeys/src/app/modules/block/button/button.resolver.spec.ts
```diff
@@ -29,6 +30,7 @@ describe('Button', () => {
     startIconId: 'start1',
     endIconId: 'end1',
     action: {
+      parentBlockId: '1',
```
tataihono, line 33: not needed. The underlying resolver should do this.

tataihono (changes requested):
I've only comment on a few places but they should broadly apply to other cases that are similar.

apps/api-journeys/src/app/modules/block/button/button.resolver.spec.ts
```diff
@@ -98,9 +106,9 @@ describe('Button', () => {
   const blockService = {
     provide: BlockService,
     useFactory: () => ({
-      get: jest.fn(() => block),
-      getAll: jest.fn(() => [block, block]),
-      getSiblings: jest.fn(() => [block, block]),
+      get: jest.fn(() => blockResponse),
```
storyworks, line 109: Since `blockResponse` now contains `parentBlockId` but `block` doesn't - we need to mock the factory to return blockResponse instead of block to get `returns ButtonBlock - line 136 test to work.

Please correct my understanding if wrong, but `block` is the data stored in the db (uses key). Whilst blockResponse is what we get from the api when we try to retrieve that data (key becomes Id for some cases). Hence get / getAll should return blockResponse.

mikeallisonJS: block would be what we get from the service, blockResponse would be from the resolver after keyasid is called. Currently all calls to the service should include _key, and calls to the resolvers should use or test id

apps/api-journeys/src/app/modules/block/videoTrigger/videoTrigger.resolver.spec.ts
mikeallisonJS: looks like an oversight during copy/pasta from prisma

mikeallisonJS (approved):

PR #318, opened 2022-03-01
fix: Set Icons on Button and SignUp update

storyworks:
# Description

Api changes that allow us to make icon updates generic (not associated with any button / sign up / other block). On the client-side, on the create signUp mutation we will:

- create a signUp without icons
- create submit icon 
- update signUp with the new icon.

See 

[Basecamp](https://3.basecamp.com/3105655/buckets/26244169/todos/4656676661)

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [x] PR checks pass
- [x] Manually tested with client-side changes - we can add a sign up & button. Icon blocks display in cache tree.
- [x] Manual test - Nua journeys still all work

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged into main

PR #319, opened 2022-03-01
fix: Create SignUp and Button with Icons

storyworks:
# Description

Always create an icon placeholder block with signUp and button so icon update mutations can be called independently from its parent. Default icon is none.

Dependant on #318 

[Basecamp](https://3.basecamp.com/3105655/buckets/26244169/todos/4656676661)

- [x] PR checks pass
- [x]  Manually test we can add a sign up & button. Icon blocks display in cache tree.

# Checklist


apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab/NewButtonButton/NewButtonButton.tsx
```diff
@@ -39,43 +56,86 @@ export function NewButtonButton(): ReactElement {
...
+          iconBlockCreateInput2: {
+            id: endId,
+            journeyId,
+            parentBlockId: id,
+            name: IconName.None
+          },
+          id,
+          journeyId,
+          updateInput: {
+            startIconId: startId,
```
storyworks, line 92: Updating startIconId rather than creating it directly on create so we can validate the iconBlock exists.

apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab/NewButtonButton/NewButtonButton.stories.tsx
```diff
@@ -22,24 +27,70 @@ export const Default: Story = () => {
             query: BUTTON_BLOCK_CREATE,
             variables: {
               input: {
+                id: 'buttonBlockId',
```
storyworks, line 30: This like all current Attribute button mocks throw an error since we're using `uuid` to generate IDs and this doesn't match mocks we pass in. Can look into how to mock `uuid` in stories / fix this during cooldown..

apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.spec.tsx
```diff
@@ -220,27 +220,53 @@ describe('ControlPanel', () => {
               query: SIGN_UP_BLOCK_CREATE,
               variables: {
                 input: {
+                  id: 'uuid',
```
storyworks, line 223: Not bothering to properly mock the uuid generated id's since this is properly tested in buttons themselves. This data just needs to be correct so we don't get errors when we check that we properly switch tabs.

mikeallisonJS (approved):

PR #339, opened 2022-03-09
feature: inline edit button

storyworks:
# Description

Allow inline editing of the button label. No new line characters allowed in label.

[Basecamp](https://3.basecamp.com/3105655/buckets/26244169/todos/4709748729)

- [ ] PR checks pass
**Manual testing**
- [ ] Inline editing of button saved when clicking DONE or within card
- [ ] Text doesn't update when we click in gap between cards
- [ ] Check with Vlad it's ok that the Icon shifts to button edge
- [ ] Icon displayed properly after label saved
- [ ] Start / End Whitespace is not saved
- [ ] New line characters not saved
- [ ] Button without label keeps shape

# Checklist


apps/journeys-admin/src/components/Editor/Canvas/InlineEditWrapper/InlineEditWrapper.spec.tsx
```diff
@@ -22,50 +25,93 @@ describe('TypographyEdit', () => {
...
+            <Typography {...typographyBlock} />
+          </InlineEditWrapper>
         </EditorProvider>
       </MockedProvider>
     )
 
     fireEvent.click(getByText('test content'))
     fireEvent.click(getByText('test content'))
-    const input = getByRole('textbox')
+    const input = getByDisplayValue('test content')
```
storyworks, line 72: getByDisplayValue only returns an `input`, `textArea` or `select` component, so this is the more specific assertion.

libs/shared/ui/src/libs/themes/base/tokens/components.ts
```diff
@@ -69,6 +71,7 @@ export const baseComponents: Required<Pick<ThemeOptions, 'components'>> = {
         {
           props: { variant: 'contained', size: 'small' },
           style: {
+            minHeight: '30.75px',
```
storyworks, line 74: Even mui components don't set min-height by default once sizes classes are applied:

image

mikeallisonJS (approved):

PR #351, opened 2022-03-14
chore: self-closing eslint rule

mikeallisonJS:
# Description

Adds a rule for self-closing html & jsx tags. Additionally adds react rules to libraries and cleans up a few resulting errors.

These changes are a test

# Checklist


storyworks (dismissed):

libs/journeys/ui/.eslintrc.json
```diff
@@ -1,5 +1,10 @@
 {
-  "extends": ["plugin:@nrwl/nx/react-typescript", "../../../.eslintrc.json"],
+  "extends": [
+    "plugin:@nrwl/nx/react-typescript",
+    "../../../.eslintrc.json",
+    "next",
+    "next/core-web-vitals"
```
tataihono, line 6: this shouldn't have next dependencies as it is just a react project.

tataihono (changes requested):
not entirely sure if next and next-vitals should be included in react projects. Thoughts?

libs/journeys/ui/.eslintrc.json
mikeallisonJS: It has at least one next dependency in Image (maybe more, not sure). I'm happy to swap it (your call). Going to swap shared-ui though, since it is only react

tataihono (approved):

PR #381, opened 2022-03-18
refactor: last child on card and typography input styling

storyworks:
# Description

Refactor so removal of the bottom margin on the last-child of a card is set from the Card itself rather than the typography or button component theme. This corrects margins in future commits where "Select a block" logic will be extracted out into a BlockRendererWrapper.

Also inherit typography styles from the block itself rather than rewriting it.

VR tests should pass

- [ ] PR Checks pass

# Checklist


tataihono (dismissed):

tataihono (approved):
this should all be verifiable via VR. So happy to get this one through from a code standpoint.

PR #412, opened 2022-03-25
fix: editor access

GeronimoJohn:
# Description

Editors currently can interact with the access button when they shouldn't be. As they don't have permission to do so. The access menu should be disabled if they're an editor

[Basecamp todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4752692658)

- [x] If you're an editor of a journey. Upon clicking on the avatar icon on the JourneyList page. All the access menus should be disabled.
- [x] If you're the owner of a journey. Upon clicking on the avatar icon on the JourneyList page. All the access menus except the owner are accessible. 

# Checklist


apps/journeys-admin/src/components/AccessDialog/AccessDialog.tsx
```diff
@@ -165,15 +186,25 @@ function UserJourneyList({
...
+  userJourney: UserJourney
+  disabled: boolean
+}
+
+function ListItem({ userJourney, disabled }: ListItemProps): ReactElement {
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null)
   const open = Boolean(anchorEl)
+  const { id, role, user } = userJourney
 
   useEffect(() => () => setAnchorEl(null), [])
```
csiyang, line 209: Can we fix this callback while we're here?

apps/journeys-admin/src/components/AccessDialog/AccessDialog.tsx
```diff
@@ -66,6 +76,13 @@ export function AccessDialog({
       variables: { id: journeySlug }
     })
 
+  const { data: currentUserData } = useQuery<GetCurrentUser>(GET_CURRENT_USER)
+
+  const disable =
```
csiyang, line 81: Add some tests for disabled

csiyang (changes requested):

apps/journeys-admin/src/components/AccessDialog/AccessDialog.spec.tsx
```diff
@@ -217,6 +235,94 @@ describe('AccessDialog', () => {
...
+          <AccessDialog journeySlug="journeySlug" open={true} />
+        </MockedProvider>
+      </SnackbarProvider>
+    )
+    await waitFor(() =>
+      expect(getByRole('button', { name: 'Editor' })).toBeDisabled()
+    )
+    expect(getByRole('button', { name: 'Manage' })).toBeDisabled()
+  })
+
```
csiyang, line 325: All check that the owner is also disabled

apps/journeys-admin/src/components/AccessDialog/AccessDialog.spec.tsx
```diff
@@ -217,6 +235,94 @@ describe('AccessDialog', () => {
     )
   })
```
csiyang, line 237: If you could also add a test for when the owner is logged in that the button to edit the owners status is also disabled. You could probs just tag it on in one of your other tests tho

csiyang (changes requested):

csiyang (approved):

PR #440, opened 2022-03-31
fix: arango host mode

mikeallisonJS:
# Description

This is an attempt to fix the mac networking issue


tataihono:
I think arango just needs to be hard fixed to 3.8.6 for now until they finish those m1 capable builds.

PR #448, opened 2022-04-04
chore: update material imports

tataihono:
- feat: add no-restricted-imports to prevent incorrect importing of material modules
- chore: update all material imports
- fix: improper @mui/system/createTheme/createBreakpoints import ignored @storyworks probably a thing to refactor


apps/watch/src/components/Videos/VideoList/Carousel/VideoListCarousel.tsx
```diff
@@ -1,4 +1,4 @@
-import { Box } from '@mui/system'
+import Box from '@mui/material/Box'
```
csiyang, line 1: I think this is importing the wrong box

libs/journeys/ui/src/components/Icon/Icon.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import { Story, Meta } from '@storybook/react'
-import { Box } from '@mui/system'
+import Box from '@mui/material/Box'
```
csiyang, line 2: Wrong box here also

csiyang (changes requested):

apps/watch/src/components/Videos/VideoList/Carousel/VideoListCarousel.tsx
tataihono: nope the original was wrong.

libs/journeys/ui/src/components/Icon/Icon.stories.tsx
tataihono: all boxes should be imported from material otherwise they fully ignore theme. guaranteed this was an automatic import.

csiyang (approved):

PR #450, opened 2022-04-04
fix: typography delete when content is empty

storyworks:
# Description

Sometimes Typography blocks get updated to be empty strings and don't get deleted. 

They aren't selectable / visible but exist in the data - so other visible blocks move past them but visually it looks like no change.

(Block move is broken)[https://3.basecamp.com/3105655/buckets/26787371/todos/4786305141]

- [ ] PR Checks

# Checklist


GeronimoJohn (commented):
Need to pass in the deleteSelf prop in InlineEditWrapper
Screen Shot 2022-04-05 at 4 43 33 PM

apps/journeys-admin/src/components/Editor/Canvas/InlineEditWrapper/TypographyEdit/TypographyEdit.tsx
```diff
@@ -37,20 +40,25 @@ export function TypographyEdit({
...
-    await typographyBlockUpdate({
-      variables: {
-        id,
-        journeyId: journey.id,
-        input: { content }
-      },
-      optimisticResponse: {
-        typographyBlockUpdate: {
+
+    if (content === '') {
```
csiyang, line 44: This also means that if the user manually deletes all the text from a typog block, it will get deleted. Is this a intended feature?

storyworks: This is exactly what we want the behaviour to be - see description :)

PR #452, opened 2022-04-04
feat: link new card to previous card

csiyang:
# Description

The user having to manually connect the blocks up using next block id when creating a new card, is both confusing and unintuitive. This PR makes it so that when a new card is created, the card immediately before the created card (if it isn't already connected to a card) will get its nextBlockId set as the id of the newly created card.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4752753075)

- [ ] When creating a new card from a card that has no next connected card, connects the card with the newly created card
- [ ] When creating a card from a card that is already connected to another card, the connection should stay the same

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/CardPreview/CardPreview.spec.tsx
+++ b/apps/journeys-admin/src/components/CardPreview/CardPreview.spec.tsx
@@ -8,7 +8,10 @@ import {
 } from '../../../__generated__/GetJourney'
 import { JourneyProvider } from '../../libs/context'
 import { ThemeName, ThemeMode } from '../../../__generated__/globalTypes'
-import { STEP_AND_CARD_BLOCK_CREATE } from './CardPreview'
+import {
+  STEP_AND_CARD_BLOCK_CREATE,
+  STEP_BLOCK_NEXTBLOCKID_UPDATE
+} from './CardPreview'
 import { CardPreview } from '.'
 
 jest.mock('uuid', () => ({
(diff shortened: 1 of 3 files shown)


apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
```diff
@@ -95,13 +113,36 @@ export function CardPreview({
...
       onSelect?.(
         transformer([
           data.stepBlockCreate,
           data.cardBlockCreate
         ])[0] as TreeBlock<StepBlock>
       )
+    }
+
+    const lastStep = last(steps)
+    const nextBlockExsists =
```
storyworks, line 126: Why are check if the nextBlock exists if you've just created it? 

Could you just check `lastStep?.nextBlockId` isn't null, and if so - set nextBlockId as this new step id?

apps/journeys-admin/src/components/CardPreview/CardPreview.spec.tsx
```diff
@@ -49,10 +52,28 @@ describe('CardPreview', () => {
     expect(onSelect).toHaveBeenCalledWith(step)
   })
 
-  it('should create step and card when add button is clicked', async () => {
+  it('should create step and card when add button is clicked and set the nextBlockId for the previous step', async () => {
```
storyworks, line 55: Maybe should keep this test what it was (and test that the creation mocked result is called), then add tests `should set nextBlockId on previous step when step created` and `should keep existing nextBlockId on previous step when step created`

apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
csiyang: Sorry the name for this is kinda confusing, the check is for if there is a connected block to the last step (before the new step is created)

Currently when a stepBlock is deleted, the references to that block is not deleted. EG:
- Step 1 has no nextBlockId by default
- Create step 2, then delete step 2
- Step1 still has nextBlockId that is linked to step2

csiyang: I updated the name, and added a comment to clarify. Let me know your thoughts.

apps/journeys-admin/src/components/CardPreview/CardPreview.spec.tsx
```diff
@@ -122,4 +135,270 @@ describe('CardPreview', () => {
...
+      </MockedProvider>
+    )
+    fireEvent.click(getByRole('button'))
+    await waitFor(() => {
+      expect(cache.extract()['Journey:journeyId']?.blocks).toEqual([
+        { __ref: 'StepBlock:stepId' },
+        { __ref: 'CardBlock:cardId' }
+      ])
+    })
+    console.log(cache.extract()['Journey:journeyId'])
```
GeronimoJohn, line 212: console.log left in here

GeronimoJohn (changes requested):

GeronimoJohn (approved):

PR #455, opened 2022-04-05
fix: disable link action in editor

csiyang:
# Description

LinkAction and NavigateToJourneyAction, should not be handled when in editor mode. This PR fixes this by checking if there is a selectedBlock and sending through a boolean prop to handleAction

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4810871664)

- [x] When in editor button that has link action to URL should not redirect to link
- [x] When in editor button that has Navigate to Journey Action should not redirect to Journey
- [x] Button actions works as expected when not in editor

# Checklist

### DIFF
--- a/libs/journeys/ui/src/libs/action/handleAction.spec.ts
+++ b/libs/journeys/ui/src/libs/action/handleAction.spec.ts
@@ -12,13 +12,17 @@ jest.mock('../useBlocks/blocks', () => {
 })
 
 describe('handleAction', () => {
-  const router = {
-    push: jest.fn()
-  } as unknown as NextRouter
+  let router: NextRouter
+  beforeEach(() => {
+    router = {
+      push: jest.fn()
+    } as unknown as NextRouter
+  })
 
   const editorMode = false
 
   it('should handle empty action', () => {
+    const editorMode = false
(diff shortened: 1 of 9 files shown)


libs/journeys/ui/src/libs/action/handleAction.spec.ts
```diff
@@ -12,16 +12,22 @@ jest.mock('../useBlocks/blocks', () => {
...
+    router = {
+      push: jest.fn()
+    } as unknown as NextRouter
+  })
+
+  const editorMode = false
 
   it('should handle empty action', () => {
-    expect(() => handleAction(router)).not.toThrowError()
+    const editorMode = false
```
GeronimoJohn, line 25: can just use the editorMode on line 22

GeronimoJohn (approved):

storyworks (changes requested):
We talked about finding a solution which didn't use the editor state in the blocks, so they could remain isolated from that logic.

csiyang:
Closing because we need another solution to solve this bug as per [this comment](https://github.com/JesusFilm/core/pull/455#pullrequestreview-932715164).

PR #502, opened 2022-04-19
refactor: journey language field

tataihono:
# Description

Change to Journeys to use languageId instead of locale. Defaults to 529 if no languageId set.

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4752622705) backend pr to allow this to work

No QA required as will be tested in a future PR.

# Checklist


storyworks (approved):
Approved though I'm still unsure of the pathway we have to use locale strings with crowdin.. I thought we were going to retrieve just the languages of the top 12 languages instead of the whole list so we could ensure they would have valid bcp values..

image

PR #504, opened 2022-04-19
fix: cardBlock sets null parentOrder on coverBlocks

storyworks:
# Description

Moving blocks breaks as cover image and videos have a parentOrder and thus are considered part of the array we can reorder. 

Set parentOrder explicitly as null in the api and delete the existing coverBlocks on when cover images are replaced/removed to fix and simplify the logic on the frontend.

This is an api change prerequisite to getting move working

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4786305141)

- [x] PR checks

# Checklist


apps/api-gateway/schema.graphql
```diff
@@ -274,6 +274,11 @@ input ImageBlockCreateInput {
   ID should be unique Response UUID (Provided for optimistic mutation result matching)
   """
   id: ID
+
+  """
+  True if the coverBlockId in a parent block is set to this block id. Cover blocks should only be set at creation.
```
tataihono, line 279: The latter is already true because it is only contained in the ImageBlockCreateInput.

```suggestion
  True if the coverBlockId in a parent block should be set to this block's id.
```

tataihono (changes requested):
image and video resolvers should update the parent block with the new cover block id. Can't see that happening here.

apps/api-gateway/schema.graphql
storyworks: Added that so people won't be tempted to add the `isCover` prop to update later down the line.

tataihono: this is already true though. The prop is only available on ImageBlockCreateInput.

storyworks: I know it's already true and I'm saying I added the latter to ensure it remains true. As someone unfamiliar with the design might think we can easily toggle an image as cover or not. This could prompt them to search out more an realise we have a create & delete process for setting cover images.

apps/api-journeys/src/app/modules/block/card/card.resolver.ts
```diff
@@ -42,6 +42,19 @@ export class CardBlockResolver {
     @Args('journeyId') journeyId: string,
     @Args('input') input: CardBlockUpdateInput
   ): Promise<CardBlock> {
+    const selectedCard: CardBlock = await this.blockService.get(id)
+
+    // Delete existing coverBlock when replacing or removing
```
storyworks, line 47: @tataihono , you mentioned to remove this logic (handles replace & delete) after we make frontend changes. 

Double checking that when we click to delete (not replace) a video or image block - does that logic also move to Image/Video? Because no other block has their own custom delete mutation, and we'd need to add the `isCover` prop to the Image/Video block

apps/api-journeys/src/app/modules/block/image/image.resolver.ts
```diff
@@ -68,18 +69,43 @@ export class ImageBlockResolver {
...
+          parentBlock.coverBlockId,
+          parentBlock.journeyId
+        )
+      }
+      await this.blockService.update(parentBlock.id, {
+        coverBlockId: coverBlock.id
+      })
+
+      return coverBlock
+    } else {
```
tataihono, line 98: this else is not needed as the above statement returns a value. This simplifies the amount of code visually in a flow control block.

```
if (test) {
  return value1
} else {
  return value2
}
```
is more complex than just

```
if (test) {
  return value1
}
return value2
```

apps/api-journeys/src/app/modules/block/image/image.resolver.ts
```diff
@@ -68,18 +69,43 @@ export class ImageBlockResolver {
...
+      // Delete and replace existing coverBlock
+      if (parentBlock.coverBlockId != null) {
+        await this.blockService.removeBlockAndChildren(
+          parentBlock.coverBlockId,
+          parentBlock.journeyId
+        )
+      }
+      await this.blockService.update(parentBlock.id, {
+        coverBlockId: coverBlock.id
+      })
```
tataihono, line 95: this should be cached and reversed. That is, store the current coverBlockId in a variable, update the parent, then delete the block. This will ensure any clients that load the page always has a working coverBlock regardless of which version they get.

apps/api-journeys/src/app/modules/block/video/video.resolver.ts
```diff
@@ -22,15 +23,39 @@ export class VideoBlockResolver {
...
+      if (parentBlock.coverBlockId != null) {
+        await this.blockService.removeBlockAndChildren(
+          parentBlock.coverBlockId,
+          input.journeyId
+        )
+      }
+
+      await this.blockService.update(input.parentBlockId, {
+        coverBlockId: coverBlock.id
+      })
```
tataihono, line 45: reverse as above

apps/api-journeys/src/app/modules/block/video/video.resolver.ts
```diff
@@ -22,15 +23,39 @@ export class VideoBlockResolver {
...
+
+      return coverBlock
+    } else {
```
tataihono, line 48: else not needed as above

tataihono (changes requested):

tataihono (approved):

PR #514, opened 2022-04-21
fix: add swipe action

GeronimoJohn:
# Description

Adding the functionality to swipe to the next card in Journeys. But don't allow swiping to the previous card.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26787371/todos/4757620248)

- [ ] Swiping right on a card should jump to the next card
- [ ] Block actions should retain their functionalities
- [ ] Clicking on the ChevronRight Icon should still jump to the next card
- [ ] Swiping to the previous card should be disabled
- [ ] Swiping to the previous card should still be disabled after jumping to another journey
- [ ] Video controls should be clickable and should still be able to swipe right to the next card.

# Checklist


apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -39,20 +41,28 @@ export function Conductor({ blocks }: ConductorProps): ReactElement {
...
       if (index > -1 && swiper.activeIndex !== index) {
         swiper.slideTo(index)
       }
     }
-  }, [swiper, activeBlock, treeBlocks])
+  }, [swiper, activeBlock, index, treeBlocks])
 
   function handleNext(): void {
     if (activeBlock != null && !activeBlock.locked) nextActiveBlock()
   }
```
GeronimoJohn, line 56: I've tried having it in a way where we only have one `handleNext` function. But adding `swiper.activeIndex !== index` in the `handleNext` function makes it jump to the last card or to a different card. And this check is needed for the `onSlideChange` event. I've tried using a different swiper event such as `onTouchMove` to see if that will work. But `onTouchMove` requires tapping twice on a card before it shows the blocks visually

Still thinking of a way to try to resolve these two functions - `handleNext` and `handleSlideChange`. But so far I got it to a point where it works well!

apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -39,20 +41,28 @@ export function Conductor({ blocks }: ConductorProps): ReactElement {
...
 
+  function handleSlideChange(): void {
+    if (swiper != null && activeBlock != null && treeBlocks != null) {
+      swiper.activeIndex === treeBlocks.length - 1
+        ? setSwipePrev(true)
+        : setSwipePrev(false)
+      if (!activeBlock.locked && swiper.activeIndex !== index) nextActiveBlock()
+    }
+  }
+
```
GeronimoJohn, line 65: I've tried having it in a way where we only have one handleNext function. But adding swiper.activeIndex !== index in the handleNext function makes it jump to the last card or to a different card. And this check is needed for the onSlideChange event. I've tried using a different swiper event such as onTouchMove to see if that will work. But onTouchMove requires tapping twice on a card before it shows the blocks visually

Still thinking of a way to try to resolve these two functions - handleNext and handleSlideChange. But so far I got it to a point where it works well!

PR #527, opened 2022-04-26
fix: radio question remove inbuilt question and description - frontend

storyworks:
# Description

Removes the inbuilt question and description in RadioQuestionBlock. These will be created as seperate TypographyBlocks.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27244374/todos/4874313225)

- [ ] No heading / description as part of radio question
- [ ] Selecting the RadioQuestion shows the AddOption button, clicking on an option selects it, must click again to edit option.

# Checklist


apps/journeys-admin/src/components/Editor/Canvas/InlineEditWrapper/RadioQuestionEdit/RadioQuestionEdit.tsx
```diff
@@ -120,50 +85,22 @@ export function RadioQuestionEdit({
...
-      fullWidth
-      disableRipple
-      startIcon={
-        <AddCircleRoundedIcon sx={{ color: `${adminPrimaryColor.main}` }} />
-      }
-      onClick={handleCreateOption}
-    >
-      Add New Option
-    </StyledRadioOption>
+    <Box className="MuiButtonGroup-root MuiButtonGroup-grouped MuiButtonGroup-groupedVertical">
```
tataihono, line 89: don't do this. Just use button group. DIrectly using classes like this is very bad since these class names are not a guarantee.

tataihono (changes requested):
no using class names directly. also you have some dependency issues.

apps/journeys-admin/src/components/Editor/Canvas/InlineEditWrapper/RadioQuestionEdit/RadioQuestionEdit.tsx
storyworks: I'm doing this because it gets put within a button group (See RadioQuestion block), when the radio options render in the DOM however - each normal radio option gets wrapped in a div that has these classes on it that determine styling but this RadioOption doesn't. So I'm adding this manually to mimic what should happen. I'm still not sure what it doesn't get generated...

tataihono: Just copy in the relevant styles to sx. Don’t use underlying classes. Material ui does not guarantee these won’t change in future versions.

storyworks: That would require to use admin-only logic to style the journeys-ui RadioOption block too.

image

You would think that if the classes ever changed, the custom styles you put in would no longer match the rest of the block - so the problem you're trying to avoid would remain?

storyworks: That would require to use admin-only logic to style the journeys-ui RadioOption block too. (Box and classes removed below)

image

You would think that if the classes ever changed, the custom styles you put in would no longer match the rest of the block - so the problem you're trying to avoid would remain?

storyworks: Adding the Box mimics the wrappers around each RadioOption - ButtonGroup then applies classNames to each div and styling is fixed.

tataihono (approved):

PR #528, opened 2022-04-26
fix: title and description form reset

csiyang:
# Description
Title and description dialog textfield was not being reset when changes are made and the dialog box is canceled out of. This PR refactors both of the dialogs to use formik to handle states of the form.

[- Title](https://3.basecamp.com/3105655/buckets/27244374/todos/4874573994)
[- Description](https://3.basecamp.com/3105655/buckets/27244374/todos/4874573489)

For both title and description edit form for journey. 
- [ ] Edit the form, but cancel (instead of saving). Opening the dialog again, form should be filled with the actual value, not what was typed last.
- [ ] After saving and opening the same form again, the value should be filled with the new value, not the old value.

# Checklist


apps/journeys-admin/src/components/JourneyView/Menu/DescriptionDialog/DescriptionDialog.tsx
```diff
@@ -58,9 +55,18 @@ export function DescriptionDialog({
   }
 
   const handleClose = (): void => {
+    formik.resetForm()
     onClose()
   }
 
+  const formik = useFormik({
```
storyworks, line 62: Can you rewrite this using <Formik> like how we do in most our other components? Only the Image & VideoEditorSettings components used the hook since Mike had some trouble getting <Formik> working.

csiyang: I think we should keep this as useFormik. The submit action is handled by `Dialog` which is modifying `Formik` and is a use case to use `useFormik` [docs](https://formik.org/docs/api/useFormik)

The main trouble I had with trying to refactor was the `handleSubmit` being called by `Dialog`, the `handleSubmit` on the `Form` never gets called which makes getting the value from the `Form` a problem.

storyworks: In those docs it says:

useFormik() is a custom React hook that will return all Formik state and helpers directly. **Despite its name, it is not meant for the majority of use cases**. Internally, Formik uses useFormik to create the <Formik> component (which renders a [React Context](https://reactjs.org/docs/context.html) Provider). If you are trying to access Formik state via context, use [useFormikContext](https://github.com/formik/formik/blob/master/docs/api/useFormikContext). **Only use this hook if you are NOT using <Formik> or withFormik**.

storyworks: Oh, I see what you mean by Dialog having a submit handler too

csiyang: Resolved this by moving the dialog inside the Formik

apps/journeys-admin/src/components/Dialog/Dialog.tsx
```diff
@@ -71,11 +71,6 @@ export function Dialog({
   fullscreen,
   children
 }: DialogProps): ReactElement {
-  function handleSubmit(): void {
-    dialogAction?.onSubmit()
-    handleClose()
```
storyworks, line 76: Can you check if handleClose was used by other menu's? I have a faint memory we might've also used it to hide a dropdown menu..

csiyang: I think it was the DeleteCard menu you're thinking off, and I have checked that one already, its still working fine.

apps/journeys-admin/src/components/JourneyView/Menu/DescriptionDialog/DescriptionDialog.tsx
```diff
@@ -30,65 +29,72 @@ export function DescriptionDialog({
...
-              rows={3}
-              onChange={(e) => setValue(e.currentTarget.value)}
-            />
-          </FormControl>
-        </form>
-      </Dialog>
+      {journey != null && (
+        <Formik
+          initialValues={{ description: journey.description ?? '' }}
+          onSubmit={handleSubmit}
```
storyworks, line 69:
```suggestion
          onSubmit={handleUpdateDescription}
```

And do the same for title so don't confuse between our submit function and the one passed by formik.

tataihono (approved):

PR #541, opened 2022-04-28
fix: video duration as default value for endat

GeronimoJohn:
# Description

Currently the default value of EndAt field is all zeros. The goal of this PR is to set the video duration to be the default value for EndAt once you select a video from the video library.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27244374/todos/4883348194)

- [ ] After selecting a video from the video library, it should set the video duration as the default value for the EndAt field on JourneysAdmin
- [ ] After selecting a video from the video library, it should set the startAt value of a video to be 0
- [ ] Visit the video in its journey and the set startAt and endAt value should work as intended

# Checklist


apps/journeys-admin/src/components/Editor/VideoLibrary/VideoDetails/VideoDetails.tsx
```diff
@@ -48,7 +48,12 @@ interface VideoDetailsProps {
   open: boolean
   id: string
   onClose: () => void
-  onSelect: (videoId: string, videoVariantLanguageId?: string) => void
+  onSelect: (
+    videoId: string,
+    videoVariantLanguageId?: string,
+    startAt?: number,
```
GeronimoJohn, line 54: I found that I also need to set a default value for startAt. When the admin/user sets a value on startAt then changes the video to something else. The form keeps that startAt value which gets set on the VideoBlock. A solution that I thought of is to just set startAt value to be 0 on VideoDetails select.

tataihono (dismissed):
to make things easier can we change onSelect parameters to just block: VideoBlockUpdateInput. Thanks!

apps/journeys-admin/src/components/Editor/VideoBlockEditor/Source/Source.tsx
```diff
@@ -12,10 +12,8 @@ interface SourceProps {
 export function Source({ onChange }: SourceProps): ReactElement {
   const [open, setOpen] = useState(false)
 
-  const onSelect = async (
-    videoId: string,
-    videoVariantLanguageId?: string
-  ): Promise<void> => await onChange({ videoId, videoVariantLanguageId })
+  const onSelect = async (block: VideoBlockUpdateInput): Promise<void> =>
+    await onChange(block)
```
tataihono, line 16: I think at this point you can just pass onChange directly instead of making onSelect method.

tataihono (dismissed):

tataihono (approved):

PR #594, opened 2022-05-12
fix: set default next step after video select

GeronimoJohn:
# Description

On video create or after selecting a video from the video library. Set the default action of that video to go to the next step

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27244374/todos/4924125766)

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [ ]  After creating a new card, it should set video blocks default action to navigate to the newly created step/card
- [ ]  If there's a step after the current card, add a new video block to the current card and select a video, it should set video blocks default action to navigate to the step after
- [x]  If there's no step after, after selecting a video from the video library, it should not set anything within action

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged into main

PR #606, opened 2022-05-15
fix: set video default action after video select

GeronimoJohn:
# Description

Had to remake #594 because of commit naming errors

On video create or after selecting a video from the video library. Set the default action of that video to go to the next step

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27244374/todos/4924125766)

- [ ] After creating a new card, it should set video blocks default action to navigate to the newly created step/card
- [ ] If there's a step after the current card, add a new video block to the previous card and select a video, it should set video blocks default action to navigate to the step after
- [ ] If there's no step after, after selecting a video from the video library, it should not set anything within action

# Checklist


apps/journeys-admin/src/components/Editor/VideoLibrary/VideoDetails/VideoDetails.tsx
```diff
@@ -44,6 +47,23 @@ export const GET_VIDEO = gql`
   }
 `
 
+export const UPDATE_VIDEO_BLOCK_NEXT_STEP = gql`
```
storyworks, line 50: move this logic to VideoOptions

apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
```diff
@@ -153,6 +175,43 @@ export function CardPreview({
         }
       })
     }
+
+    // this sets video block default action to navigate to the newly created step
+    const cardBlock = lastStep?.children.find(
```
storyworks, line 180: I know you didn't create it, but may we rename lastStep to `prevStep` - since this handleClick creates a new step, so what's known as `lastStep` is technically the second to last step now.

For interest - lines 158-174 handles the case where you have cards A,B,C. You've deleted C and just created D. That code links card B to D.

apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
```diff
@@ -153,6 +175,43 @@ export function CardPreview({
...
+      (block) => block.__typename === 'CardBlock'
+    ) as unknown as TreeBlock<CardBlock>
+    const videoBlock = cardBlock?.children.find(
+      (block) => block.__typename === 'VideoBlock'
+    )
+    if (
+      !validNextBlockId &&
```
storyworks, line 187: We want to check that the action on the video. Right now, line 187 is checking if the card  has has valid next step.

Perhaps you can rewrite `const validNextBlockId =
      steps.find(({ id }) => id === lastStep?.nextBlockId) != null` as
      `const stepBlockExists = (nextBlockId: string) => { return steps.find(({ id }) => id === nextBlockId) != null}`
      
      and refactor to use `stepBlockExists(prevStep?.nextBlockId)` 
      
    then for video - check if the action is null or type "NavigateToBlockAction`. If it's NavigateToBlock action but the action blockId is invalid - then set the default action

GeronimoJohn:
Closing PR because of linting naming issues

PR #627, opened 2022-05-18
fix: no blurhash in cardpreview

GeronimoJohn:
# Description

Editing a long journey (with like 10 cards), a noticeable delay occurs. Currently, there is about a 5-10 second delay between any action in the Editor. To help solve that issue we'll stop using blurhash in card preview

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27244374/todos/4923485547)

- [ ] Adding a block in a big journey should only take 1-2 seconds

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Canvas/CardWrapper/CardWrapper.tsx
+++ b/apps/journeys-admin/src/components/Editor/Canvas/CardWrapper/CardWrapper.tsx
@@ -29,7 +29,7 @@ export function CardWrapper({ block, children }: WrapperProps): ReactElement {
         }
         return {
           ...child,
-          blurhash: null
+          blurhash: ''
         }
       }
       return child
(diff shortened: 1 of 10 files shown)


apps/journeys-admin/src/components/Editor/Canvas/CardWrapper/CardWrapper.tsx
```diff
@@ -20,8 +20,21 @@ export function CardWrapper({ block, children }: WrapperProps): ReactElement {
...
+      if (
+        child.id === block.coverBlockId &&
+        child.__typename === 'ImageBlock'
+      ) {
+        if (child.blurhash == null) {
+          return child
+        }
+        return {
+          ...child,
+          blurhash: ''
```
GeronimoJohn, line 32: Currently using an empty string value, as the blurhash field doesn't accept a null value at the moment. This is also happening at the `ImageWrapper` component. 

I was thinking of creating another PR where I change the blurhash field in the backend to be optional. Thats only if we want that

PR #653, opened 2022-05-25
feat: add events to bigquery

mikeallisonJS:
# Description

adds events to BQ export
adds executable permission on bqexport.sh


tataihono (approved):

csiyang:
I think we can remove "responses"

PR #663, opened 2022-05-26
chore: move locales to lib

tataihono:
allow rebuilding of dependencies when locales are updated.


libs/locales/af/libs-journeys-ui.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Name must be 2 characters or more": "Naam moet 2 karakters wees",
+  "Name must be 50 characters or less": "Naam moet 2 karakters of minder wees",
```
csiyang, line 3:
```suggestion
  "Name must be 50 characters or less": "Naam moet 50 karakters of minder wees",
```

csiyang (changes requested):

libs/locales/af/libs-journeys-ui.json
tataihono: I've made this update in crowdin. Will not correct here.

csiyang (approved):

PR #669, opened 2022-05-26
feat: watch crowdin general translations

mikeallisonJS:
# Description

Initial site translations for watch & translation setup

# Checklist


apps/watch/src/components/Footer/Footer.tsx
```diff
@@ -117,14 +121,16 @@ export function Footer({ isHome = false }: FooterProps): ReactElement {
           }}
         >
           <Stack direction="row" spacing="100px">
-            <Typography variant="body2">Copyright © 1995-2022</Typography>
-            <Typography variant="body2">Jesus Film Project®</Typography>
-            <Typography variant="body2">All rights reserved</Typography>
+            <Typography variant="body2">
+              {t('Copyright © 1995-2022')}
```
tataihono, line 125:
```suggestion
              {t('Copyright © 1995-{{year}}', { year: new Date().getFullYear() } )}
```

This will mean we don't have to translate this every year.

apps/watch/pages/countries/[...slug].tsx
```diff
@@ -62,10 +68,10 @@ function CountriesPage(): ReactElement {
           <Typography variant="h2">{data.country.name[0]?.value}</Typography>
           <Stack direction="row" justifyContent="space-between">
             <Typography variant="caption">
-              Part of {data.country.continent[0]?.value}
+              {t('Part of')} {data.country.continent[0]?.value}
```
Comment (tataihono) [inline apps/watch/pages/countries/[...slug].tsx:71]:
```suggestion
              {t('Part of {{country}}', { country: data.country.continent[0]?.value })}
```

Comment (tataihono) [reply apps/watch/pages/countries/[...slug].tsx:71]: think strings like this should be handled by the translators. Since syntax order is not always guaranteed.

tataihono (changes requested):

tataihono:
@mikeallisonJS can you connect about @storyworks on the format change for this. Looks like my original key definition is not something we want to continue with so this will need to be refactored to match that new key format.

PR #673, opened 2022-05-30
fix: tidy up test warnings

GeronimoJohn:
# Description

Please include a summary of the change and which todo is completed. Please also include relevant motivation and context. List any dependencies that are required for this change.

- Link to Basecamp Todo

To get a good feel of how many console warnings/errors have disappeared. Try comparing another PR's tests with the tests here

- [ ] All tests passed!
- [ ] There should be fewer test console warnings/errors - the remaining console warnings/errors should just be about first-child/first-of-type

# Checklist


GeronimoJohn:
I think we'll need to set up i18n for our test. This is to remove this warning `console.warn
    react-i18next:: You will need to pass in an i18next instance by using initReactI18next`

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/Action/data.ts
```diff
@@ -132,7 +132,7 @@ export const steps: Array<TreeBlock<StepBlock>> = [
             children: []
           },
           {
-            id: 'typographyBlockId12',
+            id: 'typographyBlockId2',
```
storyworks, line 135: What was the issue with these id's? Any reason they can be all the same now and not wrong?

PR #676, opened 2022-05-30
fix: change translation key naming and referencing

storyworks:
# Description

Talked to Tatai about not using the value as the key name but basing them on namespaces. This would provide some structure to translations as the project grows. Errors would also be more apparent (eg `t(jounrey)` would translate to `jounrey` rather than `Journey`)

How does this kind of grouping look to others?

For more info / interesting reads:
https://lokalise.com/blog/translation-keys-naming-and-organizing/
https://dev.to/omaiboroda/three-ways-to-name-i18n-translation-keys-2fed
https://phrase.com/blog/posts/ruby-lessons-learned-naming-and-managing-rails-i18n-keys/

- Link to Basecamp Todo

- [ ] Test A
- [ ] Test B

# Checklist

### DIFF
--- a/libs/journeys/ui/src/components/SignUp/SignUp.tsx
+++ b/libs/journeys/ui/src/components/SignUp/SignUp.tsx
@@ -58,7 +58,7 @@ export const SignUp = ({
   sx,
   ...props
 }: SignUpProps): ReactElement => {
-  const { t } = useTranslation('libs-journeys-ui')
+  const { t } = useTranslation('libs-journeys-ui', { keyPrefix: 'signUp' })
 
   const submitIcon = children.find((block) => block.id === submitIconId) as
     | TreeBlock<IconFields>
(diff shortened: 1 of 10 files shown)


libs/journeys/ui/src/components/SignUp/SignUp.tsx
```diff
@@ -80,12 +80,10 @@ export const SignUp = ({
   const initialValues: SignUpFormValues = { name: '', email: '' }
   const signUpSchema = object().shape({
     name: string()
-      .min(2, t('Name must be 2 characters or more'))
-      .max(50, t('Name must be 50 characters or less'))
-      .required(t('Required')),
-    email: string()
-      .email(t('Please enter a valid email address'))
-      .required(t('Required'))
+      .min(2, t('minCharacters'))
```
tataihono, line 83: I think I made an error here. Can you fix it? The number should not be fixed in the string.

```
Name must be {{count}} character or more
```

This will make this string pluralizable should do the same for the string below.

tataihono (changes requested):
did you run extract translations to generate these json files?

Just a few changes.

libs/journeys/ui/src/components/SignUp/SignUp.tsx
storyworks: Interestingly if you do `t('minCharacters, {count: 2})` it generates 
image

But if I do `t('minCharacters, {min: 2})` it generates
image

PR #687, opened 2022-06-07
feat: new journey statuses

mikeallisonJS:
# Description

This PR adds archive/delete/restore functionality to the journeys API

https://3.basecamp.com/3105655/buckets/27930172/todolists/5007994015

This will require UI components to fully test due to authentication. It should have full unit-test coverage and uses no new service calls.

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/journey/journey.graphql
+++ b/apps/api-journeys/src/app/modules/journey/journey.graphql
@@ -6,11 +6,12 @@ type Journey @key(fields: "id") {
   id: ID!
   title: String!
   language: Language!
-  lastActiveStatus: JourneyStatus
   themeMode: ThemeMode!
   themeName: ThemeName!
   description: String
   slug: String!
+  archivedAt: DateTime
+  deletedAt: DateTime
   publishedAt: DateTime
   trashedAt: DateTime
   featuredAt: DateTime
@@ -26,10 +27,10 @@ enum IdType {
 }
 
 enum JourneyStatus {
-  draft
-  published
   archived
   deleted
+  draft
+  published
   trashed
 }
 
--- a/apps/api-journeys/src/app/modules/journey/journey.graphql
+++ b/apps/api-journeys/src/app/modules/journey/journey.graphql
@@ -87,9 +87,11 @@ extend type Mutation {
   journeyUpdate(id: ID!, input: JourneyUpdateInput!): Journey!
   journeyPublish(id: ID!): Journey
   journeyArchive(id: ID!): Journey
+  journeyArchiveAllActive: [Journey]
   journeyDelete(id: ID!): Journey
   journeyTrash(id: ID!): Journey
   journeyRestore(id: ID!): Journey
+  journeyTrashAllArchived: [Journey]
 }
 
 extend type NavigateToJourneyAction {
--- a/apps/api-journeys/src/app/modules/journey/journey.graphql
+++ b/apps/api-journeys/src/app/modules/journey/journey.graphql
@@ -6,10 +6,12 @@ type Journey @key(fields: "id") {
   id: ID!
   title: String!
   language: Language!
+  lastActiveStatus: JourneyStatus
   themeMode: ThemeMode!
   themeName: ThemeName!
   description: String
   slug: String!
+  deletedAt: DateTime
   publishedAt: DateTime
   featuredAt: DateTime
   createdAt: DateTime!
@@ -26,6 +28,9 @@ enum IdType {
 enum JourneyStatus {
   draft
   published
+  archived
+  deleted
+  removed
 }
 
 input JourneysFilter {
@@ -80,6 +85,10 @@ extend type Mutation {
   journeyCreate(input: JourneyCreateInput!): Journey!
   journeyUpdate(id: ID!, input: JourneyUpdateInput!): Journey!
   journeyPublish(id: ID!): Journey
+  journeyArchive(id: ID!): Journey
+  journeyDelete(id: ID!): Journey
+  journeyRemove(id: ID!): Journey
+  journeyRestore(id: ID!): Journey
 }
 
 extend type NavigateToJourneyAction {
--- a/apps/api-journeys/src/app/modules/journey/journey.graphql
+++ b/apps/api-journeys/src/app/modules/journey/journey.graphql
@@ -86,31 +86,31 @@ extend type Mutation {
   journeyCreate(input: JourneyCreateInput!): Journey!
   journeyUpdate(id: ID!, input: JourneyUpdateInput!): Journey!
   """
-   Sets journey status to published
+  Sets journey status to published
   """
   journeyPublish(id: ID!): Journey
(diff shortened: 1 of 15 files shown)


apps/api-journeys/src/app/modules/journey/journey.graphql
```diff
@@ -80,6 +85,10 @@ extend type Mutation {
   journeyCreate(input: JourneyCreateInput!): Journey!
   journeyUpdate(id: ID!, input: JourneyUpdateInput!): Journey!
   journeyPublish(id: ID!): Journey
+  journeyArchive(id: ID!): Journey
```
csiyang, line 88: I wonder this could take an array of id's to accomodate for the "all" mutations? (archive all, delete all etc...)

mikeallisonJS: Building a 2nd set of endpoints for that, based off the current user

csiyang (approved):

apps/api-journeys/src/app/modules/journey/journey.graphql
```diff
@@ -80,6 +86,12 @@ extend type Mutation {
...
+  journeyArchiveAllActive: [Journey]
+  journeyDelete(id: ID!): Journey
+  journeyTrash(id: ID!): Journey
+  journeyRestore(id: ID!): Journey
+  journeyTrashAllArchived: [Journey]
```
tataihono, line 94: can these get some comments on how they are expected to be used?

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -160,6 +182,65 @@ export class JourneyResolver {
...
+
+    if (
+      result.status === JourneyStatus.draft ||
+      result.status === JourneyStatus.published
+    ) {
+      throw new UserInputError('Journey is already active')
+    }
+    return await this.journeyService.update(id, {
+      status: lastActiveStatus
+    })
```
tataihono, line 241: small nit. not worth calculating lastActiveStatus unless the if statement is false.

```suggestion
    if (
      result.status === JourneyStatus.draft ||
      result.status === JourneyStatus.published
    ) {
      throw new UserInputError('Journey is already active')
    }
    
    const lastActiveStatus =
      result.publishedAt == null ? JourneyStatus.draft : JourneyStatus.published

    return await this.journeyService.update(id, {
      status: lastActiveStatus
    })
```

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -160,6 +182,65 @@ export class JourneyResolver {
...
+  }
+
+  async journeyTrashAllArchived(
+    @CurrentUserId() userId: string
+  ): Promise<Journey[]> {
+    const results: Journey[] = (await this.journeyService.getAllByOwner(userId))
+      .filter((j) => j.status === JourneyStatus.archived)
+      .map((j) => ({
+        ...j,
```
tataihono, line 218: small nit. Not a huge fan of shorthand.
```suggestion
      .filter((journey) => journey.status === JourneyStatus.archived)
      .map((journey) => ({
        ...journey,
```

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -160,6 +182,65 @@ export class JourneyResolver {
...
+    return await this.journeyService.update(id, {
+      status: JourneyStatus.trashed,
+      trashedAt: new Date().toISOString()
+    })
+  }
+
+  async journeyTrashAllArchived(
+    @CurrentUserId() userId: string
+  ): Promise<Journey[]> {
+    const results: Journey[] = (await this.journeyService.getAllByOwner(userId))
```
tataihono, line 215: wonder if we could pass a filter to getAllByOwner? if the user only has hundreds of journeys these are all going to be loaded into memory just to be filtered down.

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -178,11 +259,6 @@ export class JourneyResolver {
     return await this.userJourneyService.forJourney(journey)
   }
 
-  @ResolveField()
-  status(@Parent() { publishedAt }: Journey): JourneyStatus {
-    return publishedAt == null ? JourneyStatus.draft : JourneyStatus.published
-  }
-
```
tataihono, line 185: what happens if the status field is not set in the db? I assume this is going to fail? Should be a default if status is undefined or null.

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -47,19 +47,36 @@ export class JourneyService extends BaseService {
...
+      FOR journey in ${this.collection}
+          FILTER userJourney.journeyId == journey._key && userJourney.userId == ${userId}
+           && userJourney.role == ${UserJourneyRole.owner}
+          RETURN journey
+    `)
+    return await result.all()
+  }
+
+  @KeyAsId()
+  async getAllByOwnerEditor(
```
tataihono, line 70:
```suggestion
  async getAllByOwnerOrEditor(
```

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -47,19 +47,36 @@ export class JourneyService extends BaseService {
...
+    userId: string,
+    status: JourneyStatus[]
+  ): Promise<Journey[]> {
     const result = await this.db.query(aql`
     FOR userJourney in userJourneys
       FOR journey in ${this.collection}
           FILTER userJourney.journeyId == journey._key && userJourney.userId == ${userId}
            && (userJourney.role == ${UserJourneyRole.owner} || userJourney.role == ${UserJourneyRole.editor})
+           && journey.status IN ${status}
```
tataihono, line 79: can status be optional here?

tataihono (changes requested):

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
mikeallisonJS: Status is a required field, I'm not sure there is any case it could

apps/api-journeys/src/app/modules/journey/journey.service.ts
mikeallisonJS: This is the original name prior to this PR, do you want it changed?

tataihono: oh. leave it as is.

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
tataihono: right. Looks like the db is already populated.

tataihono (approved):

csiyang (approved):

PR #695, opened 2022-06-14
fix: current nextblockid logic

GeronimoJohn:
# Description

Currently we auto-set the nextBlockId to the next step by default (was a fix for a button navigation issue) so this needs to be reverted

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/27930061/todos/5031346158)

- [ ] Adding a new card doesn't set the previous cards nextBlockId to the newly created cards ID

_Under Next Card Properties drawer_
- [ ] If a step has null nextBlockId and has a step after, we should see the thumbnail, heading, and the text '**default next step in journey**'. The next step is also selected in preview
- [ ] If a step has null nextBlockId and is the last step in the journey, we should see the thumbnail, heading, and the text '**no next step in journey**'. No step should be selected
- [ ] If the step has an explicit nextBlockId, we should see the thumbnail, heading, and the remove icon. The step with the id of nextBlockId should be selected

# Checklist


apps/journeys-admin/src/components/CardPreview/CardPreview.tsx
```diff
@@ -157,28 +153,6 @@ export function CardPreview({
...
-            __typename: 'StepBlock',
-            id: prevStep.id,
-            nextBlockId: stepId
-          }
-        }
-      })
-    }
-
     // this sets video block default action to navigate to the newly created step
     const prevCard = prevStep?.children.find(
```
mikeallisonJS, line 157: Should this affect the video block action as well?

mikeallisonJS (approved):

PR #702, opened 2022-06-16
fix: More accurately type getSiblingsInternal return type

storyworks:
# Description

As I write tests for the blockDuplicate, I've discovered we incorrectly type the return object for `getSiblingsInternal`. This matters I need to know when the `_key` property is present or not (arrango allows us to use a user defined `_key` to specify the `_id` - we cannot directly set `_id`). This is important for the duplication api since I'm generating id's for each block.

More accurately we get something like when we call `getSiblingsInternal`
image

- [x] Manually tested move blocks and delete blocks still works
- [x] Unit tests pass

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/block/block.service.ts
+++ b/apps/api-journeys/src/app/modules/block/block.service.ts
@@ -32,7 +32,7 @@ export class BlockService extends BaseService {
   async getSiblingsInternal(
     journeyId: string,
     parentBlockId?: string | null
-  ): Promise<Block[]> {
+  ): Promise<Array<Block & { _key: string }>> {
     // Only StepBlocks should not have parentBlockId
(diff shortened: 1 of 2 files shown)


apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -32,7 +32,7 @@ export class BlockService extends BaseService {
   async getSiblingsInternal(
     journeyId: string,
     parentBlockId?: string | null
-  ): Promise<Block[]> {
+  ): Promise<Array<Block & { _key: string }>> {
```
storyworks, line 35: Actual return type is

image
(generated when moving a block in local env)

mikeallisonJS (approved):

PR #703, opened 2022-06-17
feat: add saveAll in base service

storyworks:
# Description

Allows us to use saveAll to create multiple blocks at once. Used for the duplication cycle 5 project.

The unit test, dbMock and function itself is very similar to `updateAll` and `save`.

- [ ] Unit tests pass
- [ ] Tested it works to save multiple blocks with duplication code branch

# Checklist

### DIFF
--- a/apps/api-journeys/db/seeds/nua1.ts
+++ b/apps/api-journeys/db/seeds/nua1.ts
@@ -420,8 +420,7 @@ export async function nua1(): Promise<void> {
     __typename: 'IconBlock',
     parentBlockId: button2._key,
     name: 'ContactSupportRounded',
-    size: 'md',
-    parentOrder: 4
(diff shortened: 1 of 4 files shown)


apps/api-journeys/db/seeds/nua1.ts
```diff
@@ -420,8 +420,7 @@ export async function nua1(): Promise<void> {
     __typename: 'IconBlock',
     parentBlockId: button2._key,
     name: 'ContactSupportRounded',
-    size: 'md',
-    parentOrder: 4
```
storyworks, line 424: Also found invalid parentOrder left on IconBlocks (will always be null parentOrder since ButtonBlock sets where it is displayed)

mikeallisonJS (approved):

PR #705, opened 2022-06-17
feat: block duplication api

storyworks:
# Description

Allows us to duplicate blocks (inc cards). `getDuplicateBlockAndChildren` will be reused for each StepBlock on a journey when duplicating journeys. 

- [Duplicate Block API](https://3.basecamp.com/3105655/buckets/27930137/todos/5008338102)
- [Duplicate Card API](https://3.basecamp.com/3105655/buckets/27930137/todos/5008354498)

Manually tested locally - I created a duplicate button and works for the following scenarios
- [x] Duplicate a block with no children (typography / radio option / sign up / image / video) & duplicated block appears after original. Visually VideoBlock which takes up the entirety of the card so duplicate VideoBlocks are on stacked on top.
- [x] Duplicate a block with children (radio question / button) & duplicated block appears after original
- [x] Duplicate a step / card & duplicated block appears at end of list

# Checklist


apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,140 @@ export class BlockService extends BaseService {
     return []
   }
 
+  async duplicateBlock(
```
storyworks, line 90: Calls `getDuplicateBlockAndChildren` and positions the newly created duplicate block in the correct place.

Would appreciate thoughts on the name `getDuplicateBlockAndChildren` because this just retrieves the new block objects we want to save. Line 103 takes this data and actually creates the blocks.

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,140 @@ export class BlockService extends BaseService {
...
+  ): Promise<Block[]> {
+    const block = idAsKey(await this.get(id)) as Block & { _key: string }
+
+    if (block.journeyId === journeyId) {
+      const blockAndChildrenData = await this.getDuplicateBlockAndChildren(
+        id,
+        journeyId,
+        block.parentBlockId ?? null
+      )
+      const duplicateBlockAndChildren = await this.saveAll(blockAndChildrenData)
```
storyworks, line 103: We want the `blockAndChildrenData` to have the `_key` defined so the `_id` will be generated based off that. Otherwise the values we generated for `id` in `getDuplicateBlockAndChildren` get overwritten by arrango's autogenerated `_id`

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,140 @@ export class BlockService extends BaseService {
...
+      const duplicatedBlock = {
```
storyworks, line 104: `.getSiblingsInternal` returns block objects with keys, so adding key to this to keep the siblings array consistent after inserting this in.

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,140 @@ export class BlockService extends BaseService {
...
+      return reorderedBlocks.concat(duplicatedChildren)
+    }
+    return []
+  }
+
+  async getDuplicateBlockAndChildren(
+    id: string,
+    journeyId: string,
+    parentBlockId: string | null,
+    duplicateId?: string
```
storyworks, line 130: A block may generate a duplicateId for it's child - this is needed in the cases where we need to set a child's id in the parent (eg for `endIconId, posterBlockId ...`)

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,141 @@ export class BlockService extends BaseService {
...
+
+      const siblings = await this.getSiblingsInternal(
+        journeyId,
+        block.parentBlockId
+      )
+      const insertIndex =
+        parentOrder != null
+          ? Math.min(Math.max(parentOrder, 0), siblings.length + 1)
+          : siblings.length + 1
+      siblings.splice(insertIndex, 0, duplicatedBlock)
```
storyworks, line 120: Today we've just found an issue where the duplicated block was being returned in the data twice (doesn't impact visuals, but the same block is repeated in the returned data. Fixing it now)

apps/api-journeys/src/app/modules/block/block.graphql
```diff
@@ -14,5 +14,9 @@ extend type Mutation {
   blockDelete returns the updated sibling blocks on successful delete
   """
   blockDelete(id: ID!, journeyId: ID!, parentBlockId: ID): [Block!]!
+  """
+  blockDuplicate returns the updated block, it's children and sibling blocks on successful duplicate
+  """
+  blockDuplicate(id: ID!, journeyId: ID!, parentOrder: Int): [Block!]!
```
tataihono, line 20: can you document the parent order arg here? something communicating that this is where the duplicated block should be placed otherwise where it goes if null?

tataihono (changes requested):

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -86,6 +87,144 @@ export class BlockService extends BaseService {
...
+            _key: duplicateBlockId,
+            id: duplicateBlockId,
+            parentBlockId,
+            nextBlockId: null
+          }
+        : {
+            ...block,
+            _key: duplicateBlockId,
+            id: duplicateBlockId,
+            parentBlockId
```
storyworks, line 190: In a subsequent PR (journeys duplication api) I extract this out into `defaultDuplicateBlock` and spread it over the explicitly defined blocks above to simplify.

apps/api-journeys/src/app/modules/block/block.graphql
tataihono: https://www.apollographql.com/docs/apollo-server/schema/schema/#descriptions-docstrings

```
blockDuplicate(
	id: ID!
	journeyId: ID!
	"Description for an argument"
	parentOrder: Int
): [Block!]!
```

tataihono (changes requested):

tataihono (approved):

storyworks:
Will update this api to look through a list of all generated child ids, and map to an attribute if it's key ends with "Id". This removes the need to know about the internals of a particular block.

PR #747, opened 2022-06-30
fix: firebase private key in env variable too large

tataihono:
# Description

This PR removes PRIVATE_FIREBASE_PRIVATE_KEY from Vercel deploys and stores its value in the repo for the various environments. It takes the value of that environment variable and instead encrypts it and places it into the repo (as FIREBASE_PRIVATE_KEY_ENC). The key and iv are then stored as environment variables instead of the full private key.

This should save a whole 2-3 kb which historically has been stored in the environment variables. This would often cause deployment failures when the system variables were sometimes > 1 kb breaching the 4kb max limit in Vercel.

Crypto convenience functions have been added in case future encryption of firebase private keys or other such values are required.

- [ ] Journeys admin login should still work

# Checklist


libs/shared/ui/src/libs/crypto/decrypt.ts
```diff
@@ -0,0 +1,19 @@
+import crypto from 'crypto'
+
+export function decrypt(
+  encrypt: [key: string, iv: string, encrypted: string]
+): string {
+  const [key, iv, encrypted] = encrypt
+  const decipher = crypto.createDecipheriv(
+    'aes-256-cbc',
```
storyworks, line 8: Mainly for interest - why this kind of encryption over others?

tataihono: https://vercel.com/support/articles/how-do-i-workaround-vercel-s-4-kb-environment-variables-limit they use aes-128-cbc. Arguably this is stronger than that.

storyworks: Seems like it's the default one used in documentation

PR #764, opened 2022-07-06
fix: firefox cardlist margin

csiyang:
# Description

Found that Firefox wasn't correctly scaling the iframes in card list, which then left the full height of the card on the y axis, then an overflow: hidden was positioning the cards so that any spacing on top of the card is lost.  The solution was to add an additional overflow: hidden so the full height is not passed to horizontal select.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/28141725/todos/5055669888)

**This solution uses negative margins, so should thoroughly QA'd to ensure it is still working as intended**

- [ ] Drag should show and work, cards should have correct outline when selected:
> - [ ] on Chrome web view
> - [ ] on Chrome mobile view
> - [ ] on Firefox web view
> - [ ] on Firefox mobile view
> - [ ] on Safari web view
> - [ ] on Safari mobile view

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/CardPreview/CardList/CardList.tsx
+++ b/apps/journeys-admin/src/components/CardPreview/CardList/CardList.tsx
@@ -123,7 +123,8 @@ const CardItem = ({
         width: 95,
         height: 140,
         position: 'relative',
-        top: provided != null ? -24 : undefined
+        // top: provided != null ? -24 : undefined,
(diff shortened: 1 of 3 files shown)


mikeallisonJS:
With this branch, I am seeing blanks occasionally in chrome now

mikeallisonJS (approved):

PR #783, opened 2022-07-15
fix: open the first property automatically

GeronimoJohn:
# Description

When activating a block, we need to activate or highlight the first property automatically

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/28141725/todos/5056201318)

- [ ] Clicking on a block automatically opens its first property

# Checklist


apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/Attributes.tsx
```diff
@@ -90,15 +88,6 @@ interface AttributesProps {
 }
 
 export function Attributes({ selected, step }: AttributesProps): ReactElement {
-  const { dispatch } = useEditor()
```
csiyang, line 93: Nice one for moving the logic for this into selectable wrapper!

There is a few cases that was being covered by this that you still need to add though:
1. When you switch to a new card the last selected attribute is still open, you shouldn’t be able to edit properties on a card that isn’t open, maybe select the the current card properties?
2. When you delete a block, the drawer for that block is still open, should probs also open card properties since it gets selected.
3. Not sure if this PR is meant to cover this, but when you create a new block, it should show the first property of the newly created block. Currently still previous selected property is persisting.

PR #769, opened 2022-07-11
feat: replace video url with language

mikeallisonJS:
# Description

This branch replaces the video url with current video language

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/28141725/todos/5055608394)

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [ ] Check that video blocks display language instead of url

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged into main

PR #798, opened 2022-07-27
fix: journeys language overflow

mikeallisonJS:
# Description

Currently, journeys are pulling variant language for each video block. In the case of The Jesus Film, this is 1900 per block. This is likely affecting performance. I have moved the variant call to the relevant admin section and made it non-blocking.

This is a likely unintended side effect of https://github.com/JesusFilm/core/pull/769

The majority of the changes in this PR are just removing the variantLanguages from specs and generated files.

- [ ] Ensure language is still present in video block properties

# Checklist


storyworks (changes requested):
image

Works but the language is now not being displayed, happy to approve pending that.

storyworks (approved):

PR #800, opened 2022-07-28
feat: fetch public templates

csiyang:
# Description

Extend and refactor filter to fetch all published templates

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/28496234/todos/5164988174)

- [x] API

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
@@ -10,6 +10,7 @@ import { keyAsId } from '@core/nest/decorators/KeyAsId'
 import { AqlQuery } from 'arangojs/aql'
 import {
   JourneyStatus,
+  Role,
   ThemeMode,
   ThemeName
 } from '../../__generated__/graphql'
@@ -124,6 +125,31 @@ describe('JourneyService', () => {
         return await mockDbQueryResult(db, [journey])
       })
       await service.getAllPublishedJourneys({ featured: false })
+      expect(db.query).toHaveBeenCalled()
+    })
+
+    it('should filter by template', async () => {
+      const templateJourney = {
+        ...journey,
+        template: true
+      }
+
+      db.query.mockImplementationOnce(async (q) => {
+        const { query, bindVars } = q as unknown as AqlQuery
+        expect(query).toEqual(
+          aql`FOR journey IN undefined
+          FOR user in userRoles
+          FILTER journey.template == true
+            AND user.roles IN @value0
+          RETURN journey
+      `.query
+        )
+        expect(bindVars).toEqual({
+          value0: Role.publisher
+        })
+        return await mockDbQueryResult(db, [templateJourney])
+      })
+      await service.getAllPublishedJourneys({ featured: true, template: true })
       expect(db.query).toHaveBeenCalled()
     })
   })
--- a/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
@@ -86,7 +86,9 @@ describe('JourneyService', () => {
     it('should return template query', async () => {
       const filter: JourneysFilter = { featured: true, template: true }
       const response = await service.journeyFilter(filter)
-      expect(response.query).toEqual(aql`AND journey.template == true`.query)
+      expect(response.query).toEqual(
+        aql`AND journey.template == true AND journey.featuredAt != null`.query
+      )
     })
     it('should return featured query', async () => {
       const filter: JourneysFilter = { featured: true, template: false }
--- a/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
@@ -121,10 +121,10 @@ describe('JourneyService', () => {
     })
   })
 
-  describe('getAllPublishedJourneys', () => {
+  describe('getAllFilteredJourneys', () => {
     it('should return published journeys', async () => {
       db.query.mockReturnValueOnce(mockDbQueryResult(db, [journey]))
-      expect(await service.getAllPublishedJourneys()).toEqual([journeyWithId])
+      expect(await service.getAllFilteredJourneys()).toEqual([journeyWithId])
     })
 
     it('should filter by featured', async () => {
@@ -143,7 +143,7 @@ describe('JourneyService', () => {
         })
         return await mockDbQueryResult(db, [journey])
       })
-      await service.getAllPublishedJourneys({ featured: true })
+      await service.getAllFilteredJourneys({ featured: true })
       expect(db.query).toHaveBeenCalled()
     })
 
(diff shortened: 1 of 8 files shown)


storyworks (dismissed):

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -12,31 +13,37 @@ import {
...
+    } else {
+      query = aql`FILTER journey.status == ${JourneyStatus.published}
+          AND journey.template != true`
     }
+
+    return query
+  }
+
+  @KeyAsId()
+  async getAllFilteredJourneys(filter?: JourneysFilter): Promise<Journey[]> {
```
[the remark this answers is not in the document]
csiyang: Reverted

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -12,31 +13,37 @@ import {
...
-            RETURN journey
-        `)
-      ).all()
+  journeyFilter(filter?: JourneysFilter): AqlQuery {
+    const { featured, template } = filter ?? {}
+
+    let query: AqlQuery
+
+    if (template === true) {
+      query = aql`FILTER journey.template == true
```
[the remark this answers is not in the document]
csiyang: Refactored

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -12,31 +13,32 @@ import {
...
-      ).all()
-    }
+    const search = this.journeyFilter(filter)
+
     return await (
-      await this.db.query(aql`
-        FOR journey IN ${this.collection}
+      await this.db.query(aql`FOR journey IN ${this.collection}
           FILTER journey.status == ${JourneyStatus.published}
+          ${search}
```
storyworks, line 41: Have you been able to test this? From memory, passing in a variable in place of an entire line didn't work. I think the filter string needs to start with a string instead of a variable..

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -12,31 +13,32 @@ import {
 
 @Injectable()
 export class JourneyService extends BaseService {
+  journeyFilter(filter?: JourneysFilter): AqlQuery {
+    const { featured, template } = filter ?? {}
+
+    return aql.join(
+      [
+        template === true
+          ? aql`AND journey.template == true`
```
storyworks, line 22: This can be simplified `AND journey.template == ${template}`

apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -12,31 +13,32 @@ import {
...
+          : aql`AND journey.template != true`,
+        featured === true &&
+          template !== true &&
```
storyworks, line 25: Why does the featured string depend on the template condition?

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: No, assumed it would work since its done like this in video service

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: This suggestion misses the option if template is null

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: because when we get all the templates, we dont want to filter by featured

apps/api-journeys/src/app/modules/journey/journey.service.ts
storyworks: `template ?? false` means if template is null, use false.

csiyang: The aql query will then be either true or false. So then on the actual query it misses the null case

apps/api-journeys/src/app/modules/journey/journey.service.ts
storyworks: all the variables should default to false if not passed so checks can be isolated from one another

apps/api-journeys/src/app/modules/journey/journey.service.ts
storyworks: How is the `null` case treated differently to the `false` case?

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: When we filter by templates, we should not filter by featured. So if `{featured: true, template: true}` the above check makes sure it doesnt filter featured in the above case.

Also since featured and templates props are nullable, defaulting to false could cause side effects so the check just makes sure we get what we want.

I don't mind changing it to default to false if its algood though.

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: template is nullable on journeys. If we filter by `{template == false}` it doesnt get {template == null}. But we need both since they're both non templates

storyworks: Have you checked that? https://www.arangodb.com/docs/stable/aql/fundamentals-document-data.html documentation says:

image

So it seems that if we don't set template on a journey/it's null, it would be returned..

storyworks: Have you checked that? https://www.arangodb.com/docs/stable/aql/fundamentals-document-data.html documentation says:

image

So it seems that if we don't set template on a journey/it's null, it would be returned..

apps/api-journeys/src/app/modules/journey/journey.service.ts
storyworks: You can test if it works using the staging environment, will @ you in the slack message that describes how to do this.

apps/api-journeys/src/app/modules/journey/journey.service.ts
csiyang: That only applies to checking for lesser than

image

storyworks: After discussion was recommended to test out `AND journey.template == ${template === true}`

apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
```diff
@@ -81,6 +82,33 @@ describe('JourneyService', () => {
     })
   })
 
+  describe('journeyFilter', () => {
+    it('should return template query', async () => {
```
storyworks, line 86: After you isolate the conditions, you can just test the use cases:
`filter = { featured: true}`
`filter = {template: true}`
`filter = {}`

apps/api-journeys/src/app/modules/journey/journey.service.spec.ts
```diff
@@ -126,6 +152,30 @@ describe('JourneyService', () => {
       await service.getAllPublishedJourneys({ featured: false })
       expect(db.query).toHaveBeenCalled()
     })
+
+    it('should filter by template', async () => {
```
storyworks, line 158: And you can replace "should filter by featured / should filter by not featured / should filter by template" to `should filter by filter query`. (As you already tested the filter permutations in `journeyFilter` now just need to check that `getAllPublishedJourneys` uses that function)

csiyang:
Query has been tested in stage

storyworks (approved):

PR #805, opened 2022-08-02
fix: extend role guard

GeronimoJohn:
# Description

Added new roles to our role guard. We've added the ability to check if the user is currently a publisher and if a journey is currently a template.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/28496234/todos/5168827856)

- [ ] API only
- [ ] Run any mutation call on the front end and it should still as expected

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
+++ b/apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
@@ -291,53 +291,76 @@ describe('RoleGuard', () => {
     })
   })
 
-  // describe('custome Roles', () => {
-  //   it('should return true if role is publisher and template true and status published', async () => {
-  //     const gqlContext = gqlContextMockFactory(
-  //       { id: '2' },
-  //       {
-  //         headers: { 'user-id': '1' },
-  //         req: createMock<Request>()
-  //       }
-  //     )
-  //     const RoleGuardClass = RoleGuard(
-  //       'id',
-  //       [
-  //         {
-  //           role: Role.publisher,
-  //           attributes: { template: true, status: JourneyStatus.published }
-  //         }
-  //       ],
-  //       fetchUserJourney,
-  //       fetchUserRole,
-  //       fetchJourney
-  //     )
-  //     const roleGuard = new RoleGuardClass(gqlContext)
-  //     expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
-  //   })
+  describe('custom Roles', () => {
+    it('should return true if role is publisher and template true', async () => {
+      const gqlContext = gqlContextMockFactory(
+        { id: '2' },
+        {
+          headers: { 'user-id': '1' },
+          req: createMock<Request>()
+        }
+      )
+      const RoleGuardClass = RoleGuard(
+        'id',
+        [
+          {
+            role: Role.publisher,
+            attributes: { template: true }
+          }
+        ],
+        fetchUserJourney,
+        fetchUserRole,
+        fetchJourney
+      )
+      const roleGuard = new RoleGuardClass(gqlContext)
+      expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
+    })
 
-  //   it('should return true if role is public and template true', async () => {
-  //     const gqlContext = gqlContextMockFactory(
-  //       { id: '2' },
-  //       {
-  //         headers: { 'user-id': '1' },
-  //         req: createMock<Request>()
-  //       }
-  //     )
-  //     const RoleGuardClass = RoleGuard(
-  //       'id',
-  //       [
-  //         {
-  //           role: 'public',
-  //           attributes: { template: true }
-  //         }
-  //       ],
-  //       fetchUserJourney,
-  //       fetchUserRole,
-  //       fetchJourney
-  //     )
-  //     const roleGuard = new RoleGuardClass(gqlContext)
-  //     expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
-  //   })
-  // })
+    it('should return true if role is an array and includes publisher', async () => {
+      const gqlContext = gqlContextMockFactory(
+        { id: '2' },
+        {
+          headers: { 'user-id': '1' },
+          req: createMock<Request>()
+        }
+      )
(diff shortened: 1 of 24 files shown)


apps/api-journeys/src/app/lib/roleGuard/roleGuard.ts
```diff
@@ -7,59 +7,192 @@ import {
...
       const ctx = GqlExecutionContext.create(context).getContext()
       const userId = get(ctx.headers, 'user-id')
 
       if (userId == null) return false
 
       const args = context.getArgByIndex(1)
+      const access = Array.isArray(permissions) ? permissions : [permissions]
       const journeyId = get(args, journeyIdArgName)
       if (journeyId == null)
         throw new AuthenticationError('No journeyId provided')
```
storyworks, line 166: let's add:

`const journeyIds = [journeyId].flat()`

This allows us to remove the if/else `Array.isArray(journeyId)` check and just loop through `journeyIds`
Move the `const userRole = await fur(this.userRoleService, userId)` outside the loop, as we fetching it more than once will be redundant. This will save us some db calls making the api faster.

Then for the loop:
```
journeyIds.forEach((journeyId)=>{ 
// fetch journey
// fetch userJourney
// checkAllowedAccess
})
```

storyworks (changes requested):

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -324,7 +345,13 @@ export class JourneyResolver {
   }
 
   @Mutation()
-  @UseGuards(RoleGuard('ids', UserJourneyRole.owner))
+  @UseGuards(
+    RoleGuard('ids', [
+      UserJourneyRole.owner,
+      Role.publisher,
```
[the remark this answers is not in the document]
GeronimoJohn: Oh good find! I didn't realised I left this

apps/api-journeys/src/app/lib/roleGuard/roleGuard.ts
storyworks: You can still convert `journeyId` to always be an array and remove the extra if/else check. Most of the code is unnecessarily duplicated. 

At the very least, move out `userRole` fetching

apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
```diff
@@ -150,9 +277,128 @@ describe('RoleGuard', () => {
...
+        fetchUserRole,
+        fetchJourney
+      )
+      const roleGuard = new RoleGuardClass(gqlContext)
+      expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
+    })
+  })
+
+  describe('custom Roles', () => {
+    it('should return true if role is publisher and template true', async () => {
```
storyworks, line 333: `should return true for 1 role and attribute`

apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
```diff
@@ -150,9 +277,128 @@ describe('RoleGuard', () => {
...
+        ],
+        fetchUserJourney,
+        fetchUserRole,
+        fetchJourney
+      )
+      const roleGuard = new RoleGuardClass(gqlContext)
+      expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
+    })
+
+    it('should return true if role is an array and includes publisher', async () => {
```
storyworks, line 357: `should return true for multiple roles`

apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
```diff
@@ -150,9 +277,128 @@ describe('RoleGuard', () => {
...
+        ],
+        fetchUserJourney,
+        fetchUserRole,
+        fetchJourney
+      )
+      const roleGuard = new RoleGuardClass(gqlContext)
+      await expect(roleGuard.canActivate(gqlContext)).resolves.toEqual(true)
+    })
+
+    it('should return true if role is public and template true and status published', async () => {
```
storyworks, line 380: `should return true for role with attributes`

apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
```diff
@@ -150,9 +277,128 @@ describe('RoleGuard', () => {
...
+        fetchUserRole,
+        fetchJourney
+      )
       const roleGuard = new RoleGuardClass(gqlContext)
       expect(await roleGuard.canActivate(gqlContext)).toEqual(false)
     })
   })
+
+  describe('user Role', () => {
+    it('should return true if user role includes publisher', async () => {
```
storyworks, line 293: `should return true for single user role`

apps/api-journeys/src/app/lib/roleGuard/roleGuard.spec.ts
```diff
@@ -150,9 +277,128 @@ describe('RoleGuard', () => {
...
+        Role.publisher,
+        fetchUserJourney,
+        fetchUserRole,
+        fetchJourney
+      )
+      const roleGuard = new RoleGuardClass(gqlContext)
+      expect(await roleGuard.canActivate(gqlContext)).toEqual(true)
+    })
+
+    it('should return true if role is an array and includes publisher', async () => {
```
storyworks, line 312: should return true for user role in array

apps/api-journeys/src/app/lib/roleGuard/roleGuard.ts
```diff
@@ -7,32 +7,141 @@ import {
...
+      @Inject(UserRoleService)
+      private readonly userRoleService: UserRoleService,
+      @Inject(JourneyService)
+      private readonly journeyService: JourneyService
     ) {}
 
+    checkAttributes(journey: Journey, attributes?: Partial<Journey>): boolean {
+      if (attributes == null || attributes === {}) return true
+      return Object.keys(attributes).every(
+        (key: any) => attributes[key] === journey[key]
```
storyworks, line 78: Linting

storyworks (approved):

PR #812, opened 2022-08-05
fix: publisher templates fetch

csiyang:
# Description

When fetch with `template: true` and user not a `publisher`, the query returns a list of journeys owned by the user. But it should not return anything 

- Link to Basecamp Todo

- [x] Tested you can still fetch journeys locally
- [x] Tested you can still fetch templates locally
- [x] Tested the described use case is fixed

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/journey/journey.service.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.service.ts
@@ -96,19 +96,24 @@ export class JourneyService extends BaseService {
       status != null ? aql`&& journey.status IN ${status}` : aql`&& true`
 
     const roleFilter =
-      template === true && includes(user.roles, Role.publisher)
-        ? aql`FOR journey in ${this.collection}
+      template === true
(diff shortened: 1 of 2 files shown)


apps/api-journeys/src/app/modules/journey/journey.service.ts
```diff
@@ -96,19 +96,24 @@ export class JourneyService extends BaseService {
...
           FOR journey in ${this.collection}
             FILTER userJourney.journeyId == journey._key && userJourney.userId == ${user.userId}
               && (userJourney.role == ${UserJourneyRole.owner} || userJourney.role == ${UserJourneyRole.editor})`
 
-    const result = await this.db.query(aql`${roleFilter}
-        ${statusFilter}
-          RETURN journey
-    `)
-    return await result.all()
+    if (roleFilter !== false) {
```
csiyang, line 108: roleFilter has type `false | GeneratedAqlQuery`

storyworks (approved):

PR #806, opened 2022-08-03
feat: template list/card

csiyang:
# Description

- Template List
- Add template to main nav
- Template Card
- Add admin mode for template card (when used for templatesAdmin)
- Loading skeleton
- Template library can be hidden via launch darkly

[- Template Card](https://3.basecamp.com/3105655/buckets/28496234/todos/5147215755)
[- Template List](https://3.basecamp.com/3105655/buckets/28496234/todos/5147216115)
[- Menu nav](https://3.basecamp.com/3105655/buckets/28496234/todos/5147215539)

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [x] Clicking on templates from main nav menu should take you to "Journey Templates"
- [x] Clicking the back button from "Journey Templates" page should take you back to the home page
- [x] Template card should have title, date, description, language and image (from social share image)
- [x] Clicking on card should take you to template details (empty page at the moment)

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged into main

PR #814, opened 2022-08-05
feat: add admin template list

csiyang:
# Description

Please include a summary of the change and which todo is completed. Please also include relevant motivation and context. List any dependencies that are required for this change.

[- Templates status tab panel](https://3.basecamp.com/3105655/buckets/28496234/todos/5147241616)
[- Add button to navigation bar to templates admin](https://3.basecamp.com/3105655/buckets/28496234/todos/5151173647)
- /templates/admin redirects to /templates if current user is not a publisher
- Template List button can be hidden via launch darkly
- Updated template card can be scrolled to when template is duplicated

- [ ] Publisher button should appear in main nav menu and takes you to /publisher
- [ ] Manually visiting /publisher without publisher role should redirect you to /templates
- [ ] Sort options should correctly sort
- [ ] Tabs should show correct templates
- [ ] Click on a card should take you to template details (currently blank)

# Checklist


csiyang:
**Don't merge** until https://github.com/JesusFilm/core/pull/806 is merged

apps/journeys-admin/src/components/TemplateList/TemplateCard/TemplateCard.tsx
```diff
@@ -0,0 +1,224 @@
...
+  refetch?: () => Promise<
+    ApolloQueryResult<
+      | GetActivePublisherTemplates
+      | GetArchivedPublisherTemplates
+      | GetTrashedPublisherTemplates
+    >
+  >
+}
+
+export function TemplateCard({
```
storyworks, line 40: Move TemplateCard out into its own component - it will be used by `TemplateList` and `TemplateLibrary`.

Since we have `JourneyList` - a list of your own journeys, rename `TemplateAdmin` to `TemplateList` - a list of the publishers own templates.

What you now have has `TemplateList` change to `TemplateLibrary`. Keep in mind that we don't need to 100% follow the titles used in the design. Those titles are user-centric - names of our components should be dev-centric and follow the naming conventions we have already in the code base.

apps/journeys-admin/src/components/TemplatesAdmin/TemplateStatusTabs/TemplateStatusTabs.tsx
```diff
@@ -0,0 +1,219 @@
...
+    tabIndex: 1
+  },
+  {
+    queryParam: 'trashed',
+    displayValue: 'Trash',
+    tabIndex: 2
+  }
+]
+
+export function TemplateStatusTabs({
```
storyworks, line 46: This looks and functions almost identical to the journey list display.. in this case in a seperate PR, extract out `StatusTabPanel` and rename to `TemplatesDisplay` (or something that could capture that the component displays all the different types of templates, includes the filtering / sort components and could be changed from list to grid view in the future..).

Update the component to take a prop 'template?: boolean'. `JourneyList` won't pass it, but `TemplateList` will pass in `template`.

If it's a template, pass in the Active/Archived/Trashed templates else pass in Active/Archived/Trashed journeys

apps/journeys-admin/pages/templates/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { NextSeo } from 'next-seo'
```
storyworks, line 1: In the future, `pages/index.ts` will be the dashboard. `pages/journeys/index.ts` will be a user journey list (doesn't yet exist). `pages/templates/index.ts` should be the publishers own templates. 

`pages/library/index.ts` is the list of all public templates and `pages/library/[journeyId].tsx` is the single template view page :)

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
```diff
@@ -65,8 +67,8 @@ describe('NavigationDrawer', () => {
     expect(getByTestId('LeaderboardRoundedIcon')).toHaveStyle(` color: '#fff'`)
   })
 
-  it('should have avatar menu', async () => {
-    const { getByRole, getByText } = render(
+  it('should show publisher button', async () => {
```
storyworks, line 70: Let's clean up these tests a bit now that we've got NavigationListItem.. We don't need to test that the button renders a certain way anymore - just need to know that it shows and this can be done via VR.

So can delete this test and just check that the selection properly works.

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
```diff
@@ -105,12 +120,122 @@ describe('NavigationDrawer', () => {
...
+                email: 'amin@email.com',
+                signOut
+              } as unknown as AuthUser
+            }
+          />
+        </FlagsProvider>
+      </MockedProvider>
+    )
+    await waitFor(() =>
+      expect(getByTestId('ShopTwoRoundedIcon')).toHaveStyle(` color: '#fff'`)
```
storyworks, line 180: Check that a button is selected via it's `aria-selected` prop. Doing it via colour changes leaves tests prone to breaking.

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -14,20 +14,43 @@ import Box from '@mui/material/Box'
...
 import Link from 'next/link'
+import { GetActivePublisherTemplates } from '../../../__generated__/GetActivePublisherTemplates'
+import { GetArchivedPublisherTemplates } from '../../../__generated__/GetArchivedPublisherTemplates'
+import { GetTrashedPublisherTemplates } from '../../../__generated__/GetTrashedPublisherTemplates'
 import { GetPublishedTemplates_journeys as Journey } from '../../../__generated__/GetPublishedTemplates'
 import { JourneyCardMenu } from '../JourneyList/JourneyCard/JourneyCardMenu'
 import { StatusChip } from '../JourneyList/JourneyCard/StatusChip'
 
 export interface TemplateCardProps {
   journey?: Journey
```
storyworks, line 27: Could this be make non-optional? I get that it might be optional if you're using `useJourney` but if you're passing it in, then you're not restrained the same way.

apps/journeys-admin/src/components/TemplateCard/TemplateCard.spec.tsx
```diff
@@ -64,7 +64,7 @@ describe('TemplateCard', () => {
     const { getByText, getByRole } = render(
       <MockedProvider>
         <SnackbarProvider>
-          <TemplateCard journey={defaultTemplate} admin />
+          <TemplateCard journey={defaultTemplate} isPublisher />
         </SnackbarProvider>
       </MockedProvider>
     )
```
storyworks, line 70: For line 72, what is button supposed to be testing? Not clear which button we're talking about..

Could we be more specific about what a publisher only template card looks like? Eg `it should link to publisher template details`? 

I'm not sure about testing that it's draft - since publisher templates can also be "Published" mode. Are you intending to test that the templates are draft on creation?

apps/journeys-admin/pages/publisher/index.tsx
```diff
@@ -8,16 +9,47 @@ import {
...
 import i18nConfig from '../../next-i18next.config'
+import JourneyListMenu from '../../src/components/JourneyList/JourneyListMenu/JourneyListMenu'
+import { GET_USER_ROLE } from '../../src/components/JourneyView/JourneyView'
 
 function TemplateIndex(): ReactElement {
   const AuthUser = useAuthUser()
+  const router = useRouter()
+  const [listEvent, setListEvent] = useState('')
+
+  const handleClick = (event: string): void => {
```
storyworks, line 26: I'm sure I'll discover later when I look down the files but on first glance it's not obvious what ListEvent does or is for... A Menu seemingly would have many onClick events attached to the list items so I'm not sure when that 1 onClick triggers.

[EDIT] - ok I never found out xD

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const Loading = Template.bind({})
+Loading.args = {
+  props: {
+    onLoad: noop,
+    event: ''
+  },
+  mocks: []
+}
+
+export const ArchiveDialog = Template.bind({})
```
storyworks, line 84: Maybe at some point later down the line we can refactor so ArchiveDialog is tested only once since it doesn't really differ visually whether we're dealing with a journey or template.

We could later refactor so the logic of Active/Archive/Delete is separated from the visual component and passed in by a parent or something.

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.tsx
```diff
@@ -0,0 +1,255 @@
...
+            >
+              <Typography variant="subtitle1" align="center" gutterBottom>
+                No templates to display.
+              </Typography>
+            </Card>
+          )}
+        </>
+      ) : (
+        <>
+          <TemplateCard isPublisher />
```
storyworks, line 212: Use `[0,1,2].map` to render TemplateCards. Same with Archive/Trash

apps/journeys-admin/src/components/TemplateList/TrashedTemplates/TrashedTemplates.tsx
```diff
@@ -0,0 +1,261 @@
...
+      case 'deleteAllTrashed':
+        setOpenDeleteAll(true)
+        break
+      case 'refetchTrashed':
+        void refetch()
+        break
+    }
+  }, [event, refetch])
+
+  // calculate 40 days ago. may later be replaced by cron job
```
storyworks, line 171: cron job?

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
csiyang: Reports and templates still had a test for it so thought we still wanted this in

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
csiyang: It doesn't have aira-selected since its a color change

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
csiyang: The journey prop? Its optional because of the query for the journey in the parent component. We most of our components that gets a journey passed in like this...

apps/journeys-admin/src/components/TemplateCard/TemplateCard.spec.tsx
csiyang: i'll change it to be more specific. I don't even remember why "draft" was in there.

apps/journeys-admin/pages/publisher/index.tsx
csiyang: This was something Mike wrote, it was so that the journeys on each tab gets refreshed after an action (archive, trash, delete, restore)

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.stories.tsx
csiyang: Sounds good

apps/journeys-admin/src/components/TemplateList/TrashedTemplates/TrashedTemplates.tsx
csiyang: This is Mike, just kept the comment from `TrashedJourneyList.tsx`

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.tsx
csiyang: Can do!

apps/journeys-admin/pages/publisher/index.tsx
storyworks: So list event does nothing but he just wanted to force change the state of something to trigger a refresh?

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
storyworks: So in the parent component, you can make a check - if journey is not null, then show TemplateCard. 

There are some cases in each component where we still do the journey check, but that's only the case when we use `useJourney` to get the journey.

apps/journeys-admin/pages/publisher/index.tsx
csiyang: Yeah, that happens in `StatusTabPanel.tsx` line 85

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
csiyang: It's also needed to show when the journeys are loading, pretty similar to how journeyCard works

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
storyworks: Yeah, people have copy pasted "it should render" type tests so there are a lot scattered about. Lots aren't needed because we have VR. TBH, I've been evaluating it on a case-by-case basis too.

My current thinking is - for any component, if it's got a couple of different elements on it - do a basic "it should render correctly" type test. Then test only the logic ("it shows X if Y", "it calls X on some action").

In my head, "it shows publisher button" could come under a general "it shows all navigation icons", or if our render test just tests the default - then this test should be "it should show publisher button when user is publisher". 

Sometimes I think it's ok to not even bother with this kind of test in this case because you also have a check for selection, which inadvertently tests that the button exists.

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
storyworks: This is a prompt to add aria-selected={selected} to the NavigationListItem button :) Also just noticed NavigationListItem - change the `handleClick` to `onClick`. We use on___ as the prop name, but handle____ as the function name within a component.

storyworks: Can we also test to see if we can remove the `e?` in the handleClick: (e?)=> void ? Let's just pass in the event if it's explicitly needed

apps/journeys-admin/src/components/TemplateList/TrashedTemplates/TrashedTemplates.tsx
storyworks: xD oh no.. Who reviewed his work - did they ask?

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
csiyang: Thanks for explaining! My thinking was that since there were 3 cases (not a publisher, not selected, selected) I had these tests. Im all for simplifying these down though.

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
storyworks: Ok understood.

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
csiyang: Removed test

apps/journeys-admin/pages/publisher/index.tsx
csiyang: Will rename for the template ones in this PR, and update main in another PR for the journeyList ones

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.tsx
csiyang: Also do this for journeyList ones in a future PR

apps/journeys-admin/src/components/TemplateList/TrashedTemplates/TrashedTemplates.tsx
csiyang: Its for automating the deletion of journeys after 40 days

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
csiyang: Will add aria-selected in a future PR

And the e is necessary since the handleClick can take an event.

apps/journeys-admin/pages/publisher/index.tsx
csiyang: On second look it was handling events for other things too, going to leave the naming as is for now

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.tsx
```diff
@@ -0,0 +1,255 @@
...
+          closeButton: true
+        }}
+        dialogAction={{
+          onSubmit: archiveAll,
+          submitLabel: 'Archive',
+          closeLabel: 'Cancel'
+        }}
+      >
+        <Typography>
+          Are you sure you would like to archive all active templates
```
storyworks, line 232: Could we add translations for these messages? Same with line 205 and 250

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
storyworks: Ok cool, this is just pending the aria-selected check then.

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.tsx
csiyang: Added translations for this PR

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
```diff
@@ -31,16 +33,61 @@ describe('NavigationDrawer', () => {
...
+                signOut
+              } as unknown as AuthUser
+            }
+          />
         </FlagsProvider>
       </MockedProvider>
     )
     expect(getByText('Templates')).toBeInTheDocument()
     expect(getByText('Reports')).toBeInTheDocument()
+    await waitFor(() => expect(getByText('Publisher')).toBeInTheDocument())
```
csiyang, line 90: Added the test for the publisher button here. No longer testing for selected since we're adding aria-selected to NavListItem, the test for selection there

storyworks: Could someone argue that the NaviListItem selected test, tests that the selected logic is on the button. But the selected test here would test that we properly passed the onClick to the Publisher button?

csiyang: Added test for Publisher button selection

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationListItem/NavigationListItem.tsx
```diff
@@ -32,7 +32,6 @@ export function NavigationListItem({
       <ListItemButton
         onClick={handleClick}
         aria-selected={selected}
-        aria-label={label}
```
csiyang, line 35: Removed this since it was redundant

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.spec.tsx
```diff
@@ -105,18 +226,13 @@ describe('NavigationDrawer', () => {
...
               } as unknown as AuthUser
             }
-            title="Journeys"
           />
         </FlagsProvider>
       </MockedProvider>
     )
     await waitFor(() =>
-      expect(getByRole('img', { name: 'Amin One' })).toBeInTheDocument()
+      expect(getByTestId('ShopTwoRoundedIcon')).toHaveStyle(` color: '#fff'`)
```
storyworks, line 234: Not gonna hold up this review but this looks strange too - are we checking that it's selected? Or just that the icon is there?

storyworks (approved):

apps/journeys-admin/src/components/PageWrapper/PageWrapper.stories.tsx
```diff
@@ -37,6 +39,19 @@ const Template: Story = ({ ...args }) => (
             }
           }
         }
+      },
+      {
+        request: {
```
csiyang, line 44: Had to add this so the template admin button shows up in the complete version

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
```diff
@@ -184,7 +184,7 @@ export function NavigationDrawer({
 
         {authUser != null && data?.me != null && (
           <>
-            <Divider sx={{ m: 6, mt: 0, borderColor: 'secondary.main' }} />
+            <Divider sx={{ mb: 2, mx: 6, borderColor: 'secondary.main' }} />
```
csiyang, line 187: Vlad asked for the spacing to be the same around the divider

apps/journeys-admin/src/components/TemplateCard/TemplateCard.stories.tsx
```diff
@@ -25,14 +25,20 @@ Default.args = {
...
+
 export const Complete = Template.bind({})
 Complete.args = {
   template: descriptiveTemplate,
   isPublisher: true
 }
 
-export const Loading = Template.bind({})
-Loading.args = {
+export const CompleteLoading = Template.bind({})
```
csiyang, line 40: We had originally agreed that we only needed one story to test loading, but I wrote a story when tweaking the styling and thought it was different enough to keep it in. But let me know if you want me to remove

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
             maxWidth: '129px',
             minWidth: '129px',
             borderRight: '1px solid',
-            borderColor: 'divider'
+            borderColor: 'divider',
+            backgroundColor: 'background.default'
```
csiyang, line 126: Changed colour to be the same as other similar components (the one used for social share image)

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
-            ? `/${isPublisher === true ? 'publisher' : 'templates'}/${
-                journey.id
-              }`
-            : ''
-        }
-        passHref
+      <Stack
+        direction="column"
+        sx={{
+          width: `calc(100% - 129px)`,
```
csiyang, line 136: Minus space for the card media

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
+                  </Typography>
+                  <Typography variant="body2" noWrap sx={{ fontSize: 12 }}>
+                    {date}
+                    {journey?.description != null
+                      ? ` - ${journey.description}`
+                      : ''}
+                  </Typography>
+                </>
+              ) : (
+                <Box sx={{ height: '44px' }}>
```
csiyang, line 165: The skeleton was changing the height when loading, so this just stops layout shift when the component loads

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
+              justifyContent: 'flex-start',
+              width: '100%'
+            }}
+          >
+            {isPublisher === true && (
+              <Box sx={{ pr: 3, display: 'flex' }}>
+                {journey != null ? (
+                  <StatusChip status={journey.status} />
+                ) : (
+                  <Stack direction="row" alignItems="center" spacing={1.5}>
```
csiyang, line 191: Added this stack to mimic how StatusChip is styled. Might be good in the future if Status chip can handle its own loading skeleton though

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
+            </CardContent>
+          </CardActionArea>
+        </Link>
+
+        <CardActions>
+          <Stack
+            direction="row"
+            alignItems="center"
+            sx={{
+              mt: isPublisher !== true ? 2 : 0,
```
csiyang, line 179: The menu MoreVertIcon doesn't render when its not publisher, which was causing layout shift for CardActions

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
+              mx: 2,
+              display: 'flex',
+              justifyContent: 'flex-start',
+              width: '100%'
```
csiyang, line 183: Sets the width for the `noWrap` property to work properly on the language

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
-                <Typography variant="subtitle1" noWrap>
-                  {journey.title}
-                </Typography>
-                <Typography variant="body2" noWrap sx={{ pb: 4, fontSize: 12 }}>
-                  {date}
-                  {journey?.description != null
-                    ? ` - ${journey.description}`
-                    : ''}
+                <TranslateRoundedIcon sx={{ fontSize: 13, pl: 0 }} />
+                <Typography variant="body2" noWrap sx={{ ml: 1 }}>
```
csiyang, line 202: Using the spacing prop on the parent `Stack` was messing with the spacing between the icon and text, so I removed it from the stack and let each component handle its own spacing

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
-                  >
-                    <TranslateRoundedIcon sx={{ fontSize: '14px', mr: 1 }} />
-                    {displayLanguage}
-                  </Typography>
-                </Stack>
               </>
             ) : (
               <>
-                <Skeleton variant="text" width={120} />
+                <TranslateRoundedIcon sx={{ fontSize: 13 }} />
```
csiyang, line 208: To match the icon size used in the actual component

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
                 <Skeleton
                   variant="text"
-                  sx={{
-                    display: 'flex',
-                    justifySelf: 'flex-end',
-                    mr: '20%',
-                    mb: 4
-                  }}
+                  width={50}
+                  height={20}
```
csiyang, line 212: The default height was causing layout shift when loaded

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -121,113 +122,122 @@ export function TemplateCard({
...
-            />
-          ) : (
-            <>
-              <IconButton disabled>
-                <MoreVertIcon />
-              </IconButton>
-            </>
-          )}
+            {isPublisher === true && (
+              <Box sx={{ display: 'flex', ml: 'auto' }}>
```
csiyang, line 219: `ml: 'auto'` was the magic line that fixed everything. Also found that `Box` doesn't inherit `display: 'flex'` from the parent

apps/journeys-admin/src/components/TemplateList/ActiveTemplates/ActiveTemplates.stories.tsx
```diff
@@ -1,24 +1,36 @@
...
 const ActiveTemplatesStory = {
-  ...journeysAdminConfig,
   component: ActiveTemplates,
   title: 'Journeys-Admin/TemplatesList/ActiveTemplates',
   parameters: {
     ...journeysAdminConfig.parameters,
     layout: 'fullscreen'
-  }
+  },
+  decorators: [
```
csiyang, line 23: Stories with modals were getting cut short. See [FAQ](https://www.chromatic.com/docs/faq)`Why are my components that render in a portal getting cut off?`

Also removed line 15, since both `decorators` and `parameters` where getting overwritten

apps/journeys-admin/src/components/TemplateList/ArchivedTemplates/ArchivedTemplates.tsx
```diff
@@ -187,7 +187,7 @@ export function ArchivedTemplates({
               <Box width="100%" sx={{ textAlign: 'center' }}>
                 <Typography variant="caption">
                   {t(
-                    'Archived journeys are hidden from your active templates list for better organization.'
+                    'Archived templates are delisted from the Template Library.'
```
csiyang, line 190: Vlad asked for this change

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
```diff
@@ -171,13 +171,12 @@ export function TemplateCard({
           </CardActionArea>
         </Link>
 
-        <CardActions>
+        <CardActions sx={{ px: 4, py: 2 }}>
```
csiyang, line 174: When not set the default padding on the right was getting pushed out by the child components in `CardActions`

apps/journeys-admin/src/components/TemplateCard/TemplateCard.tsx
storyworks: Good idea

apps/journeys-admin/src/components/TemplateCard/TemplateCard.stories.tsx
storyworks: Talking to Tatai, we should prioritise the TextField work now :o so I reckon, so far the pattern has been to keep to 1 loading state. Let's continue to just go with this - but use "CompleteLoading" renamed to "Loading".

Do this in another PR when you have time after finishing TextField stuff. Also look into why the loading image is not a blurred image but the fallback image.

Consider this PR otherwise approved :)

PR #826, opened 2022-08-24
fix: publish only available to journey owners and publishers if template

storyworks:
# Description

Currently publish was disabled when it was already a published journey. Add owner / publisher if template check

[Basecamp Todo](https://3.basecamp.com/3105655/buckets/28980303/todos/5258086196)

- [ ] Be an editor, see that you cannot publish a draft journey
- [ ] Be an owner, see that you can publish draft journeys
- [ ] Be a publisher, see that you can publish draft templates (Only testable later on when template details page is made)
- [ ] Any role, cannot publish already published journeys.

# Checklist


apps/journeys-admin/src/components/MenuItem/MenuItem.tsx
```diff
@@ -0,0 +1,39 @@
...
+  icon,
+  disabled,
+  openInNew,
+  onClick,
+  // implicitly pass in href
+  ...menuItemProps
+}: MenuItemProps): ReactElement {
+  const newTabProps =
+    openInNew === true
+      ? // TODO: Doesn't redirect to new page if menuitem element bug
```
GeronimoJohn, line 25: Not quite sure what the comments mean here

storyworks: The html element MuiMenuItem renders down to is `menuitem`, this takes an href but doesn't support the `target` attribute. The below fixes this by turning the menuitem into an anchor `a` element. But maybe there's another way to solve - this was just a quick fix.

GeronimoJohn (approved):

PR #841, opened 2022-09-01
fix: journeylink to be used in journeyview

GeronimoJohn:
# Description

Enabled JourneyLink component to be reused. Currently being reused in JourneyView. This was done in order to get a cleaner version of the JourneyView component

[- Link to Basecamp Todo](url)

- [ ] JourneyLink should render the Journey URL, Edit URL button, and EmbedJourney Button
- [ ] Copying the JourneyURL should work as expected
- [ ] Changing the JourneyURL should work as expected
- [ ] Clicking on the Embed Journey should work as expected

# Checklist


apps/journeys-admin/src/components/JourneyView/JourneyLink/JourneyLink.tsx
```diff
@@ -30,6 +37,16 @@ export function JourneyLink(): ReactElement {
               }/${journey.slug}`
             : undefined
         }
+        label={!smUp ? t('Journey URL') : undefined}
```
storyworks, line 40: Can we please double check this logic with UX or keep this logic exactly the same as the logic extracted from `JourneyView` ?

GeronimoJohn: Doing this allows us to retain the current functionality of JourneyLink on production. Please see the screenshots below of the production and storybook. (Storybook referring to the change mentioned above)

**Production**
Desktop Production
Mobile Production

**Storybook**
Desktop Storybook
Mobile Storybook

@lumberman Please let us know if these images or changes stay true to the current functionality in production

storyworks: It's quite uncommon to display labels based on device size - it's unfortunate there's no good server-side rendering compatible way to use this component to achieve such an effect. If this is what's in production now, I will pick my battles and save a discussion about labelling for another day xD

storyworks (approved):

PR #849, opened 2022-09-01
feat: journey view fab

GeronimoJohn:
# Description

Created the JourneyViewFab component. This component is made to be a common component under JourneyView. Rendering different buttons depending on the user and journey (either a journey or a template)

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29147486/todos/5294634630)

JourneyView
- [ ] If it's just a journey, users should see the Edit button
- [ ] If it's a template and the user is not a publisher, users see the UseTemplate button
- [ ] If it's a template and the user is a publisher, users should see the Edit button 

# Checklist


apps/journeys-admin/src/components/JourneyView/JourneyViewFab/JourneyViewFab.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { useJourney } from '@core/journeys/ui/JourneyProvider'
+import Fab from '@mui/material/Fab'
+import EditIcon from '@mui/icons-material/Edit'
+import CheckRoundedIcon from '@mui/icons-material/CheckRounded'
+import NextLink from 'next/link'
+import { useRouter } from 'next/router'
+import { ConvertTemplate } from '../../../../__generated__/ConvertTemplate'
+
+export const CONVERT_TEMPLATE = gql`
+  mutation ConvertTemplate($id: ID!) {
```
storyworks, line 12: I think our convention is snakeCase mutation names

apps/journeys-admin/src/components/JourneyView/JourneyViewFab/JourneyViewFab.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export function JourneyViewFab({
+  isPublisher
+}: JourneyViewFabProps): ReactElement {
+  const { journey } = useJourney()
+  const router = useRouter()
+  const [ConvertTemplate] = useMutation<ConvertTemplate>(CONVERT_TEMPLATE)
+
+  const smUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('sm'))
```
GeronimoJohn, line 32: Ended up using useMediaQuery for a mobile view check. I've tried looking into the [alternatives](https://mui.com/material-ui/react-use-media-query/#server-side-rendering). But none of them really talks about displaying text on specific mobile views. Well unless we install another library.

GeronimoJohn: Another solution is to keep the text 'Use Template' or just an icon. But we need to have a discussion with Vlad about it.

storyworks (approved):

PR #854, opened 2022-09-02
feat: templates page

GeronimoJohn:
# Description

This 'templates page' will be the base PR for our template components. This should either be merged first to the features/templates branch or we slowly merged the template components to this.

- Link to Basecamp Todo

This is just the single template page layout

# Checklist


apps/journeys-admin/src/components/JourneyView/JourneyView.tsx
```diff
@@ -47,12 +41,24 @@ export function JourneyView(): ReactElement {
   const [showSlugDialog, setShowSlugDialog] = useState(false)
   const [showEmbedDialog, setShowEmbedDialog] = useState(false)
 
+  const smUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('sm'))
+
   return (
     <Box sx={{ mr: { sm: '328px' }, mb: '80px' }}>
-      <TitleDescription isPublisher={isPublisher} />
+      <Stack
```
storyworks, line 48: Instead of using Stack smUp, you could use Box with responsive `flexDirection`

apps/journeys-admin/src/components/JourneyView/JourneyView.tsx
```diff
@@ -47,12 +41,24 @@ export function JourneyView(): ReactElement {
...
+        {/* if template: SocialImage */}
+        <Stack direction="column" spacing={6} sx={{ width: '100%' }}>
+          {/* if template: DatePreview */}
+          <TitleDescription isPublisher={isPublisher} />
+        </Stack>
+      </Stack>
       <Properties />
 
-      {reports && journey != null && (
+      {reports && journey != null && journey.template !== true && (
```
storyworks, line 61: Maybe could simplify this into a `showReports` variable?

apps/journeys-admin/src/components/JourneyView/JourneyView.spec.tsx
```diff
@@ -55,23 +55,6 @@ describe('JourneyView', () => {
     userJourneys: [],
     template: null
   }
-  it('should have edit button', () => {
```
storyworks, line 58: Could skip the test for now instead of removing?

apps/journeys-admin/src/components/JourneyView/JourneyView.tsx
GeronimoJohn: Line 75 of JourneyView still needs the journey null check at line 66. It's not happy with me passing the journey null check as a part of the variable

Though I found this to work
`const showReports = reports && journey?.template`
`showReports && journey != null`

storyworks (approved):

PR #915, opened 2022-09-23
fix: access templates

storyworks:
# Description

If trying to access template admin or library as a new user (with no journeys), display the following error page:

image

[Basecamp](https://3.basecamp.com/3105655/buckets/29147486/todos/5358874874)

New user (no journeys):
- [x] If they go to templates library, they will see access denied message
- [x] Cannot go to templates admin (redirected)
- [x] If they go to journey list, they will see access denied message

For a new publisher (no journeys):
- [ ] If they go templates library, they see the library.
- [ ] If they go to templates admin, they see the templates list
- [ ] If they go to their journey list ?? (currently they'll see an access denied message)

# Checklist


apps/journeys-admin/src/components/TemplateList/TemplateList.tsx
```diff
@@ -1,22 +1,31 @@
...
-  event: string | undefined
+  journeys?: Journey[]
   router?: NextRouter
+  event: string | undefined
   authUser?: AuthUser
 }
 
 export function TemplateList({
-  event,
+  journeys,
```
csiyang, line 26: All the changes on this file is not needed. This parent page of this will automatically redirect if not publisher.

apps/journeys-admin/src/components/TemplateLibrary/TemplateLibrary.tsx
```diff
@@ -1,35 +1,76 @@
...
+              </>
+            )}
+          </Container>
+        ))}
+      {journeys?.length === 0 && (
+        <Container maxWidth="sm" sx={{ mt: 20 }}>
+          <Stack direction="column" spacing={8} alignItems="center">
+            <NewReleasesRounded sx={{ fontSize: 60 }} />
+            <Typography variant="h1" align="center">
+              You need to be invited to create the first template
```
csiyang, line 53:
```suggestion
              You need to be invited to use your first template
```

apps/journeys-admin/src/components/TemplateLibrary/TemplateLibrary.stories.tsx
```diff
@@ -13,18 +13,23 @@ const TemplateLibraryStory = {
...
 }
 
-const Template: Story = ({ ...args }) => (
-  <TemplateLibrary journeys={args.templates} />
-)
+const Template: Story = ({ ...args }) => <TemplateLibrary {...args} />
 
 export const Default = Template.bind({})
 Default.args = {
+  journeys: [{}],
```
csiyang, line 20: why are the curlies needed?
```suggestion
  journeys: [],
```

csiyang (changes requested):

apps/journeys-admin/src/components/TemplateList/TemplateList.tsx
storyworks: This is a publisher, but a publisher who doesn't have any journeys

storyworks: It's a different kind of error you're thinking about

apps/journeys-admin/src/components/TemplateLibrary/TemplateLibrary.stories.tsx
storyworks: Need to pass in something for journeys so we know they have access

apps/journeys-admin/src/components/TemplateList/TemplateList.tsx
csiyang: Oh was it asked to also restrict publishers to also have a personal journey first. I was thinking since all publishers share templates, they may never have their own journey, so thought it was weird we would restrict template access with personal journies?

storyworks: That makes sense, put the push back in the templates thread?

jianwei1:
> # How should this PR be QA Tested?
> New user (no journeys):
> 
> * [x]  If they go to templates library, they will see access denied message
> * [x]  Cannot go to templates admin (redirected)
> * [x]  If they go to journey list, they will see access denied message
> 

Have tested the new user checks. All pass but I did notice that on iphone/Ipad the templates admin will display for a brief second before being redirected. I've made a comment (with video) on the [to do](https://3.basecamp.com/3105655/buckets/29494894/todos/5369802313#__recording_5383394118)

csiyang (approved):

apps/journeys-admin/src/components/TemplateLibrary/TemplateLibrary.tsx
```diff
@@ -1,35 +1,58 @@
...
 export function TemplateLibrary({
-  journeys
+  isPublisher = false,
+  journeys,
+  templates
 }: TemplateLibraryProps): ReactElement {
+  const { t } = useTranslation('apps-journeys-admin')
+
+  // journey == null is journey loading
+  const showLibrary = journeys == null || journeys?.length > 0 || isPublisher
```
csiyang, line 23: This is the bit thats causing the flash. Can we move this check out into its own check? 

```suggestion
  const showLibrary = journeys == null || journeys?.length > 0
```

csiyang (changes requested):
Don't mind this, just a comment to request change

apps/journeys-admin/src/components/TemplateLibrary/TemplateLibrary.tsx
csiyang: oops this is not the problem

csiyang (approved):

PR #933, opened 2022-10-03
test: update journey-ui vr

storyworks:
# Description

Removing a bunch of todos in stories + fixing others

[Fix TextResponse story](https://3.basecamp.com/3105655/buckets/29494894/todos/5369620172)

- [ ] Checks are green
- [ ] New VR standards agreed on by devs (will link to docs)

# Checklist


libs/journeys/ui/src/components/GridContainer/GridContainer.stories.tsx
```diff
@@ -20,12 +20,13 @@ import { GridContainer } from './GridContainer'
 const Demo = {
   ...journeyUiConfig,
   component: GridContainer,
-  title: 'Journeys-Ui/Grid/Container'
+  title: 'Journeys-Ui/Grid/Container',
+  parameters: {
+    chromatic: { disable: true }
```
storyworks, line 25: Disabled Grid stories since we aren't using this block, also renaming all templates to "Template" by default.

libs/journeys/ui/src/components/Image/Image.stories.tsx
```diff
@@ -11,30 +11,31 @@ const Demo = {
...
-  DefaultTemplate.bind({})
-NoImageSource.args = {
-  ...Default.args,
+const Template: Story<TreeBlock<ImageFields>> = ({ alt, ...props }) => (
+  <StoryCard>
+    <Image {...imageProps} {...props} alt={alt} />
+  </StoryCard>
+)
+
+export const Default: Story<TreeBlock<ImageFields>> = Template.bind({})
```
storyworks, line 29: The actual default image block is the empty source one. Have updated this in image and video. 

This makes more sense as we will have different sources of image / video in the future.

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.stories.tsx
```diff
@@ -129,30 +104,20 @@ const DefaultTemplate: Story<TreeBlock<RadioQuestionFields>> = ({
...
-  DefaultTemplate.bind({})
+export const Default: Story<TreeBlock<RadioQuestionFields>> = Template.bind({})
 Default.args = {
   id: 'Default',
   children,
   parentOrder: 1,
   parentBlockId: 'Step1'
 }
 
-export const Long: Story<TreeBlock<RadioQuestionFields>> = DefaultTemplate.bind(
```
storyworks, line 149: Don't need a `long` example anymore - RadioQuestion is just a group of options. The `Long` scenario is tested there.

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
```diff
@@ -73,20 +88,32 @@ const Template: Story<TreeBlock<SignUpFields>> = ({
...
+            />
+          )}
+        </StoryCard>
+      </SnackbarProvider>
+    </JourneyProvider>
   </MockedProvider>
 )
 
 export const Default = Template.bind({})
+Default.args = { ...signUpProps }
```
storyworks, line 110: This could be directly added to the SignUp component in line 95 - but not sure if people had a preference to put it here or there.

Putting it on line 95 is less lines of code but requires people to know how props override one another + look closer into the template to see the default build.

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
```diff
@@ -73,20 +88,32 @@ const Template: Story<TreeBlock<SignUpFields>> = ({
...
 
-export const CustomButton = Template.bind({})
-CustomButton.args = {
+export const Complete = Template.bind({})
```
storyworks, line 112: The complete example for forms should now include all the customisable options featured.
Also it has a `complete` flag which toggles on blocks which show how it sits in relation to others. 

Could have put the other blocks on the default story too, but wondered if that distracts / makes it confusing which is the actual block

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
```diff
@@ -120,5 +150,17 @@ const LoadingTemplate: Story<TreeBlock<SignUpFields>> = (): ReactElement => (
 )
 
 export const Loading = LoadingTemplate.bind({})
+Loading.play = () => {
+  const submitButtons = screen.getAllByRole('button')
+  const fields = screen.getAllByRole('textbox')
```
storyworks, line 155: For some strange reason, the name of the first Name field was "Name Name" and the second was "". Same for the 2 email fields. So referencing like this for now...

libs/journeys/ui/src/components/TextField/TextField.stories.tsx
```diff
@@ -39,11 +40,19 @@ const Template: Story<TextFieldProps> = () => (
       {() => (
         <Form>
           <Stack>
-            <TextField name="default" label="Default" />
+            <Stack direction="row" spacing={4}>
```
storyworks, line 43: Stacking these examples side by side as multiple people have now confused the default textfield as the textfield with placeholder (rather than an unfocused textfield with label)

apps/journeys-admin/src/components/Editor/Canvas/InlineEditWrapper/ButtonEdit/ButtonEdit.stories.tsx
```diff
@@ -163,14 +163,4 @@ Default.args = {
   selectedBlock: contained
 }
 
-export const Filled = Template.bind({})
```
storyworks, line 166: Button is no different whatever is in it and screenshot captured looks the same.. (since we had to hide the cursor due to blinking) Possibly could still keep the text variant as viewing it in storybook could show something different.

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/Attribute/Attribute.stories.tsx
```diff
@@ -24,18 +24,4 @@ export const Default: Story = () => {
   )
 }
 
-export const Selected: Story = () => {
```
storyworks, line 27: This test done in Button.tsx - not needed here

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/TextResponse/TextField/TextField.spec.tsx
```diff
@@ -1,9 +0,0 @@
-import { render } from '@testing-library/react'
-import { TextField } from './TextField'
-
-describe('TextField', () => {
```
storyworks, line 4: TextField was the original block for TextResponse - it's unused, our mistake for not deleting earlier

apps/journeys-admin/src/components/Editor/ControlPanel/BlocksTab/NewButtonButton/NewButtonButton.stories.tsx
```diff
@@ -1,140 +0,0 @@
-import { MockedProvider } from '@apollo/client/testing'
```
storyworks, line 1: All these "New___Button" snaps are captured in the "Add blocks" panel story. There aren't different variants of this button, and if there were - it would be captured in "Button" stories.

libs/journeys/ui/src/components/GridContainer/GridContainer.stories.tsx
```diff
@@ -154,8 +154,9 @@ Default.args = {
   ]
 }
 
-export const OffsetRight: Story<TreeBlock<GridContainerFields>> =
-  DefaultTemplate.bind({})
+export const OffsetRight: Story<TreeBlock<GridContainerFields>> = Template.bind(
+  {}
+)
```
csiyang, line 159:
```suggestion
export const OffsetRight: Story<TreeBlock<GridContainerFields>> = Template.bind({})
```

libs/journeys/ui/src/components/Image/Image.stories.tsx
```diff
@@ -11,30 +11,29 @@ const Demo = {
...
-  chromatic: { delay: 300 }
-}
 
-export const NoImageSource: Story<TreeBlock<ImageFields>> =
-  DefaultTemplate.bind({})
-NoImageSource.args = {
-  ...Default.args,
+const Template: Story<TreeBlock<ImageFields>> = ({ alt, ...args }) => (
+  <StoryCard>
+    <Image {...imageProps} {...args} alt={alt} />
```
csiyang, line 25: Why does the alt prop need to be seperated?

```suggestion
    <Image {...imageProps} {...args} />
```

libs/journeys/ui/src/components/Image/Image.stories.tsx
```diff
@@ -11,30 +11,29 @@ const Demo = {
...
+  </StoryCard>
+)
+
+// Throttle network to see loading image
+export const Default: Story<TreeBlock<ImageFields>> = Template.bind({})
+Default.args = {
   src: null
 }
```
csiyang, line 33: Can we keep it so each template passes all of its own props? I feel like this is more clear.
```suggestion
Default.args = {
  ...ImageProps,
  src: null
}
```

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.stories.tsx
```diff
@@ -118,41 +93,23 @@ const DefaultTemplate: Story<TreeBlock<RadioQuestionFields>> = ({
   >
     <StoryCard>
       <Typography {...typographyProps} />
-      {props.id === 'Long' && (
-        <Typography
-          {...typographyProps}
-          content="Have you declared that you want to accept Jesus in your life as your Lord and savior?"
-          variant={TypographyVariant.body2}
-          parentOrder={1}
-        />
```
csiyang, line 127: I think we should add this back so the story can match what a radio question block looks like

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
```diff
@@ -73,20 +88,31 @@ const Template: Story<TreeBlock<SignUpFields>> = ({
       }
     ]}
   >
-    <SnackbarProvider>
-      <StoryCard>
-        <SignUp {...signUpProps} {...props} uuid={() => 'uuid'} />
-      </StoryCard>
-    </SnackbarProvider>
+    <JourneyProvider>
```
csiyang, line 91: How come journeyProvider is needed now?

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
csiyang: Why do some blocks have surrounding typogs to show how the block fits and some others don't? 

Was kinda confused at first why they were there. Could we add a comment somewhere in the story to say why we have the typogs?

Also could we rename the complete prop to something else? Might get confusing since we're starting to call a lot of our stories Complete too. Maybe something like `inContext`?

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
```diff
@@ -27,6 +29,18 @@ const Demo = {
   }
 }
 
+const typographyProps: TreeBlock = {
```
csiyang, line 32:
```suggestion
const typographyProps: TreeBlock<TypographyBlock> = {
```

csiyang (changes requested):

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
storyworks: `Complete` is meant to show the component with all the customisable options changed.

I thought it'd be good to add context related story to these texts to test out margins. 
I added it to complete since I thought it made sense that, that story would have all the bells and whistles. I could be it's own story, but if it made sense to others - we could combine.

libs/journeys/ui/src/components/Image/Image.stories.tsx
storyworks: image

We get this error without it..

libs/journeys/ui/src/components/Image/Image.stories.tsx
storyworks: Can do - I was wondering about this.

Would you want this convention even if it meant adding in an extra line in some cases?
eg doing:
```
Default.args = { ...mockedProps }
```

Whereas if the basic mocked data was already in the template, you wouldn't need the second line?

libs/journeys/ui/src/components/Image/Image.stories.tsx
```diff
@@ -11,30 +11,29 @@ const Demo = {
...
 
+export const WebImage: Story<TreeBlock<ImageFields>> = Template.bind({})
+WebImage.parameters = {
+  chromatic: { delay: 300 }
```
storyworks, line 37: Oops, will change this - no longer need the delay

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.stories.tsx
storyworks: This is another question to raise - I removed it so it's agnostic / we focus on the margins and how the block interacts with what's above and below it. When we first started building this - I thought it might be good for Storybook to also document how we use blocks and be documentation for journey-builders.

If we were still able to do that - I reckon those stories could be omitted from Chromatic screenshots? And just in for documentation "real life example" purposes?

libs/journeys/ui/src/components/SignUp/SignUp.stories.tsx
storyworks: Will double test and get back

libs/journeys/ui/src/components/Image/Image.stories.tsx
storyworks: @csiyang will do, and explained https://jfp-digital.slack.com/archives/C041QM21FDL/p1665025846147379?thread_ts=1664925026.334719&cid=C041QM21FDL

libs/journeys/ui/src/components/GridContainer/GridContainer.stories.tsx
storyworks: This is cut off from prettier... so won't change.

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.stories.tsx
storyworks: @csiyang , double checked the code - I didn't actually remove the question. It now reads "How can we help you know more about Jesus?" So still gives context to how RadioQuestion block might be used.

libs/journeys/ui/src/components/Button/Button.stories.tsx
```diff
@@ -20,22 +18,33 @@ const ButtonDemo = {
...
-}
-
-const Template: Story<ButtonStoryProps> = ({ ...args }) => (
-  <StoryCard>
-    <MockedProvider>
-      <Button {...args} id="id" />
-    </MockedProvider>
-  </StoryCard>
+const Template: Story<
+  Parameters<typeof Button>[0] & { variants: ButtonVariant[] }
```
storyworks, line 22: Use `Story` when we need to extend the base props `Parameters<typeof Button>[0]` with something else (in this case `variants`)

libs/journeys/ui/src/components/Card/Card.stories.tsx
```diff
@@ -151,52 +150,52 @@ const video: TreeBlock<VideoFields> = {
...
+export const Default = Template.bind({})
 Default.args = {
   themeMode: null,
   themeName: null,
   children: content
 }
 
 // Expanded - override content background
-export const Custom: Story<TreeBlock<CardFields>> = Template.bind({})
+export const Custom = Template.bind({})
```
storyworks, line 161: Don't need to provide type here - typescript can infer it from Template.

Keeping it would be the same as writing
`const myName: string = "Ziwei"`
`const name: string = myName.toUpperCase()`

The second line is not needed.

GeronimoJohn (approved):

GeronimoJohn (approved):

tataihono (approved):

PR #944, opened 2022-10-10
feat: rtl language check

storyworks:
# Description

Helper function to check that a locale is RTL.

[Basecamp Todo](https://3.basecamp.com/3105655/buckets/29688713/todos/5413370739)

See unit tests but in general:

- [x] it should be false for non-rtl locale
- [x] it should be false for numeric string
- [x] it should be false for empty string
- [x] it should be false for string which includes rtl locale substring (not separated by "-" separator)
- [x] it should be true for any of the codes above
- [x] it should be true for any of the code above in uppercase
- [x] it should be true for any of the codes above with suffix of "-" 

# Checklist

### DIFF
--- a/libs/shared/ui/src/libs/rtl/rtl.ts
+++ b/libs/shared/ui/src/libs/rtl/rtl.ts
@@ -1,0 +1,22 @@
+export function isRtl(locale: string): boolean {
+  // locale should be bcp47, else iso3
+  const formattedLocale = locale.split('-')[0].toLowerCase()
+
+  switch (formattedLocale) {
+    case 'ar':
+    case 'arc':
+    case 'dv':
+    case 'fa':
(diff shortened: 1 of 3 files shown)


libs/shared/ui/src/libs/rtl/rtl.ts
```diff
@@ -0,0 +1,22 @@
+export function isRtl(locale: string): boolean {
+  // locale should be bcp47, else iso3
+  const formattedLocale = locale.split('-')[0].toLowerCase()
```
storyworks, line 3: In api-languages all languages that don't have bcp47 have at least iso3

image

bcp47 uses "-" to seperate language from region 
image

GeronimoJohn (approved):

PR #955, opened 2022-10-12
fix: add rtl to stories

GeronimoJohn:
# Description

Add RTL logic to the theme provider. By doing so we're fixing the missing configuration for the set up of MUI RTL. And we're also setting up the Stories to be supporting RTL.

A continuation/fix for this todo. It's missing the RTL config on the ThemeProvider component
[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29688713/todos/5413044377)

- [ ] Default theme to include an RTL story

After running the storybook server. And in any Journey Blocks stories within Journeys UI
1. Create a new story with a proper RTL description
2. Set rtl to be true on the new story parameters e.g. `RightToLeftButton.parameters = { rtl: true}`
3. Check story on storybook

- [ ] Created story to be in RTL

# Checklist


GeronimoJohn:
An example of how it can be used in the story once the right props have been set. 
Screen Shot 2022-10-13 at 9 47 39 AM

libs/shared/ui/src/libs/sharedUiConfig/sharedUiConfig.tsx
```diff
@@ -9,6 +9,7 @@ export const sharedUiConfig = {
       <ThemeDecorator
         mode={context.parameters.theme ?? context.globals.theme}
         layout={context.parameters.layout}
+        isRTL={context.args.isRTL}
```
storyworks, line 12: Let's add this to parameters instead

libs/shared/ui/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -24,12 +25,15 @@ export const ThemeProvider = ({
   themeName,
   themeMode,
   children,
+  rtl,
```
storyworks, line 28: `rtl = false`

Then remove the `rtl === true` check

libs/shared/ui/src/components/ThemeDecorator/ThemeDecorator.tsx
```diff
@@ -45,42 +55,51 @@ const ThemeContainer = ({
...
-          }}
-        >
-          <ThemeContainer mode="light" layout={layout}>
-            {children}
-          </ThemeContainer>
-          <ThemeContainer mode="dark" layout={layout}>
-            {children}
-          </ThemeContainer>
-        </div>
+        <CacheProvider value={emotionCache}>
```
storyworks, line 65: Move the CacheProvider inside the div. Same with line 87. 

We want CacheProvider only to influence the elements we're testing, not the div layout components.

libs/shared/ui/src/components/ThemeDecorator/ThemeDecorator.tsx
```diff
@@ -13,12 +16,18 @@ const themeMode = globalTypes.theme.toolbar.items
 interface ThemeDecoratorProps extends Pick<Parameters, 'layout'> {
   mode: typeof themeMode[number]
   children: ReactNode
+  emotionCache?: EmotionCache
+  isRTL?: boolean
```
storyworks, line 20: Also let's change this to `rtl` instead of isRTL. I know we have some props in other places called `is___` but let's stick to just the state in props. And use `is___` within components as variable names if needed.

libs/shared/ui/src/components/ThemeDecorator/ThemeDecorator.tsx
```diff
@@ -58,13 +70,16 @@ export const ThemeDecorator = ({
             margin: layout === 'fullscreen' ? '0px' : '-16px',
             overflowX: 'hidden'
           }}
+          dir={rtl === true ? 'rtl' : 'ltr'}
```
storyworks, line 73: Same thing as in other file - make rtl=false by default on line 59 to remove === true check

storyworks (approved):

libs/shared/ui/src/libs/themes/base/theme.stories.tsx
```diff
@@ -402,4 +406,69 @@ Viewport.parameters = {
   }
 }
 
+export const StyledButton = styled(Button)(({ theme }) => ({
```
storyworks, line 409: No need to export Styled button.

PR #961, opened 2022-10-13
fix: storybook emotion cache

storyworks:
# Description

Allows us to toggle between RTL and LTR stories in Storybook and the styling stays correct

[Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/29688713/todos/5413044377)

- [ ] Toggle between conductor RTL and LTR stories - it should be correctly styled
- [ ] Staging journeys should remain with correct styling when going from RTL to LTR journeys. Can use https://your-stage.nextstep.is/rtl-journey-demo-2 journeys or your own.

# Checklist


libs/shared/ui/src/components/ThemeDecorator/ThemeDecorator.tsx
```diff
@@ -16,13 +15,9 @@ const themeMode = globalTypes.theme.toolbar.items
 interface ThemeDecoratorProps extends Pick<Parameters, 'layout'> {
   mode: typeof themeMode[number]
   children: ReactNode
-  emotionCache?: EmotionCache
```
storyworks, line 19: Remove since I don't think we need a different cache for each story decorator..

GeronimoJohn (approved):

PR #962, opened 2022-10-13
feat: remove arango from k8s

mikeallisonJS:
# Description

Removes arangodb from kubernetes.
Note: already run


mikeallisonJS:
bad pr, wrong base

PR #965, opened 2022-10-14
feat: multichoice rtl

GeronimoJohn:
# Description

Confirm that the Right to Left (RTL) functionalities are being applied to the radio question and radio option component or the multichoice component. This can be done or checked by adding a RTL visual test for the multichoice component

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29688713/todos/5413251191)

Visit the storybook link: https://612c2a83fdc2b2003a5c2eb7-jtvqfltbiq.chromatic.com/?path=/story/journeys-ui-radioquestion--rtl
- [x] it displays RTL text when the journey is RTL
- [x] it displays the checkboxes on the right when the journey is RTL
- [ ] it uses the RTL font style in RTL journeys for labels
- [x] it doesn't break the style of a radio question/radio option when changing between LTR (LeftToRight) to RTL stories
- [x] it doesn't have changes to any of the LTR story

# Checklist


libs/journeys/ui/src/components/RadioOption/RadioOption.stories.tsx
```diff
@@ -47,4 +47,13 @@ Disabled.args = {
   disabled: true
 }
 
+export const RTL = Template.bind({})
+RTL.args = {
+  id: 'NestedOptions',
```
storyworks, line 52: You can replace lines 52-55 with `...Default.args,` Let's use args composition where possible. https://storybook.js.org/docs/react/writing-stories/args#args-composition

Same with RadioQuestion.stories

storyworks (approved):

PR #969, opened 2022-10-14
feat: image rtl

GeronimoJohn:
# Description

Confirm that the Right to Left (RTL) functionalities are being applied to the Image component. This can be done or checked by adding an RTL visual test for the Image component

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29688713/todos/5413262233)

Visit the storybook link: 

- [ ] it displays image as mirrored or flipped
- [ ] it doesn't have changes to any of the LTR story

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.stories.tsx
+++ b/apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.stories.tsx
@@ -620,7 +620,20 @@ const Template: Story = (args) => {
           journey: {
             id: 'journeyId',
             themeMode: ThemeMode.dark,
-            themeName: ThemeName.base
+            themeName: ThemeName.base,
+            language: {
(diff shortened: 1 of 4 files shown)


apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.stories.tsx
```diff
@@ -620,7 +620,20 @@ const Template: Story = (args) => {
           journey: {
             id: 'journeyId',
             themeMode: ThemeMode.dark,
-            themeName: ThemeName.base
+            themeName: ThemeName.base,
+            language: {
```
storyworks, line 624: I ended up finding myself applying this same change - had a case in admin where I needed to get the Journey theme of the journey we're editing. So I had to use `getJourneyRTL` since `useTheme()` would have given me admin theme.

PR #976, opened 2022-10-17
chore: update events table query

csiyang:
# Description

We need to update some fields in the events table due to changes being made for visitor activity. DONOT merge this, its just for updating the database. Close after query is ran

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29689022/todos/5426570068)

- [ ] Test A
- [ ] Test B

# Checklist

### DIFF
--- a/apps/api-journeys/db/populateNewFields.ts
+++ b/apps/api-journeys/db/populateNewFields.ts
@@ -1,0 +1,25 @@
+import { aql } from 'arangojs'
+import { ArangoDB } from './db'
+
+const db = ArangoDB()
+
+export async function populateNewFields(): Promise<void> {
+  await db.query(aql`
+    FOR e in events
+      FOR b in blocks
+        FILTER e.blockId == b._key
+        UPDATE e._key WITH {journeyId: b.journeyId} IN events
+  `)
+
+  await db.query(aql`
+    FOR e in events
+      FILTER e.createdAt == null
+      UPDATE e._key with { createdAt: ${new Date(0).toISOString()}} IN events
+  `)
+
+  await db.query(aql`
+    FOR e in events
+      FILTER e.journeyId == null
+      REMOVE e in events
+  `)
+}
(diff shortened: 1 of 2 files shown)


apps/api-journeys/db/populateNewFields.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ArangoDB } from './db'
+
+const db = ArangoDB()
+
+export async function populateNewFields(): Promise<void> {
+  await db.query(aql`
+    FOR e in events
+      FOR b in blocks
+        FILTER e.blockId == b._key
+        UPDATE e._key WITH {journeyId: b.journeyId} IN events
```
tataihono, line 11:
```suggestion
        UPDATE e._key WITH { journeyId: b.journeyId } IN events
```

apps/api-journeys/db/populateNewFields.ts
```diff
@@ -0,0 +1,25 @@
...
+  `)
+
+  await db.query(aql`
+    FOR e in events
+      FILTER e.createdAt == null
+      UPDATE e._key with { createdAt: ${new Date(0).toISOString()}} IN events
```
tataihono, line 17:
```suggestion
      UPDATE e._key with { createdAt: ${new Date(0).toISOString()} } IN events
```

apps/api-journeys/db/populateNewFields.ts
```diff
@@ -0,0 +1,25 @@
...
+  `)
+
+  await db.query(aql`
+    FOR e in events
+      FILTER e.journeyId == null
+      REMOVE e in events
+  `)
```
tataihono, line 24: we probably shouldn't worry about including this for now since events without journeyIds won't be discoverable.

tataihono (changes requested):

apps/api-journeys/db/populateNewFields.ts
csiyang: I thought we were planning on making journeyId a non nullable field on events? Won't it break if we try to query journeyId if its still null?

tataihono: how would a user get access to events that aren't connected to any journey?

PR #998, opened 2022-10-25
fix: canvas styling 

GeronimoJohn:
# Description

RTL styling is not getting applied to our blocks within an iframe

[- Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/29688713/todos/5431701777)

Check between Admin and Storybook
- [ ]  RTL styling to be applied on a card within an iframe
- [ ]  LTR styling to be applied on a card within an iframe

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/FramePortal/FramePortal.tsx
+++ b/apps/journeys-admin/src/components/FramePortal/FramePortal.tsx
@@ -1,6 +1,8 @@
 import { createPortal } from 'react-dom'
 import createCache from '@emotion/cache'
 import { CacheProvider } from '@emotion/react'
+import { prefixer } from 'stylis'
+import rtlPlugin from 'stylis-plugin-rtl'
 import {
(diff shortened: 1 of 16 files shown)


apps/journeys-admin/src/components/FramePortal/FramePortal.tsx
```diff
@@ -86,8 +94,11 @@ export const FramePortal = memo(function FramePortal({
       <StyledFrame onLoad={dispatch} ref={frameRef} {...rest} />
       {iframeLoaded &&
         document != null &&
+        dir != null &&
```
storyworks, line 97: Do we need dir here? since we add dir directly onto document...

GeronimoJohn: Passing dir prop to be used on the content component

storyworks (approved):

PR #1003, opened 2022-10-28
fix: watch page wrapper

GeronimoJohn:
# Description

Updated the page wrapper to be taking in a header, footer, and children components. 

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29906873/todos/5476468158)

Visit [vercel watch ](https://watch-git-22-08-jg-fix-page-wrapper-jesusfilm.vercel.app/)
- [x] It should render the header component
- [x] it should render the footer component
- [x] it should render the children component

Storyboook
- [x] it should show the layout of the page on the page wrapper story
- [x] it should show a story of the header component

# Checklist


apps/watch/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -1,114 +1,45 @@
...
-      sx={{ width: '240px' }}
-      role="presentation"
-      onClick={toggleDrawer('left', false)}
-      onKeyDown={toggleDrawer('left', false)}
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100vh',
+        width: '100%',
+        backgroundColor: isStory ? '#EFEFEF' : ''
```
storyworks, line 24: The main principle for doing this was:
In header / body / footer - devs should pass in a Container which adds left/right spacing.

In our PageWrapper story, we want to see how this would be done. So the area where content sits is one color, the left/right spacing is another.

The problem you had before is that backgroundColor on Container coloured more than just its content area.
Instead of changing border-box / playing around with PageWrapper backgrounds another simple solution is from:
https://mui.com/material-ui/react-container/#main-content

image

Container backgroundColor is a set color (like grey), and use a <Box/> to represent content. Box has a different backgroundColor.

storyworks: [EDIT]
After chatting, we clarified the following:
- Grey shows the extra space when the screen size exceeds Container maxWidth (XL) (implemented with isStory)
- Some other colour will show the content of the Container (implemented with different colours atm, need to be 1 colour.)
- Different colour boxes will show the content bounds within a container (to be implemented)

storyworks (approved):

PR #1058, opened 2022-11-14
fix: primary image block edited on other journeys

GeronimoJohn:
# Description

All duplicated journeys are currently pointing to the same primary image block. This fix removes the primary image block from a journey unless they have the same journeyId

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/29689605/todos/5437618333)

- [ ] it shouldn't duplicate the primaryImageBlock on the duplicated journey

# Checklist


apps/api-journeys/src/app/modules/journey/journey.resolver.spec.ts
```diff
@@ -842,6 +842,21 @@ describe('JourneyResolver', () => {
...
+        createdAt: new Date().toISOString(),
+        status: JourneyStatus.draft,
+        slug: `${journey.title}-copy`,
+        title: `${journey.title} copy`,
+        publishedAt: undefined,
+        template: false,
+        primaryImageBlock: null
+      })
+    })
+
```
tataihono, line 859: this test doesn't actually test the journey resolver primary image block function.

Essentially you need to test three situations (some which might already be tested):

1. no primary image block set should return null
2. primary image block id set but does not belong to current journey, should return null
3. primary image block id set and belongs to current journey, should return block.

tataihono (dismissed):

tataihono (approved):

PR #1068, opened 2022-11-17
ci: fix chromatic untraced

storyworks:
# Description

Untraced configuration was set incorrectly and there were bugs with older versions of Chromatic so we were still getting full rebuilds for storybook

# How should this PR be QA Tested?

- [ ] Turbosnap ignores `untraced` files
For commit https://github.com/JesusFilm/core/pull/1068/commits/37e7d42ac1a87834018e5663ffdf2a93cf1dac77
We changed `storybook/main.js` and `main.yml` files. Only yml files detected
image

- [ ] Turbosnap disabled when `forceRebuild` is true , then make sure is `forceRebuild: 'main'` 
For commit https://github.com/JesusFilm/core/pull/1068/commits/c49cffcc67cf11b878f6b7fa052b06f410ef53c9
The same change to `storybook/main.js` and `main.yml` triggered Turbosnap to be disabled. 
image

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged into main

PR #856, opened 2022-09-05
chore: use turbosnap for chromatic

storyworks:
# Description

The `--preserve-missing` flag is now becoming deprecated for Chromatic. This left stories that weren't changed, untouched (instead of deleted). 

Now instead of manually building stories for affected libraries and using that flag, it's recommended we build all stories in a monorepo and use Turbosnap to build the affected stories only.

https://www.chromatic.com/docs/turbosnap#full-rebuilds

There are some caveats to using TurboSnap. See the documentation above.

[Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/29147486/todos/5294405053)

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [x] No VR tests when no changes made
image

- [x] Should see turbosnap status in UI Review
image

- [x] Only a subset of stories should have been tested when I make a change
image

# Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged into main

PR #1073, opened 2022-11-17
ci: build only affected libraries so full rebuilds have less impact

storyworks:
# Description

Turbosnap continues to pick up package.json changes from main. Always ignore these changes by default and manually rebuild the entire project when we expect it (or storybook config changes) to cause a story update.

I have exhausted all other options on our end that we know could possibly be causing the issue... fixing git history errors (via incorrect branch fetching), incorrect configurations...
#1071 #1072 #1068 

image

Tatai has suggested building only the affected libraries. I've added back the previous logic + steps we used in #856 

But first we're testing that untraced works in all circumstances (even when merging in latest main + package.json changes)

To test after merged in.
- [ ] PRs which have this change + package.json changes merged in from latest main don't trigger full rebuild

# Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -150,8 +150,6 @@ jobs:
           exitOnceUploaded: 'true'
           autoAcceptChanges: 'main'
           onlyChanged: 'true'
-          # Uncomment forceRebuild if you're making a package and storybook config change which you expect to change stories. Recomment after the visual regression build is done.
-          # forceRebuild: 'main'
           skip: '@(dependabot/**|00-00-CI-chore-update-translations)'
           untraced: |
             - '**/package*.json'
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -150,7 +150,8 @@ jobs:
           exitOnceUploaded: 'true'
           autoAcceptChanges: 'main'
           onlyChanged: 'true'
-          forceRebuild: 'main'
+          # Uncomment forceRebuild if you're making a package and storybook config change which you expect to change stories. Recomment after the visual regression build is done.
```
csiyang, line 153:
```suggestion
          # Uncomment forceRebuild when making a package or storybook config change that is expected to change stories. Recomment after the visual regression build is done.
```

csiyang (changes requested):

tataihono (approved):

PR #1082, opened 2022-11-23
fix: facebook/twitter cards not displaying

jernst7:
# Description

Fixes the Facebook and Twitter previews not displaying when a link is shared. The main changes are adding NextSeo to the video's display page and changing how the share dialog gets the route to share.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/30300130/todos/5565198965)

- [x] Visual inspection that a shared link on FB/Twitter displays a preview of the page. Note: The shared link must come from the watch preview on this branch so that the NextSeo is present.
- [x] Checks that the sharing links were constructed properly were already present, so they have been updated to work with the new routing. Running those tests will check that the code constructs the correct links.

# Checklist


apps/watch/pages/_app.tsx
```diff
@@ -26,10 +26,7 @@ export default function WatchApp({
 }: AppProps & { emotionCache?: EmotionCache }): ReactElement {
   const auth = getAuth(firebaseClient)
   const [user] = useAuthState(auth)
-  const client = useApolloClient(
```
storyworks, line 29: I think this change was put new in main. Can you revert back to using it?

apps/watch/pages/videos.tsx
```diff
@@ -1,36 +1,8 @@
...
-import { createApolloClient } from '../src/libs/client'
 
-interface VideosPageProps {
-  initialApolloState: NormalizedCacheObject
-}
 function VideosPage(): ReactElement {
   return <Videos />
 }
 
-export const getStaticProps: GetStaticProps<VideosPageProps> = async () => {
```
storyworks, line 19: Revert all these changes too - your PR has nothing to do with this.

apps/watch/src/components/ShareDialog/ShareDialog.tsx
```diff
@@ -46,7 +47,7 @@ export function ShareDialog({
       ? `${
           process.env.NEXT_PUBLIC_WATCH_URL ??
           'https://watch-jesusfilm.vercel.app'
-        }/${routes?.join('/')}`.trim()
+        }/${slug}/${variant?.language?.name[0]?.value.toLowerCase() ?? ''}`
```
storyworks, line 50: Since this page is used by our page and the `Series` pages - our slugs are different. 
Has this been tested to work on both pages? I'm told our pages contain the language as part of the slug, but the `series` pages don't. So this logic will work for one but not the other.

Let's continue to use the routes string array to pass in the route. You can update the route passed in via our part3 `index.ts` page. This will also require updating tests and stories.

apps/watch/src/components/VideosPage/SubHero/VideosSubHero.tsx
```diff
@@ -7,11 +7,11 @@ function VideosSubHeroStats(): ReactElement {
   return (
     <Stack direction="row" justifyContent="space-around" spacing={10}>
       <div>
-        <Typography variant="h3">724</Typography>
+        <Typography variant="h3">134</Typography>
```
storyworks, line 10: What's this change about?

apps/watch/src/components/VideosPage/VideosPage.tsx
```diff
@@ -38,7 +37,7 @@ export function VideosPage(): ReactElement {
   const languageContext = useLanguage()
   const [isEnd, setIsEnd] = useState(false)
   const [previousCount, setPreviousCount] = useState(0)
-  const [filter] = useState<VideosFilter>(defaultFilter)
+  const [filter] = useState<VideosFilter>({})
```
storyworks, line 40: I think you've accepted the wrong changes when merging main... please revert these changes and the ones for the 2 folders below as well

apps/watch/pages/_app.tsx
jernst7: All reverts done

libs/locales/en/apps-journeys_old.json
```diff
@@ -0,0 +1,4 @@
+{
```
GeronimoJohn, line 1: Can delete this file. we don't usually keep old translations

PR #1083, opened 2022-11-23
feat: new api video data

mikeallisonJS:
# Description

This adds subType to api-videos, which mirrors the subType in arclight. This gives us more information about videos & container types. Additionally, it removes tags from graphql, as they are replaced by the collection subtype (which contains more appropriate data).

https://3.basecamp.com/3105655/buckets/30316909/todos/5574989750
https://3.basecamp.com/3105655/buckets/30316909/todos/5574990159
https://3.basecamp.com/3105655/buckets/30316909/todos/5574991395
https://3.basecamp.com/3105655/buckets/30316909/todos/5574992247

- [ ] Test subType exists on video or videos query and that collection is returned (id: JFM1) on https://graphql-stage.jesusfilm.org/graphql
- [ ] Test tags removed (visible in schema) on https://graphql-stage.jesusfilm.org/graphql

# Checklist


apps/api-videos/src/app/modules/video/video.graphql
```diff
@@ -40,8 +50,8 @@ type Video @key(fields: "id primaryLanguageId") {
 input VideosFilter {
```
storyworks, line 50: I can't comment above but was it possible to rename `episodeIds` and `episodes`?

apps/api-videos/src/app/modules/video/video.graphql
```diff
@@ -13,9 +13,19 @@ enum VideoType {
   playlist
```
storyworks, line 13: Can we add comments about what each VideoType is now that there are subtypes?

apps/api-videos/src/app/modules/video/video.graphql
mikeallisonJS: will follow up Monday. I have an idea for this, but would like feedback

storyworks (approved):

PR #1119, opened 2022-12-05
refactor: split out container and content pages

storyworks:
# Description

Splits out VideoContainer into the 2 different pages - one for containers and other for content (where we can play and download videos). This reduces merge conflicts between the 2 teams and makes it explicit the type of page we expect to land on for each route.

- [ ] Watch still correctly navigates to the correct looking page

# Checklist


apps/watch/src/components/VideoContainerPage/VideoContainerPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+export function VideoContainerPage({
+  content
+}: VideoContainerPageProps): ReactElement {
+  const [openShare, setOpenShare] = useState(false)
+
+  return (
+    <PageWrapper hero={<SimpleHero video={content} />}>
+      {content != null && (
+        <>
+          <Stack
```
storyworks, line 28: Copy pasted from old page - feel free to totally remove and simplify when you work on it.

apps/watch/pages/[part1]/[part2].tsx
```diff
@@ -21,7 +23,15 @@ interface Part2PageProps {
 }
 
 export default function Part2Page({ content }: Part2PageProps): ReactElement {
-  return <VideoContainer content={content} />
+  return (
+    <LanguageProvider>
+      {content.variant?.hls != null ? (
```
Comment (csiyang) [inline apps/watch/pages/[part1]/[part2].tsx:28]: I think this logic should be kept out of the page so that it can be tested

Comment (csiyang) [reply apps/watch/pages/[part1]/[part2].tsx:28]: Why don't we do something like, 
- rename `VideoContainer` to `VideoPage`
- `VideoPage` can have 2 sub components `Content` and `Container`

csiyang (changes requested):

apps/watch/pages/[part1]/[part2].tsx
Comment (storyworks) [reply apps/watch/pages/[part1]/[part2].tsx:28]: My original plan was to rename `VideoContainer` to `VideoPage`. But decided to directly use `VideoContentPage` and `VideoContainerPage`. Mainly because this check only needs to be done for `part2`. 

This logic could be moved to a `VideoPage`, but it seemed potentially unnecessary as we'd be adding an extra generation in the component tree for one test and we have put simple logic in the pages in the past to determine what page should render. 
image
Or see journey-admin publisher edit.

apps/watch/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -34,9 +35,8 @@ export function PageWrapper({
             {hero}
           </ThemeProvider>
         </Container>
-        <Container maxWidth="xxl" sx={{ flexGrow: 1 }}>
-          {children}
-        </Container>
+
+        <Box sx={{ flexGrow: 1 }}>{children}</Box>
```
storyworks, line 39: Removed the container here since SingleVideoPage has a carousel which needs to stretch to full width of the page. I have added the Container to the pages themselves.

apps/watch/src/components/ShareDialog/ShareDialog.tsx
```diff
@@ -35,6 +35,13 @@ export function ShareDialog({
     setValue(newValue)
   }
 
+  const shareDescription =
+    video.description != null && video.description.length > 0
```
storyworks, line 39: No description needed on Series/Collection pages. But also arclight data not always complete.

csiyang (approved):

PR #1128, opened 2022-12-07
refactor: move language select to shared ui

csiyang:
# Description

- moved the languageSelect component to sharedUi in preparation to be used by the watch project.
- Added functionality to shared ui project to make generated files

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/30317980/todos/5615264742)

- [x] Language select for journey should still work
- [x] Language select for video should still work

# Checklist


apps/journeys-admin/src/components/JourneyView/Menu/LanguageDialog/LanguageDialog.tsx
```diff
@@ -4,8 +4,8 @@ import { useSnackbar } from 'notistack'
 import { Formik, Form, FormikValues } from 'formik'
 import { Dialog } from '@core/shared/ui/Dialog'
 import { useJourney } from '@core/journeys/ui/JourneyProvider'
+import { LanguageSelect } from '@core/shared/ui/LanguageSelect'
```
csiyang, line 7: Was wondering if it was possible to update the query to use `LanguageFields` (line 29), but wasn't sure how to do it with the variables that was being passed in.

libs/shared/ui/project.json
```diff
@@ -38,6 +38,22 @@
           "quiet": true
         }
       }
+    },
+    "codegen": {
```
csiyang, line 42: added codgen to sharedUi project

storyworks:
Are you able to make a post in the design system chat proposing this become a design system wide component?

mikeallisonJS (dismissed):

libs/shared/ui/src/components/LanguageAutocomplete/LanguageAutocomplete.tsx
```diff
@@ -4,27 +4,36 @@ import Stack from '@mui/material/Stack'
...
+export interface LanguageOption {
   id: string
   localName?: string
   nativeName?: string
 }
 
-interface LanguageSelectProps {
-  onChange: (value?: LanguageSelectOption) => void
-  value?: LanguageSelectOption
+interface LanguageAutocompleteProps {
```
storyworks, line 24: Please also expose optional `renderInput` and `renderOption`

`renderInput` is a function that takes the type of `params` (see line 75) and returns a ReactNode.
`renderOption` is a function takes the same props as line 95 and also returns a ReactNode.

Then add in the logic `renderInput={renderInput ? renderInput : <TextField ..... /> }`
Do the same for renderOption

Then create tests to check renderInput and renderOption are called when they are passed in

This will give us more flexibility for the look of LanguageAutocomplete. It also means this shared-ui will mainly be to control the sorting and searching logic, autocomplete interactions (dropdown onBlur, esc to exit, etc)

libs/shared/ui/src/components/LanguageAutocomplete/LanguageAutocomplete.stories.tsx
```diff
@@ -89,8 +80,8 @@ const Template: Story = ({ onChange }) => {
 
 export const Default = Template.bind({})
```
storyworks, line 81: Can you add another story here same as default but called `Website theme` and just change the parameters to take `themeName: ThemeName.website` ?

From that you'll notice that we need to add `borderWidth: '2px'` to `components.ts` in website theme. 
image

csiyang: updated the website theme, but didn't find a good way to update `themName` without overriding `themeMode`, but I don't think its super important since each project that uses it will take its own snapshots. Especially since its taking in custom renderInput and renderOption it can look pretty different to any story we make for it.

storyworks: For shared-ui components - I don't think it's a good practice to omit VR for a theme, relying on it's usage on a page to generate the VR test. 

Those components that use it as you say, may pass in custom renderInputs and renderOptions. 

This story would be the equivalent of adding the LanguageAutocomplete to the website theme `components.stories.ts` - except I'm wondering if we can add it in the LanguageAutocomplete story instead of in `components.stories.ts` since that folder is just the default styled mui components. And it's not as easy to remember to keep adding each new shared-ui component to the `components.stories.tsx` file.

libs/shared/ui/src/libs/themes/website/tokens/components.stories.tsx
```diff
@@ -126,18 +159,12 @@ const Template: Story<ComponentProps<typeof Button>> = (args) => {
...
-        options={audioLanguages}
-        renderInput={(params) => (
-          <TextField
-            {...params}
-            label="Audio"
-            helperText="6 options available"
-          />
-        )}
+      {/* LANGUAGE AUTOCOMPLETE */}
+      <LanguageAutocomplete
```
storyworks, line 163: Have you checked there is no other instance of a regular autocomplete being used in watch?

storyworks (approved):

PR #1130, opened 2022-12-07
feat: create video context

storyworks:
# Description

Use VideoProvider to get values of video content and video container instead of prop drilling. Using render props pattern here would make each page too large so not appropriate here.

[Basecamp Todo](https://3.basecamp.com/3105655/buckets/30300130/todos/5616332686)

- [ ] All part2 and part3 pages render correctly

# Checklist

### DIFF
--- a/apps/watch/pages/[part1]/[part2].tsx
+++ b/apps/watch/pages/[part1]/[part2].tsx
@@ -7,6 +7,7 @@ import { createApolloClient } from '../../src/libs/client'
 import { VideoContentPage } from '../../src/components/VideoContentPage'
 import { VideoContainerPage } from '../../src/components/VideoContainerPage/VideoContainerPage'
 import { LanguageProvider } from '../../src/libs/languageContext/LanguageContext'
+import { VideoProvider } from '../../src/libs/videoContext'
 import { VIDEO_CONTENT_FIELDS } from '../../src/libs/videoContentFields'
 
 export const GET_VIDEO_CONTENT = gql`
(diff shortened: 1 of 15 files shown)


apps/watch/pages/[part1]/[part2].tsx
```diff
@@ -25,11 +26,13 @@ interface Part2PageProps {
...
-      {content.variant?.hls != null ? (
-        <VideoContentPage content={content} />
-      ) : (
-        <VideoContainerPage content={content} />
-      )}
+      <VideoProvider value={{ content }}>
+        {content.variant?.hls != null ? (
+          <VideoContentPage />
+        ) : (
+          <VideoContainerPage content={content} />
```
Comment (storyworks) [inline apps/watch/pages/[part1]/[part2].tsx:33]: I haven't removed passing in `content` for the Container page like I have with `VideoContentPage` since it may cause significant refactors for PRs already up. This PR is intended to introduce the state and you can slowly adopt the `useVideo` hook as you go.

apps/watch/src/libs/videoContext/VideoContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+  value: VideoPageProps
+}
+
+export function VideoProvider({
+  children,
+  value
+}: VideoProviderProps): ReactElement {
+  return (
+    <VideoContext.Provider
+      value={{ ...value.content, container: value.container, isPlaying: false }}
```
storyworks, line 37: Pass in the `content` and `container` as we do now. 

But I'm thinking it's more valuable to destructure `content` so we can directly access it's fields (most our components use `content` fields)

To distinguish these from `container` fields, we keep `container` as is.

apps/watch/src/libs/videoContext/VideoContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+  isPlaying?: boolean
+}
+
+const VideoContext = createContext<Context | undefined>(undefined)
+
+export function useVideo(): Context {
+  const context = useContext(VideoContext)
+
+  if (context === undefined) {
+    throw new Error('useVideos must be used within a VideoProvider')
```
storyworks, line 20: We should do this to `JourneyProvider` too so we don't have to keep doing `journey != null` checks

GeronimoJohn (approved):

GeronimoJohn (approved):

PR #1139, opened 2022-12-08
feat: watch videos preload

mikeallisonJS:
# Description

This PR allows getStaticProps to hydrate the apollo cache. It is used to pre-fill the 'All Videos' page.

- [x] No perceptible impact on site
- [x] /videos page load no longer has skeleton on initial load

# Checklist


apps/watch/src/libs/client/client.ts
```diff
@@ -4,31 +4,51 @@ import {
...
+): ApolloClient<NormalizedCacheObject> => {
+  const store: ApolloClient<NormalizedCacheObject> = useMemo(
+    () => createApolloClient(token, initialState),
+    [token, initialState]
+  )
+  return store
 }
 
 const client = new ApolloClient({
+  ssrMode: typeof window === 'undefined',
```
csiyang, line 51: What does this line do?

csiyang (commented):
Anyway to make this deploy so we can test it out on stage?

apps/watch/src/libs/client/client.ts
mikeallisonJS: sets the apollo client to headless mode when called from node. Taken from apollo docs when re-hydrating the cache.

csiyang (approved):

PR #1240, opened 2023-01-12
fix: all videos page remove multiple filters per type

storyworks:
# Description

Doesn't completely fix filters but limits them to one per filter type to reduce confusion whilst the autocomplete doesn't include check boxes.

https://3.basecamp.com/3105655/buckets/30278084/todos/5715541189

- [ ] Changing between filters of the same type replaces the previous filter chip + videos filter by the new value only.
- [ ] Deleting a chip resets the video search (but autocomplete value remains - this is a bug see https://3.basecamp.com/3105655/buckets/30278084/todos/5724905360)

# Checklist


apps/watch/src/components/VideosPage/CurrentFilters/CurrentFilters.tsx
```diff
@@ -38,46 +46,44 @@ export function CurrentFilters({
...
-              }`}
-              deleteIcon={<CloseRoundedIcon fontSize="small" />}
-              onDelete={() => onDelete('sl', language.id)}
-              sx={{
-                mr: 3,
-                border: '1px solid #DCDAD2'
-              }}
-            />
-          ))}
+        {audioLanguages.value.map((language) => (
```
storyworks, line 49: Changed CurrentFilters to loop through the string array of chip labels instead of language objects so we don't need to filter through all languages each time a filter is altered.

onDelete now also triggers the same function as onChange so we can reuse a function.

apps/watch/src/components/VideosPage/VideosPage.spec.tsx
```diff
@@ -112,12 +112,12 @@ describe('VideosPage', () => {
       fireEvent.click(option)
       const chip = getByText('audio: English')
       expect(chip).toBeInTheDocument()
-      fireEvent.click(getAllByTestId('CloseRoundedIcon')[0])
+      fireEvent.click(getByTestId('delete audio English filter'))
```
storyworks, line 115: More verbose test id easier to target and maintain in future.

apps/watch/src/components/VideosPage/VideosPage.tsx
```diff
@@ -56,22 +56,20 @@ export function VideosPage(): ReactElement {
...
   const [isEnd, setIsEnd] = useState(false)
   const [previousCount, setPreviousCount] = useState(0)
-  const [languageFilter, setLanguageFilter] = useState<string[]>([])
+  const [audioLanguageFilter, setAudioLanguageFilter] = useState<string[]>([])
   const [subtitleLanguageFilter, setSubtitleLanguageFilter] = useState<
     string[]
   >([])
-  const [filter, setFilter] = useState<VideosFilter>({
-    availableVariantLanguageIds:
-      languageFilter.length > 0 ? languageFilter : undefined,
```
storyworks, line 65: Initial filter is always empty so this logic isn't needed.

apps/watch/src/components/VideosPage/VideosPage.tsx
```diff
@@ -102,57 +100,30 @@ export function VideosPage(): ReactElement {
...
-    setStateFunction: (value: string[]) => void
-  ): void {
-    const updatedFilters = union(selectedFilter, [selectedLanguage])
-    setStateFunction(updatedFilters)
-    setFilter({
-      ...filter,
-      [field]: updatedFilters
+  }, [appliedFilters, refetch, languageContext])
+
+  function handleChange({
```
storyworks, line 110: Consolidated handleChange and handleRemove into one function.

storyworks:
Closing, it looks like they've recreated this page now.

PR #1244, opened 2023-01-13
fix: card selection panel

DanChae03:
# Description

Selecting, duplicating or deleting a card from the journeys should not switch you to the properties drawer anymore, instead should remain on the card selection drawer for better user friendliness.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31245472/todos/5847664068)

- [x] Clicking a particular card should keep the cards panel open
- [x] + Adding a new card should keep the cards panel open
- [x] Clicking "Edit cards" on Journey Details page brings you to first card with cards panel open
- [ ] Deleting / Duplicating cards should not change tab panel

# Checklist


apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.spec.tsx
```diff
@@ -94,11 +94,11 @@ describe('ControlPanel', () => {
         </JourneyProvider>
       </MockedProvider>
     )
-    // Default start on Properties
+    // Default start on Cards
+    expect(getByRole('tabpanel', { name: 'Cards' })).toBeInTheDocument()
+    fireEvent.click(getByRole('tab', { name: 'Properties' }))
```
storyworks, line 99: Add a line where you select a card in the card tab panel, then add a line show it stays on the card tab.

csiyang: I've tested 1 and 2.

4 isn't really suitable for unit testing since its only responsible for selecting the next block. The drawer should be set when a step block is selected (which is being tested)

storyworks: Fair enough, it seems more appropriate as an E2E - but for these cases, let's manually test it well? I tested this morning and point 4 hasn't been implemented yet so let's put that in?

lumberman:
Great job! Works great for me.

csiyang:
@lumberman just pushed an update for point 4 (deleting duplicating cards, won't change tab) so if you're on the properties tab and you delete or duplicate a card it should stay on the properties tab for the newly selected card

apps/journeys-admin/src/components/Editor/ControlPanel/ControlPanel.spec.tsx
csiyang: oops I did not understand the requirement properly!

apps/journeys-admin/src/components/Editor/ControlPanel/Attributes/blocks/Step/Step.tsx
```diff
@@ -33,22 +33,6 @@ export function Step({
       ? getStepHeading(nextStep.id, nextStep.children, steps, t)
       : 'None'
 
-  useEffect(() => {
```
csiyang, line 36: @storyworks  Turns out no dispatch call is needed here at all. If people are changing cards then they're already on the cards tab which satisfies 1 and 2 and ends up working for 4. 

I did a few tests and it seems to work good, welcome any thoughts about edge cases I might've missed though.

storyworks: What about when they just select a block - was that supposed to trigger a panel change?

csiyang: Gets handled in canvas

storyworks (approved):

PR #1279, opened 2023-01-25
feat: add user invite api

storyworks:
# Description

Creates and updates a userInvite to set journey permissions

[Basecamp Todo](https://3.basecamp.com/3105655/buckets/31245893/todos/5763468706)

Stage graphql correctly returns expected data
- [x] userInviteCreate - returns new userInvite
- [x] userInvites - returns all user invites for a particular journey
- [x] userInviteRemove - removes invite by id
- [x] userInviteAcceptAll - returns updated userInvite (only allows changing `accepted` value for now), userJourney also created in database (side effect)

image
image

# Checklist

### DIFF
--- a/apps/api-gateway/schema.graphql
+++ b/apps/api-gateway/schema.graphql
@@ -791,6 +791,9 @@ type Mutation
 
   """Updates template"""
   journeyTemplate(id: ID!, input: JourneyTemplateInput!): Journey! @join__field(graph: JOURNEYS)
+  userInviteCreate(journeyId: ID!, input: UserInviteCreateInput): UserInvite! @join__field(graph: JOURNEYS)
+  userInviteRemove(id: ID!, journeyId: ID!): UserInvite! @join__field(graph: JOURNEYS)
+  userInviteAcceptAll: [UserInvite!]! @join__field(graph: JOURNEYS)
   userJourneyApprove(id: ID!): UserJourney! @join__field(graph: JOURNEYS)
   userJourneyPromote(id: ID!): UserJourney! @join__field(graph: JOURNEYS)
   userJourneyRemove(id: ID!): UserJourney! @join__field(graph: JOURNEYS)
@@ -908,6 +911,7 @@ type Query
   adminJourney(id: ID!, idType: IdType): Journey @join__field(graph: JOURNEYS)
   journeys(where: JourneysFilter): [Journey!]! @join__field(graph: JOURNEYS)
   journey(id: ID!, idType: IdType): Journey @join__field(graph: JOURNEYS)
+  userInvites(journeyId: ID!): [UserInvite!] @join__field(graph: JOURNEYS)
   getUserRole: UserRole @join__field(graph: JOURNEYS)
 
   """A list of visitors that are connected with a specific team."""
@@ -1395,6 +1399,23 @@ type UserAgent
   browser: Browser!
   device: Device!
   os: OperatingSystem!
+}
+
+type UserInvite
+  @join__type(graph: JOURNEYS, key: "id")
+{
+  id: ID!
+  journeyId: ID!
+  senderId: ID!
+  email: String!
+  acceptedAt: DateTime
+  removedAt: DateTime
+}
+
+input UserInviteCreateInput
+  @join__type(graph: JOURNEYS)
+{
+  email: String!
 }
 
 type UserJourney
--- a/apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
+++ b/apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
@@ -1,0 +1,22 @@
+type UserInvite @key(fields: "id") {
+  id: ID!
+  journeyId: ID!
+  senderId: ID!
+  email: String!
+  acceptedAt: DateTime
+  removedAt: DateTime
+}
+
+extend type Query {
+  userInvites(journeyId: ID!): [UserInvite!]
+}
+
+input UserInviteCreateInput {
+  email: String!
+}
+
+extend type Mutation {
+  userInviteCreate(journeyId: ID!, input: UserInviteCreateInput): UserInvite!
+  userInviteRemove(id: ID!, journeyId: ID!): UserInvite!
+  userInviteAcceptAll: [UserInvite!]!
+}
--- a/apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.spec.ts
@@ -1,0 +1,273 @@
+import { Test, TestingModule } from '@nestjs/testing'
+import { Role, UserJourneyRole } from '../../__generated__/graphql'
+import { JourneyService } from '../journey/journey.service'
+import { UserJourneyResolver } from '../userJourney/userJourney.resolver'
+import { UserJourneyService } from '../userJourney/userJourney.service'
+import { UserRoleService } from '../userRole/userRole.service'
+
+import { UserInviteResolver } from './userInvite.resolver'
+import { UserInviteService } from './userInvite.service'
+
+describe('UserInviteResolver', () => {
+  let resolver: UserInviteResolver,
+    service: UserInviteService,
+    ujResolver: UserJourneyResolver
+
+  const createInput = {
+    email: 'test@email.com'
+  }
+
+  const userInvite = {
+    id: '1',
+    journeyId: 'journeyId',
+    senderId: 'senderId',
+    email: 'test@email.com',
+    acceptedAt: null,
+    removedAt: null
+  }
+
+  const acceptedInvite = {
+    ...userInvite,
+    id: '2',
+    journeyId: 'acceptedJourneyId',
+    acceptedAt: '2021-02-18T00:00:00.000Z'
+  }
+
+  const removedInvite = {
+    ...userInvite,
+    id: '3',
+    journeyId: 'removedJourneyId',
+    removedAt: '2021-02-18T00:00:00.000Z'
+  }
+
+  const userInviteService = {
+    provide: UserInviteService,
+    useFactory: () => ({
+      save: jest.fn((input) => {
+        return { ...input, acceptedAt: null, removedAt: null }
+      }),
+      update: jest.fn((id, input) => {
+        return { ...userInvite, ...input }
+      }),
+      remove: jest.fn((id) => userInvite),
+      getAllUserInvitesByEmail: jest.fn((email) => {
+        if (email === userInvite.email) {
+          return [userInvite, acceptedInvite, removedInvite]
+        }
+        return []
+      }),
+      getAllUserInvitesByJourney: jest.fn((journeyId) => {
+        return [{ ...userInvite, journeyId }]
+      }),
+      getUserInviteByJourneyAndEmail: jest.fn((journeyId, email) => {
+        if (
+          journeyId === removedInvite.journeyId &&
+          email === removedInvite.email
+        ) {
+          return removedInvite
+        } else if (
+          journeyId === acceptedInvite.journeyId &&
+          email === acceptedInvite.email
+        ) {
+          return acceptedInvite
+        }
+        return null
+      })
+    })
+  }
+
+  const userJourney = {
+    id: 'userJourneyId',
+    userId: 'userId',
+    journeyId: 'journeyId',
+    role: UserJourneyRole.editor
+  }
+
+  const userJourneyResolver = {
+    provide: UserJourneyResolver,
+    useFactory: () => ({
+      userJourneyRequest: jest.fn((journeyId, idType, userId) => {
+        return { ...userJourney, journeyId, userId, role: 'inviteRequested' }
+      }),
+      userJourneyApprove: jest.fn((id, userId) => {
+        return userJourney
+      })
+    })
+  }
+
+  const userJourneyService = {
+    provide: UserJourneyService,
+    useFactory: () => ({ get: jest.fn((key) => userJourney) })
+  }
+
+  const userRole = {
+    id: 'userRoleId',
+    userId: 'userId',
+    roles: [Role.publisher]
+  }
+
+  const userRoleService = {
+    provide: UserRoleService,
+    useFactory: () => ({
+      getUserRoleById: jest.fn(() => userRole)
+    })
+  }
+
+  const journeyService = {
+    provide: JourneyService,
+    useFactory: () => ({
+      get: jest.fn((id) => ({ id }))
+    })
+  }
+
+  const user = {
+    id: 'userId',
+    email: 'test@email.com',
+    firstName: 'Test'
+  }
+
+  beforeAll(() => {
+    jest.useFakeTimers('modern')
+    jest.setSystemTime(new Date('2021-02-18'))
+  })
+
+  afterAll(() => {
+    jest.useRealTimers()
+  })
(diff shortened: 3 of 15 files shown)


apps/api-journeys/src/app/modules/userInvite/userInvite.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { aql } from 'arangojs'
+import { DocumentCollection } from 'arangojs/collection'
+import { UserInvite } from '../../__generated__/graphql'
+
+@Injectable()
+export class UserInviteService extends BaseService {
+  collection: DocumentCollection = this.db.collection('userInvites')
+
+  @KeyAsId()
+  async getUserInviteByInviteId(id: string): Promise<UserInvite> {
```
csiyang, line 13: any reason we couldn't use the get function from base service?

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
```diff
@@ -0,0 +1,30 @@
+type UserInvite @key(fields: "id") {
+  id: ID!
+  inviteId: ID!
+  journeyId: ID!
+  sentBy: ID!
```
csiyang, line 5: This only ties invitations with the person who sends them right? What happens if someone else is also looking at the invitees, they won't be able to see the list of those that have been invited right?

apps/api-gateway/schema.graphql
```diff
@@ -261,7 +261,6 @@ type Country
   image: String
 }
 
-"""DateTime custom scalar"""
```
csiyang, line 264: how come this comment disappeared?

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,54 @@
...
+    return await this.userInviteService.getAllUserInvitesBySender(userId)
+  }
+
+  @Mutation()
+  @UseGuards(GqlAuthGuard)
+  async userInviteCreate(
+    @Args('journeyId') journeyId: string,
+    @Args('input') input: UserInviteCreateInput
+  ): Promise<UserInvite | null> {
+    const inviteId = uuidv4()
```
csiyang, line 29: why is this needed? Can't we just use the id generated by arrango?

csiyang (changes requested):

apps/api-gateway/schema.graphql
storyworks: It's a generated file, possibly the last person who made the change didn't update the schema

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
storyworks: UX haven't replied about whether they want the invite code to be more user friendly than uuid. I can use the userInvite for now and add this back in later, but at least now we have a record of the change :)

apps/api-journeys/src/app/modules/userInvite/userInvite.service.ts
storyworks: Since we're using the `inviteId` to fetch the userInvite rather than the `id` itself. 

After confirming with Tatai there are no security risks, let's revert to `id` and shift back to `inviteId` only if UX has a strong suggestion to use some other type of code.

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
csiyang: Umm kinda... so currently the user invited list shows everyone thats been invited. But that won't be the case when this gets added.

- say me and you are editors
- you invite someone to our journey and we're waiting for them to respond
- When I look on my user list for our journey, i'll see that they haven't been invited yet

storyworks: Interesting use case! 

We could make it that only the owner can invite people by email, so there would also only be 1 sender id.

We could also make the front end fetch all the userInvites linked to all the owners (if we make multiple owners in the future)

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
```diff
@@ -0,0 +1,30 @@
+type UserInvite @key(fields: "id") {
+  id: ID!
+  inviteId: ID!
```
csiyang, line 3: Lets not add this until we're sure we want it. It's way more annoying to remove redundant data than to add a new field
```suggestion
```

csiyang (changes requested):

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
storyworks: Oops, thought I removed - will do.

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      await resolver.userInviteCreate('journeyId', { ...createInput, expireAt })
+
+      expect(service.save).toHaveBeenCalledWith(
+        expect.objectContaining({
+          journeyId: 'journeyId',
+          senderId: 'senderId',
+          name: createInput.name,
+          email: createInput.email,
+          accepted: false
```
csiyang, line 116: expire at needs to be tested here too. I think you need to mock the timer to make it work

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+    // TODO: Get email from user in db
+    if (input.email === userInvite.email) {
+      await this.userJourneyResolver.userJourneyApprove(
+        userJourney.id,
+        userInvite.senderId
+      )
+
+      console.log('userJourney - approve', userJourney)
+
+      return await this.userInviteService.update(id, { accepted: true })
```
csiyang, line 90: can we make this conditional on the approve being successful?

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+  async userInviteAccept(
+    @Args('id') id: string,
+    @CurrentUserId() userId: string,
+    @Args('input') input: UserInviteAcceptInput
+  ): Promise<UserInvite> {
+    const userInvite = await this.userInviteService.get<UserInvite>(id)
+
+    console.log('JOURNEY ID', userInvite)
+
+    const userJourney = await this.userJourneyResolver.userJourneyRequest(
```
csiyang, line 74: thoughts on a new resolver for creating a approved userJourney?

csiyang (changes requested):

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+  @Mutation()
+  @UseGuards(GqlAuthGuard)
+  async userInviteAccept(
+    @Args('id') id: string,
+    @CurrentUserId() userId: string,
+    @Args('input') input: UserInviteAcceptInput
+  ): Promise<UserInvite> {
+    const userInvite = await this.userInviteService.get<UserInvite>(id)
+
+    console.log('JOURNEY ID', userInvite)
```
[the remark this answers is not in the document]
storyworks: Oh my lol, thanks

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const userJourney = await this.userJourneyResolver.userJourneyRequest(
+      userInvite.journeyId,
+      IdType.databaseId,
+      userId
+    )
+    console.log('userJourney - request', userJourney)
+
+    // TODO: Get email from user in db
```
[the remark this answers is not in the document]
storyworks: No, this is still needed - we can't reference user related data between api's purely server-side. Tatai said it might make sense later down the line to allow user to be referenced by other apis. When it can, we no longer want to pass in email but get it from the user via userId

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.spec.ts
storyworks: Just saw your test, can do.

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
storyworks: ? We're reusing userJourneyApprove and that's already tested, so what would the new resolver do?

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
storyworks: That's already the case, see line 91 below is the end of the `if` statement

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
```diff
@@ -0,0 +1,32 @@
...
+}
+
+input UserInviteAcceptInput {
+  email: String!
+}
+
+input UserInviteCreateInput {
+  email: String!
+  name: String!
+  senderId: ID!
```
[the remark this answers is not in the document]
storyworks: By nature of it being the senderId, it cannot be the current user. 

We call userInviteCreate only when the newly invited user is logged in, so the current user is the invitee. Tatai wants senderId to be stored for auditing, we don't need to expose it on the front end.

csiyang:
Could not QA via stage, something about firebase token being wrong.

Locally the get and the create worked, but the accept did not. It throws the circular structure error, but after trying some different auth headers the call succeeded but the but the "accepted" value did not change

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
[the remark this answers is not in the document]
csiyang: Doesn't this get created when the inviter creates the invite, how will it get created when the invitee logs in? User id's are exposed so, so if we expose senderId, bad actors can fake journey invites.

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
csiyang: it would create a userJourney that was already approved, instead of requesting and approving. Just a little cleaner I guess, but just wondering if it was something considered, not a change request

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  @Mutation()
+  @UseGuards(GqlAuthGuard)
+  async userInviteAccept(
+    @Args('journeyId') journeyId: string,
+    @CurrentUserId() userId: string,
+    @Args('input') input: UserInviteAcceptInput
+  ): Promise<UserInvite> {
+    const userInvite =
+      await this.userInviteService.getUserInviteByJourneyAndEmail(
```
storyworks, line 70: We don't don't use the invite code anymore, navigating directly to the page will trigger the userInviteAccept call

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,95 @@
...
+        journeyId,
+        input.email
+      )
+
+    // TODO: Get email from user in db when we can call api-users
+    if (userInvite != null && input.email === userInvite.email) {
```
storyworks, line 76: Forgot to add accepted and expired checks, doing that now + tests.

apps/api-journeys/src/app/modules/userInvite/userInvite.graphql
```diff
@@ -0,0 +1,34 @@
...
+}
+
+input UserInviteCreateInput {
+  email: String!
+}
+
+extend type Mutation {
+  userInviteCreate(
+    journeyId: ID!
+    senderId: ID!
```
storyworks, line 25: Oops, I think we can remove these now that we're getting via CurrentUserId...

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,130 @@
...
+        userInvite.id,
+        {
+          accepted: true
+        }
+      )
+
+      return updatedInvite
+    }
+
+    return userInvite
```
csiyang, line 109: Do you mind if we do this just so it's a bit more clear?
```suggestion
    else { return userInvite }
```

csiyang (changes requested):

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
storyworks: I would do this, but previous times Tatai has told me just to return.

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Mutation()
+  @UseGuards(
+    GqlAuthGuard,
+    RoleGuard('journeyId', [UserJourneyRole.owner, UserJourneyRole.editor])
+  )
+  async userInviteRemove(
+    @Args('id') id: string,
+    @Args('journeyId') journeyId: string
```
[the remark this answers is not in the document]
storyworks: Needed so the RoleGuard can reference it.

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
```diff
@@ -0,0 +1,139 @@
...
+      await this.userInviteService.getUserInviteByJourneyAndEmail(
+        journeyId,
+        input.email
+      )
+
+    // Create invite if doesn't exist else re-activate removed invite
+    if (userInvite == null) {
+      const journey = await this.journeyService.get<Journey>(journeyId)
+
+      if (journey == null) throw new UserInputError('journey does not exist')
```
csiyang, line 60: can we test this error?

csiyang (changes requested):

csiyang:
Getting circular structure error when trying to test accept, will try to test the accept on https://github.com/JesusFilm/core/pull/1293#pullrequestreview-1282103498

storyworks:
> Getting circular structure error when trying to test accept, will try to test the accept on [#1293 (review)](https://github.com/JesusFilm/core/pull/1293#pullrequestreview-1282103498)

csiyang:
> Getting circular structure error when trying to test accept, will try to test the accept on [#1293 (review)](https://github.com/JesusFilm/core/pull/1293#pullrequestreview-1282103498)

Have tested the acceptAll is working

csiyang (approved):

csiyang (approved):

apps/api-journeys/src/app/modules/userInvite/userInvite.resolver.ts
storyworks: Better to reuse than to create a new one I think.

PR #1295, opened 2023-02-01
feat: add journeyProfile table

storyworks:
# Description

Add a user profile table for api-journeys. 

[Basecamp Todo](https://3.basecamp.com/3105655/buckets/31245893/todos/5783931883)

# How should this PR be QA Tested?

- Would need to run locally and comment out "AuthGuard", and manually pass userId (instead of useCurrentUserId) to test.
- Run seed.
- [ ] Able to make a user profile for userId without profile with `journeyProfileCreate`
- [ ] Able to retrieve a user profile via the userId with `getJourneyProfile`

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged into main

PR #1316, opened 2023-02-10
feat: terms and conditions

PShuckle:
# Description

Add a terms and conditions screen.

[- Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/31245893/todos/5755113861)

- [x] When creating a new account, user is redirected to terms and conditions screen.
- [x] The Terms, EULA, and Community Guidelines buttons all lead to the relevant pages
- [x] The "Next" button is initially deactivated and becomes activated when the checkbox is ticked
- [x] The "Next" button leads to the main page with a list of journeys when clicked
- [x] After logging out and then logging back in, the terms and conditions page is skipped and user is taken straight to the journeys page

# Checklist


apps/api-gateway/schema.graphql
```diff
@@ -549,6 +549,14 @@ input JourneyCreateInput
   slug: String
```
storyworks, line 549: Checkout the feature branch and merge main to it (once Siyang approves https://github.com/JesusFilm/core/pull/1295), to remove these backend changes.

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,148 @@
...
+`
+
+export function TermsAndConditions(): ReactElement {
+  const [accepted, setAccepted] = useState(false)
+  const [journeyProfileCreate] = useMutation<JourneyProfileCreate>(
+    JOURNEY_PROFILE_CREATE
+  )
+
+  return (
+    <Box
```
storyworks, line 39: Use `Stack` mui component whenever you need a container with flex

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,148 @@
...
+              }}
+            />
+          </ListItem>
+        </List>
+        <NextLink href="/" passHref>
+          <Fab
+            variant="extended"
+            size="large"
+            disabled={!accepted}
+            onClick={async () => await journeyProfileCreate()}
```
storyworks, line 115: do `onClick={()=> handleJourneyProfileCreate()}` instead. Generally bad practice to put async functions in onClick functions directly. 

Then take a look to how `RadioQuestionEdit.tsx` to call the await function. You may not need the part within  `update(cache, {data})`. We do that mainly when we need to use the updated data immediately (mainly to display new data).

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.spec.tsx
```diff
@@ -0,0 +1,26 @@
...
+      expect(getByRole('link', { name: 'Next' })).not.toHaveAttribute(
+        'aria-disabled'
+      )
+    )
+    fireEvent.click(getByRole('checkbox'))
+    await waitFor(() =>
+      expect(getByRole('link', { name: 'Next' })).toHaveAttribute(
+        'aria-disabled'
+      )
+    )
```
storyworks, line 24: Add a test that journeyProfileCreate is called next button click. Take a look at the `adds an option on click` test in `RadioQuestionEdit.spec.tsx` to see how we mocked the api call.

storyworks:
Also call `nx run journeys-admin:extract-translations` to update translations files and fix linting.

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,148 @@
...
+              }
+            }}
+          >
+            <ArrowForwardIcon />
+            Next
+          </Fab>
+        </NextLink>
+      </Stack>
+
+      <Typography
```
storyworks, line 133: Use mui `Link` with no underline variant

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.spec.tsx
storyworks: Let's also add a test checking the href's are passed correctly for each button. Even though there's no complex logic with passing - this is the main feature of the page so good to have a test showing the button labeled X matches the href ____/X

apps/journeys-admin/src/components/TermsAndConditions/TermsListItem/TermsListItem.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function TermsListItem({
+  link,
+  icon,
+  text
+}: TermsListItemProps): ReactElement {
+  return (
+    <ListItem disablePadding>
+      <NextLink href={link} passHref>
+        <Button
```
storyworks, line 21: Use ListItemButton? https://mui.com/material-ui/react-list/#basic-list

apps/journeys-admin/src/libs/firebaseClient/initAuth.ts
```diff
@@ -73,10 +73,10 @@ export function initAuth(): void {
         ]
         const allowed = allowedHosts.includes(new URL(redirectUrl).host)
         if (allowed) {
-          return redirectUrl
+          return '/users/terms-and-conditions'
         }
       }
-      return '/'
+      return '/users/terms-and-conditions'
```
storyworks, line 79: This would need manual testing to check no other redirection is broken... I'm a bit short on time atm, but if you've already manually tested - would be good to get a list of what you've checked still works.

PShuckle: After testing the redirects, I realise there are actually ways the user can bypass the terms screen.

Example:
- user signs in and gets redirected to terms screen
- user closes the tab without accepting the terms
- user opens the app again or clicks a link to a journey they've been added to
- since they're already signed in the terms screen doesn't show up and they have full access to the app

I'm going to change this redirect back to the home page and then make every page check if the user has accepted terms and redirect them to the terms screen if they haven't

apps/journeys-admin/src/components/TermsRedirectWrapper/TermsRedirectWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+    getJourneyProfile {
+      id
+      userId
+      acceptedTermsAt
+    }
+  }
+`
+
+interface TermsRedirectWrapperProps {
+  router: NextRouter
```
storyworks, line 18: You can call `useRouter` directly within this component so no need to pass it.

apps/journeys-admin/src/components/TermsRedirectWrapper/TermsRedirectWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+export function TermsRedirectWrapper({
+  router,
+  children
+}: TermsRedirectWrapperProps): ReactElement {
+  const [termsAccepted, setTermsAccepted] = useState(true)
+  const { data, loading } = useQuery<GetJourneyProfile>(GET_JOURNEY_PROFILE)
+
+  useEffect(() => {
+    !termsAccepted && router.push('/users/terms-and-conditions')
+  })
```
storyworks, line 31: Add dependency on `termsAccepted`, `useEffect(()=>{....},[termsAccepted])` so you don't unnecessarily run this check on each render.

apps/journeys-admin/pages/journeys/[journeyId].tsx
```diff
@@ -46,44 +47,46 @@ function JourneyIdPage(): ReactElement {
...
-            value={{ journey: data?.journey ?? undefined, admin: true }}
-          >
-            <PageWrapper
-              title={t('Journey Details')}
-              showDrawer
-              backHref="/"
-              menu={<Menu />}
-              authUser={AuthUser}
-              router={router}
+      <TermsRedirectWrapper router={router}>
```
Comment (storyworks) [inline apps/journeys-admin/pages/journeys/[journeyId].tsx:50]: I like how you're now redirecting via the router and have thought to extract this logic into a wrapper. 

We can also handle this kind of logic in a hook. See the `useUserJourneyOpen` example on line 42 above. 
Similar to the logic you have in the wrapper but it returns void. You call it at the start of each page instead of wrapping and write tests for it. A hook with query example is `useCurrentUser`.

apps/journeys-admin/src/libs/useTermsRedirect/useTermsRedirect.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+`
+
+export function useTermsRedirect(): boolean {
+  const router = useRouter()
+  const { data, loading } = useQuery<GetJourneyProfile>(GET_JOURNEY_PROFILE)
+
+  useEffect(() => {
+    !loading &&
+      data?.getJourneyProfile === null &&
```
storyworks, line 22: I know we said let's not create a journeyProfile until they've accepted so we're always only created profiles with accepted at dates. However, we do leave the option for `acceptedTermsAt` to be null in the model (in case later we want to allow progression to some other page without accepting terms). 

Let's also catch that use case here. So check if `acceptedTermsAt` also isn't null.

apps/journeys-admin/src/libs/useTermsRedirect/useTermsRedirect.ts
```diff
@@ -0,0 +1,27 @@
...
+  query GetJourneyProfile {
+    getJourneyProfile {
+      id
+      userId
+      acceptedTermsAt
+    }
+  }
+`
+
+export function useTermsRedirect(): boolean {
```
storyworks, line 16: Test file for useTermsRedirect()?

It can test if router was or wasn't pushed based on different getJourneyProfile results.

apps/journeys-admin/src/libs/useTermsRedirect/useTermsRedirect.ts
```diff
@@ -0,0 +1,27 @@
...
+  const router = useRouter()
+  const { data, loading } = useQuery<GetJourneyProfile>(GET_JOURNEY_PROFILE)
+
+  useEffect(() => {
+    !loading &&
+      data?.getJourneyProfile === null &&
+      router.push('/users/terms-and-conditions')
+  }, [router, loading, data])
+
+  return !loading && data?.getJourneyProfile !== null
```
storyworks, line 26: What happens if you return nothing here and this Hook just returns `void` ? 

I'm expecting the redirection to still trigger, but I don't know if you get a flash of the destination page occurring before it redirects. If redirection works - even with the flash, make this function return void and remove check `termsAccepted` check around the pages. 

We don't have an error state defined, so better to just display the page normally if `getJourneyProfile` is loading for ages.

PShuckle: Yep, it shows a flash which is why I had the check there but I'll change it back

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <ListItemText
+              primary={
+                <Typography variant="body1" color="secondary.dark">
+                  I agree with listed above conditions and requirements
+                </Typography>
+              }
+            />
+          </ListItem>
+        </List>
+        <Fab
```
storyworks, line 105: Use Button not Fab, our design doesn't have elevation. But also Fab should be used in these use cases: https://m2.material.io/components/buttons-floating-action-button

apps/journeys-admin/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,139 @@
...
+            width: '100%',
+            borderRadius: '12px',
+            bgcolor: 'secondary.dark',
+            color: 'secondary.contrastText',
+            '&:hover': {
+              bgcolor: 'secondary.dark'
+            }
+          }}
+        >
+          <ArrowForwardIcon />
```
storyworks, line 121: Flip the order of arrow / text and fix spacing between them.
Design:
image

Current implementation:
image

storyworks (approved):

storyworks:
We can optimise in another PR if we want to remove the flashes.

PR #1320, opened 2023-02-13
feat: add useCurrentUser hook

storyworks:
# Description

Extract out and create a useCurrentUser hook so we don't need to pass down AuthUser or write out the getCurrentUser query each time.

Used as part of permissions feature.

- [ ] Checks green
- [ ] AccessDialog still works

# Checklist


apps/journeys-admin/src/libs/useCurrentUser/useCurrentUser.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { data, loading } = useQuery<GetCurrentUser>(GET_CURRENT_USER)
+
+  if (!loading) {
+    if (data?.me == null) {
+      throw new Error('Current user cannot be found')
+    }
+    return data.me
+  }
+
+  return { __typename: 'User', id: '', email: '' }
```
storyworks, line 26: In production / testing locally we don't display these data values but rather skeletons when it's loading. Even if did display data, this would appear blank until data is loaded.

 So using this placeholder purely so we can type this hook as always returning a User. This enables us to avoid all the user != null checks that we currently use for journey.

apps/journeys-admin/src/libs/useCurrentUser/useCurrentUser.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }
+`
+
+export function useCurrentUser(): User {
+  const { data, loading } = useQuery<GetCurrentUser>(GET_CURRENT_USER)
+
+  if (!loading) {
+    if (data?.me == null) {
+      throw new Error('Current user cannot be found')
```
storyworks, line 21: This is also highly unlikely ever to throw, since we create a new User with whatever userId is sent to this query (if the user doesn't already exist).

Would only throw if the .save() function in `users.resolvers.ts` returned null.

PR #1376, opened 2023-02-27
fix: access denied edit report

tanflem:
# Description
Before FIX: 
journeys/[id]/edit - 501 error if the user did not have access
journeys/[id]report - allowed the user to view even without access

After FIX:

journeys/[id]/edit - Redirects to journeys/[id] when the user does not have access
journeys/[id]report - Redirects to journeys/[id] when the user does not have access

-   [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31245472/todos/5855598322)

- [ ] Navigate to journeys/[id]/edit on a journey that the current user does not have access to. Should Redirect to journeys/[id] and Request Access.
- [ ] Navigate to journeys/[id]/reports on a journey that the current user does not have access to. Should Redirect to journeys/[id] and Request Access.

# Checklist


apps/journeys-admin/pages/journeys/[journeyId]/edit.tsx
```diff
@@ -67,29 +66,13 @@ function JourneyEditPage(): ReactElement {
...
-        </>
-      )}
     </>
   )
 }
 
 export const getServerSideProps = withAuthUserTokenSSR({
   whenUnauthed: AuthAction.REDIRECT_TO_LOGIN
-})(async ({ AuthUser, locale, query }) => {
+})(async ({ AuthUser, locale, query}) => {
```
Comment (csiyang) [inline apps/journeys-admin/pages/journeys/[journeyId]/edit.tsx:75]: Can we also remove the error check on line 42, its not needed anymore since we added a try catch

csiyang (changes requested):

apps/journeys-admin/pages/journeys/[journeyId]/edit.tsx
Comment (tanflem) [reply apps/journeys-admin/pages/journeys/[journeyId]/edit.tsx]: Pushed

csiyang (approved):

PR #1381, opened 2023-02-28
fix: publisher/[journeyId] no access redirect 

tanflem:
# Description

I am not sure about this solution. From what I could see the reason the component wasn't showing up was a redundant check on `data?.publisherTemplate != null` This check seems redundant to me because if the user doesn't have access does it matter if there is a template or not? This change also does not redirect access denied edit page to /publisher/[journeyId] which was requested in the original to-do.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31245472/todos/5845452183)

- [ ] Navigate to publisher/[journeyId] with a user that does not have access should show the no access component.
- [ ] Navigate to publisher/[journeyId]/edit with a user that does not have access should show no access component.

# Checklist


csiyang:
This is apart of a larger issue where we don't handle non-existent journeyId's being entered into the URL, journeyId's that does exist are working fine.

PR #1396, opened 2023-03-07
feat: embed discovery journeys

csiyang:
# Description

Adds discovery journeys to journey list page. Added some bottom padding to the bottom of the page too

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31822575/todos/5909527009)

- [ ] Should show the discovery journeys embed on the journey list page (vision, how to, feedback)
- [ ] Clicking on each of the 3 embedded cards should open the journey full screen

# Checklist


apps/journeys-admin/src/components/DiscoveryJourneys/DiscoveryJourneys.tsx
```diff
@@ -0,0 +1,37 @@
...
+        justifyContent: 'center',
+        backgroundColor: 'white',
+        border: '1px solid',
+        borderColor: 'divider',
+        borderRadius: 3,
+        mt: 16,
+        p: 4
+      }}
+    >
+      <Stack sx={{ alignItems: 'center' }}>
```
csiyang, line 21: Might make a common component out of these stacks if they get any more complex

storyworks:
Fix the bottom margin on the page then send the demo link to Steven to help inform design?

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -0,0 +1,66 @@
...
+              md: '125%'
+            }
+          }}
+        >
+          <div
+            style={{
+              position: 'relative',
+              width: '100%',
+              overflow: 'hidden',
+              paddingTop: '150%',
```
storyworks, line 42: Where do all of these values come from?

apps/journeys-admin/src/components/JourneyList/JourneyList.tsx
```diff
@@ -71,9 +73,11 @@ export function JourneyList({
               sortOrder={sortOrder}
               router={router}
             />
+            <DiscoveryJourneys />
             {!['archived', 'trashed'].includes(
               (router?.query?.tab as string) ?? ''
             ) && <AddJourneyButton variant="fab" />}
+            <Box sx={{ p: 8 }} />
```
storyworks, line 80: I'll actually fix this with my PageWrapper fix so no need for this here anymore.

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Box
+          sx={{
+            width: {
+              xs: '252%',
+              sm: '168%',
+              md: '125%'
+            }
+          }}
+        >
+          <div
```
storyworks, line 37: `Box` and `sx`, makes it easier to style / maintain in future.

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <Box
+      aria-label={`${slug}-embedded`}
+      sx={{
+        position: 'relative',
+        width: '33%',
+        height: 0
+      }}
+    >
+      <Box
+        sx={{
```
storyworks, line 19: You can remove this `Box` and transfer all of the `sx` within it to the `Box` below with width %s

apps/journeys-admin/src/components/DiscoveryJourneys/DiscoveryJourneys.tsx
```diff
@@ -0,0 +1,21 @@
+import Stack from '@mui/material/Stack'
+import { ReactElement } from 'react'
+import { EmbedJourney } from './EmbedJourney'
+
+export function DiscoveryJourneys(): ReactElement {
+  return (
+    <Stack
+      direction="row"
+      sx={{
+        justifyContent: 'space-evenly',
```
storyworks, line 10: Use `space-between`, `space-evenly` inserts a gap based on screen size around each component so mis-alignment is worse on some screen sizes than others.

apps/journeys-admin/src/components/DiscoveryJourneys/DiscoveryJourneys.tsx
```diff
@@ -0,0 +1,21 @@
...
+        pt: '15px',
+        ml: '-30px',
```
storyworks, line 12: Remove these 2 negative margins and use the iframe fix described below.

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <div
+            style={{
+              position: 'relative',
+              width: '100%',
+              overflow: 'hidden',
+              paddingTop: '150%',
+              paddingLeft: '94%'
+            }}
+          >
+            <iframe
```
storyworks, line 46: image

On some device sizes, the cards are still not aligned, you can fix this by adjusting the `left` value based on the card. 

`left: { xs: 0, sm: slug === 'admin-left' ? -32 : slug === 'admin-right' ? 32 : 0 },` 
No need to use ternary chain if you don't want but you get the idea. 
This also makes mobile padding adjustable if need be.

`const StyledIframe = styled('iframe')(({ theme }) => ({}))` to expose `sx` on iframe. 

So whole file would be:
image

This also relies on `space-between` in DiscoveryJourney.

csiyang: Thanks for the suggestion! Works great, but the gap between the cards is a little big, but will check with design to see what they think

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -0,0 +1,104 @@
+import { styled } from '@mui/material/styles'
+import Box from '@mui/material/Box'
+import { ReactElement } from 'react'
+
+interface Props {
+  slug: 'admin-left' | 'admin-right' | 'admin-center'
+}
+
+export function EmbedJourney({ slug }: Props): ReactElement {
+  const width = {
```
storyworks, line 10: Hey, well done for getting this to look decent with embeds and writing out the code like this so styling can be as simple as possible. I do agree that it is faster to just use the page itself and am ok to approve if UX is fine with the blinking.

I think for next time, if we are not so pushed for time - a more stable implementation would have been to research using EmbedPreview directly. This would look like:

1) In the DiscoveryJourneys component, we query the 3 journeys needed and get their blocks.
2) Pass the blocks down to EmbedPreview components to render them.
3) Extend EmbedPreview with a `noPadding` type of prop which removes the border around each card
4) Style blocks accordingly.

Even though this is an extra 3 api calls on first load, I think the trade off would be worth it. In that we wouldn't have such complex styling (no hard-coded widths / spacing per breakpoint) and possibly wouldn't have the blinking issue.

storyworks (approved):
Just let the team know embeds are going to be massive until template side panel gets added in.

And have we created cooldown todos for the blinking / span?

PR #1400, opened 2023-03-09
fix: title of journey from template

storyworks:
# Description

We get "copy 0" whenever we use a template from the template library. 
We want it just to display the template name, or add copy only if another journey with the template name already exists in our journey list.
image

[- Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/31822575/todos/5909445092)

- [x] Go to the template library and use a template to create a journey. The created journey should have the same name as the template.
- [x] Repeat the step above, this time the created journey should have the template name + "copy" at the end.

# Checklist

### DIFF
--- a/apps/api-journeys/src/app/modules/journey/journey.resolver.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.resolver.ts
@@ -284,8 +284,12 @@ export class JourneyResolver {
       journey.title
     )
     const duplicateNumber = this.getFirstMissingNumber(duplicates)
-    const duplicateTitle = `${journey.title} copy ${
-      duplicateNumber === 1 ? '' : duplicateNumber
+    const duplicateTitle = `${journey.title}${
+      duplicateNumber === 0
+        ? ''
+        : duplicateNumber === 1
+        ? ' copy'
+        : ` copy ${duplicateNumber}`
     }`.trimEnd()
 
     const slug = slugify(duplicateTitle, {
--- a/apps/api-journeys/src/app/modules/journey/journey.resolver.spec.ts
+++ b/apps/api-journeys/src/app/modules/journey/journey.resolver.spec.ts
@@ -69,6 +69,13 @@ describe('JourneyResolver', () => {
     createdAt
   }
 
+  const template: Journey = {
+    ...journey,
+    title: 'template',
+    id: 'templateJourneyId',
+    template: true
+  }
+
   const draftJourney = {
     ...journey,
     id: 'draftJourney',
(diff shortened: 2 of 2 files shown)


apps/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -284,8 +284,12 @@ export class JourneyResolver {
       journey.title
     )
     const duplicateNumber = this.getFirstMissingNumber(duplicates)
-    const duplicateTitle = `${journey.title} copy ${
-      duplicateNumber === 1 ? '' : duplicateNumber
+    const duplicateTitle = `${journey.title}${
```
csiyang, line 287: Not the biggest fan of chaining ternaries like this, makes it kinda hard to read. Thoughts on splitting the logic out abit?

storyworks: Personally I think 2 levels formatted like this is as readable as a switch statement. Eg first line is the case, next line the result, then case, then result, last line is default.

What would you change it to?

csiyang: Like just splitting out one of the ternaries

```
  const x = duplicateNumber > 1 ? duplicateNumber : ''
  const duplicateTitle = duplicateNumber === 0 ? '' : ` copy ${x}`
```

storyworks: Your one you've added an extra " " when duplicate number is 1. Even if you rewrite it to:

` const x = duplicateNumber > 1 ? ` ${duplicateNumber}` : ''`
I think that's still more complex that what I have :/

csiyang: Your trim should handle any trailing spaces right? Its easier to read and understand, but its just what I would do.

storyworks: Agree to disagree?

csiyang: sure lol

csiyang (approved):

apps/api-journeys/src/app/modules/journey/journey.resolver.ts
storyworks: Also I take back, I have nested the ternaries but only 1 level. We haven't enabled `/*eslint no-nested-ternary: "error"*/` so I've not violated our own linting standards. 

If we want to impose this rule, then it's quite easy to do but has not been done.
I do get where people are going when they say don't nest excessively and would agree that 2 levels or 1 level but super complex logic would make it difficult to read. But I think 1 level especially for something so simple is better than splitting up logic just for the sake of never nesting ternaries.

PR #1406, opened 2023-03-13
refactor: nav drawer remove redundant router and title

storyworks:
# Description

Fixing this before I add in my PageWrapper changes. Starting PageWrapper changes so we can easily add in the side panel on the landing page.

2 Changes:
- Use useRouter directly in NavigationDrawer instead of passing `router` down 2 levels
- remove unused `title` prop

- [ ] Checks are green
- [ ] Drawer still highlights correct page
- [ ] Drawer still shows correctly on desktop and mobile

# Checklist

### DIFF
--- a/apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
+++ b/apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
@@ -157,7 +157,9 @@ export function NavigationDrawer({
         <NavigationListItem
           icon={<ViewCarouselRoundedIcon />}
           label="Discover"
-          selected={selectedPage === 'journeys' || selectedPage === ''}
+          selected={
+            selectedPage === 'journeys' || selectedPage === '' || router == null
+          }
           link="/"
           tooltipText={journeyTooltip}
         />
--- a/apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
+++ b/apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
@@ -16,7 +16,7 @@ import ShopRoundedIcon from '@mui/icons-material/ShopRounded'
 import ShopTwoRoundedIcon from '@mui/icons-material/ShopTwoRounded'
 import Backdrop from '@mui/material/Backdrop'
 import Image from 'next/image'
-import { NextRouter } from 'next/router'
+import { useRouter } from 'next/router'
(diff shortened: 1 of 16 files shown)


apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
```diff
@@ -160,7 +157,9 @@ export function NavigationDrawer({
         <NavigationListItem
           icon={<ViewCarouselRoundedIcon />}
           label="Discover"
-          selected={selectedPage === 'journeys' || selectedPage == null} // null for when page is index. UPDATE when we add the actual index page
```
csiyang, line 163: can we keep the comment? I still don't think its super clear why `selectedPage == null` is there for otherwise.

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
```diff
@@ -160,7 +157,9 @@ export function NavigationDrawer({
...
+          selected={
+            selectedPage === 'journeys' || selectedPage === '' || router == null
```
csiyang, line 161: why check `router == null` here? If router is null it would be an error right?

csiyang (changes requested):

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
storyworks: Actually, I tested and when the page is index, it was correctly returning "/".

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
storyworks: Router is null in stories. I did look into using https://github.com/scottrippey/next-router-mock but it didn't work initially. Possibly I'd need more time to debug to get it to work.

If the router was jacked in main, I'm not sure we'd even be able to get to this page :/ So mainly added this here to say, if it's null for whatever reason - make "journeys" the selected icon by default.

But true, maybe this is better to push through more thoroughly after mocking properly works.

apps/journeys-admin/src/components/PageWrapper/NavigationDrawer/NavigationDrawer.tsx
csiyang: oh sounds good then

PR #1412, opened 2023-03-14
refactor: page wrapper

storyworks:
# Description

As we add more pages, spacing between content & layout of content has started to diverge and unnecessarily duplicated.

We want to create a PageLayout wrapper which handles all the responsive layout / spacing concerns so devs can focus on creating the components in the page.

We would insert new components into the PageWrapper as children, bottomPanelChildren or SidePanelChildren. 
 SidePanelContainer has also been created to manage spacing and the divider between side panel children.

This PR just creates the new page wrapper and refactors the landing page to use it.

Future PRs:
Refactor more pages to use the new page wrapper. 

[Refactor layout](https://3.basecamp.com/3105655/buckets/29494894/todos/5369618526)
[Add bottom padding to JourneyList](https://3.basecamp.com/3105655/buckets/29494894/todos/5373733263)
[Improve JourneyDetails layout](https://3.basecamp.com/3105655/buckets/29494894/todos/5373737029)

General
- [ ] The whole page itself is fixed height / position and cannot be scrolled when content overflows
- [ ] We can configure page to have side panel and bottom panel
- [ ] Bottom panel has a top border 
- [ ] Side panel containers are stacked vertically (doesn't take up full height of screen)
---
Desktop
- [x] NavBar is fixed width of 72px
- [ ] SidePanel is fixed width of 327px
- [x] MainBody takes up the rest of the screen width
- [x] Main body has own scroll for overflow content
- [x] Main header has  12px left/right padding
- [ ] Main header allows back button and action menu
- [ ] Side panel has own scroll for overflow content
- [ ] There's a divider / border between main and side panel
- [ ] Side panel header has  12px left/right padding
---
Mobile - portrait
- [x] AppBar is fixed to top of screen
- [x] Main header is fixed beneath app header
- [ ] Side panel content appears in drawer
- [x] 24px left/right padding on AppHeader and MainHeader

Deprecated:
- [ ] Main content and side panel content display stacked with no divider between them
- [ ] We have 1 scrollbar for main content + side panel content  (no seperate scrolls like desktop)
---
MainPanelBody
- [x] Has 32px left/right padding on desktop 
- [x] Has 36px top/bottom padding on desktop 
- [x] Has 24px padding on mobile

BottomPanel
- [ ] Has no padding

SidePanelContainer
- [ ] Has 24px left/right padding on desktop & mobile 
- [ ] Has 16px top/bottom padding on desktop & mobile

# Checklist


apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+        data-testid="side-header"
+        position="sticky"
+        color="default"
+        sx={{
+          display: { xs: 'none', sm: 'flex' }
+        }}
+      >
+        <Toolbar variant="dense">
+          <Typography variant="subtitle1" component="div" noWrap>
+            {title}
```
csiyang, line 35: what happens if title is null?

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { FlagsProvider } from '@core/shared/ui/FlagsProvider'
+import { PageWrapper } from '.'
+
+jest.mock('@mui/material/useMediaQuery', () => ({
+  __esModule: true,
+  default: jest.fn()
+}))
+
+describe('PageWrapper', () => {
+  describe('MainPanel', () => {
```
csiyang, line 13: Can we move each of the secondary describes into test files in their respective components?

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -123,7 +123,6 @@ export function PageWrapper({
         open={open}
         onClose={setOpen}
         authUser={authUser}
-        title={title}
```
csiyang, line 126: why is this change being made in this PR

csiyang (changes requested):
Great job! Excited to use this for all the pages when it gets in!. Mostly just wondering why all the tests got grouped together?

apps/journeys-admin/src/components/NewPageWrapper/MainPanelHeader/MainPanelHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <AppBar
+        color="default"
+        sx={{
+          position: { xs: 'fixed', sm: 'sticky' },
+          top: { xs: toolbarStyle.height, sm: 0 }
+        }}
+      >
+        <Toolbar variant="dense">
+          <Stack direction="row" flexGrow={1} alignItems="center">
+            {backHref != null && (
```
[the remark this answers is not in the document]
storyworks: It's tested in PageWrapper

apps/journeys-admin/src/components/NewPageWrapper/MainPanelHeader/MainPanelHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+            )}
+            <Typography
+              variant="subtitle1"
+              component="div"
+              noWrap
+              sx={{ flexGrow: 1 }}
+            >
+              {title}
+            </Typography>
+            {menu != null && menu}
```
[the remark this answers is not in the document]
storyworks: It's tested in PageWrapper

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+          </Typography>
+        </Toolbar>
+      </AppBar>
+
+      {/* SidePanelBody - children must be wrapped in SidePanelContainer */}
```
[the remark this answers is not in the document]
storyworks: There could be multiple children needing to be individually wrapped. 
Easier to get people to wrap SidePanelContainer as needed themselves than to refactor all the components to pass an array of ReactNodes which we then parse and wrap ourselves.

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
storyworks: It just doesn't show, I can refactor so header has no children. But we still need the header here as a placeholder.

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
storyworks: I thought about that but realised it was redundant. 

For instance if I test that the title shows in MainPanel, I'm checking that "title" does render when passed in.
If I write another test in PageWrapper checking that title is correctly passed to MainPanel, I check it works by seeing that the title renders. In this case, I'm also confirming that `title` does render when passed in. Hence, the first test is made redundant.

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
storyworks: Left over from when I initially was replacing PageWrapper. Tis a small change that removed a redundant property that seems not worth the effort to revert.

apps/journeys-admin/src/components/NewPageWrapper/AppHeader/AppHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <Box id="app-header" sx={{ display: { sm: 'none' } }}>
+      <AppBar
+        role="banner"
+        position="fixed"
+        sx={{
+          display: { xs: 'flex', sm: 'none' },
+          backgroundColor: 'secondary.dark'
+        }}
+      >
+        <Toolbar variant={toolbarStyle.variant}>
```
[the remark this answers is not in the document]
storyworks: Tested via VR

apps/journeys-admin/src/components/NewPageWrapper/AppHeader/AppHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <Stack direction="row" flexGrow={1} alignItems="center">
+            <IconButton
+              size="large"
+              edge="start"
+              color="inherit"
+              aria-label="open drawer"
+              onClick={onClick}
```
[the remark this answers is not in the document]
storyworks: Tested in PageWrapper `should show the app header`

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
storyworks: Originally, all these components lived within the PageWrapper component. An argument could be made that they should all stay within that component since none of these sub components are reused or have any extra special functionality beyond rendering. 

I've split it out just because PageWrapper would get too big having them all in there.

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
csiyang: oh yup, i think its fine as is then

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
csiyang: Hmm I see your point. I have a few concerns though (feel free to push back on these)

- Because the components are all split out, its kinda starting to look like a integration test. 
- Usually a component that has this many sub components, each of the subcomponents have their own unit tests. 
- It makes this entire component kinda hard to update/fix in general

Thoughts on keeping the unit tests in their own components, and just test that page wrapper can render the sub components correctly. Eg test title in main panel, and test page wrapper renders main panel (so no retest in pagewrapper)

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
csiyang: I think in your other PR you also tidy up the need for it to be passed in on the pages. I think best to keep these changes separate

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
storyworks: Agreed this has not been done widely and that yes usually a structure like this would include unit tests for each component.

Not sure I see what you mean in the 3rd point.

> just test that page wrapper can render the sub components correctly. Eg test title in main panel, and test page wrapper renders main panel (so no retest in pagewrapper)

What would be the assertion that you make to "test page wrapper renders main panel" - would this also test that we're correctly passing the all the props through?

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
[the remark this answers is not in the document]
csiyang: oh, yeah def not ideal to pass in an array of components. How do you distinguish when each child needs to be individually wrapped? Maybe the comment should also detail that?

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
csiyang: Third point as in since the structure is different to how we build other stuff, it may be confusing for someone else coming into it for the first time. And also if we ever did need to reuse parts of it, it would be more work to refactor.

For the test you could just do `expect(getByTestId("main-body")).toBeInTheDocument()` anything that has logic applied to it in page wrapper would still need to be tested though, but that only looks true for `toolbarStyle`, everything else looks like its being passed straight through (couldn't find your `.toUpperCase()` example)

storyworks: Yes agreed the structure is different, happy to move it all back into the same file to preserve our current pattern. I thought this would be a worthy exception to the rule. That way it’s also more clear we don’t need duplicate tests.

Your suggestion would capture that the main element renders but it wouldn’t capture any changes to how we pass in the props. Eg right now, we pass the title in directly, but if we make it all uppercase in the future, that won’t be picked up. That’s an easy example which can be picked up via VR but you get to point.

csiyang: As I said `anything that has logic applied to it in the page wrapper would still need to be tested`, if more logic gets added then that logic should get tested in the place it got added

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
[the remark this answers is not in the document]
storyworks: The container itself is adds the padding and bottom border. It depends on the design when it’s needed. In my storybook demo, I show a couple of children components wrapped in the container. One container doesn’t have a border for demo purposes.

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
storyworks: `anything that has logic applied to it in the page wrapper would still need to be tested` - exactly and in this case we would be double testing the rendering of the sub-component hence it's redundant to test there, when you can more throughly test each prop at the PageWrapper level once. 

In this case, you also avoid having to update the tests if logic gets added because it's already tested.

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
storyworks: Oh actually, NavigationWrapper no longer takes `title` - my new PageWrapper also references this Nav wrapper and doesn't pass it in. Need to remove here too else we get a linting error.

csiyang: what about all the other pages that are still passing in title to page wrapper? If you really want to remove it here, then update the other pages too?

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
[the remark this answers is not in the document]
csiyang: It's not clear from the code and comment when the container needs to be used, should it wrap all children or each individual children and when. Just update your comment to explain, or atleast point towards storybook for an example. 

Your comment makes it sound like just the children needs to be wrapped, kinda confusing

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
storyworks: This change only occurs because we've removed `title` from NavigationDrawer. It doesn't remove `title` from all the old PageWrapper components since that component will be deleted during cooldown and all other pages which use that will have `title` removed then.

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
csiyang: Interesting video by Kent C. Dodds talking about this very thing https://www.youtube.com/watch?v=0qmPdcV-rN8

tldr: its not that big of a deal (lol), test in parent if you can, but if there is logic in the child, then the child should have a unit test

Are we happy to follow this pattern? If we do, I think just `MainPanelHeader` and `MainPanelBody` needs unit tests for their conditional rendering

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
csiyang: mhm, i guess its technically related, can resolve this then. 

But it would be good to keep changes like this seperate in the future

apps/journeys-admin/src/components/NewPageWrapper/PageWrapper.spec.tsx
storyworks: Commented in Slack but I said I ended up adding tests for all child components. If we want to optimise in the future, we can - but this way it's less confusing for new people - some components having tests and others not.

This makes it consistent with SidePanelContainer which has a "renders children" test. We still haven't made a decision whether we want to include these types of "render props" tests so have added it since it's easier to remove later if we decide so.

storyworks: ^ Followed this and also added to `SidePanel` since I'm now conditionally rendering the `Typography` component. 

I still think it's redundant in this case but it's easier to follow this blanket rule and have redundancies. I guess if testing does start to take ages, these "render children" type of tests (which have their logical conditions tested in VR) can be first on the chopping block :)

csiyang:
QA'd! Ticked off everything that I could test through the stage link. Everything else I tested through storybook (but leaving it unticked since I couldn't test in stage)

Just one issue where on mobile where the bottom panel is covering other content. 
image

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+            width: '100%',
+            height: `calc(100% - ${toolbar.height}px)`
+          }
+        }}
+      >
+        <DrawerContent
+          title={title}
+          action={
+            <IconButton
+              onClick={handleClose}
```
storyworks, line 120: After removing `useMediaQuery` from `Drawer.tsx` we no longer have a way to test mobile implementation in unit tests. So can't check that the Drawer closes on icon click.

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+      </AppBar>
+      <Stack
+        data-testid="side-body"
+        border="hidden"
+        sx={{
+          overflow: 'none',
+          overflowY: { sm: 'auto' }
+        }}
+      >
+        {children}
```
csiyang, line 57: Is there no more need to individually wrap some of the children now?

storyworks: See the story, we should still wrap the children if we want the default padding  / border added to it. But some children components (eg ListItemButton) might take up the full width. So in those cases you won't need to wrap. The comment below

``` 
/**
   * Add default side panel padding and border by wrapping components with `SidePanelContainer`
   */
   ```
   
   Is still generic enough to handle this use case.

csiyang: There still a comment in the stories about needing it (`PageWrapper.stories` line 134)

csiyang:
main panel content is still being cut off by the bottom panel 
image

apps/journeys-admin/src/components/NewPageWrapper/SidePanel/SidePanel.tsx
storyworks: Ah good catch!

csiyang (approved):

PR #1418, opened 2023-03-16
fix: visitor creation id

csiyang:
# Description

id was manually getting created while _key was being automatically created, this was desyncing the id's causing the frontend to have the incorrect visitorId. 

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31822546/todos/5955732959)

- [x] Visit a journey in the journeys project, (using a journey that allows you to enter a name can help you to find the correct visitor object)
- [x] Check database, a new visitor should get created, the new visitor should not have an id field in the table

# Checklist


apps/api-journeys/src/app/modules/visitor/visitor.resolver.ts
```diff
@@ -9,10 +9,13 @@ import {
...
 import { ForbiddenError, UserInputError } from 'apollo-server-errors'
 import { IResult, UAParser } from 'ua-parser-js'
-import { Event, VisitorsConnection } from '../../__generated__/graphql'
+import { Event, VisitorsConnection, Visitor } from '../../__generated__/graphql'
 import { EventService } from '../event/event.service'
 import { MemberService } from '../member/member.service'
-import { VisitorService, VisitorRecord } from './visitor.service'
+import {
+  VisitorService
+  // VisitorRecord
```
csiyang, line 17: remove comment

apps/api-journeys/src/app/modules/visitor/visitor.service.ts
```diff
@@ -95,7 +80,6 @@ export class VisitorService extends BaseService<VisitorRecord> {
       ).next()
 
       visitor = await this.collection.save({
-        id: uuidv4(),
```
csiyang, line 98: this was the line causing the problem. We can also set _key here, but thought it would be better to just use the generated type.

apps/api-journeys/src/app/modules/visitor/visitor.service.ts
```diff
@@ -19,20 +15,9 @@ interface ListParams {
   sortOrder?: 'ASC' | 'DESC'
 }
 
-export interface VisitorRecord {
```
csiyang, line 22: Removed this since the refactor to use the generated type. Not too sure if there was any intention making this a custom interface rather than using the generated type. Just want to check with @tataihono and see if removing this is ok?

tataihono: Please no.

apps/api-journeys/src/app/modules/visitor/visitor.service.spec.ts
```diff
@@ -145,6 +145,7 @@ describe('VisitorService', () => {
       await service.getByUserIdAndJourneyId('user.id', 'team.id')
       expect(service.collection.save).toHaveBeenCalledWith({
         ...visitorWithId,
+        _key: 'newVisitor.id',
```
csiyang, line 148: From what I can gather from the [docs](https://www.arangodb.com/docs/stable/data-modeling-documents-document-methods.html#insert--save) we can just add it here

apps/api-journeys/src/app/modules/visitor/visitor.service.ts
```diff
@@ -94,8 +94,10 @@ export class VisitorService extends BaseService<VisitorRecord> {
       `)
       ).next()
 
+      const id = uuidv4()
       visitor = await this.collection.save({
-        id: uuidv4(),
+        _key: id,
```
csiyang, line 99: From what I can gather from the [docs](https://www.arangodb.com/docs/stable/data-modeling-documents-document-methods.html#insert--save) we can just add the key here

apps/api-journeys/src/app/modules/visitor/visitor.service.ts
```diff
@@ -94,8 +94,10 @@ export class VisitorService extends BaseService<VisitorRecord> {
...
+        id,
```
tataihono, line 100: do we need this then?

csiyang: yeah, need it so that in event service (line25) the return type won't have the return type still returns with an id since all the different event resolvers still using `visitor.id` (i guess if we really wanted to we can replace all those instances with `visitor._key`)

tataihono: I don't think thats the case. That's was @keyAsId does.

tataihono (approved):

PR #1421, opened 2023-03-19
feat: common icon button

tanflem:
# Description

Added a common component for the icon button. Created stories for said component.
image

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/31822575/todos/5935534867#__recording_5945605020)

The component is currently not used anywhere but the stories can be tested and the tests can be run. Ideally component is tested when added to the template side bar.

- [ ] Run tests
- [ ] Visually check story book

# Checklist


apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import Typography from '@mui/material/Typography'
+import AddRoundedIcon from '@mui/icons-material/AddRounded'
+import Card from '@mui/material/Card'
+import CardActionArea from '@mui/material/CardActionArea'
+import Stack from '@mui/material/Stack'
+import SvgIcon from '@mui/material/SvgIcon'
+import { ImageBlockThumbnail } from '../Editor/ImageBlockThumbnail'
+
+interface ContainedIconProps {
+  icon?: typeof SvgIcon
```
tanflem, line 12: Here I changed from ReactNode type to SvgIcon in order to use the ImageBlockThumnail component. However, I have a reworked version that does not use that component. I was not sure which would be better.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,39 @@
+import { ReactElement } from 'react'
+import Box from '@mui/material/Box'
+import Typography from '@mui/material/Typography'
+import AddRoundedIcon from '@mui/icons-material/AddRounded'
+import Card from '@mui/material/Card'
+import CardActionArea from '@mui/material/CardActionArea'
+import Stack from '@mui/material/Stack'
+import SvgIcon from '@mui/material/SvgIcon'
+import { ImageBlockThumbnail } from '../Editor/ImageBlockThumbnail'
```
tanflem, line 9: Should I be using this component?

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
```diff
@@ -0,0 +1,35 @@
+import { render, fireEvent } from '@testing-library/react'
```
tanflem, line 1: This was my first time making tests. I looked at some other files as examples. If there are things I need to add I would appreciate a bit of explanation of the workflow when creating these tests to cover all bases.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { Meta, Story } from '@storybook/react'
```
tanflem, line 1: The same goes for stories as tests. This was my first time, and I am unsure if I covered all the bases. Would appreciate a short process explanation.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
storyworks: Not directly, but you'd probably be using much of the same logic. Take a look at https://docs.core.jesusfilm.org/docs/basics/frontend/component-library/#our-components for some context :) 

Essentially we have an existing component 
image

And in our designs, we have a new button which looks and functions almost exactly the same. But we can't use the existing component since it has some built in logic specific to it's current use case (eg ImageBlockThumbnail requires a selectedBlock to display the image source.)

So we want to built a common component that can render both our new design and the existing one.

Sorry I forgot to add loading / error icon use cases to the todo like how they have in ImageBlockThumbnail.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
storyworks: I suggested `ReactNode` so in the future to use it, we can pass <OurCustomIcon/> and it will just render whatever we pass it. <OurCustomIcon/> will actually be a component we build to standardise how we use icons. 

We could reuse image thumbnail, but I'd recommend refactoring it first in another PR. eg:
1) Move ImageBlockThumbnail outside of Editor and as a direct component within journey-admins components. (since it's used outside of the Editor now too)
2) Rename to ImageThumbnail since it displays images not just for blocks
3) split out `selectedBlock` field into `imageSrc` and `imageAlt` props. 

I'd still like to use `ReactNode` in this refactor just to support future icon use.

Tests / stories would then also need to be updated.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
storyworks: https://docs.core.jesusfilm.org/docs/basics/frontend/storybook/ 

Let me know if there are any questions after that :) Would be nice to refine the docs.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
storyworks: We don't have docs for this yet but looks like you're on the right track :) 

Generally we test interaction logic like onClick or any additional display logic in the component (eg disabled ? items[0]: items => "it shows only the first item in the list when disabled")

The goal is to have basic interactions, low level logic tested by Unit tests. CSS and other simple display logic (eg that all elements appear on the page when passed) tested by VR and complex inter-component interactions tested via E2E (when it's up).

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onClick: () => void
+}
+
+export function ContainedIconButton({
+  icon,
+  label,
+  onClick
+}: ContainedIconProps): ReactElement {
+  return (
+    <>
```
storyworks, line 23: React only allows 1 child node to be returned for the component function. 

So if we had something like 
```
Component(){
return (<div/><div/>)
}
```

It would error saying we're passing 2 nodes and it only accepts one. Fragments (`<> </>`) can be used to wrap these 2 divs to tell the compiler to treat them as 1.

In this case the entire component is encompassed by `<Card/>` so the extra fragments are redundant. There are a couple of places in our code base that I've seen redundant fragments. Feel free to remove them during cooldown, they should not be in there.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
csiyang: Also noticed the existing button only has the icon clickable, but the common component, the entire button is clickable. Did UX mention what behaviour they wanted?

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+              sx={{
+                textOverflow: 'ellipsis',
+                whiteSpace: 'nowrap',
+                overflow: 'hidden'
+              }}
+            >
+              {label}
+            </Typography>
+
+            {description !== undefined && (
```
csiyang, line 55: The check below will check both null and undefined https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality

```suggestion
            {description != null && (
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  label: 'label',
+  description: undefined,
+  imageSrc: undefined,
+  imageAlt: undefined,
+  loading: false,
+  onClick: () => console.log('Clicked')
+}
+
+export const Image = Template.bind({})
+Image.args = {
```
csiyang, line 40:
```suggestion
export const Complete = Template.bind({})
Complete.args = {
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const Default = Template.bind({})
+Default.args = {
+  thumbnailIcon: <NoteAddIcon />,
+  actionIcon: <AddIcon />,
+  label: 'label',
+  description: undefined,
+  imageSrc: undefined,
+  imageAlt: undefined,
+  loading: false,
+  onClick: () => console.log('Clicked')
```
csiyang, line 36: we can use noop in this case since we don't need to worry about what the on click is doing. You can import it from the lodash library (its in our packages already)

```suggestion
  onClick: noop
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+Image.args = {
+  thumbnailIcon: <NoteAddIcon />,
+  actionIcon: <EditIcon />,
+  label: 'Video Title',
+  description: 'descripition',
+  imageSrc:
+    'https://d1wl257kev7hsz.cloudfront.net/cinematics/2_Acts7302-0-0.mobileCinematicHigh.jpg',
+  imageAlt: undefined,
+  loading: false,
+  onClick: () => console.log('Clicked')
```
csiyang, line 49:
```suggestion
  onClick: noop
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const Loading = Template.bind({})
+Loading.args = {
+  thumbnailIcon: <NoteAddIcon />,
+  actionIcon: <AddIcon />,
+  label: 'label',
+  loading: true,
+  onClick: () => console.log('Clicked')
```
csiyang, line 58:
```suggestion
  onClick: noop
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+)
+
+export const Default = Template.bind({})
+Default.args = {
+  thumbnailIcon: <NoteAddIcon />,
+  actionIcon: <AddIcon />,
+  label: 'label',
+  description: undefined,
+  imageSrc: undefined,
+  imageAlt: undefined,
```
csiyang, line 34: Not needed
```suggestion
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ContainedIconButton } from './ContainedIconButton'
+
+describe('ContainedIconButton', () => {
+  const onClick = jest.fn()
+  it('should render button', async () => {
+    const { getByRole } = render(
+      <ContainedIconButton
+        thumbnailIcon={<NoteAddIcon />}
+        label="Label"
+        onClick={onClick}
```
csiyang, line 13:
```suggestion
        onClick={noop}
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+  //         thumbnailIcon={<NoteAddIcon />}
+  //         label="Label"
+  //         onClick={onClick}
+  //         actionIcon={<AddIcon />}
+  //       />
+  //     )
+  //     expect(getByRole('imageSrc')).toHaveTextContent(
+  //       'https://d1wl257kev7hsz.cloudfront.net/cinematics/2_Acts7302-0-0.mobileCinematicHigh.jpg'
+  //     )
+  //   })
```
csiyang, line 48: Remove comment
```suggestion
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+        label="Label"
+        onClick={onClick}
+        actionIcon={<AddIcon />}
+      />
+    )
+    fireEvent.click(getByRole('button'))
+
+    expect(onClick).toHaveBeenCalled()
+  })
+
```
csiyang, line 34: Add a test for the description since its conditionally rendered

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface ContainedIconProps {
+  thumbnailIcon: ReactNode
+  actionIcon: ReactNode
+  label: string
+  description?: string
+  imageSrc?: string
+  imageAlt?: string
+  loading?: boolean
+  onClick: () => void
```
csiyang, line 18: Small nit pick, but this should be called handleClick, also update for the rest of the document
```suggestion
  handleClick: () => void
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+import Typography from '@mui/material/Typography'
+import Card from '@mui/material/Card'
+import CardActionArea from '@mui/material/CardActionArea'
+import Stack from '@mui/material/Stack'
+import Icon from '@mui/material/Icon'
+import { ImageThumbnail } from '../ImageThumbnail'
+
+interface ContainedIconProps {
+  thumbnailIcon: ReactNode
+  actionIcon: ReactNode
```
csiyang, line 12: Lets make these optional, and fill it with some icons by default
```suggestion
  thumbnailIcon?: ReactNode
  actionIcon?: ReactNode
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  description?: string
+  imageSrc?: string
+  imageAlt?: string
+  loading?: boolean
+  onClick: () => void
+}
+
+export function ContainedIconButton({
+  thumbnailIcon,
+  actionIcon,
```
csiyang, line 23: The ImageThumbnail already handles a default icon if one is not given. Once you do this

- on stories, for the default story, you should no longer need to pass in thumbnailIcon and actionIcon
- on tests, add a test to see if a custom `actionIcon` will show.

```suggestion
  thumbnailIcon,
  actionIcon = <AddIcon />,
```

csiyang (changes requested):
Good job, almost there! Just a few changes to be made =)

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: Current component uses React Node.
ImageBlockThumbnail refactored here.
https://github.com/JesusFilm/core/pull/1442

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: Refactored here.
https://github.com/JesusFilm/core/pull/1442

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
tanflem: fixed in d8e8bf0109b1e043cd7b28720c33560c89eab24c

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
tanflem: fixed in d8e8bf0109b1e043cd7b28720c33560c89eab24c

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: fixed in cc28faa3eac2c914812f6794c2f77c31032c2471

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: fixed in 1dca4540ebe0a2853e4ff46a2f5f024b0153a5d0

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: fixed in 1dca4540ebe0a2853e4ff46a2f5f024b0153a5d0

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: fixed in 1dca4540ebe0a2853e4ff46a2f5f024b0153a5d0

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
tanflem: added in 5f8047e2ba7d0fc3cd7c2ee05a0ffeb9bfadc894

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <ContainedIconButton
+        thumbnailIcon={<NoteAddIcon />}
+        label='Label'
+        handleClick={onClick}
+        actionIcon={<AddIcon />}
+      />
+    )
+    fireEvent.click(getByRole('button'))
+
+    expect(onClick).toHaveBeenCalled()
```
tanflem, line 33: I was getting an error here when trying `expect(noop).toHaveBeenCalled`

Is there a way to do this with noop or is it correct to use the jest function?

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+                  textOverflow: 'ellipsis',
+                  whiteSpace: 'nowrap',
+                  overflow: 'hidden'
+                }}
+              >
+                {description} &nbsp;
+              </Typography>
+            )}
+          </Box>
+          <Icon color="primary" sx={{ pr: 8 }}>
```
tanflem, line 69: This is the information I have about the padding. Does this look correct? Is this something that is worth posting in the shared channel and asking the designers?

Screenshot 2023-03-28 at 1 19 55 PM

Screenshot 2023-03-28 at 8 41 25 AM

This is how I currently have it. 
Screenshot 2023-03-28 at 8 46 51 AM

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.spec.tsx
csiyang: Yes this one needs to stay as a jest function, noop is for when we don't really care about what the function is doing

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
csiyang: Just checked the design, yep the padding is good. Its off by like 1px, but thats ok since we should use spacing tokens (1 spacing token = 4px)

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const Template: Story<ComponentProps<typeof ContainedIconButton>> = ({
+  ...args
+}) => (
+  <Paper elevation={0} sx={{ p: 2 }}>
+    <Stack direction="row">
+      <ContainedIconButton {...args} />
+    </Stack>
+  </Paper>
```
csiyang, line 25: We usually use paper for stories that need the background/ needs it to provide context for multiple components.
```suggestion
  <Box sx={{maxWidth: '300px'}}>
      <ContainedIconButton {...args} />
    </Box>
```

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+)
+
+export const Default = Template.bind({})
+Default.args = {
+  thumbnailIcon: <NoteAddIcon />,
```
storyworks, line 30: Only pass in the required props in default. Default should show us the state of the component when we pass in the bare minimum / don't customise it in any extra way.

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+  loading: false,
+  onClick: noop
+}
+
+export const Complete = Template.bind({})
+Complete.args = {
+  thumbnailIcon: <NoteAddIcon />,
+  actionIcon: <EditIcon />,
+  label: 'Video Title',
+  description: 'descripition',
```
storyworks, line 42: typo :)

tanflem: fixed in b391cafd3f6e38079bdd51e9ac4f5d7f20714d4b

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.tsx
tanflem: back to onClick in 057027edb949c85f092aaa1c431a7a891cc24854

apps/journeys-admin/src/components/ContainedIconButton/ContainedIconButton.stories.tsx
tanflem: fixed in f05592296ac42d6ff6667e8ed9e238d75e4b83c1

csiyang (approved):

PR #1449, opened 2023-03-29
fix: language filter refactor

csiyang:
# Description

Flakey test causing PRs to fail when going through github actions. Upon investigation, it looks like the whole `filters` is not being tested due to the expected videos being passed in.

The FilterList has been refactored to simplify the relationship between the FilterList and the VIdeosPage.

- [ ] videos from homepage show by default
- [ ] changing audio language changes list and URL
- [ ] changing subtitle language changes list and URL
- [ ] changing title search changes list and URL
- [ ] language filter shows by default
- [ ] loading more still works

# Checklist


apps/watch/src/components/VideosPage/VideosPage.spec.tsx
```diff
@@ -124,19 +125,19 @@ describe('VideosPage', () => {
           <VideosPage videos={videos} />
```
csiyang, line 125: @tataihono This is the test that is sometimes timing out in github actions. The language filter isn't actually being test since `videos` is being passed in here (this is the same with the other 2 filter tests)

Can we skip this flakey test for now since its holding up cycle PR's and this whole describe need to have its tests rewritten.

tataihono (approved):

PR #1524, opened 2023-05-03
fix: redirect user on invalid journey id

Kneesal:
# Description

### <samp>🤖 Generated by Copilot at 0913d8f</samp>

Improved validation and redirection for `JourneyIdPage`. Added a hook to check journey data and redirect to root page if invalid in `journeys/[journeyId].tsx`.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/32318254/todos/6051672098)

- [x] it should redirect user to main page if user edits the url to a journey that doesn't exist on journeys and journeys edit pages
- [x] it should redirect user to main page if user edits the url to a journey that doesn't exist on publisher and publisher edit pages
- [x] it should redirect user to main page if user edits the url to a journey that doesn't exist on templates pages

# Walkthrough

### <samp>🤖 Generated by Copilot at 0913d8f</samp>

* Import `useEffect` hook from React to enable a side effect in `JourneyIdPage` component ([link](https://github.com/JesusFilm/core/pull/1524/files?diff=unified&w=0#diff-3873b898541aff5b78e0bc4aba59192c3d0e767d86c2e57185f5c88dd3f02063L1-R1))
* Use `useEffect` hook to redirect to root page if queried journey data is null or undefined ([link](https://github.com/JesusFilm/core/pull/1524/files?diff=unified&w=0#diff-3873b898541aff5b78e0bc4aba59192c3d0e767d86c2e57185f5c88dd3f02063R55-R61))


apps/journeys-admin/src/libs/useInvalidJourneyRedirect/useInvalidJourneyRedirect.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect } from 'react'
+import { useRouter } from 'next/router'
+
+export const useInvalidJourneyRedirect = (data): void => {
+  const router = useRouter()
+  useEffect(() => {
+    const checkValidJourney = (): void => {
+      data?.journey == null && data?.journey !== undefined && router.push('/')
```
mikeallisonJS, line 8: This style of conditional code is a little unconventional outside of jsx rendering. Could you use an if statement to simplify?

mikeallisonJS: Aditionally, `journey === null` should match both conditions with less code (unless I'm missing another possible condition here).

mikeallisonJS (changes requested):

apps/journeys-admin/src/libs/useInvalidJourneyRedirect/useInvalidJourneyRedirect.ts
Kneesal: Thanks for the feedback, Mike. I will implement those changes and ping you again for a review :)

mikeallisonJS (approved):

PR #1612, opened 2023-06-11
23 04 nc feat name and location component

Kneesal:
# Description

copilot:summary

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/33034387/todos/6224831586)

- [ ] It should render name and location component as per design
- [ ] it should concat long name and locations to ellipsis
- [ ] it should not display location if one is not selected
- [ ] it should not display name and location if name is not passed in

# Walkthrough

copilot:walkthrough


libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+      ...mockProps,
+      rtl: true
+    }
+
+    const { getByText } = render(<NameAndLocation {...props} />)
+
+    expect(getByText('Student Life · Edmond Shen')).toBeInTheDocument()
+  })
+
+  it('renders only the name when location is not provided', () => {
```
storyworks, line 30: Good pickup

libs/journeys/ui/src/components/StepFooter/StepFooter.tsx
```diff
@@ -54,7 +58,8 @@ export function StepFooter(): ReactElement {
               flexGrow: 1,
               flexDirection: 'row',
               alignItems: 'center',
-              justifyContent: 'space-between'
+              justifyContent: 'space-between',
+              overflow: 'clip'
```
storyworks, line 62: Don't need this.. Take a look at to constrain width of the text.
image

Currently on your branch I'm still seeing:
image
image

But we want the location to show fully if there is enough space.

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,67 @@
...
+                  ? '260px '
+                  : maxWidthAmount,
+              lg: '100%'
+            },
+            overflow: 'clip',
+            textOverflow: 'ellipsis',
+            color: 'secondary.light'
+          }}
+        >
+          {!rtl && name != null && `${name} `}
```
storyworks, line 51: Can rewrite it like 
```
{rtl ? `${location}${location && \u00A0\u00B7\u00A0}${name}`: `${name}${location && \u00A0\u00B7\u00A0}${location}`}
```

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,67 @@
+import Box from '@mui/material/Box'
+import Stack from '@mui/material/Stack'
+import Typography from '@mui/material/Typography'
+
+import { ReactElement } from 'react'
+
+interface NameAndLocationProps {
+  name?: string
```
storyworks, line 8: Name should never be null since it's required in the backend

libs/journeys/ui/src/components/StepFooter/StepFooter.tsx
```diff
@@ -69,6 +72,12 @@ export function StepFooter(): ReactElement {
               }}
             >
               {journey?.seoTitle ?? journey?.title}
+              <NameAndLocation
+                name={name}
+                location={location}
+                rtl={rtl}
```
storyworks, line 78: Let's use the useJourney and getJourneyRTL functions directly in NameAndLocation. 

It's different to the host info since we need to do an api call to get host info (want to reduce calling the api multiple times so better to pass in name and location from StepFooter)

useJourney and getJourneyRTL on the other hand are built to be called quickly and the beauty of hooks is that we don't need to keep passing down their data - can call the hook directly in the child component.

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,71 @@
...
+        flexDirection: 'row',
+        alignItems: 'center',
+        flexWrap: 'none'
+      }}
+    >
+      <Typography
+        variant="body2"
+        sx={{
+          whiteSpace: 'nowrap',
+          width: {
```
storyworks, line 40: I've just updated the footer selection PR again after seeing this. Seems too complex to manually calculate it. 
I've fixed it in that PR, so try merging in the latest version, remove this width prop and see if it works (similar to journey title)

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,71 @@
...
+            })`,
+            lg: `calc(100vw - ${
+              !admin && src1 == null && src2 == null
+                ? '180px'
+                : maxWidthAmountDesktop
+            })`
+          },
+          overflow: 'clip',
+          textOverflow: 'ellipsis',
+          color: 'secondary.light'
```
storyworks, line 52: Color looks off.. Confirm with Lyuba what color this should be for mobile and desktop? I've seen inconsistent designs..

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
Kneesal: perfect

storyworks (dismissed):

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const NameAndLocation = ({
+  name,
+  location = ''
+}: NameAndLocationProps): ReactElement => {
+  const { journey } = useJourney()
+  const { rtl } = getJourneyRTL(journey)
+
+  return (
+    <Stack
+      className="name-and-location"
```
GeronimoJohn, line 22: is this meant to be `data-testid`

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,49 @@
...
+        flexWrap: 'none'
+      }}
+    >
+      <Typography
+        variant="body2"
+        sx={{
+          whiteSpace: 'nowrap',
+          overflow: 'clip',
+          textOverflow: 'ellipsis',
+          color: { xs: 'primary.main', lg: 'white' },
```
GeronimoJohn, line 35:
```suggestion
          color: { xs: 'primary.main', lg: 'background.default' },
```

if it's just meant to be white, can use token?

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,49 @@
...
+          opacity: 0.7
+        }}
+      >
+        {rtl
+          ? `${location}${
+              (location !== '' && '\u00A0\u00B7\u00A0') || ''
```
GeronimoJohn, line 41: `'\u00A0\u00B7\u00A0'`

I think other developers coming to this code will take some time to figure out what this means. Are we able to give it a variables name or something?

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+import {
+  JourneyStatus,
+  ThemeMode,
+  ThemeName
+} from '../../../../__generated__/globalTypes'
+import { NameAndLocation } from './NameAndLocation'
+
+describe('NameAndLocation', () => {
+  const mockProps = {
+    name: 'Edmond Shen',
```
GeronimoJohn, line 13: NIT: as much as I love Edmond and not tryna kill being fun, I think it might be worth adding generic values to the code base instead. Just the pattern that we've set or follow in Core - generic values for testing. Though just a NIT

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { ReactElement } from 'react'
+import { useJourney } from '../../../libs/JourneyProvider'
+import { getJourneyRTL } from '../../../libs/rtl'
+
+interface NameAndLocationProps {
+  name: string
+  location?: string
+}
+
```
GeronimoJohn, line 12: Question: Is this information not coming from Journeys? or some sort of API call?

Kneesal: api call being made in step footer and being passed to children

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
GeronimoJohn: ignore this one actually

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
GeronimoJohn: but can it be called through `useJourney`?

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
storyworks: Yeah, our convention uses data-testid

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
storyworks: I realised this while integrating 😅, that we're just getting host data from journey. Can keep it like this, we can argue it's more obvious whether this component will show or not if I'm making the check here. Instead of conditionally rendering it within NameAndLocation...
image

I also ended up renaming after hearing about some backend naming conventions...
image

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
Kneesal: keeping this consistent with how the step footer title is made since that PR and colors has been approved by UI

libs/journeys/ui/src/components/StepFooter/NameAndLocation/NameAndLocation.tsx
storyworks: Actually you know what, looking at `HostAvatar` - I think that one is better if we just get src data from `journey.host.src1` from `const { journey } = useJourney()` in the component itself

Let's update this to be the same.

storyworks (approved):

storyworks (dismissed):

storyworks (approved):

PR #1615, opened 2023-06-12
feat: avatar component

Kneesal:
# Description

### <samp>🤖 Generated by Copilot at 19f90b5</samp>

This pull request adds a new HostAvatars component to the journeys UI library, which displays the images of the journey host or a placeholder. The pull request includes the component file, the test file, the story file, and the index file for the component. The component uses Material UI, custom styles, and hooks to access the journey and feature flag data.

- [Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/33034387/todos/6224807465)

Journeys Admin:

- [ ] it should show gray placeholder icon when no image available
- [ ] it should show one avatar image and one placeholder image in journeys editor if user selects only one image
- [ ] it should show both avatar images if user selects two avatar images

Journeys:
- [ ] it should show nothing in journeys is no avatar is selected
- [ ] it should only show one avatar in journeys if only on is selected
- [ ] it should show both in journeys if two are selected 

# Walkthrough

### <samp>🤖 Generated by Copilot at 19f90b5</samp>

* Create a new HostAvatars component to display the journey host images or a placeholder ([link](https://github.com/JesusFilm/core/pull/1615/files?diff=unified&w=0#diff-6a574f96c8d2458f5e25a78cb54b285ef8a2efd6e720a3e725d586aafa61b8bdR1-R79), [link](https://github.com/JesusFilm/core/pull/1615/files?diff=unified&w=0#diff-9334f9b91f662fc72540f497af7f20a73bde70e4ac3b0c077f28b151bc2cfa8aR1))
* Add stories for the HostAvatars component to showcase its different props and sizes in Storybook ([link](https://github.com/JesusFilm/core/pull/1615/files?diff=unified&w=0#diff-5b88ca2c2782aed7a2e094be73df5c789e611ec9836d9dfe1785bb87f31aa9cdR1-R112))
* Add tests for the HostAvatars component to verify its rendering logic and RTL support ([link](https://github.com/JesusFilm/core/pull/1615/files?diff=unified&w=0#diff-eba3abb22ff18193b93a844753cc35005fd3ab8453f0a3f004a08b8b317bcc23R1-R147))


libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const { journey, admin } = useJourney()
+  const { rtl } = getJourneyRTL(journey)
+  console.log(rtl)
+  return (
+    <>
+      {(src1 != null || src2 != null) && !admin && (
+        <AvatarGroup
+          spacing="small"
+          data-testid="journeys-avatars"
+          sx={{
```
storyworks, line 23: Split out layout styling to StepFooter - possibly that Box you have wrapping the HostAvatars. HostAvatars should just display the avatars going edge to edge in the component (no margins / padding around it)

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export const HostAvatars = ({ src1, src2 }: HostAvatarsProps): ReactElement => {
+  const { journey, admin } = useJourney()
+  const { rtl } = getJourneyRTL(journey)
+  console.log(rtl)
+  return (
+    <>
+      {(src1 != null || src2 != null) && !admin && (
+        <AvatarGroup
```
storyworks, line 20: Should be able to use the same Avatar group for all display variants.

The children inside would render conditionally based on:
- `src1 != null ? <Avatar/> : admin && editableStepFooter ? <Placeholder/> : null`
- `src2 != null ? <Avatar/> : admin && editableStepFooter ? <Placeholder/> : null`

`editableStepFooter` is the flag we set if the footer is able to be selected. For now, if we can't edit the step footer (like can't even click to select it), let's hide the placeholder since it will confuse users.

Canvas.tsx has an example of how you'd add the flag.

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+      {src1 == null && src2 == null && admin ? (
+        <Avatar
+          sx={{
+            pr: rtl ? 2 : 0,
+            mr: 2,
+            ml: rtl ? 4 : 0,
+            height: '44px',
+            width: '44px',
+            color: 'secondary.light',
+            backgroundColor: 'rgba(0, 0, 0, 0)',
```
storyworks, line 42: I think background color is transparent for a placeholder avatar?

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+            outline: 'none',
+            border: '3px dashed ',
```
storyworks, line 44: Extra space at end `3px dashed `

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+export const HostAvatars = ({ src1, src2 }: HostAvatarsProps): ReactElement => {
+  const { journey, admin } = useJourney()
+  const { rtl } = getJourneyRTL(journey)
+  console.log(rtl)
+  return (
+    <>
+      {(src1 != null || src2 != null) && !admin && (
+        <AvatarGroup
+          spacing="small"
+          data-testid="journeys-avatars"
```
storyworks, line 22: Let's use `host-avatars`

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,110 @@
...
+            width: '44px',
+            color: 'secondary.light',
+            backgroundColor: 'rgba(0, 0, 0, 0)',
+            outline: 'none',
+            border: '3px dashed ',
+            borderColor: (theme) => theme.palette.grey[700]
+          }}
+        >
+          <UserProfile3
+            data-testid="account-circled-out-icon"
```
storyworks, line 49: `host-avatar-placeholder` - use names which convey the component purpose rather than implementation where you can, they may decide to switch the placeholder icon we use in the future and this would still make sense.

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Avatar from '@mui/material/Avatar'
+import AvatarGroup from '@mui/material/AvatarGroup'
+import UserProfile3 from '@core/shared/ui/icons/UserProfile3'
+import { useFlags } from '@core/shared/ui/FlagsProvider'
+import { useJourney } from '../../../libs/JourneyProvider'
+import { getJourneyRTL } from '../../../libs/rtl'
+
+interface HostAvatarsProps {
+  src1?: string
+  src2?: string
```
storyworks, line 11: Let's build this component so we can also reuse it for the host list in admin
image

- You can add a `hasPlaceholder` check here (StepFooter will pass `hasPlaceholder={admin}`  - but that's not part of this PR)
- You can also add a size: "small" | "large" since the host list avatar is larger. Make small the default for now..

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.spec.tsx
```diff
@@ -0,0 +1,62 @@
+import { render } from '@testing-library/react'
+import { FlagsProvider } from '@core/shared/ui/FlagsProvider'
+import { JourneyProvider } from '../../../libs/JourneyProvider'
+import { HostAvatars } from './HostAvatars'
+
+describe('HostAvatars', () => {
+  it('renders AvatarGroup with src1 and src2 avatars when NOT admin', () => {
```
storyworks, line 7: Update these test cases after HostAvatar logic is refactored. Also remember to describe tests / names by the purpose rather by implementation detail

"it renders both avatars if images are set" - is more descriptive to someone with no idea of the inside mechanics of the component. This is helpful as we will have logs / dashboards in the future which will show all test cases and being able to understand the scope of each one at a glance makes it easier to debug.
Also again, if for some reason we decide to rename the `src1` prop to `avatar1Src` or we decide we need our own custom group to combine avatars, then this test case would need to be renamed.

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+const Template: Story<
+  ComponentProps<typeof HostAvatars> & {
+    editableStepFooter: boolean
+    admin: boolean
+  }
+> = ({ src1, src2, editableStepFooter, admin }) => (
+  <FlagsProvider flags={{ editableStepFooter }}>
```
storyworks, line 24: By default let's set this to true so you don't need to pass it in. 

It has no impact on the number of visual variants we produce. (Eg editableStepFooter=false is already captured by the one and two avatar states)

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <JourneyProvider value={{ admin }}>
+      <Stack direction="row">
+        <HostAvatars src1={src1} src2={src2} />
+      </Stack>
+    </JourneyProvider>
+  </FlagsProvider>
+)
+
+export const Default = Template.bind({})
```
storyworks, line 33: I think in my last comment I mentioned to use:

"Empty" (admin placeholder)
"1 avatar + placeholder"
"1 avatar"
"2 avatars"

Sorry made it unclear since adding "Empty" first makes it seem like the default.

If you want to use "Default", the "1 avatar" variant would be the default. And we'd get:
"Default" - 1 avatar
"Empty" - 1 placeholder
"WithPlaceholder" - 1 avatar with placeholder 
"2 avatars" - 2 avatars

Our storybook convention is to use shorter names to describe variants. Once we know the 1 avatar is the default variant, we don't need to say "OneAvatarWithPlaceHolder" - WithPlaceholder is enough to convey this is the difference between the default case and this new one.

In my first suggestion, there was no default so the names were a bit longer. But I like that you did set a default.

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <AvatarGroup
+        spacing="small"
+        data-testid="host-avatars"
+        sx={{ flexDirection: rtl ? 'row' : 'row-reverse' }}
+      >
+        {src1 != null && !admin && <Avatar src={src1} />}
+        {src2 != null && !admin && <Avatar src={src2} />}
+        {src1 != null && admin && <Avatar src={src1} />}
+        {src2 != null && admin && <Avatar src={src2} />}
+        {(src1 == null || src2 == null) && admin && editableStepFooter && (
```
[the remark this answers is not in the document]
storyworks: Oh wait! Nvm! double checked the designs and we only ever see 1. This is good then.

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,56 @@
...
+  src1?: string
+  src2?: string
+}
+
+export const HostAvatars = ({ src1, src2 }: HostAvatarsProps): ReactElement => {
+  const { journey, admin } = useJourney()
+  const { rtl } = getJourneyRTL(journey)
+  const { editableStepFooter } = useFlags()
+  return (
+    <>
```
storyworks, line 19: What's the purpose of this Fragment?

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useFlags } from '@core/shared/ui/FlagsProvider'
+import { useJourney } from '../../../libs/JourneyProvider'
+import { getJourneyRTL } from '../../../libs/rtl'
+
+interface HostAvatarsProps {
+  src1?: string
+  src2?: string
+}
+
+export const HostAvatars = ({ src1, src2 }: HostAvatarsProps): ReactElement => {
```
storyworks, line 14: For the default component, our convention is `export function HostAvatars(): ReactElement {`

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+    blocks: [],
+    primaryImageBlock: null,
+    userJourneys: [],
+    template: null,
+    seoTitle: 'My awesome journey',
+    seoDescription: null,
+    host: hostData
+  }
+
+  it('renders both avatars if images are set in journeys', () => {
```
storyworks, line 62: Rename the above to
- `renders both avatars if both images are set` - this won't just be used in journeys, but also journeys-admin in the host avatar list.
- `renders nothing if no images are set` - and don't pass "hasPlaceholder to HostAvatars (add this one)
- `renders with avatar if one image is set` - and don't pass "hasPlaceholder to HostAvatars (add this one)

Then within an inner `describe("hasPlaceholder")`
- `renders placeholder if no images are set`
- `renders with avatar and placeholder if one image is set`

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const TwoAvatars = Template.bind({})
+TwoAvatars.args = {
+  ...Empty.args,
+  journey: {
+    ...journey,
+    host: { ...hostData, src2: 'https://tinyurl.com/2nxtwn8v' }
+  }
+}
```
storyworks, line 103: Also, let's add a `Large` story for when the Avatars are large...

libs/journeys/ui/src/components/StepFooter/HostAvatars/HostAvatars.spec.tsx
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    const avatars = getByRole('img')
+    expect(avatars).toBeInTheDocument()
+    expect(avatars.getAttribute('src')).toEqual('avatar1.jpg')
+    const adminPlaceholderElement = getByTestId('host-avatar-placeholder')
+    expect(adminPlaceholderElement).toBeInTheDocument()
+  })
+
+  it('renders with two avatars in journey edit page when both images are provided', () => {
```
storyworks, line 115: Not needed, the display of 2 avatars doesn't depend on `admin` check at all - just `src1 / src2` checks. And you're already testing for this above.

storyworks (approved):

storyworks (approved):

PR #1715, opened 2023-07-20
feat: background color iphone

csiyang:
# Description

### <samp>🤖 Generated by Copilot at 2f14a02</samp>

Added dark mode support for the app by setting the theme-color and background-color in `_document.tsx`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color
https://stephenradford.me/removing-the-white-bars-in-safari-on-iphone-x/

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/33281894/todos/6319404840)

Journeys-Admin
 - [ ] Everywhere cards are seen, styling should remain unchanged. (Test on devices with and without notch in portrait and landscape.)
 
Journeys (visitor)
- [ ] On devices with notch, **notch background color**, should match the **desktop background color**
- [ ] Mobile devices with **notch** in **horizontal** while **in a journey** should have card stretch to the edge of the screen, but should not cover the content
- [ ] Every other instance of the card, styling should not change. (Test on devices with and without notch in portrait and landscape.)

Embed
 - [ ] The initial embed card should have no changes mobile with notch horizontal.
 - [ ] While fullscreened, mobile with notch horizontal should stretch the card to the edges of the devices, but notch should not cover any content.

# Walkthrough

### <samp>🤖 Generated by Copilot at 2f14a02</samp>

* Add meta tag and body style for dark mode color ([link](https://github.com/JesusFilm/core/pull/1715/files?diff=unified&w=0#diff-a0c7c05bd2236691d8382533f6778b7cd63ef4e302cb96bf259538df88d469adL59-R61))


apps/journeys/pages/_app.tsx
```diff
@@ -84,7 +90,16 @@ function JourneysApp({
       <Head>
         <meta
           name="viewport"
-          content="minimum-scale=1, initial-scale=1, width=device-width"
+          content="minimum-scale=1, initial-scale=1, width=device-width viewport-fit=cover"
+        />
+        {/* Always sets background color to dark theme when on a single journey */}
```
storyworks, line 95: You've set the theme as the light theme, but the comment says you want to use the dark theme?

apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -163,6 +163,19 @@ export function Conductor({ blocks }: ConductorProps): ReactElement {
     }
   }, [swiper, blockHistory])
 
+  // Shared position styling for stepHeader and stepFooter with notch calculations when applicable.
+  const headerFooterPosition: SxProps = {
```
storyworks, line 167: maybe rename to `mobileNotchStyling` or a variable name that describes this accounts for the mobile notch so you don't need to add the comment? It's a given it's shared styling for the header/footer since we can see it used in those places.

libs/journeys/ui/src/components/Card/OverlayContent/OverlayContent.tsx
```diff
@@ -38,13 +43,32 @@ export function OverlayContent({
       }
     : {}
 
+  // Adds padding with notch calculations when applicable
+  const horizontalPadding: SxProps =
```
storyworks, line 47: rename to `mobileNotchPadding` ? We want this styling block just to address the notch issue (or use normal padding if it doesn't exist). If there are additional horizontal margins needed for other styling reasons, we want to make it more clear to people to add it in a seperate block instead of adding it to this one. (Mainly because there are so many styles applied to this, keeping the properties seperate by feature helps us in debugging / understanding the component when we return to it)

apps/journeys/pages/_app.tsx
csiyang: It still needs to default to the light theme for the homepage, just explicitly setting it now since I have the ternary for it. 

`theme.palette.grey` is pulling from the color tokens irrespective of the theme. Its similar to how you did it in `conductor.tsx`. I can remove the comment if it's confusing though

libs/journeys/ui/src/components/StepFooter/StepFooter.tsx
```diff
@@ -45,6 +45,7 @@ export function StepFooter({
         position: { xs: 'absolute', lg: 'relative' },
         zIndex: 1,
         bottom: 0,
+        width: { xs: '100%', lg: 'auto' },
```
csiyang: This component still needs a default width to be added since it's not always used from conductor (also called in `canvas`) and cannot rely on the shared styling to be passed in from there.

libs/journeys/ui/src/components/StepHeader/StepHeader.tsx
```diff
@@ -31,12 +36,13 @@ export function StepHeader(): ReactElement {
...
       sx={{
-        width: { xs: '100%', lg: 'unset' },
-        mt: { xs: 1, lg: 0 },
         position: { xs: 'absolute', lg: 'relative' },
+        mt: { xs: 1, lg: 0 },
         zIndex: 1,
         top: 0,
-        alignItems: 'flex-end'
+        alignItems: 'flex-end',
+        width: { xs: '100%', lg: 'auto' },
```
csiyang, line 44: https://github.com/JesusFilm/core/pull/1715#discussion_r1287795284

apps/journeys/pages/_app.tsx
storyworks: Oh I see what you mean. 'Always use a dark background color when on a single journey' could work? Since you're setting the background regardless of the journey theme for that journey. 

Was Vlad ok with you hard-coding the notch color like that? If you have pageProps.journey - can you not use pageProps.journey.themeMode to get the background color specific to the journey?

libs/journeys/ui/src/components/Card/ExpandedCover/ExpandedCover.tsx
```diff
@@ -32,6 +34,13 @@ export function ExpandedCover({
     }
   }
 
+  const background =
+    backgroundColor != null
+      ? imageBlock?.src != null
```
csiyang, line 39: only adds transparency if there is a image

apps/journeys/pages/_app.tsx
csiyang: This is now moved to `_document.tsx`. This move allows us to override the background color in the component level. It has to have a default value to be set (so the query selector that updates the color can work) so am always setting it as `light - background.defaut`

storyworks (approved):

PR #1776, opened 2023-08-02
feat: add user team api

tataihono:
# Description

### <samp>🤖 Generated by Copilot at 973a7e1</samp>

This pull request adds a new feature to the `api-gateway` and `api-journeys` schemas that allows querying user teams associated with a team, using a connection type with pagination support. It also updates the generated code and resolver functions to match the schema changes. It uses partial schema files and the `@join__type` directive to modularize and link the schemas.

### DIFF
--- a/apps/api-journeys/src/app/modules/team/team.resolver.ts
+++ b/apps/api-journeys/src/app/modules/team/team.resolver.ts
@@ -1,4 +1,11 @@
-import { Resolver, Query, Mutation, Args } from '@nestjs/graphql'
+import {
+  Resolver,
+  Query,
+  Mutation,
+  Args,
+  ResolveField,
+  Parent
+} from '@nestjs/graphql'
 import { UseGuards } from '@nestjs/common'
 import { Team, Prisma } from '.prisma/api-journeys-client'
 import { subject } from '@casl/ability'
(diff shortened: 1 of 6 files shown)


apps/api-journeys/src/app/modules/team/team.resolver.ts
```diff
@@ -83,4 +91,43 @@ export class TeamResolver {
...
+  ): Promise<UserTeamsConnection> {
+    const result = await this.prismaService.userTeam.findMany({
+      where: { AND: [accessibleUserTeams, { teamId: team.id }] },
+      cursor: after != null ? { id: after } : undefined,
+      orderBy: { role: 'asc' },
+      skip: after == null ? 0 : 1,
+      take: first + 1
+    })
+
+    const sendResult = result.length > first ? result.slice(0, -1) : result
```
Kneesal, line 110: correct me if I am wrong, 
is this meant to be this: 

` const sendResult = result.length > first ? result.slice(0, first) : result`

apps/api-journeys/src/app/modules/team/team.resolver.ts
```diff
@@ -83,4 +91,43 @@ export class TeamResolver {
...
+    return {
+      edges: sendResult.map((userTeam) => ({
+        node: {
+          ...userTeam,
+          role: userTeam.role as UserTeamRole,
+          createdAt: userTeam.createdAt.toISOString(),
+          updatedAt: userTeam.createdAt.toISOString(),
+          user: { id: userTeam.userId }
```
Kneesal, line 118: doesn't the userTeam object already have the user object within it? Is there a reason we need to set the user id this way?

PR #1784, opened 2023-08-03
fix: remove old api teams

tataihono:
# Description

### <samp>🤖 Generated by Copilot at 1b7c028</samp>

No summary available (Limit exceeded: required to process 118127 tokens, but only 50000 are allowed per call)


Kneesal (approved):
there's just a couple of lint issues to fix. but approved otherwise

Kneesal (approved):

PR #1926, opened 2023-09-27
feat: category templates

GeronimoJohn:
# Description

### <samp>🤖 Generated by Copilot at de4e92f</samp>

This pull request adds a new feature to the journeys-admin app, which allows the user to view templates by category based on the tags. It introduces a new component `CategoryTemplates`, which fetches the tags and journeys data from the server and renders them using the existing `TemplateSection` component. It also adds the types and tests for the new component and the `GetTags` query.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/34356579/todos/6557668486)

- [ ] it should render all the child tag categories with 5 templates and more from the server
- [ ] it should render all the tag categories relating to the tagIds in the query param
- [ ] it should render all the templates within a category if filtered is true, regardless of how much templates is in it
- [ ] it should not render parent tag id's as a section
- [ ] it should not render a category if it doesn't have any templates

# Walkthrough

### <samp>🤖 Generated by Copilot at de4e92f</samp>

*  Add the `CategoryTemplates` component to render templates grouped by categories based on tags ([link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-e1b03f9220b6b898c3bddfbdde4d80a874daf64865e7207dab78501074bdcb30R1-R78), [link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-04a64f74df8dd85e371fe3dbbc8a0701a834ec7182ff4ecb9a864e30adccc4fcR1), [link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-8528e7b4c06335c3a765fccde78d68fff718a8b69d7384bea6782deeb45efb8aL6-R11), [link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-8528e7b4c06335c3a765fccde78d68fff718a8b69d7384bea6782deeb45efb8aR29))
*  Add the unit tests for the `CategoryTemplates` component using `MockedProvider` and `useRouter` ([link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-cd641987d96a96777c2085e0ace7459c3bad3df086af98d6ea29a5702df79757R1-R248))
*  Add the TypeScript types for the `GetTags` query using Apollo codegen ([link](https://github.com/JesusFilm/core/pull/1926/files?diff=unified&w=0#diff-cbadcf12d65d25a2e89e6689beec13dc482912c315bc2c0139127f6a531281fcR1-R24))


Kneesal (dismissed):

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
```diff
@@ -0,0 +1,78 @@
+import { gql, useQuery } from '@apollo/client'
+import { useRouter } from 'next/router'
+import { ReactElement, useMemo } from 'react'
+
+import { GetTags } from '../../../../__generated__/GetTags'
+import { useJourneysQuery } from '../../../libs/useJourneysQuery'
+import { TemplateSection } from '../TemplateSection'
+
+export const GET_TAGS = gql`
```
storyworks, line 9: create a reusable getTagsQuery hook for this - we'll be using it for multiple elements on this page + using it in the settings dialog :)

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.spec.tsx
```diff
@@ -0,0 +1,248 @@
...
+import { CategoryTemplates } from '.'
+
+jest.mock('next/router', () => ({
+  __esModule: true,
+  useRouter: jest.fn()
+}))
+
+const mockUseRouter = useRouter as jest.MockedFunction<typeof useRouter>
+
+describe('CategoryTempaltes', () => {
```
storyworks, line 18: spelling

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <TemplateSection
+        category={queryTags != null ? queryTagName : serverTagName}
+        journeys={data?.journeys}
+      />
+    )
+  }
+
+  return (
+    <>
+      {tagDetails?.map(({ name, id }) => (
```
storyworks, line 68: Reminder there was another requirement for this! Don't render the templates if there's less than 5 for a tag.

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.spec.tsx
GeronimoJohn: sadge react

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
GeronimoJohn: good reminder

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  }
+`
+
+type TagDetailsProps = Array<{ id: string; name?: string }> | undefined
+
+export function CategoryTemplates(): ReactElement {
+  const router = useRouter()
+  const queryTags = router.query.tags as string[]
```
storyworks, line 25: rename `paramTags` - since queryTags will be the tags specified in the filters

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
```diff
@@ -0,0 +1,78 @@
...
+        value
+        primary
+      }
+    }
+  }
+`
+
+type TagDetailsProps = Array<{ id: string; name?: string }> | undefined
+
+export function CategoryTemplates(): ReactElement {
```
storyworks, line 23: I reckon, TemplateSections should handle the global state of which tags appear on the page (derived from the url parameters and autocomplete filters) and pass these down to the components which need them.

This component should just take a `tags` array and render templates for the tags it's given. Unless there are less than 5 templates in that tag list (keep that logic here).

Later on when we add the autocomplete filter, it will also take that initial tags list and prepopulate the autocomplete, then any changes it makes would update the TemplateSections tag state which will update this component too.

apps/journeys-admin/src/libs/useTagsQuery/useTagsQuery.ts
```diff
@@ -0,0 +1,21 @@
...
+import { GetTags } from '../../../__generated__/GetTags'
+
+export const GET_TAGS = gql`
+  query GetTags {
+    tags {
+      id
+      name {
+        value
+        primary
+      }
```
Kneesal, line 12: don't need parent tag id?

Kneesal (approved):

apps/journeys-admin/src/libs/useTagsQuery/useTagsQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+  }
+`
+
+export function useTagsQuery(): QueryResult<GetTags> {
+  const query = useQuery<GetTags>(GET_TAGS)
+
+  return query
```
storyworks, line 22: Take a look at `useUserTeamsAndInvitesQuery` - see how they extract out the email data separately as a helper. Would this be helpful for you to do as well? To distinguish between parent and child tags?

apps/journeys-admin/src/components/TemplateSections/TemplateSections.tsx
```diff
@@ -1,22 +1,53 @@
...
+      const tagsArray = Array.isArray(paramTags) ? paramTags : [paramTags]
+      return tagsArray.map((id) => ({
+        id,
+        name: serverTags?.tags
+          .find((tag) => tag.id === id)
+          .name.find(({ primary }) => primary)?.value
+      }))
+    }
+
+    return serverTags?.tags.map((tag) => ({
```
storyworks, line 32: Are you rendering every server tag at the moment? Have a think - do we actually want every single tag to have it's own section?

apps/journeys-admin/src/components/TemplateSections/TemplateSections.tsx
```diff
@@ -1,22 +1,62 @@
...
+  name: string
+}
+
 export function TemplateSections(): ReactElement {
   const router = useRouter()
-  const tags = router.query.tags
+  const paramTags = router.query.tags
+  const { childTags } = useTagsQuery()
+
+  const [tagsFilter, setTagsFilter] = useState<string[]>()
```
storyworks, line 21: tagsFilter might be null initially, but later on we can set filters then delete them all - leaving an empty array. Then all your checks would break since you're just checking for null.

apps/journeys-admin/src/components/TemplateSections/TemplateSections.spec.tsx
```diff
@@ -18,29 +29,228 @@ describe('TemplateSections', () => {
...
+        </MockedProvider>
+      )
+      await waitFor(() =>
+        expect(getByRole('heading', { name: 'Acceptance' })).toBeInTheDocument()
+      )
+      expect(getByRole('heading', { name: 'Addiction' })).toBeInTheDocument()
+      expect(getByRole('heading', { name: 'Anger' })).toBeInTheDocument()
+    })
+
+    it('should render categories related to tagIds in the query param', async () => {
```
storyworks, line 234: "should render categories by url tag queries by default" - maybe helpful to call it url here since this query could be confused with api query call. 

And add default since later there will be tests "should render categories by autocomplete tag filters" (and these would override the url queries)

storyworks:
Also the index file of useTagsQuery is empty.

apps/journeys-admin/src/components/TemplateSections/TemplateSections.tsx
```diff
@@ -1,17 +1,69 @@
...
+  const paramTags = router.query.tags
+  const { childTags } = useTagsQuery()
+
+  const [tagsFilter, setTagsFilter] = useState<string[]>([])
+
+  useEffect(() => {
+    // add logic to update state with tags from filter component
+    if (paramTags != null) {
+      const tagsArray = Array.isArray(paramTags) ? paramTags : [paramTags]
+      setTagsFilter(tagsArray)
```
storyworks, line 27: Wouldn't this update the tags filter each time the paramTags are changed? In the meeting we discussed that we only wanted paramTags to set the initial filter.

apps/journeys-admin/src/components/TemplateSections/CategoryTemplates/CategoryTemplates.tsx
```diff
@@ -22,8 +22,12 @@ export function CategoryTemplates({
     }
   })
 
+  function getJourneysLength(length): boolean {
```
storyworks, line 25: getJourneysLength implies you're expecting a number (the length) back. I see what you're trying to do though.
Could rename to `hasMinimumJourneys()`?

storyworks (approved):

storyworks (approved):

PR #2064, opened 2023-11-20
fix: reduce embed weight

tataihono:
# Description

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 30033c6</samp>

Refactored the `DiscoveryJourneys` and `EmbedJourney` components to remove the dependency on GraphQL data and simplify the testing and storybook setup. Deleted the unused `data.ts` file and cleaned up the imports and props in the related files.

Incorporated are a number of playwright fixes and optimizations as I had to work through some failures in the journeys screenshots.


apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -87,22 +85,19 @@ export function EmbedJourney({
               >
                 <Box
                   sx={{
-                    height: '100%',
-                    width: '100%',
-                    borderRadius: 4,
```
csiyang, line 92: add this back

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
```diff
@@ -87,22 +85,19 @@ export function EmbedJourney({
...
-                    position: 'relative',
-                    cursor: 'pointer'
+                    display: 'flex',
+                    justifyContent: 'center',
+                    flexDirection: 'column',
+                    height: '100%'
                   }}
                 >
-                  <BlockRenderer block={block} />
-                  <StepFooter title={discoveryJourney?.seoTitle ?? ''} />
```
csiyang, line 99: No more stepFooter? It seems like they were using the host name?

apps/journeys-admin/src/components/DiscoveryJourneys/DiscoveryJourneys.spec.tsx
```diff
@@ -1,15 +1,10 @@
...
 import { DiscoveryJourneys } from '.'
 
 describe('DiscoveryJourneys', () => {
   it('should show embed journeys', () => {
-    const { getByLabelText } = render(
-      <MockedProvider>
-        <DiscoveryJourneys />
-      </MockedProvider>
-    )
+    const { getByLabelText } = render(<DiscoveryJourneys />)
```
csiyang, line 7: Just to cover the warning from `DiscoverJourneys.tsx`?

```suggestion
    const { getByLabelText } = render(<DiscoveryJourneys />)
    expect(getByTestId('JourneysAdminDiscoveryJourneys').toBeInTheDocument()
```

csiyang (commented):
- Do we need ti get this look exactly the same as the current ones? If so, some of the spacing within the card is wrong. (Button)
-  run translations

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
tataihono: not needed.

apps/journeys-admin/src/components/DiscoveryJourneys/EmbedJourney/EmbedJourney.tsx
tataihono: fixed this.

csiyang (approved):

apps/journeys-e2e/src/e2e/journeys.spec.ts
tataihono, line 1: @kiran-redhat got rid of the video playback in this PR. This was causing unexpected failures preventing us from shipping. We should reintroduce this behaviour once we have a test journey built out.

csiyang (approved):

mikeallisonJS (approved):

PR #2157, opened 2023-12-19
fix: social share previews

nicholasyao974:
# Description

copilot:summary

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/35391220/todos/6827013906)

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough


apps/journeys-admin/src/components/Editor/ImageEdit/ImageEdit.tsx
```diff
@@ -114,12 +115,16 @@ export function ImageEdit({
...
   >(JOURNEY_IMAGE_BLOCK_ASSOCIATION_UPDATE)
+  const { setPrimaryImageBlock } = useSocialPreview()
   const { journey } = useJourney()
   const [open, setOpen] = useState<boolean | undefined>()
   const targetImageBlock =
     target === 'primary'
       ? journey?.primaryImageBlock
       : journey?.creatorImageBlock
+  useEffect(() => {
+    setPrimaryImageBlock(targetImageBlock)
```
csiyang, line 126: Can we do this when the image is updated instead of a use effect?. Also please test this

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/ImageEdit/ImageEdit.tsx
nicholasyao974: The use effect is needed to make it show the image by default when you create a journey from a template (since it just starts with that image, and doesn't get updated)

nicholasyao974: I can't really test it right now, since all the test cases are already failing on main :/

PR #2161, opened 2023-12-20
fix: copy to team

Kneesal:
# Description

copilot:summary

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/35391220/todos/6827014607)

- [ ] it should copy journey to a new team 
- [ ] it should correctly display existing journeys in the copied team

# Walkthrough

copilot:walkthrough


apps/journeys-admin/src/components/Team/CopyToTeamDialog/CopyToTeamDialog.tsx
```diff
@@ -40,16 +43,24 @@ export function CopyToTeamDialog({
...
+  const [updateLastActiveTeamId] = useMutation<UpdateLastActiveTeamId>(
+    UPDATE_LAST_ACTIVE_TEAM_ID
+  )
+
   async function handleSubmit(
     values: FormikValues,
     { resetForm }: FormikHelpers<FormikValues>
   ): Promise<void> {
-    // submitAction runs first so loading state can be shown
+    await updateLastActiveTeamId({
```
nicholasyao974, line 54: Do you think we need to wait for the journey duplication to succeed first? Otherwise, if the copying fails, it will still switch teams.

nicholasyao974 (commented):
There is another case where similar issues happen. On creating a new team, it shows the name of the new team at the top, but the journey list is still from the old team. Can you take a look at this too?

Kneesal:
> There is another case where similar issues happen. On creating a new team, it shows the name of the new team at the top, but the journey list is still from the old team. Can you take a look at this too?

This ticket solves that issue

apps/journeys-admin/src/components/Team/CopyToTeamDialog/CopyToTeamDialog.tsx
Kneesal: without updateLastActiveTeamIdI(). It would switch teams if duplicate fails regardless, added intentionally as user is already expecting a switch operation

Kneesal:
> There is another case where similar issues happen. On creating a new team, it shows the name of the new team at the top, but the journey list is still from the old team. Can you take a look at this too?

Out of scope for this PR

csiyang (approved):

csiyang (approved):

PR #2211, opened 2024-01-15
fix: access denied 500 error

ax00ng:
# Description
fixed access denied 500 error to be redirected to access denied page

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/35391220/todos/6910467386)


apps/journeys-admin/pages/journeys/[journeyId].tsx
```diff
@@ -52,7 +53,9 @@ function JourneyEditPage(): ReactElement {
     }
   )
 
-  return (
+  return status === 'noAccess' ? (
+    <AccessDenied />
```
Comment (csiyang) [inline apps/journeys-admin/pages/journeys/[journeyId].tsx:57]: NextSeo should be wrapped around this

apps/journeys-admin/src/components/AccessDenied/AccessDenied.tsx
```diff
@@ -1,28 +1,43 @@
...
+    userJourneyRequest(journeyId: $journeyId, idType: databaseId) {
+      id
+    }
+  }
+`
+export function AccessDenied(): ReactElement {
+  const [userJourneyRequest] =
+    useMutation<UserJourneyRequest>(USER_JOURNEY_REQUEST)
+  const router = useRouter()
+  const [requestedAccess, setRequestReceived] = useState(false)
```
csiyang, line 32: what happens if you revisit the request access page after already requesting access?

csiyang (dismissed):

apps/journeys-admin/src/components/AccessDenied/AccessDenied.tsx
csiyang: The idea was to try get the userJourney and see if the user has already requested access and not display the button a second time.  But realised we don't actually have a query to get a userJourneys, so will opt to keep the current behaviour

apps/journeys-admin/src/components/AccessDenied/AccessDenied.spec.tsx
```diff
@@ -1,18 +1,60 @@
...
+      }
+    }))
+
+    const mockUserJourneyRequest = {
+      request: {
+        query: USER_JOURNEY_REQUEST,
+        variables: { journeyId: 'mockedJourneyId' }
+      },
+      result: mockUserJourneyResult
+    }
```
Kneesal, line 35:
```suggestion
    const mockUserJourneyRequest: MockedResponse<UserJourneyRequest> = {
      request: {
        query: USER_JOURNEY_REQUEST,
        variables: { journeyId: 'mockedJourneyId' }
      },
      result: mockUserJourneyResult
    }
```

Kneesal: just give it some typing in case some  one updates this query. Then they will need to update the tests accordingly too

Kneesal (changes requested):

Kneesal (approved):

PR #2234, opened 2024-01-23
fix: zindex for video details drawer

csiyang:
# Description

Fix z index of a drawer that opened over another drawer

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/26802573/todos/6967665571)

### DIFF
--- a/apps/journeys-admin/src/components/Editor/VideoLibrary/VideoDetails/VideoDetails.tsx
+++ b/apps/journeys-admin/src/components/Editor/VideoLibrary/VideoDetails/VideoDetails.tsx
@@ -5,7 +5,7 @@ import Button from '@mui/material/Button'
 import Drawer from '@mui/material/Drawer'
 import IconButton from '@mui/material/IconButton'
 import Stack from '@mui/material/Stack'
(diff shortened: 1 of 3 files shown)


apps/journeys-admin/src/components/Editor/VideoLibrary/VideoDetails/VideoDetails.tsx
```diff
@@ -121,6 +122,7 @@ export function VideoDetails({
         elevation={smUp ? 1 : 0}
         hideBackdrop
         sx={{
+          zIndex: zIndex.modal,
```
csiyang, line 125: I'm a bit unsure if we should keep using `zIndex.modal` for the drawers, that seems like the standard thats in the code currently... thoughts @tataihono?

tataihono, line 125:
```suggestion
          zIndex: (theme) => theme.zIndex.modal,
```

tataihono (approved):

tataihono (approved):

PR #2244, opened 2024-01-29
fix: load more button on video library state

tanflem:
# Description

Changed state management of the videos array when clicking the load more button.

- Link to Basecamp Todo

Check that the load more button works in the video library.

- [ ] Test A
- [ ] Test B

# Considerations
There is most likely another fix for this involving settings with Apollo cache or a more refined pagination implementation.
https://www.apollographql.com/docs/react/pagination/core-api/#the-fetchmore-function

The load was broken because the data from useQuery was overwritten instead of appended. With our current settings in the Apollo cache to videos, this should not happen as far as  I can tell. 

I did remove `notifyOnNetworkStatusChange` from the query. So the query component does not rerender on refetch. The initial data is set by the initial query and subsequent data is appended by the fetchMore query. The rerender was causing both sets of data to be added to videos.

# Walkthrough

copilot:walkthrough

### DIFF
--- a/apps/journeys-admin/src/components/Editor/VideoLibrary/VideoFromLocal/VideoFromLocal.tsx
+++ b/apps/journeys-admin/src/components/Editor/VideoLibrary/VideoFromLocal/VideoFromLocal.tsx
@@ -43,8 +43,7 @@ export function VideoFromLocal({
 }: VideoFromLocalProps): ReactElement {
   const [searchQuery, setSearchQuery] = useState<string>('')
   const [videos, setVideos] = useState<VideoListProps['videos']>()
-  const { loading, data, fetchMore } = useQuery<GetVideos>(GET_VIDEOS, {
-    notifyOnNetworkStatusChange: true,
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/VideoLibrary/VideoFromLocal/VideoFromLocal.tsx
```diff
@@ -76,10 +75,25 @@ export function VideoFromLocal({
...
       variables: {
-        offset: data?.videos?.length ?? 0
+        offset: videos?.length ?? 0
       }
     })
-    if (response.data?.videos?.length === 0) setHasMore(false)
+    if (response.data?.videos?.length === 0) {
+      setHasMore(false)
+    } else {
+      console.log('Response in fetch More', response.data.videos)
```
mikeallisonJS, line 84: Remove console.log

mikeallisonJS (changes requested):

PR #2319, opened 2024-02-15
test: step updates

levi-zustiak:
# Description
- adds test to validate `stepViewEventCreate` only emits when `activeStep` is true
- adds tests to validate Seo changes based on `activeStep` and if first step

[Link](https://3.basecamp.com/3105655/buckets/35958767/todos/7065661942) to Basecamp Todo

### DIFF
--- a/libs/journeys/ui/src/components/Step/Step.spec.tsx
+++ b/libs/journeys/ui/src/components/Step/Step.spec.tsx
@@ -297,9 +297,46 @@ describe('Step', () => {
     expect(baseElement).toContainHTML('<body><div /></body>')
   })
 
-  // TODO (SWIPE): test activeStep
+  it('should not send stepViewEvent if not activeStep', async () => {
+    mockUuidv4.mockReturnValueOnce('uuid')
+    blockHistoryVar([block, { ...block, id: 'Step2' }])
 
-  // TODO (SWIPE): test seo
+    const result = jest.fn(() => ({
+      data: {
+        stepViewEventCreate: {
+          id: 'uuid',
+          __typename: 'StepViewEvent'
+        }
+      }
+    }))
+
+    render(
+      <MockedProvider
+        mocks={[
+          {
+            request: {
+              query: STEP_VIEW_EVENT_CREATE,
+              variables: {
+                input: {
+                  id: 'uuid',
+                  blockId: 'Step1',
+                  value: 'Step {{number}}'
+                }
+              }
+            },
+            result
+          }
+        ]}
+      >
+        <JourneyProvider value={{ journey }}>
+          <Step {...block} />
+        </JourneyProvider>
+      </MockedProvider>
+    )
+
+    await waitFor(() => expect(result).not.toHaveBeenCalled())
+  })
+
   it('should set seoTitle to [journey name (step name)] if activeStep and on first card', () => {
     treeBlocksVar([block])
     blockHistoryVar([block])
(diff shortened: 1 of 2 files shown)


levi-zustiak:
Are there any other tests you'd like to see for the activeStep?

libs/journeys/ui/src/components/Step/Step.spec.tsx
```diff
@@ -220,23 +283,89 @@ describe('Step', () => {
...
+          {/* eslint-disable-next-line react/no-children-prop */}
+          <Step {...block} children={[]} />
+        </JourneyProvider>
       </MockedProvider>
     )
     expect(baseElement).toContainHTML('<body><div /></body>')
   })
 
-  // TODO (SWIPE): test activeStep
+  it('should not send stepViewEvent if not activeStep', async () => {
```
csiyang, line 300: This test is failing for me locally, can you double checK?

libs/journeys/ui/src/components/Step/Step.spec.tsx
```diff
@@ -220,23 +283,89 @@ describe('Step', () => {
...
+          id: 'uuid',
+          __typename: 'StepViewEvent'
+        }
+      }
+    }))
+
+    render(
+      <MockedProvider
+        mocks={[
+          {
```
csiyang, line 316: We're in the middle of converting to a new standard of writting out these mocks. You can find a recent example of one I wrote in `SwipeNavigation.tsx` (line 84)

csiyang (changes requested):

libs/journeys/ui/src/components/Step/Step.spec.tsx
```diff
@@ -179,64 +268,65 @@ describe('Step', () => {
...
         </JourneyProvider>
       </MockedProvider>
     )
+
     await waitFor(() => expect(result).not.toHaveBeenCalled())
   })
 
-  it('should render blocks', () => {
-    const { getByText } = render(
+  it('should not send stepViewEvent if not activeStep - broken', async () => {
```
levi-zustiak, line 281: This test fails since the result mock has been called in the other tests. Should I create a new result mock like I did in the working version of this test?

I also tried to reset the mock using jest, but it caused Apollo to throw errors

libs/journeys/ui/src/components/Step/Step.spec.tsx
```diff
@@ -179,64 +268,65 @@ describe('Step', () => {
...
+    mockUuidv4.mockReturnValueOnce('uuid')
+    treeBlocksVar([])
+    blockHistoryVar([block, { ...block, id: 'Step2' }])
+
+    render(
       <MockedProvider>
```
csiyang, line 287: Does this work? If not, lets make one just for this test
```suggestion
      <MockedProvider mocks={[mockStepViewEventCreate]}>
```

libs/journeys/ui/src/components/Step/Step.tsx
```diff
@@ -47,7 +47,11 @@ export function Step({
     (variant === 'default' || variant === 'embed') && activeStep
 
   useEffect(() => {
-    if (activeJourneyStep && isActiveBlockOrDescendant(blockId)) {
+    if (
+      activeJourneyStep &&
+      isActiveBlockOrDescendant(blockId) &&
+      wrappers === undefined
```
levi-zustiak, line 53: This check was added for the wrapper test. I don't really have the context for the wrapper functionality or if it's still valid, but this now will prevent the event from firing if there are wrappers

csiyang (approved):

PR #2338, opened 2024-02-19
test: update card tests

levi-zustiak:
# Description
- Updates tests for Card component

[Link](https://3.basecamp.com/3105655/buckets/35958767/todos/7065659175) to Basecamp Todo


csiyang (commented):
Nice what you have here is good, I just realised there are some missing tests that are missing for the navigation that it would be good to add now. Most of these you should be able to get from other examples with some minor alterations.

Basically all the tests from `SwipeNavigation.spec.tsx` should get added

edmonday (approved):

PR #2387, opened 2024-02-26
feat: more prominent unmute button

levi-zustiak:
# Description
Modifies the `VideoControls` component to combine the play/pause button with the unmute button on mobile devices.

- [Link](https://3.basecamp.com/3105655/buckets/35958767/todos/6986070521) to Basecamp Todo

- [ ] Mobile devices will show the Play button first if not playing or autoplay, then transition to the unmute button if the video is muted, and then to the pause button after unmuting.
- [ ] Desktop controls should not be affected by the changes to mobile controls
- [ ] Playback controls should be kept in sync between desktop and mobile in case a user switches between the two.

# Walkthrough
Testing can be done using a muted video or muting the video on a desktop viewport and then viewing the page on a mobile viewport using developer tools.


libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -488,6 +552,23 @@ export function VideoControls({
                       }
                     }}
                   />
+                  <IconButton
```
csiyang, line 561: The position of this has moved from where it originally was. Can we shift it back?

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -456,17 +531,6 @@ export function VideoControls({
...
-                      <VolumeOffOutlined />
-                    ) : volume > 60 ? (
-                      <VolumeUpOutlined />
-                    ) : volume > 30 ? (
-                      <VolumeDownOutlined />
-                    ) : (
-                      <VolumeMuteOutlined />
-                    )}
-                  </IconButton>
                   <Slider
```
csiyang, line 540: The slider behaviour has changed now, can make it so that it can stay open again?

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -320,45 +420,18 @@ export function VideoControls({
         timeout={{ exit: 3000 }}
       >
         <Stack justifyContent="flex-end" sx={{ height: '100%' }}>
-          {/* Mute / Unmute */}
-          <Stack
```
csiyang, line 324: The removal of this has also shifted the play button up, can we bring back what's necessary to keep it at its original position?

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -13,15 +13,39 @@ import Fade from '@mui/material/Fade'
 import IconButton from '@mui/material/IconButton'
 import Slider from '@mui/material/Slider'
 import Stack from '@mui/material/Stack'
+import { useTheme } from '@mui/material/styles'
 import Typography from '@mui/material/Typography'
+import useMediaQuery from '@mui/material/useMediaQuery'
```
csiyang, line 18: Just a FYI, we're pretty tight on the use `useMediaQuery.tsx`, it causes layout shift when applied to UI elements. But it's fine in this case since it only affect the click handlers

csiyang (changes requested):
The file is getting a little big, do you think we could also refactor, and pull out the the `desktop controls`, `mobile controls` and `centre button` into their own components?

Just the rendered components themselves (if there is logic only that component uses it can be moved as well, otherwise just pass the required props in). I don't think we would need to write any new tests either, the main ones from `VideoControls.spec.tsx` should be sufficient.

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
levi-zustiak: I didn't see another example for determining the viewport width other than useMediaQuery, do you guys have an example of that somewhere?

levi-zustiak:
@csiyang couldn't really figure out a better folder structure for the mobile/desktop/general components, so if you have any recommendations I'd love to hear them.

csiyang (approved):

PR #2422, opened 2024-03-06
test: toolbar menu tests

jianwei1:
# Description

copilot:summary

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/36562260/todos/7145944356)

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough


apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
```diff
@@ -172,7 +124,6 @@ describe('Toolbar Menu', () => {
       getByRole('menuitem', { name: 'Create Template' })
     ).toBeInTheDocument()
     expect(getByRole('menuitem', { name: 'Language' })).toBeInTheDocument()
-    expect(getByRole('menuitem', { name: 'Analytics' })).toBeInTheDocument()
     expect(getByRole('menuitem', { name: 'Copy Link' })).toBeInTheDocument()
   })
```
csiyang, line 129: I don't see any logic to that requires a publisher and a template for this to show up, just template. This also means you only ned the `Template Settings` assertion
```suggestion
 it('should render template menu', async () => {
```

csiyang (changes requested):
Also add tests for the items that only show up in mobile

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
```diff
@@ -96,7 +96,9 @@ export function Menu(): ReactElement {
           <StrategyItem variant="menu-item" closeMenu={handleCloseMenu} />
         )}
         {journey != null &&
-          (journey?.template !== true || isPublisher != null) && <Divider />}
+          (journey?.template !== true || isPublisher != null) && (
+            <Divider data-testid="menu-divider" />
```
csiyang, line 100: I actually think the mui dividers have a role on them, and so shouldn't need to add a testid

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
```diff
@@ -100,15 +63,11 @@ describe('Toolbar Menu', () => {
...
-    expect(
-      getByRole('menuitem', { name: 'Duplicate Card' })
-    ).toBeInTheDocument()
-    expect(getByRole('menuitem', { name: 'Delete Card' })).toBeInTheDocument()
+    expect(getByRole('menuitem', { name: 'Manage Access' })).toBeInTheDocument()
     expect(getByRole('menuitem', { name: 'Title' })).toBeInTheDocument()
     expect(getByRole('menuitem', { name: 'Description' })).toBeInTheDocument()
     expect(getByRole('menuitem', { name: 'Language' })).toBeInTheDocument()
-    expect(getByRole('menuitem', { name: 'Analytics' })).toBeInTheDocument()
+    expect(getByTestId('menu-divider')).toBeInTheDocument()
```
csiyang, line 70: This actually shouldn't show up here, only when its not a template and accessed by a publisher. This also reveals this test is throwing a false positive here. Can you investigate and see if any of the set up for this test needs to be updated?

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
jianwei1: This isn't actually a false positive. The menu divider should show up when its not a template OR accessed by a publisher. Since the journey isn't a template, we expect the divider to be here

csiyang (approved):

PR #2434, opened 2024-03-07
test: radio option block node

levi-zustiak:
# Description
- Add tests for `RadioOptionBlockNode`

[Link](https://3.basecamp.com/3105655/buckets/36562260/todos/7145948371) to Basecamp Todo


csiyang (changes requested):
**Update**
- update component useTranslation import
- add test translation file into spec
- Test when text `Option` is used

csiyang (approved):

PR #2444, opened 2024-03-11
test: text response block node

levi-zustiak:
# Description
- Adds tests for `TextResponseBlockNode`

- [Link](https://3.basecamp.com/3105655/buckets/36562260/todos/7145949638) to Basecamp Todo


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+    parentBlockId: 'CardBlock.id',
+    parentOrder: 0,
+    label: 'TextResponseBlock',
+    hint: null,
+    minRows: null,
+    submitLabel: null,
+    submitIconId: null,
+    action: null,
+    children: []
+  } satisfies TreeBlock<TextResponseBlock>
```
csiyang, line 37: Wow, didn't know `satisfies` was a thing! I'm still looking into exactly how it works, but I dont think its necessary here.

```suggestion
  const mockTextResponseBlock: TreeBlock<TextResponseBlock> = {
    __typename: 'TextResponseBlock',
    id: 'TextResponseBlock.id',
    parentBlockId: 'CardBlock.id',
    parentOrder: 0,
    label: 'TextResponseBlock',
    hint: null,
    minRows: null,
    submitLabel: null,
    submitIconId: null,
    action: null,
    children: []
  }
```

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+  it('should render text response block node with label', () => {
+    render(
+      <ReactFlowProvider>
+        <MockedProvider>
+          <TextResponseBlockNode {...defaultProps} />
+        </MockedProvider>
+      </ReactFlowProvider>
+    )
+
+    expect(screen.getByText('Feedback')).toBeInTheDocument()
```
csiyang, line 89:
```suggestion
    const { getByText } = render(
      <ReactFlowProvider>
        <MockedProvider>
          <TextResponseBlockNode {...defaultProps} />
        </MockedProvider>
      </ReactFlowProvider>
    )

    expect(getByText('Feedback')).toBeInTheDocument()
```

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
levi-zustiak: Yeah, it's a pretty neat type! I'm fine switching to the annotation since it's really the same in this instance. The main idea is to provide 'narrow' types vs 'wider' types which is pretty nice. https://www.learningtypescript.com/articles/the-satisfies-operator this article does a great job of explaining it.

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
levi-zustiak: Is it possible that we move to using `screen` queries instead of render queries? It's recommended to use `screen` for dx and usability. And once again, if this is preferred for an outstanding reason or just preference that works too.

csiyang: I would defer to @tataihono for this decision

tataihono: yes. I agree. Was something we should have done a long time ago. If we were going to do it we should consider linting and making the update in cooldown.

csiyang (approved):

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {
+  ThemeMode,
+  ThemeName
+} from '../../../../../../../__generated__/globalTypes'
+import { mockReactFlow } from '../../../../../../../test/mockReactFlow'
+import '../../../../../../../test/i18n'
+
+import { TextResponseBlockNode } from './TextResponseBlockNode'
+
+describe('TextResponseBlockNode', () => {
```
csiyang, line 20: The title of the file also needs to be changed to this
```suggestion
describe('TextResponseBlockNode', () => {
```

csiyang, line 20: The title of the file also needs to be changed to this `TextResponseBlockNode.spec.tsx`
```suggestion
describe('TextResponseBlockNode', () => {
```

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/TextResponseBlockNode/TextResponseBlock.spec.tsx
levi-zustiak: Oops, good catch!

csiyang (approved):

PR #2468, opened 2024-03-14
test: social preview

levi-zustiak:
# Description
- Refactors tests for `SocialPreview`

[Link](https://3.basecamp.com/3105655/buckets/36562260/todos/7145955438) to Basecamp Todo

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Content/Social/SocialPreview.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Content/Social/SocialPreview.spec.tsx
@@ -1,25 +1,81 @@
-import { render } from '@testing-library/react'
+import useMediaQuery from '@mui/material/useMediaQuery'
+import { render, screen } from '@testing-library/react'
+
+import {
+  ActiveCanvasDetailsDrawer,
+  ActiveContent,
+  ActiveFab,
+  ActiveSlide,
+  EditorState,
+  useEditor
+} from '@core/journeys/ui/EditorProvider'
 
 import { ThemeProvider } from '../../../../ThemeProvider'
 
 import { SocialPreview } from './SocialPreview'
 
+jest.mock('@mui/material/useMediaQuery', () => ({
+  __esModule: true,
+  default: jest.fn()
+}))
+
+const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
+  typeof useMediaQuery
+>
+
+jest.mock('@core/journeys/ui/EditorProvider', () => {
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/Content/Social/SocialPreview.spec.tsx
```diff
@@ -1,25 +1,81 @@
...
+jest.mock('@mui/material/useMediaQuery', () => ({
+  __esModule: true,
+  default: jest.fn()
+}))
+
+const mockUseMediaQuery = useMediaQuery as jest.MockedFunction<
+  typeof useMediaQuery
+>
+
+jest.mock('@core/journeys/ui/EditorProvider', () => {
```
csiyang, line 26: Any special reason we need to mock this out? Can we use the `EditorProvider` instead? (example below)

apps/journeys-admin/src/components/Editor/Slider/Content/Social/SocialPreview.spec.tsx
```diff
@@ -1,25 +1,81 @@
...
+      dispatch: jest.fn()
+    })
+  })
+
+  it('should render Message Post and Fab for viewports larger than mobile', () => {
+    render(
       <ThemeProvider>
         <SocialPreview />
       </ThemeProvider>
     )
```
csiyang, line 59:
```suggestion
  it('should render Message Post and Fab for viewports larger than mobile', () => {
    render(
      <EditorProvider initialState={state}>
        <ThemeProvider>
          <SocialPreview />
        </ThemeProvider>
      </EditorProvider>
    )
```

csiyang (changes requested):
Can you also please test the `handleSelect` in the component

apps/journeys-admin/src/components/Editor/Slider/Content/Social/SocialPreview.spec.tsx
levi-zustiak: Definitely would have been the better option for the two original tests haha I am going to use the mock for the click test

csiyang: Do you think we can change it in this PR? For code maintainability as well as don't want someone to accidentally copy this somewhere else.

levi-zustiak: Do you have an example to mock the dispatch function for me to use then? I'm not sure how to do that without this

csiyang (approved):

PR #2540, opened 2024-03-27
feat: video block node stories

levi-zustiak:
# Description
- Adds stories for `VideoBlockNode`

[Link](https://3.basecamp.com/3105655/buckets/36562260/todos/7150978348) to Basecamp Todo

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/VideoBlockNode/VideoBlockNode.stories.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/VideoBlockNode/VideoBlockNode.stories.tsx
@@ -20,7 +20,7 @@ import { journeysAdminConfig } from '../../../../../../libs/storybook'
 import { VideoBlockNode } from '.'
 
 const VideoBlockNodeStory: Meta<typeof VideoBlockNode> = {
-  ...journeysAdminConfig,
+  ...simpleComponentConfig,
   component: VideoBlockNode,
   title: 'Journeys-Admin/Editor/Slider/JourneyFlow/nodes/VideoBlockNode'
 }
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/VideoBlockNode/VideoBlockNode.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+          <ReactFlow {...args}>
+            <Background color="#aaa" gap={16} />
+          </ReactFlow>
+        </Box>
+      </MockedProvider>
+    )
+  }
+}
+
+export const Default = {
```
csiyang, line 104: I think this is the only variant we need, the video title logic should be covered already in the unit tests, and visually these aren't different enough to have its own variant.

In the past we've ran into problems where we were exceeding our snapshot limits, so we try to limit story variants to visual changes (empty/ filled, selected etc...)

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/VideoBlockNode/VideoBlockNode.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+  ThemeMode,
+  ThemeName,
+  VideoBlockSource
+} from '../../../../../../../__generated__/globalTypes'
+import { journeysAdminConfig } from '../../../../../../libs/storybook'
+
+import { VideoBlockNode } from '.'
+
+const VideoBlockNodeStory: Meta<typeof VideoBlockNode> = {
+  ...journeysAdminConfig,
```
csiyang, line 23: I don't think this is explained anywhere, but FYI, we use `journeysAdminConfig` for when we need both mobile and and desktop snapshots. When we only need a desktop snapshot, we use `simpleComponentConfig`. Since the flow looks the same on all screen sizes, we use simple to cut down on the number of snapshots.

```suggestion
  ...simpleComponentConfig,
```

csiyang (changes requested):

csiyang (approved):

PR #2565, opened 2024-04-02
feat: sign in tabs

edmonday:
# Description

Added tabs to help user know if they are creating an account or signing in.

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/36562315/todos/7149656416)

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough


apps/journeys-admin/src/components/TemplateView/CreateJourneyButton/CreateJourneyButton.tsx
```diff
@@ -77,15 +77,16 @@ export function CreateJourneyButton({
...
     const url = window.location.origin + router.asPath
     void router.push(
       {
         pathname: '/users/sign-in',
         query: {
           redirect: url.includes('createNew')
             ? url
-            : `${window.location.origin + router.asPath}?createNew=true`
+            : `${window.location.origin + router.asPath}?createNew=true`,
+          action: action === 'login' ? 'login' : 'signup'
```
GeronimoJohn, line 89: Like the idea. Just thinking that we're never really using sign up. Thoughts on a query that only updates the url if value passed in is relevant, if possible. E.g. `handleSignIn(login: boolean): void => () ` Then can use login accordingly. We can remove the type this way too

edmonday: yup, sounds good. Made the change, just need to update tests

GeronimoJohn (approved):

PR #2582, opened 2024-04-03
feat: limit one video block per parent

mikeallisonJS:
# Description

copilot:summary

- [Link to Basecamp Todo
](https://3.basecamp.com/3105655/buckets/36562350/todos/7252796180)

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough

### DIFF
--- a/apps/api-journeys/src/app/modules/block/video/video.resolver.ts
+++ b/apps/api-journeys/src/app/modules/block/video/video.resolver.ts
@@ -140,6 +140,18 @@ export class VideoBlockResolver {
             tx
           )
       }
+
+      const existingVideosOnParent = await tx.block.count({
+        where: { parentBlockId: input.parentBlockId, typename: 'VideoBlock' }
+      })
+      if (existingVideosOnParent > 0)
+        throw new GraphQLError(
+          'Parent block already has an existing video block',
+          {
+            extensions: { code: 'BAD_REQUEST' }
+          }
+        )
+
       const block = await tx.block.create({
         data: {
           ...omit(
--- a/apps/api-journeys/src/app/modules/block/video/video.resolver.ts
+++ b/apps/api-journeys/src/app/modules/block/video/video.resolver.ts
@@ -141,10 +141,10 @@ export class VideoBlockResolver {
           )
       }
 
-      const existingVideosOnParent = await tx.block.count({
+      const existingVideoOnParent = await tx.block.findFirst({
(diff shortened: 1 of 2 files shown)


apps/api-journeys/src/app/modules/block/video/video.resolver.ts
```diff
@@ -140,6 +140,18 @@ export class VideoBlockResolver {
...
       }
+
+      const existingVideosOnParent = await tx.block.count({
+        where: { parentBlockId: input.parentBlockId, typename: 'VideoBlock' }
+      })
+      if (existingVideosOnParent > 0)
+        throw new GraphQLError(
+          'Parent block already has an existing video block',
+          {
+            extensions: { code: 'BAD_REQUEST' }
```
tataihono, line 151: This is a User input error. Can you change it to reflect that?

apps/api-journeys/src/app/modules/block/video/video.resolver.ts
```diff
@@ -140,6 +140,18 @@ export class VideoBlockResolver {
             tx
           )
       }
+
+      const existingVideosOnParent = await tx.block.count({
```
tataihono, line 144: You shouldn't use count as it unnecessarily goes through the whole table.

The better approach is to use findFirst, which is basically a LIMIT 1 on the database, so the database can stop searching for more results after the first hit.

tataihono (changes requested):

apps/api-journeys/src/app/modules/block/video/video.resolver.ts
mikeallisonJS: got it

apps/api-journeys/src/app/modules/block/video/video.resolver.ts
```diff
@@ -140,6 +140,18 @@ export class VideoBlockResolver {
...
       }
+
+      const existingVideoOnParent = await tx.block.findFirst({
+        where: { parentBlockId: input.parentBlockId, typename: 'VideoBlock' }
+      })
+      if (existingVideoOnParent != null)
+        throw new GraphQLError(
+          'Parent block already has an existing video block',
+          {
+            extensions: { code: 'USER_INPUT_ERROR' }
```
tataihono, line 151:
```suggestion
            extensions: { code: 'BAD_USER_INPUT' }
```

tataihono (changes requested):

tataihono (approved):

PR #2591, opened 2024-04-05
feat: video race condition

mikeallisonJS:
# Description

copilot:summary

- Link to Basecamp Todo

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough


tataihono (changes requested):
Apologies. We should get rid of the first check. And only perform this check after the block creation after:
```
      if (!ability.can(Action.Update, subject('Journey', block.journey)))
        throw new GraphQLError('user is not allowed to create block', {
          extensions: { code: 'FORBIDDEN' }
        })
```
we want to check if there are now two video blocks with the same parent id.

tataihono (approved):

PR #2610, opened 2024-04-09
refactor: video block editor stories

levi-zustiak:
# Description
- Updates `VideoBlockEditor` story
- Updates paths for `VideoLibrary` stories

[Link](https://3.basecamp.com/3105655/buckets/36562260/todos/7156076000) to Basecamp Todo


csiyang:
Also, resolve lint errors

levi-zustiak:
@csiyang the full Property drawer has drop downs that give the correct padding. Should the full Video property drawer be added then? or just simulate the dropdown padding? 
image

csiyang:
> @csiyang the full Property drawer has drop downs that give the correct padding. Should the full Video property drawer be added then? or just simulate the dropdown padding? image

We shouldn't add the the accordion in, I think that may be confusing since we're adding in a something that is not being rendered by the component. (I think the drawer is fine because since it mostly changes the width of the component to how its normally used).

We can either just add some padding within the story itself to keep using the drawer, or remove the drawer wrapper in the story. Removing the drawer, more accurately shows the component from a technical standpoint (how it will fit within the parent), but also looks a little weird since we never use it without the drawer component in code.

We have a mix of these 2 solutions in code, but I think for the easiest/ most accurate snapshot, we should just put some padding in the story

csiyang (approved):

PR #2615, opened 2024-04-10
feat: example email

mikeallisonJS:
# Description

copilot:summary

- Link to Basecamp Todo

- [ ] Test A
- [ ] Test B

# Walkthrough

copilot:walkthrough

### DIFF
--- a/apps/api-users/src/app/modules/user/user.service.ts
+++ b/apps/api-users/src/app/modules/user/user.service.ts
@@ -1,3 +1,5 @@
+import crypto from 'crypto'
+
 import { InjectQueue } from '@nestjs/bullmq'
 import { Injectable } from '@nestjs/common'
 import { Queue } from 'bullmq'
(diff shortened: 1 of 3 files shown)


apps/api-users/src/app/modules/user/user.service.ts
```diff
@@ -93,7 +93,10 @@ export class UserService {
     email: string,
     redirect?: string
   ): Promise<void> {
-    const token = Math.floor(100000 + Math.random() * 900000).toString() // six digit, first is not 0
+    const isExample = email.endsWith('@example.com')
+    const token = isExample
+      ? process.env.EXAMPLE_EMAIL_TOKEN ?? ''
+      : Math.floor(100000 + Math.random() * 900000).toString() // six digit, first is not 0
```
tataihono, line 99: If there is no email token it should not be possible for the token to be empty.

```suggestion
    const isExample = email.endsWith('@example.com') && process.env.EXAMPLE_EMAIL_TOKEN != null
    const token = isExample
      ? process.env.EXAMPLE_EMAIL_TOKEN ?? ''
      : Math.floor(100000 + Math.random() * 900000).toString() // six digit, first is not 0
```

Also do you think we should use `getRandomValues()` from [Crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues). Math.random is not cryptographically secure.

tataihono (changes requested):

tataihono (dismissed):

tataihono (approved):

mikeallisonJS:
atlantis apply

PR #2640, opened 2024-04-16
fix: added share menu item on mobile

jianwei1:
# Description

### Issue

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062269#__recording_7297314657)

### Solution

# External Changes

# Additional information


apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
```diff
@@ -31,243 +32,305 @@ describe('Toolbar Menu', () => {
...
-        {
-          query: { param: 'description' },
-          push,
-          events: {
-            on
-          }
-        },
-        undefined,
-        { shallow: true }
+    it('should render preview, analytics, strategy and share buttons on mobile', () => {
```
GeronimoJohn, line 292: description says it checks for these 3 things. The test really checks for everything in the menu? Are we able to update either or? Else looks good

GeronimoJohn (approved):

PR #2642, opened 2024-04-16
feat: block positioning

csiyang:
# Description

### Issue

Nodes in journey flow did not have their own positions

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062673)

### Solution

- Refactor nodes and edges algorithm - only runs if there is a stepNode with no x, y
- Update mutations and query for steps
- ActionNode is now apart of StepNode
- Fixed StepNode creation on handle click and drag
- Fixed StepNode duplication and deletion

# External Changes

N/A

# Additional information

- A lot of the StepNode create/ update/ deletes were fixed through adding additional manual cache updates
- A lot of the action nodes are now redundant and have been deleted


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -18,398 +28,293 @@ import { v4 as uuidv4 } from 'uuid'
...
-        // markerEnd: {
-        //   type: MarkerType.Arrow
-        // },
-        style: {
-          // strokeWidth: 2,
-          strokeDasharray: 4
-        },
-        type: 'smart'
-      })
+export const GET_STEP_BLOCKS_WITH_POSITION = gql`
```
GeronimoJohn, line 50: NIT: could be refactor to be a custom hook. https://docs.core.jesusfilm.org/docs/basics/frontend/custom-hooks/ thought debatable if really need to be one

GeronimoJohn (approved):
Approving as I think the rest can be picked up by the bugs team

apps/api-journeys/db/ERD.svg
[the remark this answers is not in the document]
csiyang: It probably should've gone in with the original api changes, but I think its fine to go in with this PR

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
csiyang: What should be the pattern for writing custom hooks? I think this is the only place this is being used, I don't think we'll call this anywhere else for now.

GeronimoJohn: Yea was thinking that now, not really being used anywhere

PR #2650, opened 2024-04-18
fix: remove hover icon

zwsaqx:
# Description

### Issue

plus icon appearing on mobile when tapping as well as broken tooltips

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062030)

### Solution

remove icon and tooltips on mobile resolutions

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/Button/Button.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/Button/Button.tsx
@@ -73,7 +73,7 @@ export function Button({
             ]
           }
         }}
-        sx={{ opacity: { xs: 0, sm: 1 } }}
+        sx={{ display: { xs: 'none', sm: 'flex' } }}
       >
         <Box>
           <Card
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/Button/Button.tsx
```diff
@@ -79,6 +73,7 @@ export function Button({
             ]
           }
         }}
+        sx={{ opacity: { xs: 0, sm: 1 } }}
```
GeronimoJohn, line 76: Since we're not really wanting it to show at all on mobile. Can we make change this to be rendering through the display property instead of opacity. e.g. display: { xs: 'none', ... }

Though small, `opacity: 0` still gets calculated on layout and paint process. Apply the same logic to the opacity below please

GeronimoJohn: Let me know if you run into any issues

GeronimoJohn (approved):

PR #2652, opened 2024-04-18
fix: hide template filters with no background

scottkelso:
# Description

### Issue

Filter card buttons (FeltNeedsButton.tsx) should only display if they have a background.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071027/todos/7306953317)

### Solution

Initially tried to conditionally render the ButtonBase component iff the `tagImageData` variable is defined. However this lead to uneven spacing. 

Instead, let the TagCarousel filter out a fixed set of tags that are known to have valid backgrounds as hard-coded in FeltNeedsButton. 

image

# External Changes

n/a

# Additional information

UI test added took a long time to debug. Solution: I needed to make another button assertion so that jest would wait for the render to complete.

### DIFF
--- a/apps/journeys-admin/src/components/TemplateGallery/TagCarousels/TagCarousels.spec.tsx
+++ b/apps/journeys-admin/src/components/TemplateGallery/TagCarousels/TagCarousels.spec.tsx
@@ -22,15 +22,16 @@ describe('TagCarousels', () => {
     expect(getAllByTestId('collections-button-loading')).toHaveLength(2)
 
     await waitFor(async () => {
-      await expect(
+      expect(
         within(feltNeedsCarousel).getByRole('button', {
-          name: 'Acceptance Acceptance Acceptance'
+          // The name of the tag is the alt text + tag label + tag label
+          name: 'Acceptance tag Acceptance Acceptance'
         })
       ).toBeInTheDocument()
 
-      await expect(
+      expect(
         getByRole('button', {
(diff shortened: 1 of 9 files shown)


scottkelso:
I think @tataihono is working on a fix for the failing lighthouse CI/CD step. By this stage I have already done some steps to improving the accessibility score using google chromes lighthouse as per latest commits (improved from 96% to 100% in mobile)

## Mobile
image


## Desktop
image

apps/journeys-admin/src/components/TemplateGallery/TagCarousels/TagCarousels.spec.tsx
```diff
@@ -45,22 +46,45 @@ describe('TagCarousels', () => {
...
 
+  it('should filter out felt needs tags that have no backgrounds', async () => {
+    const onChange = jest.fn()
+    const { getByRole, queryAllByText } = render(
+      <MockedProvider mocks={[getTagsMock]}>
+        <TagCarousels onChange={onChange} />
+      </MockedProvider>
+    )
+
+    // Check that carousel has loaded properly
```
GeronimoJohn, line 73: NIT: can do without the comment. I think our goal for our codebase is to write code well, that there is no need for comments. Which you're doing well here 😄

GeronimoJohn (approved):

GeronimoJohn (approved):

PR #2656, opened 2024-04-19
chore: node 20

mikeallisonJS:
# Description

Update to Node 20. Performance update & chore

### Issue

Node 20 is more performant and a logical update if there are no dependency issues, which it doesn't seem like there are.

### Solution

Updated GHAs, Docker images and dev container

# External Changes

none

# Additional information

Needs to be tested on stage first


tataihono (dismissed):
are you also able to update the github actions that are still running on node < 20?

mikeallisonJS:
> are you also able to update the github actions that are still running on node < 20?

Updated the ones I could

tataihono (approved):

PR #2659, opened 2024-04-20
feat: arclight expansion

mikeallisonJS:
# Description

### Issue

New endpoints for arclight facade

### Solution

New endpoints for arclight facade, detailing missing information and services

# External Changes

None

# Additional information

Code commented for missing information, params, services, etc


tataihono (changes requested):
can we update the links to use the location the user is visiting from? e.g if its running on localhost then use that link, if its in the preview env etc.

tataihono (approved):

PR #2660, opened 2024-04-22
fix: image and video thumbnails not displaying on node

jianwei1:
# Description

### Issue

In certain situations, the background image, video thumbnail, or video cover image aren't displaying on the journey flow nodes

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062629)

### Solution

Made changes to the functions which get the bgImage const in getCardMetadata.ts

# Additional information

## Title and Subtitle fields

While working on this, I was initially only altering the bgImage const. However as I was working on the ticket I discovered another bug where if the user selected a custom or youtube video, the title and subtitle also weren't displaying on the node. This is why I also altered the title and subtitle consts in getCardMetadata.ts

## Truncating the title text

I also added the wordBreak prop to the title typography, so that the title correctly truncates if it is one long string


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/libs/getCardMetadata/getCardMetadata.ts
```diff
@@ -15,27 +18,32 @@ interface CardMetadata {
...
+      // Use posterBlockId image or default poster image on video
+      (priorityBlock?.posterBlockId != null
+        ? (
+            priorityBlock.children.find(
+              (block) =>
+                block.id === priorityBlock.posterBlockId &&
+                block.__typename === 'ImageBlock'
+            ) as TreeBlock<ImageBlock>
+          )?.src
+        : priorityBlock?.video?.image ?? priorityBlock.image) ?? undefined
```
GeronimoJohn, line 46: Think we can put them all in one chained nullish check? so like check1 ?? check 2?? check3

GeronimoJohn (approved):

PR #2661, opened 2024-04-22
feat: update api videos schema

edmonday:
# Description

### Issue

The prisma schema for api-videos needs to be updated with new models `VideoDescription` and `VideoSnippet` to ingest translation data from Crowdin.

### Solution

`VideoDescription` and `VideoSnippet` have been converted to prisma models to ingest translation data from Crowdin. Seeding logic has been updated for api-videos.

The field `seoTitle` has been dropped from the `Video` model as we do not use that field.

# External Changes

# Additional information

This is a breaking change and will require @mikeallisonJS to reset prod and stage. All other engineers will need to be pinged about this to reset and to seed again.


tataihono:
@mikeallisonJS is doing this in a safe way that allows us to keep serving the API with no downtime.

PR #2669, opened 2024-04-24
fix: reset canvas focus when slider to journey flow

scottkelso:
# Description

### Issue

When one selects any block that results in the QuickControls to appear, and then navigates to the journey flow slide, the QuickControls remain open. We want the parent card to be focused on and the QuickControls to close.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062111)

### Solution

Add a dispatch on the slider useEffect function that sets the selected block value to the parent card block.

# External Changes

n/a

# Additional information

This fix seems to add some lagging in terms of rendering the canvas when the slider changes but the hope is that this can be resolved in another ticket. https://3.basecamp.com/3105655/buckets/37071143/todos/7307053111

### DIFF
--- a/libs/journeys/ui/src/libs/EditorProvider/EditorProvider.tsx
+++ b/libs/journeys/ui/src/libs/EditorProvider/EditorProvider.tsx
@@ -101,6 +101,10 @@ interface SetSelectedBlockAction {
   type: 'SetSelectedBlockAction'
   selectedBlock?: TreeBlock
 }
+interface SetSelectedBlockOnlyAction {
+  type: 'SetSelectedBlockOnlyAction'
+  selectedBlock?: TreeBlock
+}
 export interface SetSelectedBlockByIdAction {
   type: 'SetSelectedBlockByIdAction'
   selectedBlockId?: string
@@ -128,6 +132,7 @@ type EditorAction =
   | SetActiveSlideAction
   | SetSelectedAttributeIdAction
   | SetSelectedBlockAction
+  | SetSelectedBlockOnlyAction
   | SetSelectedBlockByIdAction
   | SetSelectedGoalUrlAction
   | SetSelectedStepAction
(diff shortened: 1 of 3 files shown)


libs/journeys/ui/src/libs/EditorProvider/EditorProvider.tsx
```diff
@@ -172,6 +177,8 @@ export const reducer = (
         activeContent: ActiveContent.Canvas,
         activeSlide: ActiveSlide.Content
       }
+    case 'SetSelectedBlockOnlyAction':
+      return { ...state, selectedBlock: action.selectedBlock }
```
scottkelso, line 181: Tried modifying this reducer case to apply to my use case but decided it made it more complicated looking than it needed to be...

```
case 'SetSelectedBlockAction':
  // Normally we assue the content view is needed
  return action.skipUpdateContent === true
    ? {
        ...state,
        selectedBlock: action.selectedBlock
      }
    : {
        ...state,
        selectedBlock: action.selectedBlock,
        activeCanvasDetailsDrawer: ActiveCanvasDetailsDrawer.Properties,
        activeContent: ActiveContent.Canvas,
        activeSlide: ActiveSlide.Content
      }


// my use case
dispatch{{
  type: 'SetSelectedBlockAction',
  selectedBlock: block,
  skipUpdateContent: true
})

// original use case
dispatch{{
  type: 'SetSelectedBlockAction',
  selectedBlock: block,
})
```

I also justified not renaming the original `SetSelectedBlockAction` because the `SetSelectedBlockOnlyAction` is altogether more rare. I.e. nearly all use cases assume you are in the canvas view when selecting blocks. Plus many other reducer cases are named in a way that doesn't fully articulate everything they are modifying so I'd reed to go on a refactoring rampage! 😆

GeronimoJohn (approved):

PR #2671, opened 2024-04-25
feat: api videos new fields p2

mikeallisonJS:
# Description

### Issue

Follow-up to move new fields to correct place and call correctly

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071156/todos/7321833284]

### Solution

Drop old tables, remap api to new fields

# External Changes

None

# Additional information

Requires part 1 and seed to be merged


tataihono (dismissed):

apps/api-videos/src/app/modules/algolia/algolia.service.ts
```diff
@@ -25,7 +25,9 @@ export class AlgoliaService {
         take: 1000,
         skip: offset,
         include: {
-          video: { include: { title: true } },
+          video: {
+            include: { title: true, description: true, imageAlt: true }
```
tataihono, line 29:
```suggestion
            include: { title: true, description: true, imageAlt: true, snippet: true }
```

tataihono (changes requested):

PR #2681, opened 2024-04-28
fix: unsplash duplication issue

GeronimoJohn:
# Description

### Issue

_Clicking load more button within the unsplash gallery component causes a duplicated datda_

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062018)

### Solution

_This is happening because we're not managing the async functionality right. Update the fetchMore calls to be async_


apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/ImageBlockEditor/UnsplashGallery/UnsplashGallery.tsx
```diff
@@ -95,75 +100,101 @@ interface UnsplashGalleryProps {
...
+  async function handleFetchMore(): Promise<void> {
+    const newPage = page + 1
+    const variables = { page: newPage, perPage: 20 }
     if (query == null) {
-      void fetchMoreList({
-        variables: { collectionId, page: page + 1, perPage: 20 }
+      await fetchMoreCollection({
+        variables: { collectionId, ...variables },
+        updateQuery: (prevResult, { fetchMoreResult }) => {
+          // bandage fix, prev result initially returns null for page 1
```
GeronimoJohn, line 168: https://www.apollographql.com/docs/react/data/queries/#queryresult-interface-previousdata

levi-zustiak (approved):
I didn't see anything out of place. As for the bandage fix, I see what you mean about it not being ideal, but that does seem to be the fix we can use for now

PR #2686, opened 2024-04-30
feat: ingest crowdin translations

edmonday:
# Description

### Issue

As we look to migrate video data from Arclight to core, translation data from Crowdin will also need to be ingested.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663169/todos/7506042571)

### Solution

Crowdin translations are retrieved via CDN, transformed and pushed to api-videos. BullMQ job is set to run every 12 hours.

Custom mapping of Crowdin language codes and WESS codes are set up via Crowdin.

# External Changes

Crowdin distribution hash will need to updated in doppler for prod api-videos.

# Additional information

- dependent on bibleBook importer portion of https://github.com/JesusFilm/core/pull/2795
- requires `crowdinId` field to be populated for `studyQuestions` on prod


apps/api-videos/src/app/modules/crowdin/crowdin.consumer.ts
```diff
@@ -0,0 +1,16 @@
...
+import { CrowdinService } from './crowdin.service'
+
+@Processor('api-videos-crowdin')
+export class CrowdinConsumer extends WorkerHost {
+  constructor(private readonly crowdinService: CrowdinService) {
+    super()
+  }
+
+  async process(job: Job): Promise<void> {
+    console.log(job.name)
```
tataihono, line 13: drop

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+interface CrowdinData {
+  resources: CrowdinResource
+}
+
+@Injectable()
+export class CrowdinService {
+  constructor(private readonly prisma: PrismaService) {}
+
+  async getCrowdinTranslations(): Promise<void> {
```
tataihono, line 36:
```suggestion
  async pullTranslations(): Promise<void> {
```

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
```diff
@@ -0,0 +1,214 @@
...
+    const hash = process.env.CROWDIN_DISTRIBUTION_HASH ?? ''
+
+    if (hash === '') throw new Error('crowdin distribution hash not set')
```
tataihono, line 39:
```suggestion
    if (process.env.CROWDIN_DISTRIBUTION_HASH == null) throw new Error('crowdin distribution hash not set')
```

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
```diff
@@ -0,0 +1,214 @@
...
+      disableStringsCache: true
+    })
+
+    const languages = await client.getTranslations()
+
+    const wessLanguageMap = mapValues(languages, (language) => {
+      const regex = /\/content\/(.*?)\.xliff/
+      const match = language[0].file.match(regex)
+      return match != null ? match[1] : ''
+    })
```
tataihono, line 52: what is this doing? I think its unclear.

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    await Promise.all(
+      map(languages, async (files, languageCode) => {
+        await Promise.all(
+          files.map(async ({ content }: { content: string }) => {
+            const data: CrowdinData = await xliff12ToJs(content)
```
tataihono, line 58: the data json response should be validated through something like yup. So that we can be sure that it has the correct format otherwise fail. I did something like this with row validation on the bq import pr. can you check that ou t with @Kneesal then apply that here.

tataihono, line 58: this seems overly complex. Why can't we just map over one set of files?

tataihono (changes requested):
now that you have a working implementation can you take a run and simplifying your code? Particularly try to remove the number of maps that you are needing to do.

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
edmonday: The WESS language code for the custom mapping from Crowdin language codes can only be retrieved from the exported file name.  So what this is doing is creating an object with key value pairs `CrowdinLanguageCode : WESS Code` e.g `ko: 3804`

I then use this later on when storing the translations so we have the correct WESS code. This is needed since the converted xliff to json object has an attribute `targetLanguage` which is still the Crowdin language code.

What would you recommend in terms of making this more clear?

edmonday: Did more thinking around this. What do you think about this?

```
 await Promise.all(
      map(languages, async (language) => {
        const data: CrowdinData = await xliff12ToJs(language[0].content)
        const languageCode = language[0].file.match(/\/content\/(.*?)\.xliff/)
        if (languageCode != null)
          await this.storeTranslations(languageCode[1], data)
      })
    )
  }
```

this will get rid of the `wessLanguageMap` and remove one map.

apps/api-videos/src/app/modules/crowdin/crowdin.service.ts
```diff
@@ -0,0 +1,247 @@
...
+        {
+          const videoId = resName
+          const videos = await this.getVideos(videoId)
+          if (videos.length === 0)
+            throw new MatchError(`no matching videoId found for ${videoId}`)
+          await this.updateVideoTitle(videos[0].id, languageId, value)
+        }
+        break
+      case '/Arclight/collection_long_description.csv':
+      case '/Arclight/media_metadata_description.csv':
```
mikeallisonJS, line 136: Is this by chance a snippet and not a description?

mikeallisonJS (dismissed):

PR #2687, opened 2024-05-01
fix: active team on signin

scottkelso:
# Description

### Issue

When you signin for the first time, all is well. After signin out (which sets activeTeam to null) and then signing back in again, it usually will default to the 'Shared with me'

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071027/todos/7337389239)

### Solution

Add the `notifyOnNetworkStatusChange: true` flag to the `TeamProvider`'s query which forces a re-render when the call is in flight. This seems to ensure that dropdown menus update when activeTeam has been correctly fetched.

# External Changes

n/a

# Additional information

n/a

### DIFF
--- a/apps/journeys-admin/src/components/Team/TeamProvider/TeamProvider.tsx
+++ b/apps/journeys-admin/src/components/Team/TeamProvider/TeamProvider.tsx
@@ -66,6 +66,7 @@ export function TeamProvider({ children }: TeamProviderProps): ReactElement {
   const query = useQuery<GetLastActiveTeamIdAndTeams>(
     GET_LAST_ACTIVE_TEAM_ID_AND_TEAMS,
     {
+      notifyOnNetworkStatusChange: true,
       onCompleted: (data) => {
         if (activeTeam != null || data.teams == null) return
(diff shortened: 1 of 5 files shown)


apps/journeys-admin/src/components/Team/TeamProvider/TeamProvider.tsx
```diff
@@ -66,6 +66,7 @@ export function TeamProvider({ children }: TeamProviderProps): ReactElement {
   const query = useQuery<GetLastActiveTeamIdAndTeams>(
     GET_LAST_ACTIVE_TEAM_ID_AND_TEAMS,
     {
+      notifyOnNetworkStatusChange: true,
```
csiyang, line 69: Oh I was saying that this used to work, but it was actually a bug that got fixed and no longer triggers the onComplete to run after the first time.

scottkelso:
Completed by https://github.com/JesusFilm/core/pull/2695/files. It is still unknown why the behaviour in dev on my local machine was not matching that of staging.

PR #2269, opened 2024-02-01
feat: journey flow

tataihono:
# Description

- Card management through journey flow
- Editor design improvements

[- Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/36562260/todosets/7145202277)

Journeys admin deployment: https://journeys-admin-2269-jesusfilm.vercel.app/

PR #2707, opened 2024-05-06
fix: update next block id

GeronimoJohn:
# Description

### Issue

_We need to get rid of null values for our nextBlockId and for our navigate actions_

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062717)

### Solution

_Create a seed file that will be run only twice when needed. The seed file is to go through all the journeys. Find all the StepBlocks and NavigateActions and to update the values to be pointing to the next StepBlock if it exists_

Depends on #2269


apps/api-journeys/db/seed.ts
```diff
@@ -27,6 +28,7 @@ async function main(): Promise<void> {
   await discoveryAdminCenter()
   await discoveryAdminRight()
   await playwrightUserAccess()
+  await processNextBlockId()
```
tataihono, line 31: no. this will just be run onetime probably by mike.

apps/api-journeys/db/seeds/processNextBlockId.ts
```diff
@@ -0,0 +1,85 @@
...
+        )
+
+        await Promise.all(
+          currentStepBlockActions.map(async (block) => {
+            if (block != null) {
+              await prisma.block.update({
+                where: { id: block.id },
+                data: {
+                  action: {
+                    update: { blockId: nextStepBlock.id }
```
tataihono, line 58: do you also need to update the gtmEventName?

tataihono (changes requested):
looks good. just a few nits.

apps/api-journeys/db/seeds/processNextBlockId.ts
```diff
@@ -0,0 +1,88 @@
...
+    (block): block is StepBlock => block.typename === 'StepBlock'
+  )
+}
+
+function getBlocksWithNavigateActions(blocks): Block[] {
+  return blocks.filter(
+    (block): boolean =>
+      block.action != null && block.action.gtmEventName === 'NavigateAction'
+  )
+}
```
tataihono, line 20:
```suggestion
  actionType(obj: Action): string {
    if (get(obj, 'blockId') != null) return 'NavigateToBlockAction'
    if (get(obj, 'journeyId') != null) return 'NavigateToJourneyAction'
    if (get(obj, 'url') != null) return 'LinkAction'
    if (get(obj, 'email') != null) return 'EmailAction'
    return 'NavigateAction'
  }

function getBlocksWithNavigateActions(blocks): Block[] {
  return blocks.filter(
    (block): boolean =>
      block.action != null && actionType(block.action) === 'NavigateAction'
  )
}
```

apps/api-journeys/db/seeds/processNextBlockId.ts
tataihono, line 1: can you type the arguments to functions

tataihono (changes requested):

PR #2719, opened 2024-05-08
fix: add back button tooltip

scottkelso:
# Description

### Issue

Add a tooltip on the back button of the journey flow view.
Also change the icon to bullet list.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062745)

### Solution

As above

# External Changes

n/a

# Additional information

Was:
image

Changed to:
image

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
@@ -15,6 +15,7 @@ import { EDIT_TOOLBAR_HEIGHT } from '../constants'
 
 import { Items } from './Items'
 import { Menu } from './Menu'
+import { Tooltip } from '@mui/material'
 
 export function Toolbar(): ReactElement {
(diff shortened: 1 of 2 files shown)


apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
```diff
@@ -42,9 +43,11 @@ export function Toolbar(): ReactElement {
         }}
       />
       <NextLink href="/" passHref legacyBehavior>
-        <IconButton data-testid="ToolbarBackButton">
-          <ChevronLeftIcon />
-        </IconButton>
+        <Tooltip title="See all journeys" placement="right" arrow>
```
GeronimoJohn, line 46: Can we add a test for this please

scottkelso: Just added :)

GeronimoJohn (approved):

PR #2720, opened 2024-05-08
feat: close icon on settings panel

jianwei1:
# Description

### Issue

Add a close icon onto the settings panel, which takes the user back to the journey map

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062706#__recording_7375869126)

### Solution

Added an onClose method to the settings drawers, which creates the close icon, and sets the activeSlide to JourneyFlow when clicked

# Additional information

Also refactored the properties block stories to include the close icon

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'next-i18next'
 import { ReactElement } from 'react'
 
 import { TreeBlock } from '@core/journeys/ui/block'
-import { useEditor } from '@core/journeys/ui/EditorProvider'
+import { ActiveSlide, useEditor } from '@core/journeys/ui/EditorProvider'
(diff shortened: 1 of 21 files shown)


GeronimoJohn (approved):

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.tsx
```diff
@@ -33,8 +34,15 @@ export function AddBlock(): ReactElement {
...
+  function onClose(): void {
+    dispatch({
+      type: 'SetActiveSlideAction',
+      activeSlide: ActiveSlide.JourneyFlow
+    })
+  }
+
   return (
-    <Drawer title={t('Add a block')}>
+    <Drawer title={t('Add a block')} onClose={onClose}>
```
scottkelso, line 45: Just a thought for the future but `onClose={onClose}` isn't that useful - it doesn't tell an engineer what is suppose to happen when the drawer closes.

Maybe a better name that describes what the function does might be `goToJourneyFlow` resulting in something like `onClose={goToJourneyFlow}` which is infinitely more useful.

PR #2730, opened 2024-05-09
fix: drop navigate action

GeronimoJohn:
# Description

**DEPENDS ON #2727**

### Issue

We're dropping the use of navigate actions

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7364834584)

### Solution

Drop navigate actions from the backend


tataihono (dismissed):

GeronimoJohn:
After chatting with Tatai, converting PR to no longer add deprecation warning. As we're dropping the use of the fields, inputs, and mutations on the client side. Though means that this will now depend on the PR: https://github.com/JesusFilm/core/pull/2727

tataihono (changes requested):
update nua seeds

apps/api-journeys/src/app/modules/action/emailAction/emailAction.resolver.ts
```diff
@@ -13,6 +12,7 @@ import { AppAbility, Action as CaslAction } from '../../../lib/casl/caslFactory'
 import { AppCaslGuard } from '../../../lib/casl/caslGuard'
 import { PrismaService } from '../../../lib/prisma.service'
 import { ACTION_UPDATE_RESET } from '../actionUpdateReset'
+import { checkBlockSupport } from '../checkBlockSupport'
```
tataihono, line 15: change name to canBlockHaveAction

tataihono (changes requested):

tataihono (approved):

PR #2734, opened 2024-05-10
fix: action same block id error message

GeronimoJohn:
# Description

### Issue

_Currently there are no backend restrictions for what navigateToBlockAction and nextBlockId can navigate to, but these should not be allowed to reference itself_

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7377801409)

### Solution

_We need to prevent users from being able to do so, through adding a restriction that if the user input id is referring to the current block we throw a message_


apps/api-journeys/src/app/modules/action/navigateToBlockAction/navigateToBlockAction.resolver.ts
```diff
@@ -67,6 +67,10 @@ export class NavigateToBlockActionResolver {
         { extensions: { code: 'BAD_USER_INPUT' } }
       )
     }
+    if (block.id === input.blockId)
```
csiyang, line 70: This needs check needs to be changed to check the parent `StepBlock`, not the block the action is on.

csiyang (changes requested):

csiyang (approved):

PR #2735, opened 2024-05-10
fix: canvas size on journeymap

jianwei1:
# Description

### Issue

Issue #1 on the below basecamp ticket. When the user shrinks the screen vertically, the canvas width and height will reduce. However, if you then bring the screen back to normal size, the canvas size does not adjust back to normal. 

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7356029597)

### Solution

Removed clientWidth from the scale calculations. 

# Additional information

When shrinking the screen vertically, the top of the card(on the canvas) would go up behind the toolbar. To fix this, I added some extra margin height when the scale value has reduced below .65.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/utils/calculateDimensions/calculateDimensions.ts
+++ b/apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/utils/calculateDimensions/calculateDimensions.ts
@@ -12,8 +12,7 @@ export function calculateScale(ref: RefObject<HTMLDivElement>): number {
   if (current.clientWidth === 0) return 1
 
   const clientHeight = current.clientHeight / CARD_HEIGHT
-  const scale = clientHeight
-  return Math.min(scale, 1)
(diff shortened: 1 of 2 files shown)


apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/utils/calculateDimensions/calculateDimensions.ts
```diff
@@ -11,10 +11,8 @@ export function calculateScale(ref: RefObject<HTMLDivElement>): number {
   // which is done intentionally for the swiper
   if (current.clientWidth === 0) return 1
 
-  const clientWidth = current.clientWidth / CARD_WIDTH
   const clientHeight = current.clientHeight / CARD_HEIGHT
-  const scale = Math.min(clientWidth, clientHeight)
-
+  const scale = clientHeight
```
GeronimoJohn, line 15: can we just pass clientHeight in the return

GeronimoJohn (approved):

PR #2738, opened 2024-05-10
feat: delete nodes and edges on keypress

levi-zustiak:
# Description

### Issue
Allow users to delete nodes and edges when pressing backspace or delete

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7364539792)

### Solution
- created a custom hook `useDeleteOnKeyPress` to perform various logic to delete a selected node or edge when pressing the backspace or delete key

# External Changes
N/A

# Additional information
- The original suggestion was to use reactflow's `onEdgesDelete` and `onNodesDelete` functions, but since some edges or nodes should not be deletable I had to use the `useKeyPress` helper as an override.


apps/journeys-admin/src/libs/useBlockDeleteMutation/useBlockDeleteMutation.ts
GeronimoJohn, line 1: Test for this one

GeronimoJohn (approved):

PR #2739, opened 2024-05-10
feat: languages bq import support part 1

mikeallisonJS:
# Description

### Issue

Name field is in a Json Blob array

### Solution

Incrementally move name to a table relation

# External Changes

Adds new model & relations, as well as seed for new model

# Additional information

Once complete, this should allow us to import languages from BQ.
Additionally it should make for better caching down the line and reduce cpu usage of filtering json. Especially once we if/when we add WESS data

### DIFF
--- a/apps/api-languages/db/schema.prisma
+++ b/apps/api-languages/db/schema.prisma
@@ -19,10 +19,30 @@ datasource db {
 }
 
 model Language {
-  id        String   @id
-  createdAt DateTime @default(now())
-  updatedAt DateTime @updatedAt
-  name      Json[]
-  bcp47     String?
-  iso3      String?
+  id           String         @id
+  createdAt    DateTime       @default(now())
+  updatedAt    DateTime       @updatedAt
+  name         Json[]
+  bcp47        String?
+  iso3         String?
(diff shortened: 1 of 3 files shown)


apps/api-languages/db/schema.prisma
```diff
@@ -19,10 +19,30 @@ datasource db {
...
-  bcp47     String?
-  iso3      String?
+  id           String         @id
+  createdAt    DateTime       @default(now())
+  updatedAt    DateTime       @updatedAt
+  name         Json[]
+  bcp47        String?
+  iso3         String?
+  languageName LanguageName[] @relation(name: "name") // TODO: rename to name after migration
+  nameLanguage LanguageName[] @relation(name: "nameLanguage")
```
tataihono, line 29: Not sure what this relation is for?

tataihono (changes requested):

apps/api-languages/db/schema.prisma
mikeallisonJS: This is the language relationship for languageId of name. It enforces a foreign key relationship on the data.

tataihono (approved):

PR #2750, opened 2024-05-14
fix: new node active slide

GeronimoJohn:
# Description

**FOR TESTING PURPOSES ONLY**

### Issue

Need to give designers a feel of what would it look like if after creating a new step it doesn't slide to canvas

### Solution

Remove the dispatch that triggers the canvas to be the active slide on New Step button click


GeronimoJohn:
Was only used for demo purposes, closing pr

PR #2756, opened 2024-05-15
feat: back to map admin changes

jianwei1:
# Description

### Issue

Frontend changes for the back to map button ticket

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7292062740)

### Solution


tataihono (changes requested):
looks good. you'll need to run extract-translations on journeys-admin.

tataihono (changes requested):
looks good. you'll need to run extract-translations on journeys-admin.

tataihono (approved):

PR #2758, opened 2024-05-15
fix: add icons after actions

GeronimoJohn:
# Description

### Issue

Ultimately the goal is to add a node that lets our users know that a button with a set action goes either towards an email or url. Currently we don't have the time do that.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7380344215)

### Solution

For now we're adding an icon after the title of the action

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/ActionButton/ActionButton.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/ActionButton/ActionButton.tsx
@@ -1,4 +1,5 @@
 import Box from '@mui/material/Box'
+import Tooltip from '@mui/material/Tooltip'
 import Typography from '@mui/material/Typography'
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/ActionButton/ActionButton.tsx
```diff
@@ -22,46 +26,71 @@ export function ActionButton({
...
+              sx={{ fontSize: 10, position: 'absolute', left: -20, top: 8 }}
+            />
+          </Tooltip>
+        )
+      default:
+        return <></>
+    }
+  }
+
+  function getSourceConnection(block): boolean {
```
edmonday: rename please, naming could be more clear about what it's doing

```suggestion
  function hasConnection(block): boolean {
```

edmonday (approved):

PR #2774, opened 2024-05-17
fix: card templates close icon

jianwei1:
# Description

### Issue

Pressing the close icon on the card templates drawer should open card properties, instead of re-directing the user to the journey map

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7396819928)

### Solution

Added a "showTemplates" state in Properties.tsx and updated the onClose function


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.spec.tsx
```diff
@@ -355,4 +366,63 @@ describe('Properties', () => {
...
+    )
+
+    await waitFor(() =>
+      expect(screen.getByTestId('CardTemplates')).toBeInTheDocument()
+    )
+
+    expect(screen.getByText('activeSlide: 1')).toBeInTheDocument()
+    await waitFor(() =>
+      expect(screen.getByTestId('X2Icon')).toBeInTheDocument()
+    )
```
GeronimoJohn, line 418: since we have a test that checks if we can click on this, we can drop this part

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.spec.tsx
```diff
@@ -355,4 +366,63 @@ describe('Properties', () => {
...
+    await waitFor(
+      async () => await userEvent.click(screen.getByTestId('X2Icon'))
+    )
```
GeronimoJohn, line 421: I can see why this was done, but just wondering if the test will still pass if we don't do the double await

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.tsx
```diff
@@ -97,16 +97,10 @@ interface PropertiesProps {
 export function Properties({ block, step }: PropertiesProps): ReactElement {
   const { t } = useTranslation('apps-journeys-admin')
   const { state, dispatch } = useEditor()
+  const [showTemplates, setshowTemplates] = useState(true)
```
GeronimoJohn, line 100: `const [showCardTemplates, setShowCardTemplates] = useState(true)`

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.tsx
```diff
@@ -166,6 +160,19 @@ export function Properties({ block, step }: PropertiesProps): ReactElement {
...
+    if (title !== 'Card Templates') {
+      dispatch({
+        type: 'SetActiveSlideAction',
+        activeSlide: ActiveSlide.JourneyFlow
+      })
+      setshowTemplates(true)
+    } else {
+      setshowTemplates(false)
+    }
+  }
```
GeronimoJohn, line 174: can further simplify

`function onClose(): void {
  const isCardTemplates = title === 'Card Templates'
  setShowTemplates(!isCardTemplates)

  if (!isCardTemplates) 
    dispatch({
      type: 'SetActiveSlideAction',
      activeSlide: ActiveSlide.JourneyFlow
    })
}`

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.spec.tsx
jianwei1: Throws errors if you try to remove double await

GeronimoJohn (approved):

PR #2786, opened 2024-05-20
refactor: pagination bullets branching

Kneesal:
# Description

### Issue

pagination bullets are do not reflect the state of the journey since the introduction of branching

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7409977847)

### Solution

pagination bullets are now agnostic of the card/parent order of a journey. 
Start card will always show pagination start, 
middle cards will animate pagination as per normal. 
If you are on a card that has no next-step - it will show end state of pagination bullets


libs/journeys/ui/src/components/StepHeader/PaginationBullets/PaginationBullets.tsx
```diff
@@ -70,3 +97,23 @@ export function PaginationBullets(): ReactElement {
...
   )
 }
+
+function removeAlreadyVisitedBlocksFromHistory(
+  blockHistory: TreeBlock[],
+  currentActiveBlock: TreeBlock<StepFields>
+): void {
+  // paginate to beginning if you are back at the starting point
+  if (currentActiveBlock?.id === blockHistory[0]?.id) {
+    blockHistory.splice(1, blockHistory.length - 1)
```
csiyang, line 107: Can we use [lodash drop right](https://lodash.com/docs/4.17.15#dropRight) instead of the splices. I think it would make it a bit easier to understand

libs/journeys/ui/src/components/StepHeader/PaginationBullets/PaginationBullets.tsx
```diff
@@ -70,3 +97,23 @@ export function PaginationBullets(): ReactElement {
...
+  // paginate back if you have already visited card
+  let prevCards: TreeBlock[] = []
+  prevCards = blockHistory.filter((block) => block.id === currentActiveBlock.id)
+  if (prevCards.length > 1) {
+    blockHistory.forEach((block, i) => {
+      if (block.id === currentActiveBlock.id) {
+        blockHistory.splice(i + 1, blockHistory.length - i)
+      }
+    })
+  }
```
csiyang, line 118: I think creating this array is redundant. You will always need to loop through all of blockHistory, except you need to do it twice if `prevCards.length > 0`

```suggestion
    blockHistory.forEach((block, i) => {
      if (block.id === currentActiveBlock.id) {
        blockHistory.splice(i + 1, blockHistory.length - i)
      }
    })
```

csiyang (changes requested):

libs/journeys/ui/src/components/StepHeader/PaginationBullets/PaginationBullets.tsx
Kneesal: lodash doesn't mutate the original array hence why I opted for using splice instead

csiyang (approved):

PR #2801, opened 2024-05-23
fix: slow delete block

scottkelso:
# Description

### Issue

The delete step dialog and quick control menu doesn't disable on database call.
[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7409941009)

### Solution

Add loading to the dialog delete and quick control menu

# External Changes

n/a

# Additional information

n/a

# How to test
Delete a step block in the journey flow. When presented with the are you sure dialog, try and mash the delete button multiple times. You should not be able to because the button should disable.

Delete a block in the canvas editor. When presented with the quick controls menu, try and mash the delete button multiple times. You should not be able to because the button should disable.


apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/QuickControls/DeleteBlock/DeleteBlock.tsx
```diff
@@ -46,8 +47,7 @@ export function DeleteBlock({
     setOpenDialog(false)
     closeMenu?.()
   }
-
-  const disableAction = currentBlock == null || disabled
+  const disableAction = currentBlock == null || disabled || result?.loading
```
csiyang, line 50: Is result nullable? I would be expecting this to throw a lint error if it was. Since its not, it looks like its not nullable.
```suggestion
  const disableAction = currentBlock == null || disabled || result.loading
```

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/QuickControls/DeleteBlock/DeleteBlock.tsx
```diff
@@ -103,6 +103,7 @@ export function DeleteBlock({
           submitLabel: t('Delete'),
           closeLabel: t('Cancel')
         }}
+        loading={disableAction}
```
csiyang, line 106:
```suggestion
        loading={result.loading}
```

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/QuickControls/DeleteBlock/DeleteBlock.tsx
scottkelso: Do we not want to disable the button if the "disabled" prop is true as well as if loading?

csiyang (approved):

PR #2803, opened 2024-05-23
test: useCreateStep

levi-zustiak:
# Description

### Issue
Needed tests for `useCreateStep`

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37071143/todos/7390523100)

### Solution
Added tests for `useCreateStep`

# External Changes
N/A

# Additional information
N/A

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useCreateStep/useCreateStep.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useCreateStep/useCreateStep.spec.tsx
@@ -114,9 +114,9 @@ const stepBlockNextBlockUpdateMock: MockedResponse<
 describe('useCreateStep', () => {
   it('should create a new step block', async () => {
     const cache = new InMemoryCache().restore({
-      'Journey:journey.id': {
+      'Journey:journey-id': {
         blocks: [{ __ref: `StepBlock:step.id` }],
-        id: 'journey.id',
+        id: 'journey-id',
         __typename: 'Journey'
       },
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useCreateStep/useCreateStep.spec.tsx
```diff
@@ -0,0 +1,273 @@
...
+  data: {
+    blockOrderUpdate: [
+      {
+        __typename: 'StepBlock',
+        id: 'blockId',
+        parentOrder: 0
+      }
+    ]
+  }
+}))
```
levi-zustiak, line 53: Do you think there is a better way to do this? I feel like this is cumbersome

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useCreateStep/useCreateStep.spec.tsx
```diff
@@ -0,0 +1,273 @@
...
+    await act(
+      async () =>
+        await hook.current({
+          x: -200,
+          y: 38
+        })
+    )
+
+    expect(result).toHaveBeenCalled()
+
```
csiyang, line 154: Can you add the cache check for blocks
```suggestion
expect(cache.extract()['Journey:journey-id']).toEqual({
    // step1 ref
    // newStep ref
    //newCard ref
})
```

csiyang (changes requested):

csiyang (approved):

PR #2817, opened 2024-05-30
fix: migrate text block actions

Kneesal:
# Description

### Issue

We are about to drop actions for text response blocks but there are existing text responses with actions in legacy journeys

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663186/todos/7442527855)

### Solution

migration script that will create a button in the journey that moves the text response action to a button, placed in the relevant card


mikeallisonJS (approved):

Kneesal:
This script has run in prod - closing pr

PR #2819, opened 2024-05-30
fix: embed disable fullscreen

tataihono:
# Description

### Issue

Some journey creators want to be able to embed without going fullscreen on click since sometimes that can be quite distracting. 

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663057/todos/7453044296)

### Solution

This enables the journey creator to adjust their snippets by adding in ?expand=false after there iframe src url. If they do so it will no longer go fullscreen when the user tries to interact with the journey.


apps/journeys/pages/[hostname]/embed/[journeySlug].tsx
```diff
@@ -76,7 +78,10 @@ function HostJourneyEmbedPage({
       `}</style>
       <JourneyProvider value={{ journey, variant: 'embed' }}>
         <ThemeProvider {...theme} rtl={rtl} locale={locale}>
-          <EmbeddedPreview blocks={blocks} />
+          <EmbeddedPreview
+            blocks={blocks}
+            disableFullscreen={query.expand === 'false'}
+          />
```
Comment (Kneesal) [inline apps/journeys/pages/[hostname]/embed/[journeySlug].tsx:84]:
```suggestion
          <EmbeddedPreview
            blocks={blocks}
            disableFullscreen={query?.expand === 'false'}
          />
```

Kneesal (changes requested):

Kneesal (approved):

Kneesal (approved):

PR #2826, opened 2024-06-03
feat: journey notification management frontend

Kneesal:
# Description

### Issue

We're about to add a new functionality where we send emails out to our creators once a visitor or a user interacts with their journey. Though that can be quite a bit of emails and they can get a bunch of emails go through

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663186/todos/7447102636)

### Solution

Add the ability for our creators to opt in/out for these emails. We are to add a toggle/switch button by the team access section


apps/journeys-admin/src/components/AccessDialog/UserList/UserListItem/UserListItem.tsx
```diff
@@ -105,14 +107,40 @@ export function UserListItem({
...
     <>
-      <ListItem
-        sx={{
-          px: 0,
-          '& > .MuiListItemSecondaryAction-root': {
-            right: 0
-          }
-        }}
-        secondaryAction={
+      <Grid container spacing={1} alignItems="center">
```
GeronimoJohn: can we move the content of this grid into a reusable component if there's not much prop passing. UserTeamListItem has pretty much the same code

apps/journeys-admin/src/components/AccessDialog/UserList/UserListItem/UserListItem.tsx
```diff
@@ -105,14 +107,40 @@ export function UserListItem({
...
+            sx={{
+              ml: 2,
+              '& > .MuiListItemText-secondary': {
+                width: { xs: '90%', sm: '90%' },
+                whiteSpace: 'nowrap',
+                overflow: 'clip',
+                textOverflow: 'ellipsis'
+              }
+            }}
+          />
```
GeronimoJohn, line 136: Use Typography instead since we're using Grid now

apps/journeys-admin/src/components/AccessDialog/AccessDialog.tsx
```diff
@@ -78,12 +91,24 @@ export function AccessDialog({
 
   const smUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('sm'))
 
-  const currentUser = useMemo(() => {
+  const currentUserJourney = useMemo(() => {
     return data?.journey?.userJourneys?.find(
       (userJourney) => userJourney.user?.email === user.email
     )
   }, [data?.journey?.userJourneys, user])
```
GeronimoJohn, line 98: not really your work but can we just use data here

apps/journeys-admin/src/components/AccessDialog/UserList/UserListItem/UserListItem.tsx
Kneesal: The content of the grid is different depending on which component it is used in, see notification switch and button

GeronimoJohn (approved):

PR #2810, opened 2024-05-28
refactor: discovery journeys templates

scottkelso:
# Description

This is an experimental branch. Likely won't get merged in it's current state. Too many changes and too many issues to fix all in one PR.

### Issue

Components needed for the resources hub need refactored out into the shared space to be used by both journeys-admin and watch.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663178/todos/7452920587)

### Solution

Refactor out needed components

# External Changes

There should be no behavioural changes to any of the projects during this refactor

# Additional information

n/a

PR #2827, opened 2024-06-05
refactor: admin theme

scottkelso:
# Description

### Issue

This is needed to access the journeys admin theme from outside the project for certain components.

### Solution

This PR refactors the admin theme into the shared/ui space while leaving the shorthand `<ThemeProvider>` in the journeys-admin project to remain being used. Note the shared theme provider now handles the ThemeMode.journeysAdmin type.

# External Changes

Admin theme move meant that storybook needed moved to the shared/ui (which is needed for future component refactors).
Extend shared/ui/ThemeProvider to handle ThemeName.journeysAdmin
updated imports for linter

# Additional information

## How To Test
There should be no behavioural impact so ensure application still performs as expected.
One should be able to use the shared/ui Theme provider as shown inplace of the admin-journeys one.
image
The shared/ui ThemeProvider is used by journeys-admin, and watch so we would need to regression test these.
The journeys-admin ThemeProvider is used in journeys-admin so we might need to regression test this.


scottkelso:
We tried moving journeysAdminConfig into journeys/ui but then there are circular dependencies between journeys/ui and shared/ui. We can't move simpleComponentConfig into journeys/ui too because then ThemeProvider would also need to move - which it shouldn't. We decided to leave it as is in PR.

GeronimoJohn (dismissed):

GeronimoJohn (dismissed):

apps/arclight/__generated__/graphql-env.d.ts
GeronimoJohn, line 1: NIT: Thought about it more, since GH checks passes with/without this can we remove it? I want to isolate these changes since they're all pretty big PRs

GeronimoJohn (approved):

GeronimoJohn (approved):

PR #2833, opened 2024-06-05
refactor: card and video wrappers

scottkelso:
# Description

**Note: only merge after https://github.com/JesusFilm/core/pull/2827!**

### Issue

We need to refactor out CardWrapper and VideoWrapper because they are needed by TemplateCardPreview which will be refactored out in a future PR.

### Solution

This PR refactors the card wrapper and video wrapper into the journeys/ui/ space. 

# External Changes

n/a

# Additional information

## How To Test
One should be able to use the CardWrapper component in another project.
One should be able to use the VideoWrapper component in another project.


libs/journeys/ui/src/components/ResourcesHub/CardWrapper/CardWrapper.spec.tsx
```diff
@@ -1,10 +1,9 @@
 import { render } from '@testing-library/react'
```
scottkelso, line 1: Having spoken to @tataihono, we would rather stuff moved into journeys/ui/src/components/... instead of journeys/ui/src/components/ResourcesHub/...

GeronimoJohn (approved):
LGTM

PR #2837, opened 2024-06-06
fix: backwards edge editing not removing old edge

DanChae03:
# Description

### Issue

Modifying the left side (source) of the connection without changing the target used to duplicat the connection - create a new one and not delete the old one. This has since been fixed in this PR.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663057/todos/7439632898)

### Solution

# External Changes

# Additional information


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useUpdateEdge/useUpdateEdge.ts
```diff
@@ -32,6 +38,15 @@ export function useUpdateEdge(): (
     let selectedStep: TreeBlock<StepBlock> | undefined
     const edgeSource = convertToEdgeSource(rawEdgeSource)
 
+    if (
```
csiyang, line 41: add test

csiyang (changes requested):

csiyang (approved):

PR #2842, opened 2024-06-06
refactor: language queries

scottkelso:
# Description

**Note: only merge after https://github.com/JesusFilm/core/pull/2827!**

### Issue

We need to refactor out getLanguages query because it is needed by HeaderAndLanguages which will be refactored out in a future PR.

### Solution

This PR refactors the getLanguages query into the journeys/ui/ space. 

# External Changes

n/a

# Additional information

## How To Test
There should be no behavioural impact so ensure application still performs as expected. These components are currently being used in arclight, journeys-admin, journeys, nexus-admin, & watch so regression test as appropriate.
One should be able to use the getLanguages query in another project.


scottkelso:
Renamed the GetLanguages query in arclight to GetLanguagesWithTags to sort duplication error. There doesn't seem to be an nx run arclight:codegen command but I have picked up a generated file change somewhere on my journeys. Wondering where this belongs to @tataihono ?

image
Currently commited here https://github.com/JesusFilm/core/pull/2842/files#diff-b9da916a00dfa7a4409c60ead2bf00f3e88b9d0a33f7a484cc9c7fd8b4560ace

GeronimoJohn (approved):

PR #2853, opened 2024-06-09
feat: watch multi page

tataihono:
# Description

### Issue

Watch only runs on /watch. But we need it to also allow /strategies etc.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663178/todos/7442353275)

### Solution

Remove basepath. Move all existing assets to /watch.

# External Changes

Cloudflare Proxy Script will need to be updated to allow _next


scottkelso:
Tested locally and all network calls seems to be going to https://www.jesusfilm.org/watch/_next....

scottkelso (approved):
LGTM! 👍

PR #2874, opened 2024-06-12
feat: pnpm

mikeallisonJS:
# Description

### Issue

Npm is slow, deployments are failing

### Solution

Replace npm with pnpm

# External Changes

# Additional information

all npm commands will work with pnpm, and npm commands should redirect you to use pnpm


mikeallisonJS:
looks like bun is going to win this one

PR #2884, opened 2024-06-14
refactor: i18next

scottkelso:
# Description

### Issue

We need to refactor out i18next query because it is needed by many stories and spec files which will be refactored out in a future PR.

### Solution

This PR refactors the i18next function into the journeys/ui/ space. 

# External Changes

n/a

# Additional information

## How To Test
One should be able to use the i18n function in another project.

### DIFF
--- a/apps/journeys-admin/setupTests.tsx
+++ b/apps/journeys-admin/setupTests.tsx
@@ -6,7 +6,7 @@ import crypto from 'crypto'
 import { configure } from '@testing-library/react'
 
 import { mswServer } from './test/mswServer'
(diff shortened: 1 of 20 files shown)


GeronimoJohn (approved):

scottkelso:
I have had a change of heart about how this should have been refactored out. I actually think this `i18n.ts` should stay in the journeys-admin project, continue to use the `apps-journeys-admin.json`. This PR should instead and a specific one for journeys-ui which instead uses `libs-journeys-ui.json`.

PR #2886, opened 2024-06-14
fix: analytics switch

csiyang:
# Description

### Issue

- Nodes and edges could still be deleted on `Backspace` press in analytics mode
- Edge selection styles was still showing up in analytics mode

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663055/todos/7502469950)

### Solution

- Add more checks based if analytics mode is shown
- Add tests

# External Changes

n/a

# Additional information

n/a


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/StepBlockNodeCard/StepBlockNodeCard.tsx
```diff
@@ -43,7 +43,7 @@ export function StepBlockNodeCard({
   } = getCardMetadata(card)
 
   function handleClick(): void {
-    if (selectedStep?.id === step?.id) {
+    if (selectedStep?.id === step?.id && !showJourneyFlowAnalytics) {
```
levi-zustiak, line 46: Looks like you moved the conditional checks in the editor provider actions to the component. That's fine for this one, but you'll also have to add this for the action when clicking the on the canvas card footer as well. That's why I had added the check on the action instead.

levi-zustiak (changes requested):
One comment but otherwise looks good

levi-zustiak (approved):

PR #2900, opened 2024-06-18
refactor: i18next

scottkelso:
# Description

### Issue

https://github.com/JesusFilm/core/pull/2884 was incorrectly refactored. It assumes that all projects consume from the same local language json which is not the case.

### Solution

Revert the journeys-admin changes to go back to using its own `i18n.ts` which uses the `apps-journeys-admin.json`.
Add a new `i18n.ts` into the journeys/ui space which instead uses `libs-journeys-ui.json`

# External Changes

n/a

# Additional information

### How To Test
There should be no behavioural impact so ensure application still performs as expected. This component is currently only being used in journeys-admin so regression test as appropriate.
One should have access to the i18n function in journeys/ui.


GeronimoJohn (approved):

GeronimoJohn:
Reminder: generate translations from Crowdin to populate libs-journeys-ui with the new translations

PR #2906, opened 2024-06-18
fix: filter visitor details page

Kneesal:
# Description

### Issue

duplicate text event details need to be omitted 

### Solution

filtered duplicate text events by block id, only return user's most recent submission


GeronimoJohn (dismissed):
lgtm

PR #2907, opened 2024-06-19
refactor: journeys templates view

scottkelso:
# Description

**Note: only merge after  https://github.com/JesusFilm/core/pull/2912!**

### Issue

We need to refactor out TemplateView components and the related data.ts files that are needed to create journeys pages in the resources hub section of the watch project.

### Solution

This PR refactors the TemplateView components into the journeys/ui/ space. 

# External Changes

n/a

# Additional information

## How To Test
One should be able to use the TemplateView and child components in another project.


GeronimoJohn (commented):
components that got move to journeys/ui - please update all the useTranslations to be using `libs-journeys-ui`

GeronimoJohn (approved):

PR #2911, opened 2024-06-19
fix: remove header divider on watch

jianwei1:
# Description

### Issue

Header divider line was showing up on watch page, which meant there was a line going through the background content

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/37663178/todos/7439735805)

### Solution

Added a check to only show divider on light theme


apps/watch/src/components/Header/Header.tsx
```diff
@@ -141,7 +143,7 @@ export function Header({
               background: 'transparent',
               boxShadow: 'none'
             }}
-            showDivider
+            showDivider={lightTheme}
```
tataihono, line 146:
```suggestion
            showDivider={themeMode === ThemeMode.light}
```

tataihono (changes requested):

tataihono (approved):

PR #2982, opened 2024-07-02
fix: plausible stats collection

csiyang:
# Description

### Issue

Stats have been consistently incorrect with the visitors 

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7585668435)

### Solution

- All events also sends the page they happened on
- Fix page path for plausible events
- Added PlausibleProvider to all [journeySlug.tsx] pages, and made sure they now all have the same props
- All footerChatButton events will be counted as a chat event (regardless of link)

# External Changes

n/a

# Additional information

n/a


apps/journeys/pages/[hostname]/embed/[journeySlug].tsx
```diff
@@ -46,7 +47,17 @@ function HostJourneyEmbedPage({
       ? getStepTheme(blocks[0] as TreeBlock<StepFields>, journey)
       : { themeName: journey.themeName, themeMode: journey.themeMode }
   return (
-    <>
+    <PlausibleProvider
```
Comment (tataihono) [inline apps/journeys/pages/[hostname]/embed/[journeySlug].tsx:50]: Can we move this into a JourneyPlausibleProvider component that takes a journeyId & a teamId prop. Seems like this is duplicated in a few places?

apps/journeys/src/components/Conductor/NavigationButton/NavigationButton.spec.tsx
```diff
@@ -79,6 +79,13 @@ const journey = {
 describe('NavigationButton', () => {
   mockUuidv4.mockReturnValue('uuid')
 
+  const mockOrigin = 'https://example.com'
+  Object.defineProperty(window, 'location', {
+    value: {
+      origin: mockOrigin
+    }
+  })
```
tataihono, line 87: this needs to reset the window object after messing with it for this test.

apps/journeys/src/components/Conductor/SwipeNavigation/SwipeNavigation.spec.tsx
```diff
@@ -56,6 +56,13 @@ describe('SwipeNavigation', () => {
   const swipeLeft = -100
   const swipeRight = 100
 
+  const mockOrigin = 'https://example.com'
+  Object.defineProperty(window, 'location', {
+    value: {
+      origin: mockOrigin
+    }
+  })
```
tataihono, line 64: same here

libs/journeys/ui/src/components/Button/Button.spec.tsx
```diff
@@ -129,6 +129,13 @@ const journey = {
 } as unknown as Journey
 
 describe('Button', () => {
+  const mockOrigin = 'https://example.com'
+  Object.defineProperty(window, 'location', {
+    value: {
+      origin: mockOrigin
+    }
+  })
```
tataihono, line 137: and here

libs/journeys/ui/src/components/Card/Card.spec.tsx
```diff
@@ -69,6 +69,13 @@ describe('CardBlock', () => {
   const leftSide = { clientX: 0 }
   const rightSide = { clientX: 1000 }
 
+  const mockOrigin = 'https://example.com'
+  Object.defineProperty(window, 'location', {
+    value: {
+      origin: mockOrigin
+    }
+  })
```
tataihono, line 77: and here. and every other place where this is done.

libs/journeys/ui/src/components/VideoTrigger/VideoTrigger.tsx
```diff
@@ -43,36 +45,39 @@ export function VideoTrigger({
           player.pause()
 
           const input = {
```
tataihono, line 47: can this arg be typed?

tataihono (changes requested):

tataihono (approved):

PR #3025, opened 2024-07-10
feat: search bar ui

scottkelso:
# Description

### Issue

We need a basic text input with the correct styling from which to work off for adding search functionality

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todos/7581208971)

### Solution

Add a text field. The styling here is unusual because the `border-radius` and `background-image-source` aren't compatible. [This is a known problem](https://github.com/w3c/csswg-drafts/issues/7550). I was able to create the same affect by adding a box behind the input that has the background set as the linear gradient. 

# External Changes

n/a

# Additional information

Story added called `Watch/SearchBar`


apps/watch/src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,12 +1,49 @@
...
+    }
+  }))
+
   return (
-    <form>
-      <input
-        type="text"
+    <Box
+      sx={{
+        borderRadius: '12px',
```
GeronimoJohn, line 29:
```suggestion
        borderRadius: 4,
```

apps/watch/src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,12 +1,49 @@
...
+        background:
+          'linear-gradient(90deg, #0C79B3 0%, #0FDABC 51%, #E72DBB 100%)',
+        p: '4px'
```
GeronimoJohn, line 32:
```suggestion
        p: 1
```

apps/watch/src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,12 +1,49 @@
...
+      }}
+      data-testid="SearchBar"
+    >
+      <ColoredTextField
         placeholder="Search by topic, occasion, or audience ..."
```
GeronimoJohn, line 37: Can we please wrap this in the `t` function from `useTranslation`

apps/watch/src/components/SearchBar/SearchBar.tsx
```diff
@@ -1,12 +1,49 @@
...
+      },
+      fieldset: {
+        border: 'none'
+      },
+      input: {
+        // Overriding the default set in components.tsx
+        transform: 'none'
+      }
+    }
+  }))
```
GeronimoJohn, line 24: Can we please define this outside SearchBar and change the name of it to be StyledTextField. Just the pattern thats around in Core

GeronimoJohn (commented):
Just a few, and add an index file as well

GeronimoJohn (approved):

PR #3026, opened 2024-07-10
fix: search bots crawl

csiyang:
# Description

### Issue

Search bots (including google) can not see card content while crawling journeys.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7586079750)

### Solution

- values used from `useBlocks` contributed to this issue, refactored to use `journey.blocks`
- Mui Fade needed the child components to forward its ref ([docs](https://v5-0-6.mui.com/guides/migration-v4/#cannot-read-property-scrolltop-of-null))
- setting video background color causing video to render with the incorrect layout on first initialization

# External Changes

n/a

# Additional information

n/a


apps/journeys/src/components/Conductor/DynamicCardList/DynamicCardList.tsx
```diff
@@ -31,55 +31,64 @@ export function DynamicCardList(): ReactElement {
...
+      <Box
+        ref={ref}
+        className={isCurrent ? 'active-card' : undefined}
+        onClick={() => setShowNavigation(true)}
+        sx={{
+          width: 'inherit',
+          position: 'relative',
+          height: '100%',
+          display: isCurrent ? 'block' : 'none'
+        }}
```
levi-zustiak, line 77: Are the other conditional styles not needed anymore?

levi-zustiak (changes requested):

apps/journeys/src/components/Conductor/DynamicCardList/DynamicCardList.tsx
csiyang: This was being used to keep the pre rendered cards in the dom but not display them. For some reason why I first worked on it , using`display: 'none'` the card blocks would not be in the DOM, but now it appears correctly.

levi-zustiak (approved):

PR #3057, opened 2024-07-16
feat: sync tags to wordpress

GeronimoJohn:
# Description

### Issue

Move our tags to wordpress so that resources or posts within wordpress can have tags


GeronimoJohn:
We're no longer bringing in tags to wordpress during this cycle. Please re-open when needed

PR #3067, opened 2024-07-17
feat: properties editor state function

Kneesal:
# Description

### Issue

a function used to set the state of the editor after an undo/redo/execute for things in the properties tab


apps/journeys-admin/src/libs/setPropertiesEditorState/setPropertiesEditorState.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    parentBlockId: null,
+    parentOrder: 0,
+    locked: false,
+    nextBlockId: null,
+    children: []
+  }
+  const dispatch = jest.fn()
+
+  beforeEach(() => jest.clearAllMocks())
+  it('should set edior state', () => {
```
edmonday, line 19: NIT:
```suggestion
  it('should set editor state', () => {
```

edmonday: @Kneesal sup, can you see this comment

edmonday (changes requested):

edmonday (approved):

PR #3079, opened 2024-07-18
fix: extend languages in algolia videos module

GeronimoJohn:
# Description

Extend Languages in algolia videos module

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todos/7619095252)


scottkelso (approved):

GeronimoJohn:
after some consideration we're no longer extending the object that we send to algolia as it currently doesn't support words in other languages

PR #3084, opened 2024-07-18
feat: use mutation link

tataihono:
# Description

### Issue

Undos and redos can be undone and redone in any order! Chaos! This should be fixed.

### Solution

use mutationQueueLink from @adobe/apollo-link-mutation-queue to execute mutations in the order they were received.


Kneesal (approved):
lgtm

PR #3090, opened 2024-07-19
fix: prevent same value submission

GeronimoJohn:
# Description

Prevent Text Response from submitting the same value


GeronimoJohn:
Had to create a state to check if the values the same or not, values.response and e.target.value updates at the same time so values always the same

libs/journeys/ui/src/components/TextResponse/TextResponse.tsx
```diff
@@ -126,7 +128,11 @@ export const TextResponse = ({
                 onChange={handleChange}
                 onBlurCapture={(e) => {
                   handleBlur(e)
-                  void onSubmitHandler(values)
+                  const currentValue = (e.target as HTMLInputElement).value
```
Kneesal, line 131:
```suggestion
                  const currentValue = values.response
```

Kneesal (approved):

PR #3103, opened 2024-07-22
feat: language select

scottkelso:
# Description

### Issue

We need a language button on the searchbar

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todos/7581215916)

### Solution

n/a

# External Changes

n/a

# Additional information

See searchbar tests for how to test


libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -62,6 +66,41 @@ export function SearchBar(props: SearchBoxConnectorParams): ReactElement {
...
+                <Globe1Icon />
+                <Typography
+                  sx={{
+                    px: 2
+                  }}
+                >
+                  English
+                </Typography>
+                <ChevronDown />
+              </Box>
```
GeronimoJohn, line 102: What happens if we use a button instead of a box? Is the design too off? Also thoughts on creating a subfolder within SearchBar called LanguageButton if this parts gets too big? it'll later open the dialog or something

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -62,6 +66,41 @@ export function SearchBar(props: SearchBoxConnectorParams): ReactElement {
             <InputAdornment position="start">
               <Search1Icon />
             </InputAdornment>
+          ),
+          endAdornment: (
+            <InputAdornment position="end">
+              <Box
```
GeronimoJohn, line 72: can we add the prop `component="button"` to the box

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
GeronimoJohn: Makes sense

libs/journeys/ui/src/components/SearchBar/SearchBar.spec.tsx
```diff
@@ -47,4 +47,21 @@ describe('SearchBar', () => {
...
+  it('should render english language by default', async () => {
+    render(<SearchBar />)
+    expect(screen.getByText('English')).toBeInTheDocument()
+  })
+
+  it('should be clickable', async () => {
+    render(<SearchBar />)
+    const languageButton = screen.getByTestId('LanguageSelect')
+    fireEvent.click(languageButton)
+  })
```
GeronimoJohn, line 66: NIT: clicking doesn't really do anything atm, so thinking that we just leave a comment after click on what it's meant to do, probs good that the description reflects that too

scottkelso:
Replaced by https://github.com/JesusFilm/core/pull/3284

PR #3114, opened 2024-07-23
fix: team refetch

csiyang:
# Description

### Issue

When logging in, active team shows "Shared with me" even though there is a lastActiveTeamId

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7585966878)

### Solution

There is a check on the page which calls refetch for the`GetLastActiveTeamIdAndTeams` query. In the `onCompleted`, the activeTeam is set. But when refeth is called, the onCompleted is not rerun, so the correct team id was not being set.

# External Changes

n/a

# Additional information

Check https://github.com/apollographql/apollo-client/issues/11151 for more details


libs/journeys/ui/src/components/TeamProvider/TeamProvider.spec.tsx
```diff
@@ -150,4 +159,27 @@ describe('TeamProvider', () => {
...
   })
+
+  it('should refetch', async () => {
+    const result = jest.fn(() => ({ ...getTeamsMock.result }))
+    const refetchMock = {
+      ...getTeamsMock,
+      result
+    }
+
+    const { getByText, getByRole } = render(
```
levi-zustiak, line 170: Could you update these to use `screen` instead?

levi-zustiak (changes requested):

levi-zustiak (approved):

PR #3133, opened 2024-07-24
fix: closing addblock drawer action

DanChae03:
# Description

### Issue

Closing the Addblock drawer used to switch the screen to the journeyflow view. This has now been fixed, and the screen will not switch, and the properties drawer will become visible. 

Also, closing the properties drawer has been fixed to redirect to the journeyflow view ONLY on desktop - closing it on mobile/ipad (smaller than medium breakpoint) now redirects to the content view.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7586174105)

### Solution

# External Changes

# Additional information


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.spec.tsx
```diff
@@ -130,14 +135,13 @@ describe('AddBlock', () => {
       </MockedProvider>
     )
 
-    expect(screen.getByText('activeSlide: 1')).toBeInTheDocument()
```
levi-zustiak, line 133: Could I have you add an assertion that `activeCanvasDetailsDrawer` is set to `AddBlock` similarly to the previous version?

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.spec.tsx
```diff
@@ -130,14 +135,13 @@ describe('AddBlock', () => {
...
     await waitFor(() =>
```
levi-zustiak, line 138: There's aren't necessarily your changes, but if we can also clean some of the extra stuff that's not needed while we're here.

```suggestion
    const closeButton = screen.getByTestId('X2Icon')
    
    expect(closeButton).toBeInTheDocument()
    
    await waitFor(() => userEvent.click(closeButton))
```

levi-zustiak (changes requested):
Just a couple small comments, but looks good man! Good stuff

levi-zustiak (approved):
Looks good 👍

csiyang (approved):

PR #3144, opened 2024-07-26
feat: resource heading on journeys page

jianwei1:
# Description

### Issue

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todos/7585580522)

### Solution

# External Changes

# Additional information


libs/journeys/ui/src/components/TemplateGallery/TemplateGallery.tsx
```diff
@@ -11,6 +11,7 @@ import { TagCarousels } from './TagCarousels'
 import { ThemeProvider } from '@core/shared/ui/ThemeProvider'
 import { ThemeMode, ThemeName } from '@core/shared/ui/themes'
 import Stack from '@mui/material/Stack'
+import { ResourceHeading } from '../../../../../../apps/watch/src/components/ResourceHeading'
```
scottkelso, line 14: As far as I am aware - this is going to create a circular dependency. Probably need to move this to journeys-ui.

scottkelso (changes requested):
I think you have a circular dependancy issue. Also I'd be keen to pull down h4 to h5. let me know if you disagree. Once resolved this should be all good!

PR #3087, opened 2024-07-18
feat: journeys search bar

scottkelso:
# Description

### Issue

We want our journeys page on the resources hub to use algolia for searching.

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todos/7581223703)

### Solution

Refactor the template gallery and related components to allow us to display journeys from algolia API instead of core

# External Changes

* Add algolia hooks to related components
* Add component tests and mocks
* Comment out failing tests to do with category & language select dropdowns
* Conditionally render the algolia <Index> element so that component tests don't fail
* Add algolia data for tests
* Duplicate the watch config into shared space (needed because template components are in journeys-ui) 

# Additional information

## How to test
* As a user I must be able to see the same journey templates listed as before from NS templates page
* As a user I should be able to search by title/description to filter the journeys shown
* Grouping of template sections should match the behaviour of journeys admin
  - No search or tags = featured & new (if more than 5 journeys in a category then show category)
  - Search with no tags = featured & new (if more than 5 journeys in a category then show category)

PR #3153, opened 2024-07-30
feat: resource hub strategies page 

jianwei1:
# Description

### Issue

Create the strategies page on the resource hub, and use Aloglia. 

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213804/todolists/7585485978)

### Solution

Created all the components related to the strategy page:

- watch: 
	- StrategyCard
	- StrategySection
	- StrategySections
	- StrategiesView 

- journeys/ui:
	- EmptySearch - renders when no search results are found
	- ResourceHeading - the "___ for every interaction" text above the search bar

Used Algolia for grabbing strategies data and implementing search

### Additional Information

Refactored TemplateGalleryCarousel to [ContentCarousel](https://github.com/JesusFilm/core/pull/3034)
Uses the [searchbar ](https://github.com/JesusFilm/core/pull/3087/files#diff-1130362ad9b18ffdd8f20302bc751874ef265080d767819d8c0fba5059f5d170)created by Josh


apps/watch/pages/strategies/index.tsx
```diff
@@ -1,149 +1,56 @@
...
-        lorem. Quisque a ipsum eu quam aliquet fringilla vel vitae sem. Quisque
-        pulvinar mi et commodo dignissim. Phasellus erat ligula, faucibus quis
-        efficitur ut, sollicitudin et libero. Duis accumsan ex metus, sed cursus
-        nisl auctor non. Aliquam erat volutpat. In hac habitasse platea
-        dictumst.
-      </p>
-    </PageWrapper>
+function StrategiesPage({ serverState }: StrategiesPageProps): ReactElement {
+  return (
+    <InstantSearchSSRProvider {...serverState}>
```
scottkelso, line 30: Could you try adding some text into the input, the `query=love` or whatever your search is should appear. Then Change that `love` to `loved` in the url and submit. Does it persist? If not, might want to sort that. Could probably do in separate PR. Solution might be similar to mine where we ignore the server state here.

apps/watch/src/components/StrategiesView/StrategiesView.spec.tsx
```diff
@@ -0,0 +1,52 @@
+import { render, screen } from '@testing-library/react'
```
scottkelso, line 1: Great job on this spec file!

apps/watch/src/components/StrategySections/StrategySections.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { StrategySection } from './StrategySection/StrategySection'
+
+interface StrategySectionsProps {
+  index?: boolean
+}
+
+export function StrategySections({
+  index
+}: StrategySectionsProps): ReactElement {
+  // TODO: update this indexes variable to use the real indexes
```
scottkelso, line 14: Could you remove this comment?

apps/watch/src/components/StrategySections/StrategySections.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const indexes = ['wp_dev_posts_passionpurpose']
```
scottkelso, line 15: Is it possible to add the `wp_dev_posts_mission-trip` into this array? And reference the first one by `indexes[0]` instead? Makes it clearer that there is 2x indexes being used here instead of 1? Then loop through the rest of the indexes using `.shift()` or `.splice(1)`?

apps/watch/src/components/StrategySections/StrategySections.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const [hasResult, setHasResult] = useState<boolean>(true)
+
+  const resultsMap = new Map<number, boolean>()
+
+  function handleItemSearch(index: number, hasResult: boolean) {
+    resultsMap.set(index, hasResult)
+    const hasTrueValue = Array.from(resultsMap.values()).some(
```
scottkelso, line 23: `isAnyResults` any better?

libs/journeys/ui/src/components/SearchBar/SearchBar.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  {
+    render: (args) => (
+      <InstantSearchTestWrapper
+        query={args.query}
+        indexName="api-journeys-journeys-dev"
+      >
+        <SearchBar />
+      </InstantSearchTestWrapper>
+    )
+  }
```
scottkelso, line 26: Now that we have got MSW working, we should probably mock all algolia calls in stories, right?

scottkelso (changes requested):
Really good work! Just a few naming things to sort.

apps/watch/src/components/StrategySections/StrategySections.tsx
jianwei1: Leaving this for now as @GeronimoJohn asked me to keep it in

apps/watch/pages/strategies/index.tsx
jianwei1: Will address in separate PR: https://3.basecamp.com/3105655/buckets/38213804/todos/7672021287

scottkelso (approved):
LGTM!

PR #3167, opened 2024-07-31
refactor: helpscout beacon

csiyang:
# Description

### Issue

Beacon when on `_app.tsx` renders the beacon on every page, but positions clash with elements on certain pages

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7586079004)

### Solution

- Each page is responsible for rendering and positioning its own beacon
- Refactor`HelpscoutBeacon.tsx`:
> - split out the beacon `Script` and `Style` from the rendered button
> - added variants for colours and button style
> - render and hide the default beacon icon*
> - use default beacon close button when on mobile
- misc clean up refactors / fixes

*We were originally trying to also have a custom opening and closing button for mobile, but due to how the dialog opens, it covers everything else, sometimes leaving users no way to close the dialog. We decided to use the default close button for the dialog on mobile, but when we hid the default beacon icon, the close button for mobile mode inconsistently shows up. So as a hack, we render the default icon (while keeping our custom opening code), and hide the default icon, to get the mobile close button to work consistently.

# External Changes

n/a

# Additional information

We don't call `const user = useUser()` at the component level since there is not a good way to mock that in storybook. So each page/component that needs the beacon will need to pass the user down.


apps/journeys-admin/pages/templates/index.tsx
```diff
@@ -49,10 +52,25 @@ function TemplateIndexPage(): ReactElement {
...
+        showAppHeader={userSignedIn}
+        showNavBar={userSignedIn}
         background="background.paper"
       >
+        <Box
+          sx={{
+            position: 'absolute',
+            right: 16,
+            top: 8,
+            display: { xs: userSignedIn ? 'none' : 'block', md: 'block' }
```
levi-zustiak, line 64: Is there a reason to only hide the beacon when not signed in on mobile?

apps/journeys-admin/pages/templates/[journeyId].tsx
```diff
@@ -54,11 +60,48 @@ function TemplateDetailsPage(): ReactElement {
           user={user}
           backHref="/templates"
           mainBodyPadding={false}
-          showMainHeader={user?.id != null}
-          showAppHeader={user?.id != null}
-          showNavBar={user?.id != null}
+          showMainHeader={userSignedIn}
+          mainHeaderChildren={
+            <Stack
```
Comment (levi-zustiak) [inline apps/journeys-admin/pages/templates/[journeyId].tsx:65]: Is there a way to create a common component out of this? or to create one source of truth in case someone has to update these in the future?

Probably difficult since there's still a few variations throughout. So it's cool if you want to keep this way, just a thought

levi-zustiak (changes requested):

apps/journeys-admin/pages/templates/index.tsx
csiyang: Yeah there is a `AppHeader.tsx` that appears mobile and tablet that will take over showing the beacon. The AppHeader only shows up when a user is signed in, but the templates pages are public and don't require logging in.

apps/journeys-admin/pages/templates/[journeyId].tsx
Comment (csiyang) [reply apps/journeys-admin/pages/templates/[journeyId].tsx:65]: Yeah its kinda hard to make the Beacon common inside of the `PageWrapper.tsx`, the pagewrapper is so flexible, I think the amount of logic would make it more confusing than just adding it at the places its needed.

levi-zustiak (approved):

PR #3196, opened 2024-08-06
feat: analytics beacon

csiyang:
# Description

### Issue

Adding new analytics feature, and want to collect user feedback

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7681138004)

### Solution

- Refactor teams popover to be common
- Add popover message when analytics switch is turned on
- Add popover message when visiting plausible embed dashboard
- Add beacon hooks to manually set beacon routes and open the beacon

# External Changes

n/a

# Additional information

- dependant on https://github.com/JesusFilm/core/pull/3167


apps/journeys-admin/src/components/HelpScoutBeacon/HelpScoutBeacon.tsx
```diff
@@ -28,29 +28,43 @@ export function HelpScoutBeacon({
...
+    }
+  }
+
+  useEffect(() => {
+    handleBeaconToggle()
+  }, [loaded, setBeaconOpen])
+
+  const handleBeaconClick = (): void => {
+    if (window.Beacon != null) {
+      handleBeaconToggle()
```
levi-zustiak, line 51: Is there a reason to set the beacon event listeners multiple times?

levi-zustiak (changes requested):

apps/journeys-admin/src/components/HelpScoutBeacon/HelpScoutBeacon.tsx
csiyang: Yeah there's some weird things going on with how `window.Beacon` was getting loaded in, so the 
- useEffect handles when the beacon is opened manually through the popover
- handleClick for when the beacon is actually clicked on

The useEffect can handle the buttonClicks sometimes too, but its just a little inconsistent. If we want to do this properly, I think we would need to do it through a provider. But I don't think we'll have this popover for too long.

levi-zustiak (approved):

PR #3210, opened 2024-08-07
fix: handle null parentorder on block restore

Kneesal:
# Description

### Issue

block restore can now handle blocks that has a null parentOrder

### DIFF
--- a/apps/api-journeys/src/app/modules/block/block.resolver.ts
+++ b/apps/api-journeys/src/app/modules/block/block.resolver.ts
@@ -11,7 +11,7 @@ import { Action, AppAbility } from '../../lib/casl/caslFactory'
 import { AppCaslGuard } from '../../lib/casl/caslGuard'
 import { PrismaService } from '../../lib/prisma.service'
 
-import { BlockService } from './block.service'
+import { BlockService, BlockWithAction } from './block.service'
 
 @Resolver('Block')
 export class BlockResolver {
@@ -215,6 +215,7 @@ export class BlockResolver {
       })
 
     return await this.prismaService.$transaction(async (tx) => {
+      let res: BlockWithAction[] | undefined
       const updatedBlock = await tx.block.update({
         where: { id },
         data: {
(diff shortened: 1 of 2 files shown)


apps/api-journeys/src/app/modules/block/block.resolver.ts
```diff
@@ -215,6 +215,7 @@ export class BlockResolver {
       })
 
     return await this.prismaService.$transaction(async (tx) => {
+      let res: BlockWithAction[] | undefined
```
tataihono, line 218:
```suggestion
      let siblings = [updatedBlock]
```

apps/api-journeys/src/app/modules/block/block.resolver.ts
```diff
@@ -247,7 +245,9 @@ export class BlockResolver {
         updatedBlock.id,
         blocks
       )
-      return [...updatedBlockAndSiblings, ...children]
+
+      if (res == null) res = [updatedBlock]
+      return [...res, ...children]
```
tataihono, line 250:
```suggestion
      return [...siblings, ...children]
```

apps/api-journeys/src/app/modules/block/block.resolver.ts
```diff
@@ -224,16 +225,13 @@ export class BlockResolver {
...
-          extensions: { code: 'INTERNAL_SERVER_ERROR' }
-        })
-
-      const updatedBlockAndSiblings = await this.blockService.reorderBlock(
-        updatedBlock,
-        updatedBlock.parentOrder,
-        tx
-      )
+      if (updatedBlock?.parentOrder != null)
+        res = await this.blockService.reorderBlock(
```
tataihono, line 229:
```suggestion
        siblings = await this.blockService.reorderBlock(
```

tataihono (changes requested):

tataihono (approved):

PR #3214, opened 2024-08-08
refactor: events numbers to visitors

csiyang:
# Description

### Issue

Step events show total numbers, but users expect them to be unique to the visitor since the rest of the stats are

### Solution

show events for unique visitors rather than total events

# External Changes

n/a

# Additional information

n/a


levi-zustiak (approved):
I thin

PR #3247, opened 2024-08-12
fix: editor mobile swipe

levi-zustiak:
# Description

### Issue
Could not swipe up on the card preview for mobile devices. Caused by the logic to set the swiper `allowTouchMove` property not handling the undefined case for `showAnalytics`

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7700139780)

### Solution
Updated the logic for the swiper property to properly handle the necessary `showAnalytics` cases

# External Changes

# Additional information

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
@@ -107,7 +107,7 @@ export function Slider(): ReactElement {
 
   useEffect(() => {
     if (swiperRef.current != null) {
-      swiperRef.current.swiper.allowTouchMove = showAnalytics === false
+      swiperRef.current.swiper.allowTouchMove = showAnalytics === false || showAnalytics == null
       swiperRef.current.swiper.setGrabCursor()
     }
   }, [showAnalytics])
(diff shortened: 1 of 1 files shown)


apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
```diff
@@ -107,7 +107,8 @@ export function Slider(): ReactElement {
 
   useEffect(() => {
     if (swiperRef.current != null) {
-      swiperRef.current.swiper.allowTouchMove = showAnalytics === false
+      swiperRef.current.swiper.allowTouchMove =
+        showAnalytics === false || showAnalytics == null
```
tataihono, line 111:
```suggestion
      swiperRef.current.swiper.allowTouchMove =
        showAnalytics != true
```

tataihono (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
levi-zustiak: @tataihono The `eqeqeq` lint rule prevents this comparison, and then we have the lint rule for explicitly handling nullish cases

levi-zustiak:
@tataihono could we add `showAnalytics` to the default editor state so we don't have to handle the nullish case going forward, or is there justification for keeping it as is?

apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
tataihono: sorry it should be showAnalytics !== true.

tataihono (approved):

PR #3256, opened 2024-08-14
perf: improve image size

scottkelso:
# Description

### Issue

Warning errors in the console complaining about needlessly fetching full resolution images when not needed.

image

### Solution

When the viewport is less than md, the image will need to render 100% viewport width
When the viewport is less than xl, the image will only be max 33% viewport width
When the viewport is larger than xl, the image will only be max 20% viewport width

Should should also in turn reduce the payload size on network calls which is already quite heavy for this page.

# External Changes

n/a

# Additional information

n/a


GeronimoJohn (approved):

scottkelso:
Closing without merging as the performance change is likely not worth it for the reduction of image quality seen everywhere across the website.

PR #3280, opened 2024-08-20
feat: use languages continents query

GeronimoJohn:
# Description

Created a use hook query that retrieves all the languages data with their relevant continents


libs/journeys/ui/src/libs/useLanguagesContinentsQuery/useLanguagesContinentsQuery.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+      }
+    }
+  }
+`
+
+export function useLanguagesContinentsQuery(): QueryResult<GetLanguagesContinents> {
+  const query = useQuery<GetLanguagesContinents>(GET_LANGUAGES_CONTINENTS)
+
+  return query
```
scottkelso, line 26:
```suggestion
  return useQuery<GetLanguagesContinents>(GET_LANGUAGES_CONTINENTS)
```

scottkelso (changes requested):
Just one suggestion then good to go!

scottkelso (approved):

PR #3283, opened 2024-08-21
feat: use sort language continents

GeronimoJohn:
Create a function that sorts our languages to their own respected regions


apps/watch/src/libs/algolia/useSortLanguageContinents/useSortLanguageContinents.spec.tsx
```diff
@@ -0,0 +1,17 @@
...
+    expect(result).toEqual({
+      Africa: ['Deutsch', 'Arabic'],
+      Asia: ['Mandarin', 'Japanese'],
+      Europe: ['Français'],
+      Oceania: ['Bislama'],
+      NorthAmerica: ['English'],
+      SouthAmerica: ['Português']
+    })
+  })
+})
```
scottkelso, line 17: Is it worth testing some unhappy paths? What if your languages is an empty list? How should this function behave?

scottkelso (dismissed):

scottkelso (approved):
Class!

PR #3284, opened 2024-08-21
feat: add language button on searchbar

scottkelso:
# Description

### Issue

Need a language button on the searchbar to trigger language dropdown

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-575/english-button-on-searchbar)

### Solution

Hidden by default
Add button that console logs when clicked

# External Changes

n/a

# Additional information
## How to test

Open searchbar in stories and ensure there is a view that has a button present


GeronimoJohn (approved):
Not sure what design wants, whether if an icon remains once the button is clicked or we show nothing at all. We can just fix it later once we have further clarification for direction

PR #3298, opened 2024-08-22
fix: card delete resetting flow

csiyang:
# Description

### Issue

When deleting a card, journey flow gets reset

### Solution

Flow is calculated using the a `step blocks` query a well as `steps` from use editor. A [recent change](https://github.com/JesusFilm/core/pull/3241) caused `steps` to be updated in later render cycle to `step blocks`, this slight mismatch of step data was causing the flow to get reset.

- The fix adds a check for `steps` and `step blocks` to be in sync (have the same length) before calculating if a flow reset is needed.
- Cache fix for duplicating steps
- Cache fix for restoring steps


tataihono (dismissed):

apps/api-journeys/src/app/__generated__/graphql.ts
tataihono, line 1: revert

tataihono (changes requested):

tataihono (dismissed):

tataihono (approved):

tataihono (dismissed):

Kneesal (approved):

PR #3301, opened 2024-08-23
feat: searchbar dropdown

GeronimoJohn:
With all the moving parts finally into the feature branch, we are to move our current dropdown closer to the requirements. We are to group the 1000 languages returned by Algolia into their own respective continents and be able to refine in their respective groups


libs/journeys/ui/src/components/SearchBar/AlgoliaLanguageDropdown/AlgoliaLanguageDropdown.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <MockedProvider mocks={[getLanguagesContinentsMock]}>
+        <AlgoliaLanguageDropdown open handleClickAway={jest.fn()} />
+      </MockedProvider>
+    )
+    await waitFor(() => {
+      fireEvent.click(screen.getByText('English'))
+    })
+    expect(refine).toHaveBeenCalled()
+  })
+})
```
scottkelso, line 60: Probably need a test to describe what the behaviour should be when there are no refinements to show. We should probably hide the continent groups and show one copy of the no refinements message.

scottkelso (changes requested):
Very good! The splitting down of this task seems to have made this much cleaner to tackle.

Just a few comments I have added. Searchbar story seems to have broken. I also added some UX suggestions on linear: i.e. mobile view and clickaway.

scottkelso (approved):

PR #3317, opened 2024-08-27
test: algolia search suggestions

scottkelso:
# Description

### Throwaway Prototype

Only for testing. 

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-674/poc)

# Additional information

Try entering a search and hitting enter. After doing this a few times you will have some search history.
You should also when searching have some suggestions that try to suggest popular searches similar to your search.


scottkelso:
Only a throwaway prototype.

PR #3330, opened 2024-08-28
feat: show selected language in button

GeronimoJohn:
To improve the user experience and to let our user know that a language is currently applied or selected. We are to show in the button the current selected languages. If there are multiple languages selected we are to show the top most selected value from the refinements


libs/journeys/ui/src/components/SearchBar/LanguageButton/LanguageButton.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    const button = screen.getByRole('button', { name: 'Language' })
+    fireEvent.click(button)
+    expect(onClick).toHaveBeenCalled()
+  })
+
+  it('should render the button with the selected language', () => {
+    render(<LanguageButton onClick={jest.fn()} selectedLanguage="English" />)
+    expect(screen.getByRole('button', { name: 'English' })).toBeInTheDocument()
+  })
```
scottkelso, line 24: Can you add at least one more test test to cover any weird selectedLanguage values? Empty string, multiple commas, no commas but multiple words eta? How should the button render those.

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.spec.tsx
```diff
@@ -111,15 +99,9 @@ describe('SearchbarDropdown', () => {
...
-    const refine = jest.fn()
-    mockUseRefinementList.mockReturnValue({
-      items: languageRefinements,
-      refine
-    } as unknown as RefinementListRenderState)
-
     render(
       <MockedProvider mocks={[getLanguagesContinentsMock]}>
-        <SearchbarDropdown open />
+        <SearchbarDropdown open refinements={refinements} />
```
scottkelso, line 104: Should probably have at least one test that tests usecases for changing language button depending on refinements. I.e. no refinements = `Language`, Spanish = `Spanish`, Spanish and English = `English`

scottkelso (changes requested):
Looking good. Noticed when you tick the Spanish refinement, there is a comma showing on the button. Probably need to sort that. Also maybe one or two additional component tests needed.

GeronimoJohn:
redundant pr

PR #3340, opened 2024-08-29
feat: multiple language selected

GeronimoJohn:
One of the main abilities that we would like to have within the search bar is to see the amount of languages our users have selected


scottkelso (approved):
Good job! Only one naming suggestion. Not a blocker. Double check that the failing e2e test isn't something needing investigation.

PR #3345, opened 2024-08-30
fix: language selection and folder structure

GeronimoJohn:
Selecting a language highlights it in all of the other continents, to make the experience nicer we are to only select the language within that continent and disable the rest.


scottkelso (approved):
Smashing!

PR #3355, opened 2024-09-02
refactor: drop api-videos and use api-media

tataihono:
# Description

move api-videos resolvers across to api-media.


tataihono:
*MUST RUN INFRASTRUCTURE COMMANDS AFTER THIS IS DEPLOYED TO PRODUCTION*

PR #3357, opened 2024-09-02
feat: eng 700 searchbar suggestions

scottkelso:
# Description

### Issue

Need to render some suggestions to educate users on how to apply filters

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-700/searchbar-suggestions)

### Solution

Create a suggestions component for the dropdown
Change the logic for opening and closing dropdown so that it is more consistent switching between both

image
image

# How to test

Should still be able to use languages dropdown as before
Should also be able to navigate between suggestions dropdown and languages dropdown
Should always see two suggestions
When a suggestion is clicked, it should apply those refinements


libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -117,6 +135,7 @@ export function SearchBar({
           open={open}
           id={open ? 'simple-popper' : undefined}
           anchorEl={anchorEl}
+          varient={dropdownVarient}
```
GeronimoJohn, line 138: Noticed this a bit, but I'm assuming this is meant to be 'variant'?

GeronimoJohn (commented):
Awesome stuff! Left 1 comment and can we please add a story for Suggestions?

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
scottkelso: Dyslexia strikes again! Good find.

GeronimoJohn (approved):

PR #3374, opened 2024-09-04
feat: step header display title

levi-zustiak:
# Description

### Changes
- Updated footer to include the display title for website mode
- Added the display title to `StepHeader`
- Test and stories for website mode


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Footer/Footer.tsx
```diff
@@ -66,6 +66,7 @@ export function Footer(): ReactElement {
       {websiteMode && <WebsiteToggle />}
       {journey?.website === true ? (
         <>
+          <DisplayTitle />
```
csiyang, line 69: Can we add a test in footer to see that the displaytitle edit is being shown?

libs/journeys/ui/src/components/StepHeader/StepHeader.tsx
```diff
@@ -39,11 +42,51 @@ export function StepHeader({
       }}
     >
       {journey?.website === true ? (
-        <>
-          {/* Logo */}
-          {/* Title */}
-          {/* Menu */}
-        </>
+        <Stack
```
csiyang, line 45: Are these styles copied from the footer? I just want to double check that all the styles are still relevant? Especially would be nice if we could simplify the stack, maybe refactor to Box if it doesn't specifically need to be a stack

csiyang (changes requested):

libs/journeys/ui/src/components/StepHeader/StepHeader.tsx
levi-zustiak: Yeah, they should be "updated" but they will likely be updated after adding the logo and menu icon

csiyang (approved):

PR #3384, opened 2024-09-05
feat: eng 844 dynamically show search query

scottkelso:
# Description

### Issue

See how it feels for users to have their suggestions dynamically updated to their query string

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-844/dynamically-show-search-query)

### Solution

n/a

# External Changes

n/a

# Additional information

Still would need to do testing but doing this as a quick test for design/QA


libs/journeys/ui/src/components/SearchBar/SearchDropdown/Suggestions/Suggestions.tsx
```diff
@@ -25,21 +28,23 @@ export function Suggestions({ refinements }: SuggestionsProps): ReactElement {
...
     const suggestionParts = suggestion.split(/\s(?:in|and)\s/)
-    refine(suggestionParts[0])
     refineLanguages(suggestionParts.slice(1))
   }
 
   return (
     <Box width="100%">
+      <Typography variant="h6" color="primary.main" marginBottom={6} pl={3}>
+        {t('Suggestions')}
+      </Typography>
```
GeronimoJohn, line 38: Would drop - it's already under a suggestions Tab, having two suggestions feels a bit much? Also seems like this is not in design

GeronimoJohn (approved):
Approved just a lint issue

PR #3413, opened 2024-09-08
feat: eng 849 language name in query

scottkelso:
# Description

### Issue

When a user types a language name and then selects a language filter or suggestion with that language, we want to strip the language in question out of the query as best we can.

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-849/language-name-in-query)

### Solution

Normalise the name of the language being refined i.e. `Spanish, Latin American` -> `spanish`
Remove any occurences of this normalised language from the query. i.e. `Jesus Chinese` -> `Jesus` after `Chinese, Traditional` is selected

# External Changes

Updated the strategies view test to match pattern established

# How To Test

- If a user types jesus spanish, then selects the Spanish, Latin American language, we should set the query string to jesus
- If a user types Jesus Chinese, (with any capitals) then selects the Chinese, Traditional language, we should set the query string to Jesus
- If a user types russian, then selects the Russian, Sign Language language, we should set the query string to empty


libs/journeys/ui/src/components/SearchBar/SearchDropdown/LanguageContinentRefinements/RefinementGroup/RefinementGroup.tsx
```diff
@@ -13,16 +14,37 @@ interface RefinementGroupProps {
   selectedContinent?: string
 }
 
+export function normalizeLanguage(language: string): string {
+  return language.toLowerCase().split(',')[0].trim()
+}
```
GeronimoJohn, line 19: Maybe we could move this into the component? We usually keep functions inside so tests can reflect the component's behavior. If we want it to be separate, we could add a utils folder for it :)

GeronimoJohn (commented):
surprised the renovate ran in this PR

libs/journeys/ui/src/components/SearchBar/SearchDropdown/LanguageContinentRefinements/RefinementGroup/RefinementGroup.tsx
scottkelso: Component tests are higher level than unit tests and therefore don't give the same confidence return on code robustness. If you have the choice, always try to unit test over component test. I'll move it into it's own file.

GeronimoJohn (approved):

PR #3437, opened 2024-09-09
feat: eng 872 refinements count

scottkelso:
# Description

### Issue

A user needs to be able to see how many language refinements are available

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-872/refinements-count)

### Solution

Add a label that is a count of the refinements returned from algolia given the search state

# How to test

- Language tab will try to show a number beside it
- The number will equal the number of language refinements available


libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
```diff
@@ -25,6 +25,46 @@ import { useSortLanguageContinents } from '../../../libs/useSortLanguageContinen
...
+  }`
+}))
+
+interface LocalTabsHeaderProps {
+  label: string
+  count?: number
+  maxCount?: number
+}
+
+function LocalTabsHeader({
```
GeronimoJohn, line 43: can we move anything that relates to LocalTabsHeader to the bottom of the file please

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
```diff
@@ -25,6 +25,46 @@ import { useSortLanguageContinents } from '../../../libs/useSortLanguageContinen
...
+    maxCount: number
+  ): string | null => {
+    if (count == null || count <= 0) return null
+    return count >= maxCount ? `${maxCount}+` : count.toString()
+  }
+
+  const displayedCount = getDisplayedCount(count, maxCount)
+
+  return (
+    <div className="tab-label">
```
GeronimoJohn, line 59: Does Material UI `Box` not play ball with the `Tab`? If so fair enough can ignore this comment

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
```diff
@@ -25,6 +25,46 @@ import { useSortLanguageContinents } from '../../../libs/useSortLanguageContinen
...
+  count,
+  maxCount = 1000
+}: LocalTabsHeaderProps): ReactElement {
+  const getDisplayedCount = (
+    count: number | undefined,
+    maxCount: number
+  ): string | null => {
+    if (count == null || count <= 0) return null
+    return count >= maxCount ? `${maxCount}+` : count.toString()
+  }
```
GeronimoJohn, line 54: convert to function please

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
```diff
@@ -25,6 +25,46 @@ import { useSortLanguageContinents } from '../../../libs/useSortLanguageContinen
 import { RefinementGroups } from './RefinementGroups'
 import { Suggestions } from './Suggestions'
 
+const StyledBox = styled(Box)(({ theme }) => ({
+  borderRadius: 32,
```
GeronimoJohn, line 29: if we change to be `borderRadius: '50%'` I think it'll become more of a circle in one digit cases

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
scottkelso: The MUI tab doesn't have a class that I can target for hiding on smaller devices. Wrapped it in a `div` so that I can add the targetable custom class `tab-label` which is hidden on small devices in the `StyledTab`

libs/journeys/ui/src/components/SearchBar/SearchDropdown/SearchbarDropdown.tsx
scottkelso: Have sent you on slack what it looks like with `50%`. Can change it but I actually think it looks better with `32`.

GeronimoJohn (approved):

PR #3440, opened 2024-09-10
fix: refinement groups to be using grid

GeronimoJohn:
Each column's width is currently defined by their content. We are to generalise the width of each column in the language tab


scottkelso (approved):
Looks good!

PR #3442, opened 2024-09-10
 fix: add language changer in title

zwsaqx:
# Description
this pr is working just needs the weird languages to not return undefined
the toolbar should display and allow editing of the language directly
### Issue

it currently doesnt

[Link to Basecamp Todo](https://3.basecamp.com/3105655/buckets/38213939/todos/7668412584)

### Solution

add language field


edmonday:
address in https://github.com/JesusFilm/core/pull/3559

PR #3443, opened 2024-09-10
fix: remove dependencies that triggers default selection

GeronimoJohn:
Some of the dependencies within the useEffect triggers the default language selection function. Removed those dependencies and have made the useEffect to only call the default function depending on `refinements`


scottkelso (approved):
Looks good. Don't be afraid to use the default description PR template. This would have helped describe the problem, separate from the solution.

PR #3446, opened 2024-09-10
feat: watch middleware

GeronimoJohn:
# Description

### Issue

The purpose of this middleware is to figure out the country the user is in. We are to later use this information to render necessary information within the watch client

### Solution

Create a middleware in watch where we save the users country as a cookie. We're using the geo property that we can call from `NextRequest`. 

Things to keep in mind is that geo property comes out of the box for projects hosted on Vercel. If by any point we choose to run watch on a different hosting platform we are to update the configuration accordingly. Read more here: https://nextjs.org/docs/pages/api-reference/functions/next-request#geo

### DIFF
--- a/apps/watch/middleware.spec.ts
+++ b/apps/watch/middleware.spec.ts
@@ -1,0 +1,26 @@
+import { NextRequest, NextResponse } from 'next/server'
+
+import { middleware } from './middleware'
+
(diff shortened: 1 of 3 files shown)


scottkelso (approved):
Awesome!

PR #3453, opened 2024-09-11
feat: use country query

GeronimoJohn:
# Description

### Issue

We need the ability to grab a country information, this information would be used to render the languages a user can choose from in the CountryLanguageSelector

### Solution

Created the a useCountryQuery hook that enables us to call data for a country

it should return the following details: 

- flagPngSrc - image link to country flag
- countryLanguages
      - returns the language(s) spoken by that country
      - returns the amount of speakers of that language

### DIFF
--- a/apps/journeys-admin/__generated__/GetCountry.ts
+++ b/apps/journeys-admin/__generated__/GetCountry.ts
@@ -1,0 +1,40 @@
+/* tslint:disable */
+/* eslint-disable */
+// @generated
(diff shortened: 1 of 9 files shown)


scottkelso (dismissed):
Apply changes from review at https://github.com/JesusFilm/core/pull/3456#pullrequestreview-2294952757

scottkelso (approved):

PR #3456, opened 2024-09-11
feat: get top spoken languages

GeronimoJohn:
# Description

Depends on #3453 

### Issue

We need the ability to get the top spoken languages available to a `Country` and the content that we have in `Algolia`

### Solution

Create a function to get the top spoken languages of a country based on the number of speakers. The logic is based on the number of speakers who speak a particular language. We then retrieve these languages and check if they exist in the `languageEnglishName` refinements.


libs/journeys/ui/src/libs/algolia/getTopSpokenLanguages/getTopSpokenLanguages.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const languageNames = sortedLanguages.map(({ language }) => {
+    const localName = language.name.find(({ primary }) => !primary)?.value
+    const nativeName = language.name.find(({ primary }) => primary)?.value
+    return localName ?? nativeName ?? ''
+  })
+
+  const topSpokenLanguages = languageNames
+    .filter(Boolean)
+    .filter((language) => availableLanguageSet.has(language))
+    .slice(0, 4)
```
scottkelso: Could you add a function parameter that defaults to 4 instead, therefore allowing anyone to override this if anyone needs to.

libs/journeys/ui/src/libs/useCountryQuery/useCountryQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+          }
+        }
+        speakers
+      }
+    }
+  }
+`
+
+export function useCountryQuery(
+  variables?: GetCountryVariables
```
scottkelso, line 24: Should variables be optional here? `const { data } = useCountryQuery()` yields empty response. Isn't this query only useful when we provide an ID like `const { data } = useCountryQuery({ countryId: 'AD' })`?

libs/journeys/ui/src/libs/useCountryQuery/useCountryQuery.mock.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { GetCountry } from './__generated__/GetCountry'
+import { country } from './data'
+import { GET_COUNTRY } from './useCountryQuery'
+
+export const getCountryMock: MockedResponse<GetCountry> = {
+  request: {
+    query: GET_COUNTRY,
+    variables: {
+      countryId: 'country.id'
```
scottkelso, line 11: Just a thought - Would it be more useful to use something that is at least the same shape as the real data? E.g. `XX`?

scottkelso (changes requested):
Nice! Just left some suggestions.

libs/journeys/ui/src/libs/algolia/getTopSpokenLanguages/getTopSpokenLanguages.tsx
GeronimoJohn: Love it, this would be following Open/Closed principle right?

scottkelso (approved):
Awesome!

PR #3467, opened 2024-09-12
fix: language buttons hydration error

GeronimoJohn:
# Description

### Issue

On developer environment, we're getting a hydration error for Language Buttons

### Solution

Update the Box that's causing the error, from Material UI to stick with being a `div` but have a role of a `button`


scottkelso (approved):
Nice solve! UI looks unchanged to me.

PR #3469, opened 2024-09-12
feat: logo edit

csiyang:
# Description

Edit logo for website mode

- add new icon
- add new types to front end `logoImageBlock`, `scale`
- logo image edit
- logo scale edit


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Footer/Logo/Logo.tsx
```diff
@@ -0,0 +1,234 @@
...
+    journeyUpdate(id: $id, input: $journeyUpdateInput) {
+      id
+      logoImageBlock {
+        id
+      }
+    }
+  }
+`
+
+export function Logo(): ReactElement {
```
csiyang, line 54: Logo Edit

libs/shared/ui/src/components/icons/Diamond.tsx
```diff
@@ -0,0 +1,10 @@
+import { createSvgIcon } from '@mui/material/utils'
+
+export default createSvgIcon(
```
csiyang, line 3: new icon

libs/journeys/ui/src/libs/JourneyProvider/journeyFields.tsx
```diff
@@ -90,5 +90,8 @@ export const JOURNEY_FIELDS = gql`
     showLikeButton
     showDislikeButton
     displayTitle
+    logoImageBlock {
```
csiyang, line 93: new `logoImageBlock` field for journey

libs/journeys/ui/src/components/Image/imageFields.ts
```diff
@@ -10,5 +10,6 @@ export const IMAGE_FIELDS = gql`
     width
     height
     blurhash
+    scale
```
csiyang, line 13: new `scale` field for imageBlocks

apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/ImageBlockEditor/CustomImage/CustomImage.tsx
```diff
@@ -11,7 +11,7 @@ import { ImageUpload } from './ImageUpload'
 interface CustomImageProps {
   onChange: (input: ImageBlockUpdateInput) => void
   setUploading?: (uploading?: boolean) => void
-  selectedBlock: ImageBlock | null
+  selectedBlock?: ImageBlock | null
```
csiyang, line 14: Had to add a few of these to satisfy some types, should have no real impact on the app since all our checks are nullish checks

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Footer/Logo/Logo.tsx
levi-zustiak, line 54: Looks like the foreign key for the logo doesn't get updated when removing the logo. It still technically works since the block itself is soft deleted, but it's probably better to update it regardless.

levi-zustiak (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Footer/Logo/Logo.tsx
csiyang: No thats not quite whats happening. The only thing here that actually deletes it is the `undo` action that happens after the initial create. Everything else is just updating the src field of the imageBlock, so the existing reference is accurate.

Although its not the cleanest way, its a lot more simple to do it this way.

levi-zustiak (dismissed):

levi-zustiak (approved):

PR #3472, opened 2024-09-12
feat: grab all the languages from algolia

GeronimoJohn:
# Description

### Issue

Currently we set a limit for the languageEnglishName refinement to only return 5 at the start, this doesn't reflect all the languages that we currently have

### Solution

Up the limit of the refinement to be 1000 and update RefinementGroup to show the rest of the languages through CSS


libs/journeys/ui/src/components/SearchBar/SearchDropdown/RefinementGroups/RefinementGroups.tsx
```diff
@@ -40,14 +40,13 @@ export function RefinementGroups({
 }: RefinementGroupsProps): ReactElement {
   const { t } = useTranslation('apps-watch')
   const theme = useTheme()
+  const [isOpen, setIsOpen] = useState(false)
```
scottkelso, line 43: Would this be a better name?
```suggestion
  const [isCollapsed, setIsCollapsed] = useState(false)
```

scottkelso (approved):
Really clean solution! Left one suggestion for a var rename but not a blocker.

PR #3477, opened 2024-09-13
fix: tidy up top bar on mobile

edmonday:
# Description

### Issue

The top bar of admin looks messy on mobile. 

### Solution

- Move the title and description into three dots menu
- Create a new 'Edit Details' menu item to allow users to edit the title and description
- Add Preview button icon to the top bar


apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
```diff
@@ -316,9 +319,12 @@ export function Toolbar({ user }: ToolbarProps): ReactElement {
           <Items />
         </>
       )}
-      <Box ref={menuRef}>
+      <Stack ref={menuRef} flexDirection="row">
+        <Box sx={{ display: { xs: 'flex', sm: 'none' }, px: 2 }}>
+          <PreviewItem variant="icon-button" />
+        </Box>
```
GeronimoJohn, line 325: Is `PreviewItem` reused? If not, can we please move the sx applied to the `box` in to the `box` in `PreviewItem`

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
```diff
@@ -80,21 +82,54 @@ export function Menu({ user }: MenuProps): ReactElement {
...
+                maxWidth: 'auto',
+                overflow: 'hidden',
+                display: '-webkit-box',
+                '-webkit-line-clamp': '2',
+                '-webkit-box-orient': 'vertical'
+              }}
+            >
+              {journey?.description}
+            </Typography>
+          </Stack>
```
GeronimoJohn, line 116: Just following the pattern set in this component, can we move this into it's own component

apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.spec.tsx
```diff
@@ -118,10 +118,12 @@ describe('Toolbar', () => {
...
-    expect(screen.getByText('My Awesome Journey Title')).toBeInTheDocument()
-    expect(
-      screen.getByText('My Awesome Journey Description')
-    ).toBeInTheDocument()
+    expect(screen.getByTestId('ToolbarTitle')).toHaveTextContent(
+      'My Awesome Journey Title'
+    )
+    expect(screen.getByTestId('ToolbarDescription')).toHaveTextContent(
+      'My Awesome Journey Description'
+    )
```
GeronimoJohn, line 126: This works but just following jest priority https://testing-library.com/docs/queries/about/#priority

Not sure if will work but can we please try getting by role if subtitle or caption and with the right text

apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
edmonday: Screenshot 2024-09-17 at 3 36 18 PM
`PreviewItem` is reused in `Items.tsx`. There is sx that hides that the `Items" component (contains the `PreviewItem` component) on mobile.

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
edmonday: extracted this out to a component called `JourneyDetails`

GeronimoJohn (approved):

PR #3488, opened 2024-09-15
feat: hide search bar dropdown

GeronimoJohn:
# Description

### Issue

On search bar click the dropdown still shows in context that it might not be needed

### Solution

We are to add a prop to the search bar which conditionally renders the dropdown depending on the case that it's in


scottkelso (approved):
Awesome!

libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/data.ts
```diff
@@ -0,0 +1,3 @@
+// The cookie is set by the watch middleware
+// Only use for tests
+export const NEXT_COUNTRY = 'NEXT_COUNTRY=00001---US'
```
scottkelso, line 3: Was this dropped in recent changes?

scottkelso (dismissed):

libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/data.ts
GeronimoJohn: yea it's weird it somehow stayed, currently in main as well. removed it

scottkelso (approved):

scottkelso (approved):

PR #3491, opened 2024-09-15
perf: eng 913 lazy load dropdown

scottkelso:
# Description

### Issue

Sometimes the dropdown doesn't open immediately and can be annoying for users. It would be better to lazy load the languages instead of dynamically loading the dropdown.

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-913/lazy-load-dropdown)

### Solution

Fetch the language continent data when a user indicates they are about to interact with the searchbar

# External Changes

Made useLanguageContinents query into a lazy query
Made useSortLanguageContinents into a utility function, didn't see any reason why it should be a hook

# How to test

The language continents should be fetched when...
* a user hovers over the searchbar
* a user's touch event enters the searchbar like scrolling to it on mobile
* a user focuses on the searchbar
* a user clicks the searchbar or language button


libs/journeys/ui/src/libs/algolia/sortLanguageContinents/sortLanguageContinents.tsx
```diff
@@ -1,18 +1,21 @@
...
+  [continent: Continent]: Language[]
 }
 
-interface useSortLanguageContinentsProps {
+interface sortLanguageContinentsProps {
   languages: Languages[]
 }
 
-export function useSortLanguageContinents({
+export function sortLanguageContinents({
```
scottkelso, line 16: Thoughts on moving this into the algolia/languageUtils @GeronimoJohn ?

GeronimoJohn: Yes for the idea! Though was reflecting on this recently. My realisation is that the functions in the global libs folder are functions that can be reused in multiple components. 

But on our case the functions in utils are all being used in `SearchBar`. Thoughts of us creating a utils folder there instead since it's not really a function that will be reused in other components in journeys/ui? Though one can argue that normaliseLanguage can be reused haha

libs/journeys/ui/src/libs/algolia/sortLanguageContinents/sortLanguageContinents.tsx
```diff
@@ -1,18 +1,21 @@
 import mapValues from 'lodash/mapValues'
 
-import { GetLanguagesContinents_languages as Languages } from '../useLanguagesContinentsQuery/__generated__/GetLanguagesContinents'
+import { GetLanguagesContinents_languages as Languages } from '../../useLanguagesContinentsQuery/__generated__/GetLanguagesContinents'
 
-interface LanguageContinentsRecord {
-  [continent: string]: string[]
+export type Continent = string
+export type Language = string
```
GeronimoJohn, line 6: A part of me thinks this should come from the SearchBarProvider instead. Provider can become like the source of truth for this type?

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -90,6 +88,29 @@ export function SearchBar({
     setOpen(!open)
   }
 
+  const emptyLanguageContinents = sortLanguageContinents({
+    languages: []
+  })
+  const [languageContinents, setLanguageContinents] =
+    useState<LanguageContinentsRecord>(emptyLanguageContinents)
```
GeronimoJohn, line 95: Just wondering how come we call the function for it to be an empty value that we use for this state? Thoughts on setting a default value instead?

scottkelso: Was a handy way to get it to be the correct type (LanguageContinentsRecord). Maybe it would be cleaner just instantiating an empty LanguageContinentsRecord.

libs/journeys/ui/src/libs/algolia/sortLanguageContinents/sortLanguageContinents.tsx
scottkelso: Had that thought to. Came to this because it is the function responsible for putting it into that logical mapping of `Continent` -> `Languages`. Don't mind either way.

libs/journeys/ui/src/libs/algolia/sortLanguageContinents/sortLanguageContinents.tsx
scottkelso: Agree in general with this idea. Will have a go at moving some of these components as you say in separate PR.

GeronimoJohn (approved):

GeronimoJohn (dismissed):

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
```diff
@@ -123,6 +141,10 @@ export function SearchBar({
             }}
             data-testid="SearchBar"
             ref={popperRef}
+            onMouseEnter={prepareDropdown}
+            onTouchStart={prepareDropdown}
+            onClick={prepareDropdown}
```
GeronimoJohn, line 146: is mouse enter not enough to prepare the data needed on desktop?

GeronimoJohn (approved):
Looking good! Just left a question

libs/journeys/ui/src/components/SearchBar/SearchBar.tsx
scottkelso: I get you. You're saying because we have onMouseEnter for desktop and onTouchStart for mobile, we _shouldn't_ need onClick? After a quick google there are some cases when you might still want an onClick as well as a fallback behaviour for when onTouchEvents aren't fired or in the case of keyboard accessibility.

GeronimoJohn (approved):

GeronimoJohn (approved):

PR #3504, opened 2024-09-16
fix: update strategies to consume excerpt

GeronimoJohn:
# Description

### Issue

Strategy card items are currently consuming all of the content for description. This won't do as it's not very descriptive

### Solution

Update the hook useAlgoliaStrategies to be consuming the excerpt from Wordpress. The value comes with `p` tags we are to ensure the tag doesn't get renderd in strategies

### DIFF
--- a/apps/watch/src/libs/algolia/useAlgoliaStrategies/useAlgoliaStrategies.spec.ts
+++ b/apps/watch/src/libs/algolia/useAlgoliaStrategies/useAlgoliaStrategies.spec.ts
@@ -5,6 +5,7 @@ import { useHits } from 'react-instantsearch'
 
 import {
   StrategyItem,
+  removeExcerptPTags,
   transformAlgoliaStrategies,
   useAlgoliaStrategies
 } from './useAlgoliaStrategies'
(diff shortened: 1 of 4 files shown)


apps/watch/src/libs/algolia/useAlgoliaStrategies/useAlgoliaStrategies.spec.ts
```diff
@@ -13,6 +14,13 @@ jest.mock('react-instantsearch')
 
 const mockUseHits = useHits as jest.MockedFunction<typeof useHits>
 
+describe('removeExcerptPTags', () => {
+  it('should remove p tags from excerpt', () => {
+    const result = removeExcerptPTags('<p>description</p>')
+    expect(result).toBe('description')
+  })
+})
+
```
scottkelso, line 23: Yeay for unit testing! Might be nice to have another unit test to ensure that multiple p tags are removed, maybe also no tags.

scottkelso (approved):
LGTM! Just added a suggestion for other unit test cases but the main happy path is covered 😄

scottkelso (approved):

PR #3512, opened 2024-09-17
fix: android video autoplay

mikeallisonJS:
# Description

### Issue

Autoplay not working on first card on Android

### Solution


mikeallisonJS:
closing to revist

PR #3514, opened 2024-09-17
fix: update country detector

GeronimoJohn:
# Description

### Issue

Current implementation of how we detect the country doesn't work as expected. As it's picking up the country where vercel is currently hosted compared to where the user actually is

### Solution

Remove the use of middleware and create an api endpoint that picks up the geolocation of the user instead


libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/CountryLanguageSelector.spec.tsx
```diff
@@ -9,6 +9,16 @@ import { languageRefinements } from '../../data'
...
 import { NEXT_COUNTRY } from './data'
 
+global.fetch = jest.fn(
+  async () =>
+    await Promise.resolve({
+      json: async () =>
+        await Promise.resolve({
+          country: 'US'
+        })
+    })
```
scottkelso, line 19: Would be super slick to have this refactored into that data.ts file.

libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/CountryLanguageSelector.tsx
```diff
@@ -52,27 +52,21 @@ export function CountryLanguageSelector({
...
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+    void fetch('/api/jf/watch.html/geolocation').then((response) => {
+      void response.json().then((data: { country?: string }) => {
+        if (data.country != null) {
+          const countryName = parseCountryName(data.country)
+          setCountry(countryName)
+          setCountryCode(data.country)
+        }
+      })
```
scottkelso, line 68: Can this be done with `await` keywords instead?

scottkelso (changes requested):
Looking good! Left just 2 things to check.

libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/CountryLanguageSelector.tsx
```diff
@@ -52,31 +52,25 @@ export function CountryLanguageSelector({
...
+  const findUserCountry = useCallback(async () => {
+    await fetch('/api/jf/watch.html/geolocation').then(async (response) => {
+      await response.json().then((data: { country?: string }) => {
+        if (data.country != null) {
+          const countryName = parseCountryName(data.country)
+          setCountry(countryName)
+          setCountryCode(data.country)
+        }
+      })
+    })
```
scottkelso, line 69: Was thinking more can it be done like...
```suggestion
    const response = await fetch('/api/jf/watch.html/geolocation');
    const data: { country?: string } = await response.json();
    
    if (data.country) {
      const countryName = parseCountryName(data.country);
      setCountry(countryName);
      setCountryCode(data.country);
    }
```

scottkelso (changes requested):
Nice refactor - one one thing I think can be improved.

scottkelso (approved):

apps/watch/pages/api/jf/watch.html/geolocation.ts
tataihono, line 1: can this move to apps/watch/pages/api/geolocation.ts

PR #3518, opened 2024-09-18
fix: move country detector to searchbar

GeronimoJohn:
# Description

### Issue

Geolocation end point keeps on getting called everytime the popper opens

### Solution

Move the country detector logic to the search bar so that the geolocation endpoint only gets called once


scottkelso (dismissed):
LGTM

scottkelso (approved):

PR #3524, opened 2024-09-18
feat: default journey accessible form the root domain

Kneesal:
# Description

### Issue

window history overrides the next/router and hence is not actually a scalable solution, especially if we want to add card routing.

###Solution

index page is now rendered via length of journey collections - this is scalable in terms of card routing but also in the future if we decide to make the index page a collection of journeys as opposed to a default journey


apps/journeys/pages/home/index.tsx
tataihono, line 1: revert

apps/journeys/pages/[hostname]/index.tsx
```diff
@@ -56,67 +97,124 @@ function HostJourneysPage({ journeys }: HostJourneysPageProps): ReactElement {
...
-                        >
-                          <Fade in timeout={(index + 1) * 1000}>
-                            <StyledIframe
-                              src={`/embed/${slug}`}
+              </Stack>
+            </Container>
+          </ThemeProvider>
+        </>
+      )}
+      {props.pageType === 'JourneyCollection' && (
```
Comment (tataihono) [inline apps/journeys/pages/[hostname]/index.tsx:105]: the collection and journey render should be unified between the hostname and home folder.

tataihono (changes requested):

tataihono (approved):

tataihono (approved):

PR #3526, opened 2024-09-18
fix: drop dev indices strategy sections

GeronimoJohn:
# Description

### Issue

It looks like that we might not be using the dev indices in the near future or not at all

### Solution

Drop the use of dev indices and enable it to be used in all environments

### DIFF
--- a/apps/watch/next.config.js
+++ b/apps/watch/next.config.js
@@ -20,6 +20,7 @@ const nextConfig = {
       // jesusfilm wordpress website
       { protocol: 'https', hostname: 'develop.jesusfilm.org' },
       { protocol: 'https', hostname: 'jesusfilm.org' },
+      { protocol: 'https', hostname: 'www.jesusfilm.org' },
       // arclight image provider - cloudfront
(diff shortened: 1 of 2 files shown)


apps/watch/next.config.js
```diff
@@ -20,6 +20,7 @@ const nextConfig = {
       // jesusfilm wordpress website
       { protocol: 'https', hostname: 'develop.jesusfilm.org' },
       { protocol: 'https', hostname: 'jesusfilm.org' },
+      { protocol: 'https', hostname: 'www.jesusfilm.org' },
```
scottkelso, line 23: Would have assumed L22 covered this? Disregard if not.

scottkelso (commented):
Failing build needing sorted, and also just double checking the next config.

apps/watch/next.config.js
GeronimoJohn: That's what I thought too, but it specifically asked for it

scottkelso (approved):

PR #3528, opened 2024-09-19
fix: text overflow on small devices

scottkelso:
# Description

### Issue

Text spills outside dropdown on small devices

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-949/text-spill-suggestions)

### Solution

Remove the problematic `<ListItemText>` to allow noWrap to work.
image


libs/journeys/ui/src/components/SearchBar/SearchDropdown/Suggestions/Suggestion/Suggestion.tsx
```diff
@@ -36,17 +35,19 @@ export function Suggestion({
...
-          {query}
-        </Typography>
-        <Typography display="inline">{` - in ${filtersLabel}`}</Typography>
-      </ListItemText>
+      <Typography component="span" fontWeight="1000">
+        {query}
+      </Typography>
+      <Typography
+        pl={1}
+        component="span"
```
GeronimoJohn, line 43: Wondering if we can achieve the same outcome without converting it to a span?

scottkelso: Ah, you can. Class!

GeronimoJohn (approved):

PR #3542, opened 2024-09-22
fix: enable journey map draggable in analytics mode

Kneesal:
# Description

### Issue

journey map should be draggable in analytics mode 

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/SocialPreviewNode/SocialPreviewNode.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/SocialPreviewNode/SocialPreviewNode.tsx
@@ -68,14 +68,15 @@ export function SocialPreviewNode(): ReactElement {
     <BaseNode
       id="SocialPreview"
       selected={activeContent === ActiveContent.Social}
-      sourceHandle={HandleVariant.Shown}
+      sourceHandle={
+        showAnalytics === true ? HandleVariant.Disabled : HandleVariant.Shown
+      }
       targetHandle={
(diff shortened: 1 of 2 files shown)


apps/journeys-admin/src/components/Editor/Slider/Slider.tsx
```diff
@@ -251,8 +251,7 @@ export function Slider(): ReactElement {
           height: {
             xs: `calc(100svh - ${EDIT_TOOLBAR_HEIGHT}px - 50px)`,
             md: '100%'
-          },
-          pointerEvents: showAnalytics === true ? 'none' : 'auto'
```
csiyang, line 255: This removal lets you create nodes via the social media node, can we get that case covered in this PR? The desired behaviour is that no blocks can be created and edited in analytics mode

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/SocialPreviewNode/SocialPreviewNode.tsx
```diff
@@ -75,6 +75,7 @@ export function SocialPreviewNode(): ReactElement {
       onSourceConnect={handleSourceConnect}
       isSourceConnected
       positionTargetHandle={false}
+      sourceHandleProps={{ isConnectableStart: showAnalytics === false }}
```
csiyang, line 78: Can you use the base node prop on line 71
```suggestion
			sourceHandle={ showAnalytics === true ? HandleVariant.Shown : HandleVariant.Disabled }
```

csiyang (changes requested):

csiyang (approved):

PR #3546, opened 2024-09-23
refactor: eng 908 searchbar provider refactor

scottkelso:
# Description

### Issue

The search bar provider needs to be told how to update it's state like which languages to remove or which to add. This means components that don't care what continent a language is part of still need to calculate this and tell the provider. 
[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-908/searchbar-provider-refactor)

### Solution

The provider should be responsible for updating it's own state. I.e. let components that want to refine by a language, only worry about refining by a language. In the case where a component wants to add a language to a specific continent, it can tell the context this - otherwise the context will update it's state accordingly

### Additional information
The expensive api call for all continent languages is triggered on a user interaction and so on complete, is given to the context. The context needs this to maintain it's state. There is a chance that while waiting for this on initial call, a user may do something to apply a refinement that cannot be applied until this call is resolved.

### How to test
Regression test all interactions that add and remove language refinements to ensure they still behave as before
* select a language from dropdown
* select another language from the dropdown
* select first suggestion to apply one language
* select second suggestion to apply two languages
* use the language buttons to remove a language
* use the clear all button to remove all languages

In all of the above cases, when a language refinement is applied
1. it should be ticked
2. other copies of that language should be disabled
3. the refinement name should appear in the url


libs/journeys/ui/src/components/SearchBar/SearchDropdown/CountryLanguageSelector/CountryLanguageSelector.stories.tsx
```diff
@@ -17,9 +21,13 @@ const CountryLanguageSelectorStory: Meta<typeof CountryLanguageSelector> = {
...
-    <SearchBarProvider>
-      <CountryLanguageSelector {...args} />
-    </SearchBarProvider>
+    <MockedProvider mocks={[getLanguagesContinentsMock]}>
+      <InstantSearchTestWrapper>
+        <SearchBarProvider>
+          <CountryLanguageSelector {...args} />
+        </SearchBarProvider>
+      </InstantSearchTestWrapper>
+    </MockedProvider>
```
GeronimoJohn, line 30: since we're already using `apolloClient` in the parameters in each of the stories. Can we please just add the mock there and drop `MockedProvider`

libs/journeys/ui/src/components/SearchBar/SearchDropdown/RefinementGroups/RefinementGroups.tsx
```diff
@@ -47,10 +45,11 @@ export function RefinementGroups({
   const { refine: clearRefinements, canRefine: canClearRefinements } =
     useClearRefinements()
 
-  const { dispatch } = useSearchBar()
+  const { state } = useSearchBar()
```
GeronimoJohn, line 48: pattern that we go for is to destructure one step further `{ state: { allContinentLanguages } }`

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
```diff
@@ -28,25 +38,36 @@ interface SelectLanguageContinentAction {
...
+function setAllContinentLanguages(
+  state: SearchBarState,
+  action: SetAllContinentLanguagesAction
+): SearchBarState {
+  const { allContinentLanguages } = action
+  return {
+    ...state,
+    allContinentLanguages
+  }
+}
```
GeronimoJohn, line 69: Can we please add test for this please

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
```diff
@@ -28,25 +38,36 @@ interface SelectLanguageContinentAction {
...
 interface RemoveAllLanguageContinentsAction {
   type: 'RemoveAllLanguageContinents'
 }
 
 type SearchBarAction =
+  | SetAllContinentLanguagesAction
   | SelectLanguageContinentAction
   | SetDefaultLanguageContinentAction
   | RemoveLanguageContinentsAction
   | RemoveAllLanguageContinentsAction
```
GeronimoJohn, line 58: Just wondering where are we using `RemoveAllLanguageContinentsAction`? Thoughts on just removing one or the other? Can one of them just cover removing languages

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
```diff
@@ -17,6 +21,12 @@ export interface SearchBarState {
    * selected languages sorted by continent
    */
   continentLanguages: Record<Continent, Language[]>
+  allContinentLanguages: Record<Continent, Language[]>
```
GeronimoJohn, line 24: Have been reflecting on the naming - thoughts on doing `selectedContinentLanguages` and `defaultContinentLanguages` or something along the lines?

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
```diff
@@ -154,15 +186,41 @@ interface SearchBarProviderProps {
...
   children,
   initialState
 }: SearchBarProviderProps): ReactElement {
   const [state, dispatch] = useReducer(reducer, {
     continentLanguages: {},
+    allContinentLanguages: {},
     ...initialState
   })
 
+  const { items } = useRefinementList(languageRefinementProps)
```
GeronimoJohn, line 206: this might require a bit of work, maybe, but thoughts on just passing this from the component instead of being defined here?

GeronimoJohn (commented):
Just a few things

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
scottkelso: Thought it might be useful in the future incase a component wants to tell the provider to unset all languages and also doesn't need to know which are currently set to do so. And it was already there too. Happy either way.

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
scottkelso: Good shout. I like `selectedContinentLanguages`. If we don't like `allContinentLanguages`, what about `continentLanguageMap` or `continentLanguageDirectory` to indicate **selectedContinentLanguages** is only for which are selected, and the **continentLanguageMap** is the full suite of languages mapped to the correct continent to lookup?

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
scottkelso: Your component would need to dispatch the new copy of items from useRefinementList every time it changes. And your context would need a new dispatch type to handle this. Not sure that would add any value for the additional overhead.

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
scottkelso: Ah, I think I've got something that should be clear. Note I have swapped the order of the values here compared to .
```suggestion
  /**
   * only the selected languages sorted by continent
   */
  selectedContinentLanguages: Record<Continent, Language[]>
  /**
   * all the languages sorted by continent
   */
  continentLanguages: Record<Continent, Language[]>
```

libs/journeys/ui/src/libs/algolia/SearchBarProvider/SearchBarProvider.tsx
GeronimoJohn: Makes sense

GeronimoJohn (approved):

PR #3550, opened 2024-09-23
refactor: remove formium front end

csiyang:
# Description

Formium seems to be no longer maintained so we are removing it from the project. This is only part 1 of the removal, targeted at removing all of the related front end


Kneesal (approved):
well done bro, this was a lot of work

kiran-redhat (approved):

PR #3561, opened 2024-09-25
feat: eng 957 resources url

scottkelso:
# Description

### Issue

URL parameters are verbose on resources page and also don't match videos page. Since https://github.com/JesusFilm/core/pull/3560 there will also be a reference to the algolia index name.

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-957/resources-url)

### Solution

Add the state mapping to the nextRouter used in the resources page

# External Changes

Refactor the nextRouter objects into their own libs class.

# How to test

* As a user interacts with the searchbar, the query param in the url should always represent the text in the searchbar
* the url query params should always be in the format... `.../resources?query=something`


apps/watch/src/libs/algolia/instantSearchRouter/instantSearchRouter.ts
```diff
@@ -0,0 +1,44 @@
...
+      stateToRoute(uiState) {
+        const indexUiState = uiState[indexName]
+
+        const stateRoute = {
+          query: indexUiState.query,
+          refinementList: {
+            languageEnglishName:
+              indexUiState.refinementList?.languageEnglishName
+          }
+        } as unknown as UiState
```
GeronimoJohn, line 33: From memory, not every page will have the same object shape for refinementList. e.g. if we add refinements to Resources. The pathing for that data would be `taxonomies.languages` and journeys might have their own way of doing things too. 

We should probably open this up for any shape?

GeronimoJohn (commented):
Great idea! This really cleans things up. Left a thought on `instantSearchRouter`

apps/watch/src/libs/algolia/instantSearchRouter/instantSearchRouter.ts
```diff
@@ -0,0 +1,50 @@
...
+  serverUrl?: string
+  mapRefinements?: (indexUiState) => any
+}
+
+export function createInstantSearchRouter({
+  indexName = process.env.NEXT_PUBLIC_ALGOLIA_INDEX ?? '',
+  serverUrl = process.env.NEXT_PUBLIC_WATCH_URL ?? '',
+  mapRefinements = (indexUiState) => ({
+    languageEnglishName: indexUiState.refinementList?.languageEnglishName
+  })
```
GeronimoJohn, line 21: NIT: can we make this a standalone function please hehe

GeronimoJohn (approved):

GeronimoJohn (approved):

levi-zustiak (approved):

PR #3570, opened 2024-09-25
fix: apollo cache rehydration

tataihono:
This pr primarily fixes apollo cache hydration on pages after the initial load. This causes the prior cache to not get wiped out as a result of new values to hydrate.

other smaller fixes:
- drawer flashing on editor initial load
- topbar on editor show navigation elements before journey load
- animate in the canvas once loaded
- don't load journey list on ssr since we are now caching between navigation events


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.tsx
```diff
@@ -176,11 +175,28 @@ export function Properties({ block, step }: PropertiesProps): ReactElement {
...
+  return (
+    <Stack
+      component={Paper}
+      elevation={0}
+      sx={{
+        height: '100%',
+        borderRadius: 3,
+        borderBottomLeftRadius: 0,
+        borderBottomRightRadius: 0,
+        display: 'flex',
```
Kneesal, line 189:
```suggestion

```

apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/Drawer.tsx
```diff
@@ -105,10 +108,10 @@ export function Drawer({
           borderRadius: 3,
           borderBottomLeftRadius: 0,
           borderBottomRightRadius: 0,
-          width: smUp ? DRAWER_WIDTH : 'auto',
+          width: { xs: 'auto', md: DRAWER_WIDTH },
           left: { xs: 0, md: 'auto' },
-          top: { xs: 0, md: 20 },
-          right: { xs: 0, md: 20 },
+          top: { xs: 0, md: 15 },
+          right: { xs: 0, md: 16 },
```
Kneesal, line 114: since we are importing the media query hook, might as well use it for other props?

Kneesal (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/Drawer.tsx
tataihono: don't use the media query hook if possible. anywhere. the reason its used here is because the anchor attribute can't be set in css.

Kneesal (approved):
lgtm

PR #3574, opened 2024-09-26
fix: select options not displaying

levi-zustiak:
# Description

### Issue
- Options for select inputs are not displaying in the properties drawer

### Solution
- Add the swiper class to prevent event capture for the drawer


scottkelso (approved):
Nice fix. I can confirm the dropdown now opens.

PR #3603, opened 2024-10-04
fix: video language graphql info

mikeallisonJS:
# Description

### Issue

Video languages are not working in journeys

### Solution

fix graphql info searching for Hive

### DIFF
--- a/apps/api-media/src/schema/video/video.ts
+++ b/apps/api-media/src/schema/video/video.ts
@@ -12,21 +12,36 @@ import { VideoLabel } from './enums/videoLabel'
 import { VideosFilter } from './inputs/videosFilter'
 import { videosFilter } from './lib/videosFilter'
 
+interface Info {
+  variableValues:
+    | Record<
+        string,
+        Array<
+          Array<{
+            id: string
+            primaryLanguageId: string
+          }>
+        >
+      >
+    | Array<{ id: string; primaryLanguageId: string }>
+}
+
 export function getLanguageIdFromInfo(
-  info: any,
+  info: unknown,
   parentId: string
 ): string | undefined {
-  return typeof info.variableValues === 'object'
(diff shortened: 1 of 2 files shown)


tanflem (dismissed):

apps/api-media/src/schema/video/video.ts
```diff
@@ -214,11 +214,23 @@ const Video = builder.prismaObject('Video', {
...
+          typeof info.variableValues === 'object'
+            ? (
+                Object.values(info.variableValues) as Array<
+                  Array<{
+                    id: string
+                    primaryLanguageId: string
+                  }>
+                >
+              ).find((inner) => inner.find(({ id }) => id === parent.id))?.[0]
+                .primaryLanguageId
```
levi-zustiak: Would it make more sense to use a `flatMap` and then a find? I think this would return the incorrect variant id similar to the issue Tatai fixed in this PR https://github.com/JesusFilm/core/pull/3312

apps/api-media/src/schema/video/video.ts
```diff
@@ -214,12 +214,28 @@ const Video = builder.prismaObject('Video', {
...
+                  Array<{
+                    id: string
+                    primaryLanguageId: string
+                  }>
+                >
+              ).find(
+                (inner) =>
+                  Array.isArray(inner) &&
+                  inner.find(({ id }) => id === parent.id)
+              )?.[0].primaryLanguageId
```
levi-zustiak, line 230: I think this would still need to find the correct language id like the other condition does. Otherwise, would it result in the error Tatai fixed in this PR? https://github.com/JesusFilm/core/pull/3312

levi-zustiak (changes requested):

levi-zustiak (approved):

tanflem (approved):

PR #3647, opened 2024-10-13
feat: mobile scroll down arrow

jianwei1:
### Issue

Show a bouncing scroll down arrow at bottom of card if there is more content below (on mobile)


libs/journeys/ui/src/components/Card/OverlayContent/DownScrollArrow/DownScrollArrow.tsx
```diff
@@ -0,0 +1,49 @@
...
+import Fade from '@mui/material/Fade'
+import { keyframes } from '@mui/material/styles'
+import { ReactElement, useEffect, useState } from 'react'
+
+import ChevronDownIcon from '@core/shared/ui/icons/ChevronDown'
+
+interface DownScrollArrowProps {
+  trigger: boolean
+}
+export function DownScrollArrow({
```
scottkelso, line 11: Would it be a better name to swap scroll and down and also not tie the naming to what icon is used?

```suggestion
interface ScrollDownIndicatorProps {
  trigger: boolean
}
export function ScrollDownIndicator({
```

You could also take the icon in as a prop so that in the future you could use this with another icon instead of that particular down arrow.

libs/journeys/ui/src/components/Card/OverlayContent/DownScrollArrow/DownScrollArrow.tsx
```diff
@@ -0,0 +1,49 @@
...
+  trigger
+}: DownScrollArrowProps): ReactElement {
+  const [shouldAnimate, setShouldAnimate] = useState(false)
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setShouldAnimate(true)
+    }, 5000)
```
scottkelso, line 19: This is a MAGIC_NUMBER so should be pulled out as a const. Even make it a prop too that defaults to `BOUNCE_DELAY = 5000`

libs/journeys/ui/src/components/Card/OverlayContent/OverlayContent.tsx
```diff
@@ -16,6 +19,16 @@ export function OverlayContent({
...
+    disableHysteresis: true,
+    target: cardOverlayRef.current,
+    threshold: 20
+  })
+  const isScrollable = (): boolean => {
+    const box = cardOverlayRef.current
+    return box != null ? box.scrollHeight > box.clientHeight : false
+  }
   const enableVerticalScroll: SxProps = {
     overflowY: 'scroll',
```
scottkelso, line 33:
```suggestion
  const cardOverlayRef = useRef<any>()
  
  const trigger = useScrollTrigger({
    disableHysteresis: true,
    target: cardOverlayRef.current,
    threshold: 20
  })
  
  const isScrollable = (): boolean => {
    const box = cardOverlayRef.current
    return box != null ? box.scrollHeight > box.clientHeight : false
  }
  
  const enableVerticalScroll: SxProps = {
    overflowY: 'scroll',
```

Add spacing to break up the code blocks for better reading

scottkelso (changes requested):
Looking good! Added a few things to think about and would also need some sort of test for this. At the very least just test that it appears when we would expect it to.

libs/journeys/ui/src/components/Card/OverlayContent/ScrollDownIndicator/ScrollDownIndicator.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { keyframes } from '@mui/material/styles'
+import { ReactElement, useEffect, useState } from 'react'
+
+import ChevronDownIcon from '@core/shared/ui/icons/ChevronDown'
+
+const BOUNCE_DELAY = 5000
+
+interface ScrollDownIndicatorProps {
+  trigger: boolean
+}
```
scottkelso, line 12: It would be a really pro move to have `bounceDelay` and `icon` be optional props that default to `BOUNCE_DELAY` and `<ChevronDownIcon />` respectfully. This better adheres to the open/closed principle.

scottkelso (changes requested):
Looks well! Left a couple of suggestions. Also need some test coverage.

tataihono:
closing due to age.

PR #3660, opened 2024-10-15
fix: add series to algolia search filter in journeys admin

Kneesal:
# Description

updated algolia filter for journeys-admin


Kneesal:
discussion around this ticket has halted, closing pr for now

PR #3667, opened 2024-10-16
feat: background image focal point

GeronimoJohn:
# Description

### Issue

Users currently face challenges uploading images with exact proportions, making background image editing difficult. To improve the user experience, we're introducing the ability for users to set a focal point for background images.

### Solution

To support this enhancement, we need to add two new fields—focalTop and focalLeft—to the Image Block. These fields will allow users to specify the focal point of the background image.

The following changes are made:

1. Add focalTop and focalLeft to the block model in Schema.Prisma within the api-journeys package.
2. Update the GraphQL schema to include these new fields in the Image Block.
3. Create a FocalPoint component that enables the users to set the focal point of an image


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundMedia/Image/FocalPoint/FocalPointInputs/FocalPointInputs.tsx
```diff
@@ -0,0 +1,37 @@
...
+          label={axis === 'x' ? t('Left') : t('Top')}
+          value={localPosition[axis].toFixed(0)}
+          onChange={(e) => handleChange(axis, e.target.value)}
+          slotProps={{
+            input: { endAdornment: '%' },
+            htmlInput: { min: MIN_VALUE, max: MAX_VALUE }
+          }}
+          sx={{ width: '45%' }}
+        />
+      ))}
```
tataihono, line 34: just make two inputs. this is not really saving much code but its more confusing.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundMedia/Image/FocalPoint/FocalPointImage/GridLines/GridLines.tsx
```diff
@@ -0,0 +1,35 @@
...
+              position: 'absolute',
+              top: `${position}%`,
+              left: 0,
+              right: 0,
+              height: '1px',
+              backgroundColor: 'rgba(255,255,255,0.5)'
+            }}
+          />
+        </>
+      ))}
```
tataihono, line 32: can we write this out by hand? just introducing a loop which makes it a bit more complex.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundMedia/Image/FocalPoint/FocalPointImage/FocalPointImage.tsx
```diff
@@ -0,0 +1,83 @@
...
+        overflow: 'hidden',
+        position: 'relative',
+        alignItems: 'center',
+        justifyContent: 'center',
+        borderRadius: 4,
+        border: (theme) => `1px solid ${theme.palette.divider}`
+      }}
+      onClick={handleClick}
+    >
+      {imageBlock?.src != null ? (
```
tataihono, line 55: i don't think this component should render at all if there is no image.

tataihono (changes requested):
I found that splitting this out into a bunch of components is making this more difficult to understand. I don't think you'll reuse any of the subcomponents anywhere else.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundMedia/Image/FocalPoint/FocalPointImage/FocalPointImage.tsx
GeronimoJohn: That makes sense

tataihono (approved):

PR #3675, opened 2024-10-17
fix: email notifications being submitted

Kneesal:
# Description

### Issue

journey notifications aren't being sent

### Solution

prisma to search the userId col for user as opposed to id col


apps/api-users/src/schema/user/user.spec.ts
```diff
@@ -200,7 +200,7 @@ describe('api-users', () => {
       })
       expect(prismaMock.user.findUnique).toHaveBeenCalledWith({
         where: {
-          id: '1'
+          userId: user.id
```
tataihono, line 203: shouldn't this be 
```suggestion
          userId: user.userId
```

but also can you edit the underlying mock so that user id and user userId aren't 1?

tataihono (changes requested):

tataihono (approved):

PR #3680, opened 2024-10-17
feat: add focal top and left to image block

GeronimoJohn:
# Description

### Issue
Users currently face challenges uploading images with exact proportions, making background image editing difficult. To improve the user experience, we're introducing the ability for users to set a focal point for background images.

### Solution
To support this enhancement, we need to add two new fields—focalTop and focalLeft—to the Image Block. These fields will allow users to specify the focal point of the background image.

The following changes are made:

1. Add focalTop and focalLeft to the block model in Schema.Prisma within the api-journeys package.
2. Update the GraphQL schema to include these new fields in the Image Block.


tataihono:
@GeronimoJohn can we set a db default of 50?

apps/api-journeys/src/app/modules/block/image/image.graphql
```diff
@@ -36,6 +38,9 @@ input ImageBlockCreateInput {
   """
   isCover: Boolean
   scale: Int
+
```
tataihono, line 41: nit: no space here

tataihono (changes requested):

tataihono (approved):

PR #3684, opened 2024-10-17
fix: toolbar icons hover spacing

storyworks:
# Description

### Issue

Toolbar icon hover backgrounds are squashed, icon spacing is off and toolbar overflows on tablet view

### Solution

Adjusted padding and gaps between icons, display mobile toolbar at mdUp breakpoints

# Additional information

See linear ticket as I've removed some lines that were included in previous fixes. Need to ensure they still work, I wasn't sure how to reproduce the issues to test myself.


apps/journeys-admin/src/components/Editor/Toolbar/Items/CommandRedoItem/CommandRedoItem.tsx
```diff
@@ -22,14 +21,12 @@ export function CommandRedoItem({
   }
 
   return (
-    <Box data-testid="CommandRedoItem">
```
storyworks, line 25: Couldn't see if this box was used anywhere. Is the "Redo" data-testid not enough for testing?

apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
```diff
@@ -38,10 +38,12 @@ export const GET_ROLE = gql`
 
 interface MenuProps {
   user?: User
+  ref: MutableRefObject<null>
```
storyworks, line 41: @csiyang , you added the original refs in, I'm not too sure what it's supposed to do - would it work to refactor it here?

apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
```diff
@@ -195,188 +195,206 @@ export function Toolbar({ user }: ToolbarProps): ReactElement {
...
-          <Typography
-            variant="caption"
-            sx={{ display: { xs: 'none', md: 'block' } }}
-          >
-            <Skeleton width="80%" />
-          </Typography>
-        </Stack>
-      )}
-      <Items />
-      <Stack ref={menuRef} flexDirection="row">
```
storyworks, line 361: menuRef used to be applied to the container that wraps the preview item and menu. I've refactored away this Stack and the Box wrapping PreviewItem, can the menuRef be moved directly to menu?

apps/journeys-admin/src/components/Editor/Toolbar/Items/CommandRedoItem/CommandRedoItem.tsx
edmonday: @storyworks yup for sure can drop the box. Not sure why I used it haha

apps/journeys-admin/src/components/Editor/Toolbar/Toolbar.tsx
csiyang: I don't think so, I needed a comparison on `menuRef` and `helpScoutRef` to anchor the popover menu.

This popover only shows up once, so you'll need to update the database to test out your version (`journeyFlow` -> `plausibleJourneyFlowViewed`)

storyworks: Refactored so it works just as well if you put it on a box around menu instead of passing it in. This box should be removed when this flag / feature is removed.

image

PR #3699, opened 2024-10-17
feat: removed focus on empty input logic

jianwei1:
# Description

Allow user to deselect empty text block when clicking elsewhere on the card, or on the canvas


jianwei1:
This was intentional behaviour, not a bug. See linear ticket

PR #3755, opened 2024-10-20
fix: disable add video button if card has background image or video

jianwei1:
### Issue

Users were able to press the "Add Video" button if the card had a background video/image. An apollo error was then thrown

### Solution

Change the hasChildBlock check on AddBlock.tsx to disable the NewVideoButton if the cardblock has any children( previously, it would only check if it had children that weren't coverblocks, which allowed the button to be enabled if the card had a background image/video). Also, update the tooltip in button.tsx to include background media


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/AddBlock/AddBlock.tsx
```diff
@@ -30,9 +30,7 @@ export function AddBlock(): ReactElement {
     (block) => block.__typename === 'CardBlock'
   ) as TreeBlock<CardBlock>
 
-  const hasChildBlock = cardBlock?.children?.some(
-    (block) => block.id !== cardBlock.coverBlockId
-  )
+  const hasChildBlock = cardBlock?.children?.length > 0
```
tataihono, line 33:
```suggestion
  const hasChildBlock = cardBlock?.children != null && cardBlock.children.length > 0
```

tataihono (changes requested):

tataihono (dismissed):

scottkelso (approved):

PR #3758, opened 2024-10-21
feat: update journey updatedAt on block update

edmonday:
# Description

### Issue

In order to sort our `JourneyList` by modified date, we need to update the `Journey` whenever a block is **updated, re-ordered, duplicated, created** or **restored**.

### Solution

- add method  to `block.service.ts` that makes a prisma call to update the Journey `updatedAt` of that block that has been updated.
- call `setJourneyUpdatedAt` on block create, block delete and block restore
- add journey update call on `blockOrderUpdate` and `blockDuplicate` in `block.resolver.ts`

**Blocks accounted for:**
- button
- card
- icon
- image
- radioOption
- signUp
- textResponse
- typography
- video

**Blocks ignored:**
- radioQuestion (ignore this as radioOption creation when radioQuestion is created overrides updatedAt)
- step (card create will handle the updatedAt)


apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -25,6 +25,18 @@ type PrismaTransation = Omit<
   | 'enableShutdownHooks'
 >
 
+export async function updateJourneyUpdatedAt(
```
GeronimoJohn, line 28: Potentially can simplify the function name to` updateJourneyAt`

apps/api-journeys/src/app/modules/block/block.service.ts
```diff
@@ -25,6 +25,18 @@ type PrismaTransation = Omit<
...
+  tx: PrismaTransation,
+  block: Block
+): Promise<void> {
+  await tx.journey.update({
+    where: {
+      id: block.journeyId
+    },
+    data: { updatedAt: block.updatedAt }
+  })
+}
```
GeronimoJohn, line 38: Rather than defining this separately from BlockService, could we move it inside the service? This way, other resolvers can access it via blockService, rather than exporting it from this file.

e.g. in ButtonResolver, we could call it like this: `this.blockService.updateJourneyUpdatedAt`.

GeronimoJohn: You will also have to mock this function in your tests now as well

apps/api-journeys/src/app/modules/block/block.service.ts
GeronimoJohn: Ignore the comment above. This is really more of personal preference. But reading update twice just doesn't feel nice in my opinion hahaha. Could do like `setJourneyUpdatedTime`. Something along the lines

edmonday: will name it `setJourneyUpdatedAt`

GeronimoJohn (approved):

apps/api-journeys/src/app/modules/block/block.resolver.ts
```diff
@@ -49,7 +49,20 @@ export class BlockResolver {
...
+      const reorderedBlocks = await this.blockService.reorderBlock(
+        block,
+        parentOrder
+      )
+      await tx.journey.update({
+        where: {
+          id: block.journeyId
+        },
+        data: { updatedAt: new Date().toISOString() }
+      })
```
GeronimoJohn, line 63: can't this just be `setUpdatedAt`?

GeronimoJohn (approved):

PR #3759, opened 2024-10-21
fix: rtl gradient direction

storyworks:
# Description

### Issue

image

CSS standards in the browser have updated so it knows to switch linear gradient direction based on rtl status. We don't need to manually change it in the code now 

### Solution

Remove the direction change based on RTL calculation, the browser will do that for us now.

# Additional information

Can test with Arabic languages - pick one with arabic characters, arabic bedouin (and possibly others) don't change rtl.


libs/journeys/ui/src/components/Card/ContainedCover/ContainedCover.tsx
```diff
@@ -255,11 +255,11 @@ export function ContainedCover({
                 pl: { sm: 50 },
                 WebkitMask: {
                   xs: overlayGradient('bottom'),
-                  sm: overlayGradient(rtl ? 'left' : 'right')
```
storyworks, line 258: Example of it working now

csiyang (approved):

PR #3785, opened 2024-10-21
refactor: include action and journey

scottkelso:
# Description

### Issue

Lots of repeated object definition on prisma service queries

### Solution

Refactor the include of action and journey into `journey.acl.ts`

# Additional information

Shouldn't have any functional changes


apps/api-journeys/src/app/modules/journey/journey.acl.ts
```diff
@@ -6,6 +6,15 @@ import {
 
 import { Action, AppAclFn, AppAclParameters } from '../../lib/casl/caslFactory'
 
+export const JOURNEY = {
```
tataihono, line 9:
```suggestion
export const INCLUDE_JOURNEY_ACL: Prisma.JourneyDefaultArgs<DefaultArgs> = {
```

tataihono (dismissed):
journey is pretty generic. can we make it more specific. Also might be worth trying to type enforce it.

tataihono (approved):

PR #3802, opened 2024-10-21
fix: add protocol check

GeronimoJohn:
# Description

### Issue

Adding URLs to the chat platforms accepts the string as is and doesn't support adding the HTTPS protocol. Causing the user to visit your.nextstep.is/[chat link]

### Solution

Add a protocol check before Chat button link update. If it already has a protocol proceed to update. Otherwise we are to append the HTTPS protocol to the URL


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/JourneyAppearance/Chat/ChatOption/Details/Details.tsx
```diff
@@ -89,7 +90,8 @@ export function Details({
   setCurrentPlatform,
   setCurrentLink,
   helperInfo,
-  enableIconSelect
+  enableIconSelect,
+  active
```
edmonday, line 94: is this ever used?

edmonday (approved):

edmonday (approved):

PR #3803, opened 2024-10-22
fix: language autocomplete in android

Kneesal:
# Description

### Issue

Language autocomplete dialog is not scrollable on android mobile devices

### Solution

Change the dialog to be more consistent with other dialogs in mobile,  - making it full screen variant on mobile.
Position of popper always opens on top to account for mobile keyboard space
Fixed popper height in mobile so as to not overflow it's container


apps/journeys-admin/src/components/Editor/Toolbar/JourneyDetails/JourneyDetailsDialog/JourneyDetailsDialog.tsx
```diff
@@ -193,12 +197,15 @@ export function JourneyDetailsDialog({
                       />
                     )}
                     popper={{
-                      modifiers: [
-                        {
-                          name: 'flip',
-                          enabled: false
+                      placement: smUp ? 'bottom' : 'top',
```
edmonday, line 200: NIT: can we change this so it's easier to follow,

```suggestion
                      placement: !smUp ? 'top' : 'bottom',
```

Kneesal:
closing in favour of https://github.com/JesusFilm/core/pull/3864

PR #3808, opened 2024-10-22
fix: reset progress

scottkelso:
# Description

### Issue

On new video upload after a previous upload, the progress bar is momentarily at 100%

[Link to Linear Todo](https://linear.app/jesus-film-project/issue/ENG-1211/video-upload-status)

### Solution

Reset progress on completed upload

# External Changes

Also refactoring some logic in AddByFile.tsx into function calls
Some of these function calls were able to be refactored into a utils folder and unit tested. 

# How to test
* Upload a video successfully
* Replace the uploaded video and the progress bar should already be at 0% before it begins to increase


apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/VideoLibrary/VideoFromCloudflare/AddByFile/AddByFileUtils.spec.ts
```diff
@@ -0,0 +1,61 @@
+import { ReadStream, readFileSync } from 'fs'
+import { join } from 'path'
+
+import { fileToCloudflareUpload, getBuffer } from './AddByFileUtils'
+
```
csiyang, line 5: We follow the pattern of wrapping the entire file in a describe with the name of the file

csiyang (changes requested):
`AddByFileUtils.ts` should be renamed, the PascalCase is generally reserved for components. Also why not just name it `utils`? Its already in the `AddByFile` folder.

Have you considered using the existing utils folder pattern?
- `/AddByFile/utils/fileToCloudflareUpload`
- - `/AddByFile/utils/getBuffer`

csiyang (changes requested):
`AddByFileUtils.ts` should be renamed, the PascalCase is generally reserved for components. Also why not just name it `utils`? Its already in the `AddByFile` folder.

Have you considered using the existing utils folder pattern?
- `/AddByFile/utils/fileToCloudflareUpload`
- `/AddByFile/utils/getBuffer`

csiyang (approved):

PR #3856, opened 2024-10-24
fix: ios touch bug

jianwei1:
# Description

### Issue

On iOS devices, tapping once on a card with no connections brings up the canvas view, not 2 taps(first tap to select, second tap to bring up canvas).

### Solution

Note: Unfortunately since it's been a few weeks, I cannot remember exactly why the behaviour was occurring. I believe it was because when the user tapped on a node with no edges, the device recognised the interaction as both a tap and a drag(meaning handleClick() on StepBlockNodeCard.tsx would be called, selecting the node and creating the IPhoneSecondTapLayer box, then calling handleClick() again on that box which would call setSelectedBlockAction bringing up the canvas).

In order to fix it, we moved the selection logic out of StepBlockNodeCard and into journeyFlow instead. Here, we created a function to check whether the interaction was a touch/click. If true, use the useStepAndBlockSelection hook to call the necessary dispatches. If false, the interaction is a drag, so update the positions.

- Extract dispatch/selection logic into it's own hook (useStepAndBlockSelection)
- Removed handleClick() from StepBlockNodeCard.tsx
- On JourneyFlow.tsx, created an isClickOrTouch function:
	- inside onNodeDragStop, if isClickOrTouch then call useStepAndBlockSelection hook, else update positions 

# Additional information

This PR removes work done on Nisal's ticket: https://github.com/JesusFilm/core/pull/3600


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -272,6 +274,7 @@ export function JourneyFlow(): ReactElement {
       let yPos = (event as unknown as MouseEvent).clientY
 
       if (!(event instanceof MouseEvent)) {
+        console.log('touch event  ')
```
Kneesal, line 277: drop

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -339,6 +342,11 @@ export function JourneyFlow(): ReactElement {
       createStepFromAction
     ]
   )
+
+  const isClickOrTouch = (endDragTimeStamp: number): boolean => {
+    return endDragTimeStamp - dragTimeStampRef.current < 150
+  }
+
   const onNodeDragStop: NodeDragHandler = (_event, node): void => {
     if (node.type !== 'StepBlock') return
```
Kneesal, line 351:
```suggestion
  const onNodeDragStop: NodeDragHandler = (event, node): void => {
    if (node.type !== 'StepBlock') return
```

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -347,31 +355,38 @@ export function JourneyFlow(): ReactElement {
...
-          x,
-          y
-        },
-        undo: {
-          x: step.x,
-          y: step.y
+    // if click or tap, go through block selection logic
+    // else go through standard positioning logic below
+
+    if (isClickOrTouch(_event.timeStamp)) {
```
Kneesal, line 361:
```suggestion
    if (isClickOrTouch(event.timeStamp)) {
```

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -491,6 +506,9 @@ export function JourneyFlow(): ReactElement {
         onEdgesChange={onEdgesChange}
         onConnect={onConnect}
         onConnectEnd={onConnectEnd}
+        onNodeDragStart={(_event) => {
```
Kneesal, line 509:
```suggestion
        onNodeDragStart={(event) => {
```

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
```diff
@@ -491,6 +506,9 @@ export function JourneyFlow(): ReactElement {
...
+          dragTimeStampRef.current = _event.timeStamp
```
Kneesal, line 510:
```suggestion
          dragTimeStampRef.current = event.timeStamp
```

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/JourneyFlow.tsx
Kneesal: The underscore syntax is used to denote that a argument of a function is not being used. Since you are using it, you need to drop the underscore

Kneesal (changes requested):

Kneesal (approved):

PR #3934, opened 2024-10-31
feat: color hex validation

DanChae03:
# Description

### Issue

Currently the HEX field in the background colour picker accepts any string as a hex. Fixed to ensure only valid HEX values result in network calls that change the colour.

### Solution

Isolate field value as a useState variable. 
Regex validation for colour string
Reset upon blur if value is invalid.


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundColor/BackgroundColor.tsx
```diff
@@ -132,6 +137,28 @@ export function BackgroundColor(): ReactElement {
     }
   }
 
+  function validateHexString(color: string): boolean {
```
scottkelso, line 140:
```suggestion
  function isValidHex(color: string): boolean {
```
would be common naming convention here

scottkelso (changes requested):
Looks good! Nice refactoring into named functions. For future reference there would be nothing stopping you refactoring out the `isValidHex` function and unit testing it. We definitely need some tests here so make sure that we have somewhere check for the invalid states.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundColor/BackgroundColor.tsx
```diff
@@ -132,6 +137,28 @@ export function BackgroundColor(): ReactElement {
...
+    }
+  }
+
+  async function handleBlur(color: string): Promise<void> {
+    if (isValidHex(color)) {
+      await handleColorChange(color)
+    } else {
+      setColor(selectedColor)
+    }
+  }
```
Kneesal, line 160:
```suggestion
  const validationSchema = object({
    color: string()
      .required(t('Required'))
      .test('valid-hex-color', t('Invalid HEX color code'), isValidHex)
  })
```

Drop these functions and use `Formik` to handle validation, see `LinkAction.tsx`

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundColor/BackgroundColor.tsx
```diff
@@ -89,6 +89,11 @@ export function BackgroundColor(): ReactElement {
   const [tabValue, setTabValue] = useState(0)
   const selectedColor =
     cardBlock?.backgroundColor ?? cardTheme.palette.background.paper
+  const [color, setColor] = useState<string>(selectedColor)
+
+  useEffect(() => {
+    setColor(selectedColor)
+  }, [selectedColor])
```
Kneesal, line 96: is this needed? You're already passing in default state for the component when it mounts in line 92 , useState will handle re-renders

DanChae03: It's so that if selectedColor changes via external means (Palette, etc) it will update in the textfield as well.

DanChae03: Not needed with formik

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/BackgroundColor/BackgroundColor.tsx
```diff
@@ -164,23 +180,22 @@ export function BackgroundColor(): ReactElement {
...
-            startAdornment: (
-              <InputAdornment position="start">
-                <Edit2Icon
-                  onClick={(e) => handleTabChange(e, 1)}
-                  style={{ cursor: 'pointer' }}
-                />
-              </InputAdornment>
-            )
-          }}
+          label=""
```
Kneesal, line 187:
```suggestion
```
drop empty label prop

libs/locales/en/apps-journeys-admin.json
```diff
@@ -235,6 +235,7 @@
   "Medium": "Medium",
   "Large": "Large",
   "Contained": "Contained",
+  "Invalid HEX color code": "",
```
Kneesal, line 238: Looks like the key isn't being translated properly, can you please look into this?

Kneesal (changes requested):

Kneesal (approved):

PR #4047, opened 2024-11-07
feat: drag to reorder blocks on canvas

DanChae03:
# Description

### Issue

Currently the only way to re-order the positioning of blocks on each card is to use the "QuickControls" toolbar that appears upon selecting each block, and you can only move blocks 1 space at a time. 

Dragging the blocks using a drag handle allows for blocks to be moved much more efficiently.

### Solution

Added a drag handle on the left of each block, which appears upon hovering the corresponding block. Each block should now be draggable, and an indicator line should show where the dragged item will be placed.

Using the dnd-kit library to initialise the drop area, the dragging of components, etc.


apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/DropArea/DropArea.tsx
```diff
@@ -0,0 +1,151 @@
...
+    >
+      <SortableContext items={itemIds} strategy={verticalListSortingStrategy}>
+        {children}
+      </SortableContext>
+      {activeItem != null && (
+        <DragOverlay dropAnimation={null}>
+          <Box
+            sx={{
+              ml: -8,
+              pl: 8,
```
csiyang, line 133: Whats this for?

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/SelectableWrapper/SelectableWrapper.tsx
```diff
@@ -1,9 +1,16 @@
+import { DragOverEvent, useDndMonitor } from '@dnd-kit/core'
```
csiyang, line 1: How come the drag logic had to be done here? It ideally should be its own component.

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/DropArea/DropArea.tsx
DanChae03: It's so that when the grab handle is clicked, the cursor has the "grabbing"  cursor shape. Since the drop preview will follow the cursor as it drags, there's no other way for the cursor to have the "grabbing" shape unless it's actively over part of the drag preview - since the handle is on the side, the drag preview needs to invisibly extend to the cursor region as well - this is the only way that I found worked

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/SelectableWrapper/SelectableWrapper.spec.tsx
```diff
@@ -236,13 +236,15 @@ describe('SelectableWrapper', () => {
...
 
     fireEvent.click(getByRole('button', { name: 'Option 1' }))
-    expect(
-      getByTestId(`SelectableWrapper-${radioQuestionBlock.id}`)
-    ).toHaveStyle({
-      outline: '2px solid',
-      zIndex: '1',
-      outlineColor: '#C52D3A'
-    })
+    await waitFor(() =>
```
csiyang, line 239: revert

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/SelectableWrapper/SelectableWrapper.tsx
```diff
@@ -125,12 +125,6 @@ export function SelectableWrapper({
             : ''
         }
         sx={{
-          '&:first-child': {
```
csiyang, line 128: Are these changes still required?

csiyang (changes requested):

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/SelectableWrapper/SelectableWrapper.tsx
DanChae03: No

csiyang (approved):

PR #4112, opened 2024-11-11
refactor: send one journey view per day

csiyang:
# Description

### Issue

JourneyViewEvent was sending every time a website page is viewed. 

### Solution

Limit website and journeys to only create 1 JourneyViewEvent per user, per journey, per 24hr period


apps/api-journeys/src/app/modules/event/journey/journey.resolver.ts
```diff
@@ -3,6 +3,7 @@
 import { UseGuards } from '@nestjs/common'
 import { Args, Mutation, Parent, ResolveField, Resolver } from '@nestjs/graphql'
 import { GraphQLError } from 'graphql'
+// import pick from 'lodash/pick'
```
Kneesal, line 6: drop
```suggestion

```

Kneesal (changes requested):

Kneesal (approved):

PR #4222, opened 2024-11-12
feat: add helper text for react flow controls

DanChae03:
# Description

React flow control icons (especially the recenter view, and reset layout, are not intuitive for users. Added tooltip indicators for users to see which buttons are which. 

### Solution

Wrap all control buttons in tooltips. 
Refactor controls component (which was abstracted as a Reactflow component) and extract individual buttons so that they could be wrapped in tooltips.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/Controls/Controls.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/Controls/Controls.spec.tsx
@@ -1,0 +1,36 @@
+import { MockedProvider } from '@apollo/client/testing'
+import { fireEvent, render, screen } from '@testing-library/react'
+import { ReactFlowProvider } from 'reactflow'
+
+import { Controls } from '.'
+
+describe('Controls', () => {
+  it('should render', () => {
+    render(
+      <MockedProvider>
+        <ReactFlowProvider>
+          <Controls handleReset={jest.fn()} />
+        </ReactFlowProvider>
+      </MockedProvider>
+    )
+
+    expect(screen.getByTestId('Plus1Icon')).toBeInTheDocument()
+    expect(screen.getByTestId('DashIcon')).toBeInTheDocument()
+    expect(screen.getByTestId('Maximise2Icon')).toBeInTheDocument()
+    expect(screen.getByTestId('ArrowRefresh6Icon')).toBeInTheDocument()
+  })
+
+  it('should call reset function on click', () => {
+    const handleReset = jest.fn()
+    render(
+      <MockedProvider>
+        <ReactFlowProvider>
+          <Controls handleReset={handleReset} />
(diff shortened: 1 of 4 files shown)


apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/Controls/Controls.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export function Controls({ handleReset }: ControlsProps): ReactElement {
+  const { zoomIn, zoomOut, fitView } = useReactFlow()
+  const { t } = useTranslation('apps-journeys-admin')
+
+  return (
+    <Control showInteractive={false} showFitView={false} showZoom={false}>
+      <ControlItem onClick={zoomIn} title={t('Zoom in')}>
+        <Plus1 />
+      </ControlItem>
```
GeronimoJohn, line 34: can we make the icon a passable prop instead of it being a children?

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/Controls/Controls.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <Controls handleReset={jest.fn()} />
+        </ReactFlowProvider>
+      </MockedProvider>
+    )
+
+    expect(screen.getByTestId('Plus1Icon')).toBeInTheDocument()
+    expect(screen.getByTestId('DashIcon')).toBeInTheDocument()
+    expect(screen.getByTestId('Maximise2Icon')).toBeInTheDocument()
+    expect(screen.getByTestId('ArrowRefresh6Icon')).toBeInTheDocument()
+  })
```
GeronimoJohn, line 21: can drop since the tooltip test checks the icons already

GeronimoJohn (commented):
Almost there! Just a few things and a test failing

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/Controls/Controls.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+function ControlItem({ onClick, title, icon }): ReactNode {
+  return (
+    <Tooltip title={title} arrow placement="right">
+      <Box>
+        <ControlButton onClick={onClick}>{icon}</ControlButton>
+      </Box>
+    </Tooltip>
+  )
+}
```
GeronimoJohn, line 24: Can we please add types to the props. 

NIT: If we're keeping this outside the main component can we change the return type to ReactElement and put it under the main component. But if we want to stick with ReactNode can we move it inside the main component. Reason - just pattern set in Core

Global search for renderInput for example

GeronimoJohn (dismissed):

GeronimoJohn (approved):

PR #4277, opened 2024-11-13
fix: seo issues on videos page

GeronimoJohn:
# Description

The Single Video/Container page SEO is impacted due to the recent Algolia changes. The issue arises because VideoCard uses `sendEvent` (from Algolia) for tracking video selections, requiring pages that uses VideoCard to be wrapped in `InstantSearchProvider` - even non-search pages.

## Solution

Added an `onClick` prop to VideoCard to handle tracking externally, removing the direct dependency on Algolia's provider. This preserves analytics on search page while fixing SEO in the Single Video/Container Page.

### DIFF
--- a/apps/watch/src/components/VideoGrid/AlgoliaVideoGrid/AlgoliaVideoGrid.tsx
+++ b/apps/watch/src/components/VideoGrid/AlgoliaVideoGrid/AlgoliaVideoGrid.tsx
@@ -20,7 +20,7 @@ export function AlgoliaVideoGrid(props: VideoGridProps): ReactElement {
 
   const handleClick =
     (videoId?: string) =>
-    (event: MouseEvent<HTMLAnchorElement>): void => {
+    (event: MouseEvent): void => {
       event.stopPropagation()
       if (videoId == null) return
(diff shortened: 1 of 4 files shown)


apps/watch/src/components/VideoGrid/AlgoliaVideoGrid/AlgoliaVideoGrid.tsx
```diff
@@ -14,15 +14,28 @@ export function AlgoliaVideoGrid(props: VideoGridProps): ReactElement {
...
   } = useAlgoliaVideos<CoreVideo>({ transformItems })
+
+  const handleClick =
+    (videoId?: string) =>
+    (event: MouseEvent): void => {
+      event.stopPropagation()
+      if (videoId == null) return
+      const item = algoliaVideos.filter((item) => item.id === videoId)
+      sendEvent('click', item, 'Video Clicked')
+      console.log('I got called', item)
```
tataihono, line 28: drop

tataihono (changes requested):
just one console log

tataihono (approved):

PR #4407, opened 2024-11-19
fix: prefetch journey

GeronimoJohn:
# Description

### Issue

Currently, when a journey is loading, we disable the button or show a loading state. While this approach works, there’s an opportunity to optimize the experience by reducing API call times and improving overall page load speed for journeys.

### Solution

Implement prefetching for the target page to minimize API call times and reduce page load times, enhancing the user experience.


GeronimoJohn:
Closing as I won't have enough time to figure out the cause of this issue

PR #4429, opened 2024-11-20
feat: upload image show loading progress and close dialog

Kneesal:
# Description

### Issue

file upload dialog was not updating image correctly

### Solution

it now updates the image and even shows loading state and handles errors


apps/videos-admin/src/libs/SnackbarProvider/SnackbarProvider.tsx
```diff
@@ -0,0 +1,14 @@
+'use client'
```
edmonday, line 1: Is this file necessary? This seems like a wrapper for SnackbarProvider which I think we can just use as is?

edmonday (changes requested):

apps/videos-admin/src/libs/SnackbarProvider/SnackbarProvider.tsx
Kneesal: Yes see https://nextjs.org/docs/14/app/building-your-application/rendering/composition-patterns#using-third-party-packages-and-providers

edmonday: tyvm for the info

edmonday (approved):

PR #4453, opened 2024-11-22
test: add tests for video view

edmonday:
# Description

### Issue

`VideoView.tsx` needed tests and stories

### Solution

- Added tests and stories for `VideoView.tsx`
- Added mock data for `apps/videos-admin/src/libs/useAdminVideo/useAdminVideo.mock.ts`
- Updated `Metadata.tsx` so that video data is passed down to children
- Updated tests for files affected by addition of mock data and changes to `Metadata.tsx`
- Updated `EditProviderProps` so the children is of type `ReactNode`. This way when we need to test useEdit, the `EditProvidor` can take more than one child.


apps/videos-admin/src/app/[locale]/(dashboard)/videos/[videoId]/_VideoView/data.mock.ts
Comment (Kneesal) [inline apps/videos-admin/src/app/[locale]/(dashboard)/videos/[videoId]/_VideoView/data.mock.ts:1]: drop this file and use adminVideoMock file, both files contain the same data

Kneesal (changes requested):

apps/videos-admin/src/app/[locale]/(dashboard)/videos/[videoId]/_VideoView/data.mock.ts
Comment (edmonday) [reply apps/videos-admin/src/app/[locale]/(dashboard)/videos/[videoId]/_VideoView/data.mock.ts:1]: dropped

Kneesal (approved):

PR #4496, opened 2024-11-25
refactor: change number of recommended languages

levi-zustiak:
# Description

### Issue
Vlad has requested to change the number of suggested languages in the language select for watch search bar

### Solution
Add limit to `CountryLanguageSelector`


tataihono:
@levi-zustiak is there a test for this?

PR #4720, opened 2024-12-16
fix: default icon text for button icon property

DanChae03:
### Issue

Instead of reading "Select an icon..." when no icon is selected, it will just say "None".


Kneesal (approved):

Kneesal (dismissed):
add test to assert value if empty, or update any existing tests

jianwei1 (approved):

PR #4727, opened 2024-12-16
fix: turn off share button display for journey creation

DanChae03:
# Description

### Issue
Turn off reaction buttons upon journey creation. They can be added later on. 

Buttons are confusing for newer creators as the purpose of the buttons is not clearly described.


Kneesal (approved):

Kneesal (changes requested):
put the front end stuff into another PR. 
Merge backend in first.

jianwei1 (approved):

PR #4757, opened 2024-12-18
chore: delete null edtions

Kneesal:
# Description

### Issue

after migration, old video editions will null fields still exist

### Solution

delete null editions so we can enforce non-nullability


mikeallisonJS (approved):

Kneesal:
ran script in prod and stage, closing PR

PR #4934, opened 2025-01-23
fix: lang attribute for watch pages

lumberman:
# Description

### Issue

Google serves users content with videos in the wrong language due to the HTML lang attribute set to 'en' for every single video page on Watch.

Screenshot 2025-01-23 at 4 36 08 PM

### Solution

This code pulls the current video language from the pageProperties, defines the lang attribute, and adds the language meta field accordingly.

# Additional information

- We can't use NextJS locale as it's not UI language but content language. More language-related fixes are coming; this is only a first step and won't fix all the language-related issues in Google SERP.


tataihono (approved):
great work!

PR #5041, opened 2025-01-29
feat: add qr code edit dialog

edmonday:
# Description

### Issue

This PR brings in the frontend UI for the QR Code work.

# Additional information

Users will not be able to navigate to this dialog for now.

There will be a PR to follow from @csiyang that will hook up the backend to make this complete.


apps/journeys-admin/src/components/Editor/Toolbar/Items/ShareItem/ShareItem.tsx
```diff
@@ -87,18 +106,7 @@ export function ShareItem({
...
-              journey?.slug != null
-                ? `${
-                    hostname != null
-                      ? `https://${hostname}`
-                      : (process.env.NEXT_PUBLIC_JOURNEYS_URL ??
-                        'https://your.nextstep.is')
-                  }/${journey.slug}`
-                : undefined
-            }
-          />
```
csiyang, line 101: Can you revert this back, we wont need to pass journeyUrl through to the qrCodeDialog. You can just leave the prop there with a empty string for the sake of this PR

apps/journeys-admin/src/components/Editor/Toolbar/Items/ShareItem/ShareItem.spec.tsx
```diff
@@ -152,6 +152,50 @@ describe('ShareItem', () => {
     ).toBeInTheDocument()
   })
 
+  it('should handle qr code', async () => {
```
csiyang, line 155: I think this test will fail right? Maybe just comment out

csiyang (changes requested):

csiyang (approved):

PR #5091, opened 2025-02-02
feat: create qr code with custom id

csiyang:
# Description

Adds option to create qr code with a given id. This allows the front end to pass in the id and generate the cache entry.


jianwei1 (approved):

csiyang:
Realised this change is redundant

PR #5197, opened 2025-02-12
feat: merge backend for spacer block

DanChae03:
# Description

### Issue

Adds backend fields for querying, adding and mutating spacer blocks. 
New resolver for spacer blocks, as well as modifications to the prisma and graphQL schemas.

## Summary by CodeRabbit

- **New Features**
  - Introduced spacer block functionality to enhance layout control within journeys. Users can now create and update spacer blocks via new API operations, offering greater flexibility in managing spacing.
- **Tests**
  - Expanded test coverage for creating and updating spacer blocks ensures reliable performance, robust error handling, and proper authorization controls.


apis/api-journeys/src/app/modules/block/spacer/spacer.graphql
```diff
@@ -0,0 +1,31 @@
...
+
+extend type Mutation {
+  spacerBlockCreate(input: SpacerBlockCreateInput!): SpacerBlock!
+  spacerBlockUpdate(
+    id: ID!
+    input: SpacerBlockUpdateInput!
+    """
+    drop this parameter after merging teams
+    """
+    journeyId: ID
```
csiyang, line 29: Is this needed? Remove if not.

apis/api-journeys/src/app/modules/block/spacer/spacer.resolver.ts
```diff
@@ -0,0 +1,89 @@
...
+        include: {
+          action: true,
+          ...INCLUDE_JOURNEY_ACL
+        }
+      })
+      await this.blockService.setJourneyUpdatedAt(tx, block)
+      if (!ability.can(Action.Update, subject('Journey', block.journey)))
+        throw new GraphQLError('user is not allowed to create block', {
+          extensions: { code: 'FORBIDDEN' }
+        })
```
csiyang, line 55: The journey update should happen after authorization
```suggestion
      if (!ability.can(Action.Update, subject('Journey', block.journey)))
        throw new GraphQLError('user is not allowed to create block', {
          extensions: { code: 'FORBIDDEN' }
        })
      await this.blockService.setJourneyUpdatedAt(tx, block)
```

apis/api-journeys/schema.graphql
```diff
@@ -688,7 +696,7 @@ type GridContainerBlock implements Block {
   journeyId: ID!
   parentBlockId: ID
   parentOrder: Int
-  spacing: Int!
+  gap: Int!
```
csiyang, line 699: This is not the right way to change a field in the schema. The old field needs to be deleted and a new field needs to be added. 

You also need to verify nothing is using the field or the deployed environments will crash. We generally do this by deprecating the field, run some sort of data migration on the database, then delete. You may not need to go through the deprecation process though since I dont think we ever used the grid block, but check this with Mike first.

Also do this change in a separate pr

csiyang (changes requested):

apis/api-journeys/schema.graphql
DanChae03: I discussed this with Nisal and he told me to go ahead with this, because grid container isn't being used in any capacity in our database.

apis/api-journeys/src/app/modules/block/spacer/spacer.graphql
```diff
@@ -0,0 +1,24 @@
...
+}
+
+input SpacerBlockUpdateInput {
+  parentBlockId: ID
+  spacing: Int
+}
+
+extend type Mutation {
+  spacerBlockCreate(input: SpacerBlockCreateInput!): SpacerBlock!
+  spacerBlockUpdate(id: ID!, input: SpacerBlockUpdateInput!): SpacerBlock
```
csiyang, line 23:
```suggestion
  spacerBlockUpdate(id: ID!, input: SpacerBlockUpdateInput!): SpacerBlock!
```

apis/api-journeys/db/migrations/20250211040352_20250211040350/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Block" ADD COLUMN     "gap" INTEGER;
```
csiyang, line 2: Is this migration still meant to be here?

csiyang (changes requested):

csiyang (approved):

PR #5357, opened 2025-02-24
fix: improve YouTube video autoplay initialisation

thomastayler:
# Description

On all networks, if the journey has a youtube video and the users navigates to the video cards, then press play, pause then play again the youtube video will break and not load.

On slower networks this bug presents itself more often as the user tries to play the video before it's ready

[[Link to Basecamp Todo](https://3.basecamp.com/3105655/projects)](https://linear.app/jesus-film-project/issue/ENG-1910/youtube-video-freezes#comment-2fed9324)

### Solution

Function is introduced to add a buffer when the user tries to play the video before it's ready. This function also includes a buffer to force the video to play after 10 seconds.

Please note this does not fix the issue for 3g network speed.

# External Changes

N/A

# Additional information

Bug fix

## Summary by CodeRabbit

- **New Features**
	- Updated video autoplay behavior to allow autoplay for YouTube sources without requiring specific conditions, enhancing user experience.
	- Improved initialization logic for the video player to ensure it initializes only once, optimizing performance.


libs/journeys/ui/src/components/Video/VideoControls/DesktopControls/DesktopControls.tsx
```diff
@@ -65,8 +71,14 @@ export function DesktopControls({
...
         onClick={handlePlay}
+        disabled={
+          source === VideoBlockSource.youTube && playing && currentTime === 0
+        }
         sx={{
-          ml: { xs: 0, lg: -1 }
+          ml: { xs: 0, lg: -1 },
+          '&.Mui-disabled': {
+            opacity: 0.5
+          }
```
Kneesal, line 81: Do you need this? Usually this gets handled for you in line 74-76: the underlying component usually handles the styling for disabled

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -229,6 +232,11 @@ export function VideoControls({
   ])
 
   function handlePlay(): void {
+    // Disables play button when video is at 0.0 for YouTube videos
+    if (isYoutube && state.playing && (player.currentTime() ?? 0) === 0) {
+      return
+    }
+
```
Kneesal, line 239: make this a const isLoading 
`const isLoading = isYoutube && state.playing && (player.currentTime() ?? 0) === 0`

```suggestion
    // Disables play button when video is at 0.0 for YouTube videos
    if (isLoading) {
      return
    }

```

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
```diff
@@ -426,6 +434,14 @@ export function VideoControls({
...
               fullscreen={fullscreen}
               handleFullscreen={handleFullscreen}
+              currentTime={player.currentTime() ?? 0}
+              source={
+                isYoutube
+                  ? VideoBlockSource.youTube
+                  : isMux
+                    ? VideoBlockSource.mux
+                    : VideoBlockSource.internal
+              }
```
Kneesal, line 444:
```suggestion
              disabled={isLoading}
```

Kneesal (changes requested):

libs/journeys/ui/src/components/Video/VideoControls/DesktopControls/DesktopControls.tsx
thomastayler: True, I've tested and removed this

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
thomastayler: Yes, this is needed. If I remove it, the small button is disabled but the screen button becomes usable

libs/journeys/ui/src/components/Video/VideoControls/DesktopControls/DesktopControls.tsx
```diff
@@ -40,6 +44,8 @@ export function DesktopControls({
   startAt,
   endAt,
   progress,
+  currentTime,
+  source,
```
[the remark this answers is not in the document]
thomastayler: Done

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
thomastayler: Done

libs/journeys/ui/src/components/Video/VideoControls/VideoControls.tsx
thomastayler: Done

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
```diff
@@ -80,22 +80,13 @@ export function InitAndPlay({
...
-    muted,
-    posterBlock,
-    setPlayer,
-    videoRef,
-    autoplay,
-    activeStep,
-    source
-  ])
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- should only run once
+  }, [muted, setPlayer, videoRef, autoplay, source])
```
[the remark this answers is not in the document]
thomastayler: do we need to add something that disposes the player?

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
```diff
@@ -80,22 +82,14 @@ export function InitAndPlay({
           muted: muted === true,
           autoplay:
             autoplay === true &&
-            activeStep &&
+            (activeStep || activeBlock) &&
```
csiyang, line 85: Active block is the current block that the user is on. Since there is no check to see if the video block belongs to it, it looks like it will always be true. It kinda of makes the check redundant, which actually is fine if this works. We should be able to just remove live 85.

csiyang (changes requested):

Kneesal (approved):

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
```diff
@@ -116,19 +116,10 @@ export function InitAndPlay({
...
-    muted,
-    posterBlock,
-    setPlayer,
-    videoRef,
-    autoplay,
-    activeStep,
-    source,
-    muxMetadata
-  ])
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- should only run once
```
csiyang, line 121: Needs to be updated

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
```diff
@@ -88,7 +90,8 @@ export function InitAndPlay({
 
   // Initiate video player
   useEffect(() => {
-    if (videoRef.current != null) {
+    // Only initialize video player if not already done
```
csiyang, line 93: I dont actually think this comment is needed. I think the comments on line 91 and 121 are already explaining the useEffect

csiyang (changes requested):

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
thomastayler: True

thomastayler: I've removed this comment

csiyang (approved):

csiyang (approved):

libs/journeys/ui/src/components/Video/InitAndPlay/InitAndPlay.tsx
thomastayler: Good catch, I've updated this now

PR #5538, opened 2025-03-12
feat(api-media): add update-slugs command to project configuration

mikeallisonJS:


mikeallisonJS:
closing, only here for reference if needed again

PR #5749, opened 2025-03-18
feat(api-journeys-modern): exporting history

levi-zustiak:
## Summary by CodeRabbit

- **New Features**
  - Introduced journey event export logging, enabling users to generate and export logs of journey interactions.
  - Enhanced filtering options now allow selection of specific event types and optional date ranges for refined log outputs.
  - Expanded API capabilities support seamless creation and retrieval of these logs, delivering improved tracking of user engagement.
  - New enumeration for event types associated with user interactions has been added.
  
- **Bug Fixes**
  - Adjusted the `dateRangeEnd` field to accept NULL values, improving data flexibility.


apis/api-journeys-modern/src/lib/graphql/subgraphGraphql.ts
```diff
@@ -0,0 +1,7 @@
+import { initGraphQLTada } from 'gql.tada'
+
+// import type { introspection } from '../../__generated__/graphql-subgraph-env.js'
+
+export const graphql = initGraphQLTada<{
+  introspection: any
```
levi-zustiak, line 6: @mikeallisonJS I'm not entirely sure how to generate the type for this.

apis/api-journeys/db/migrations/20250318223633_20250318223628/migration.sql
```diff
@@ -0,0 +1,21 @@
+-- CreateTable
```
csiyang, line 1: I think this migration needs to be api-journeys-modern

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { JourneyEventsExportLogInput } from './inputs'
+
+builder.prismaObject('JourneyEventsExportLog', {
+  fields: (t) => ({
+    id: t.exposeID('id', { nullable: false }),
+    createdAt: t.expose('createdAt', { type: 'Date', nullable: false }),
+    userId: t.exposeID('userId', { nullable: false }),
+    journeyId: t.exposeID('journeyId', { nullable: false }),
+    eventsFilter: t.exposeStringList('eventsFilter', { nullable: false }),
```
csiyang, line 14: Can we make this the a list of typenames? Each of the events are in the events module in api-journeys

csiyang (changes requested):

apis/api-journeys/db/migrations/20250318223633_20250318223628/migration.sql
levi-zustiak: The migrations are still added through api-journeys. I had to ask Mike about it yesterday to make sure

mikeallisonJS (dismissed):

apis/api-journeys-modern/src/schema/journeyEventsExportLog/inputs/JourneyEventsExportLogInput.ts
```diff
@@ -0,0 +1,17 @@
+import { builder } from '../../builder'
+import { EventType } from '../enums'
+
+export const JourneyEventsExportLogInput = builder.inputType(
+  'JourneyEventsExportLogInput',
+  {
+    fields: (t) => ({
+      journeyId: t.id({ required: true }),
+      eventsFilter: t.field({
```
csiyang, line 9: This shouldn't be required, no filter should mean the user has requested all the events. Maybe we can add a description for this?

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { prisma } from '../../lib/prisma'
+import { builder } from '../builder'
+
+import { JourneyEventsExportLogInput } from './inputs'
+
+builder.prismaObject('JourneyEventsExportLog', {
+  fields: (t) => ({
+    id: t.exposeID('id', { nullable: false }),
+    createdAt: t.expose('createdAt', { type: 'Date', nullable: false }),
```
csiyang, line 11: Can this also be date time iso?

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
```diff
@@ -0,0 +1,66 @@
...
+    journeyId: t.exposeID('journeyId', { nullable: false }),
+    eventsFilter: t.exposeStringList('eventsFilter', { nullable: false }),
+    dateRangeStart: t.expose('dateRangeStart', {
+      type: 'DateTimeISO',
+      nullable: true
+    }),
+    dateRangeEnd: t.expose('dateRangeEnd', {
+      type: 'DateTimeISO',
+      nullable: true
+    })
```
csiyang, line 22: I think each of these fields can be nullable right?

apis/api-journeys-modern/db/schema.prisma
```diff
@@ -636,3 +637,17 @@ model QrCode {
...
 }
+
+model JourneyEventsExportLog {
+  id             String    @id @default(uuid())
+  createdAt      DateTime  @default(now())
+  userId         String
+  journeyId      String
+  eventsFilter   String[]
+  dateRangeStart DateTime?
+  dateRangeEnd   DateTime  @default(now())
```
csiyang, line 648: This filed shouldnt get a default value. It should only be populated if the user actually applies it as a filter

csiyang (changes requested):

apis/api-journeys-modern/db/schema.prisma
levi-zustiak: The linear ticet specified that the dateRangeEnd field default to 'createdAt' date if there wasn't a value. Was that not intended or did I misunderstand the ask?

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
levi-zustiak: This is set as Date to match the implementations in api-media. I believe there was an issue with the types using iso, but I can try again

apis/api-journeys-modern/src/schema/journeyEventsExportLog/inputs/JourneyEventsExportLogInput.ts
levi-zustiak: I think we actually need to use an empty array as requesting all events since Prisma doesn't allow nullable lists. We could make this field nullable and back it with an empty array, but I'd probably opt for explicitly including in the input I think

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
levi-zustiak: eventsFilter is required due to the same reasons in the comment for the field type

apis/api-journeys-modern/db/schema.prisma
csiyang: Oh yeah, that's my bad, you're right I did originally ask for this. 

I think I would like it removed on reflection because
- filling it in with a default value is redundant since created at already contains that information
- makes it look like the user is applying that filter while they're actually not

apis/api-journeys-modern/src/schema/journeyEventsExportLog/journeyEventsExportLog.ts
csiyang: Oh yeah it would be good if we can this one working? I would've figured it just works since we're doing basically the same thing with the dateRanges?

apis/api-journeys/db/migrations/20250327010915_20250327010913/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
```
csiyang, line 1: Can you get this migration merged with the other one? You should be able to just delete both of the migrations and then generate/migrate again

csiyang (changes requested):

apis/api-gateway/schema.graphql
```diff
@@ -2190,6 +2196,16 @@ type Translation @join__type(graph: API_JOURNEYS)  {
...
 
+type JourneyEventsExportLog @join__type(graph: API_JOURNEYS_MODERN)  {
+  id: ID!
+  createdAt: DateTimeISO!
+  userId: ID!
+  journeyId: ID!
+  eventsFilter: [String!]!
+  dateRangeStart: DateTimeISO
+  dateRangeEnd: DateTimeISO
+}
```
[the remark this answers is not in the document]
levi-zustiak: We aren't storing the event typenames in an enum hence the use of the [String!]!

csiyang (approved):

PR #5887, opened 2025-03-26
feat: export data button

levi-zustiak:
## Summary by CodeRabbit

- **New Features**
  - Introduced a CSV export option on the visitor reports page, allowing users to download journey event data.
  - Added an "Export Data" button to initiate the export process.
  - Enhanced error feedback for the export process, ensuring users are notified if data retrieval issues occur.
  - The `FilterDrawer` component now accepts a `journeyId` prop for improved data context.

- **Documentation**
  - Updated localization entries for improved user messaging related to export functionality.


apps/journeys-admin/pages/journeys/[journeyId]/reports/visitors.tsx
```diff
@@ -288,7 +296,8 @@ export const getServerSideProps = withUserTokenSSR({
 
   return {
     props: {
-      ...translations
+      ...translations,
+      journey
```
Comment (csiyang) [inline apps/journeys-admin/pages/journeys/[journeyId]/reports/visitors.tsx:300]: I don't think we should fetch the journey here, I think just passing in the journeyId from  the parent component will be sufficient

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
```diff
@@ -8,14 +10,45 @@ import Radio from '@mui/material/Radio'
...
+    'languageId',
+    'email',
+    'position',
+    'source',
+    'progress',
+    'updatedAt',
+    'journeyName',
+    'visitorId',
+    'visitorName'
+  ]
```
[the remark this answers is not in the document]
csiyang: I think this is fine, I've made updates to it on my hook pr https://github.com/JesusFilm/core/pull/5906 for the more specific changes from design

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
```diff
@@ -27,7 +60,121 @@ interface FilterDrawerProps {
...
+          # }
+          # ... on VideoStartEvent {
+          #   position
+          #   source
+          # }
+          # ... on VideoProgressEvent {
+          #   position
+          #   source
+          #   progress
+          # }
```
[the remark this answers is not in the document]
csiyang: Yeah this was an intentional change

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
```diff
@@ -39,8 +186,86 @@ export function FilterDrawer({
...
+      window.URL.revokeObjectURL(url)
+
+      await createEventsExportLog({
+        variables: {
+          input: {
+            journeyId: journey.id,
+            eventsFilter: []
+            // eventsFilter: ['JourneyViewEvent', 'ButtonClickEvent'],
+            // preiodRangeStart: '2025-03-01T00:00:00Z',
+            // periodRangeEnd: '2025-03-25T00:00:00Z'
```
csiyang, line 256: Can we add this back in?

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
```diff
@@ -25,9 +57,124 @@ interface FilterDrawerProps {
...
+            }
+          }
+          email
+          blockId
+          position
+          source
+          progress
+          typename
+          visitorId
+          # May not be needed since we have access to journey?
```
csiyang, line 99: I think the changes im making with the backend improvements makes it so we dont lose any performance from fetching the journey. Can we add it in the query for now, i'll keep an eye on it and trial it in another PR to test the performance.

csiyang (changes requested):

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
levi-zustiak: Well I was saying we already have the journey data passed in so is there a need to retrieve the journey again for the events? It makes sense for queries that don't pertain to this journey but we are retrieving these in the context of the one journey so it could save us a query.

apps/journeys-admin/pages/journeys/[journeyId]/reports/visitors.tsx
Comment (levi-zustiak) [reply apps/journeys-admin/pages/journeys/[journeyId]/reports/visitors.tsx:300]: Yeah I guess I'm not opposed. I just passed it in since we were already fetching the journey

csiyang (approved):

PR #5942, opened 2025-03-31
fix: revert behavior of submit button being set to false during duplication

Ur-imazing:
Reverting change made from previous PR - https://github.com/JesusFilm/core/pull/5916

## Summary by CodeRabbit

- **Refactor**
  - Updated button duplication behavior to allow duplicated submit buttons to retain their enabled state.
- **Tests**
  - Modified test suite to enable the submit button for testing purposes.


apis/api-journeys/src/app/modules/block/block.service.spec.ts
```diff
@@ -131,7 +131,6 @@ describe('BlockService', () => {
     parentBlockId: cardBlock.id,
     startIconId: null,
     endIconId: 'icon',
-    submitEnabled: true,
```
Kneesal, line 134: Leave this in

apis/api-journeys/src/app/modules/block/block.service.spec.ts
```diff
@@ -619,7 +618,6 @@ describe('BlockService', () => {
           id: 'specificButtonId',
           parentBlockId: 'specificCardId',
           endIconId: 'specificIconId',
-          submitEnabled: false,
```
Kneesal, line 622: Leave this test in and change it to true. 
This way we are asserting that the button will be duplicated properly.

Kneesal (changes requested):

Kneesal (approved):

PR #5976, opened 2025-04-01
feat: update ui to differentiate a normal button and a submit button

jianwei1:
- For submit buttons (ie submitEnabled = true or null), sets it's label to "Submit", and it's properties title to "Submit Button Properties"
- For a normal button, sets label to "Button", and properties title to "Button Properties"

### Additional information
 - Set submitenabled: false on the optimistic response of DuplicateBlock, so that the optimistic response returns a normal button when duplicating a submit button


libs/journeys/ui/src/components/Button/Button.tsx
```diff
@@ -258,7 +259,9 @@ export function Button({
             ? editableLabel
             : label !== ''
               ? label
-              : t('Submit')}
+              : submitEnabled
```
edmonday, line 262: Can you move this out so it's a little bit more readable?

edmonday (changes requested):

edmonday (approved):

PR #6055, opened 2025-04-06
feat: journey event loading bar front end

levi-zustiak:


apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.tsx
```diff
@@ -1,10 +1,12 @@
+import LoadingButton from '@mui/lab/LoadingButton'
 import Box from '@mui/material/Box'
 import Button from '@mui/material/Button'
 import Checkbox from '@mui/material/Checkbox'
 import Divider from '@mui/material/Divider'
 import FormControlLabel from '@mui/material/FormControlLabel'
 import FormGroup from '@mui/material/FormGroup'
 import IconButton from '@mui/material/IconButton'
+import LinearProgress from '@mui/material/LinearProgress'
```
csiyang, line 9: remove

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.spec.tsx
```diff
@@ -199,11 +205,17 @@ describe('FilterDrawer', () => {
...
+      expect(screen.queryByRole('progressbar')).not.toBeInTheDocument()
+
       await user.click(screen.getByRole('button', { name: 'Export Data' }))
 
       await waitFor(() => {
-        expect(getJourneyEventsMock.result).toHaveBeenCalled()
+        expect(mockGetJourneyEventsCountQuery.result).toHaveBeenCalled()
       })
+
+      expect(screen.getByRole('progressbar')).toBeInTheDocument()
```
csiyang, line 216: does this still work?

csiyang (changes requested):

apps/journeys-admin/src/components/JourneyVisitorsList/FilterDrawer/FilterDrawer.spec.tsx
levi-zustiak: Yessir. The loading button uses the CircularProgress component with role `progressbar`

csiyang (approved):

csiyang (dismissed):

apps/journeys-admin/src/libs/useJourneyEventsExport/useJourneyEventsExport.tsx
```diff
@@ -116,16 +123,21 @@ interface JourneyEvent
   email?: string | null
 }
 
-export function useJourneyEventsExport(): {
+export function useJourneyEventsExport(): [
```
csiyang, line 126: oh wait, just saw this, how come you changed this into an array?

levi-zustiak: I changed it to an array since I added the state values in the return. That way it matches similar action/state hooks like useState or useMutation.

csiyang: Can we keep it as an object? It's just easier to access all the values without requiring the previous ones. We have a mix of both arrays and objects, but I think objects suits a bit better here.

csiyang (approved):

csiyang (approved):

PR #6058, opened 2025-04-07
fix: remove succesful submission notification

Ur-imazing:


libs/journeys/ui/src/components/Card/Card.spec.tsx
```diff
@@ -786,61 +786,5 @@ describe('CardBlock', () => {
...
-      expect(
-        mockTextResponseSubmissionEventCreate.result
-      ).not.toHaveBeenCalled()
-    )
-
-    await waitFor(() =>
-      expect(
-        queryByText('Thank you for your response!')
-      ).not.toBeInTheDocument()
-    )
```
Kneesal, line 844: leave test in. It should still handle empty submissions, just not show the snackbar

Kneesal (changes requested):

PR #6443, opened 2025-04-28
feat: create mutation to do the translation

mikeallisonJS:


apis/api-journeys-modern/src/schema/aiTranslateJourney/aiTranslateJourney.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface AiTranslateJourneyJob {
+  userId: string
+  journeyId: string
+  name: string
+  textLanguage: string
+  videoLanguage: string
+}
+
+builder.mutationField('aiTranslateJourneyCreate', (t) =>
```
edmonday, line 17:
```suggestion
builder.mutationField('journeyTranslateCreate', (t) =>
```

something like this, @tataihono mentioned would should try to namespace well. So keeping all Journey related mutations together.

edmonday (changes requested):

apis/api-journeys-modern/src/schema/journeyAiTranslate/journeyAiTranslate.ts
```diff
@@ -0,0 +1,40 @@
...
+    input: {
+      journeyId: t.input.id({ required: true }),
+      name: t.input.string({ required: true }),
+      textLanguageId: t.input.id({ required: true }),
+      videoLanguageId: t.input.id({ required: false })
+    },
+    type: 'ID',
+    nullable: false,
+    resolve: async (_root, { input }, { user }) => {
+      const job = (await queue.add('api-media-transcode-video', {
```
edmonday, line 28: this should be named something related to translating journey

edmonday (changes requested):

edmonday (approved):

PR #6614, opened 2025-05-12
feat: bible quotes

levi-zustiak:


apps/watch/src/libs/videoContentFields.ts
```diff
@@ -19,6 +19,17 @@ export const VIDEO_CONTENT_FIELDS = gql`
     studyQuestions(languageId: $languageId, primary: true) {
       value
     }
+    bibleCitations {
```
csiyang, line 22: Does bible citations have translations? Can you add this if it has it?
```suggestion
    bibleCitations(languageId: $languageId, primary: true) {
```

apps/watch/src/components/NewContentPage/BibleCitations/BibleCitations.tsx
```diff
@@ -0,0 +1,134 @@
...
+import Button from '@mui/material/Button'
+import Stack from '@mui/material/Stack'
+import { alpha, styled } from '@mui/material/styles'
+import SvgIcon from '@mui/material/SvgIcon'
+import Typography from '@mui/material/Typography'
+import { ReactElement } from 'react'
+import { A11y, FreeMode, Mousewheel } from 'swiper/modules'
+import { Swiper, SwiperSlide } from 'swiper/react'
+
+import { VideoContentFields_bibleCitations } from '../../../../__generated__/VideoContentFields'
```
csiyang, line 11: We usually type cast these so its easier to read
```suggestion
import { VideoContentFields_bibleCitations as BibleCitations} from '../../../../__generated__/VideoContentFields'
```

csiyang (changes requested):

apps/watch/src/libs/videoContentFields.ts
levi-zustiak: They do not

csiyang (approved):

PR #6670, opened 2025-05-15
feat(watch): download button

levi-zustiak:


apps/watch/src/components/NewContentPage/NewContentPage.tsx
```diff
@@ -164,26 +181,28 @@ export function NewContentPage(): ReactElement {
...
-                      color: 'common.white'
-                    },
-                    transition: 'colors 0.2s'
-                  }}
-                >
-                  {t('Share')}
-                </Button>
+                <Stack direction="row" spacing={2}>
+                  <Button
+                    size="xsmall"
```
csiyang, line 186: Does this exist? The smallest size from the docs says "small"

levi-zustiak: It's defined in the components theme definition

csiyang (approved):

PR #6720, opened 2025-05-21
fix: add from template field

Kneesal:


apis/api-journeys/src/app/modules/journey/journey.graphql
```diff
@@ -223,7 +224,7 @@ input JourneyTemplateInput {
 
 extend type Mutation {
   journeyCreate(input: JourneyCreateInput!, teamId: ID!): Journey!
-  journeyDuplicate(id: ID!, teamId: ID!): Journey!
+  journeyDuplicate(id: ID!, teamId: ID!, fromTemplate: Boolean): Journey!
```
csiyang, line 227: Is this a string or boolean?

csiyang (changes requested):

apis/api-journeys/src/app/modules/journey/journey.graphql
Kneesal: boolean

PR #6812, opened 2025-06-13
fix: remove duplicate name label from visitor card details

Ur-imazing:
## Summary by CodeRabbit

- **Refactor**
  - Visitor names are no longer displayed as a separate detail in visitor cards.
- **Tests**
  - Updated test cases to remove assertions and props related to visitor names.


thomastayler (dismissed):

apps/journeys-admin/src/components/JourneyVisitorsList/VisitorCard/VisitorCardDetails/VisitorCardDetails.tsx
```diff
@@ -66,11 +64,8 @@ export function VisitorCardDetails({
         </>
       ) : (
         <>
-          {name != null && (
-            <DetailsRow label={t('Name')} value={name} loading={loading} />
-          )}
-
           {eventsToRender.map((event) => {
+            if (event.label === 'Name') return null
```
thomastayler, line 68: Can this we removed rather than return null if there is a name? It appears to completely remove the name response from the details

Ur-imazing: Good idea. UX also agrees. Let's do that!

thomastayler (approved):

PR #7184, opened 2025-07-13
Jacobusbrink/eng 3055 update docs

jaco-brink:


jaco-brink (commented):
@tataihono Not 100% on the branch naming convention change I have here.
- Is the approach of using Linear to generate branch name desired over the current format described in docs?
- Does the proposed description for `feature/` branch format line up with how/when we use the convention?

apps/docs/docs/04-engineering-practices/01-workflow/index.md
```diff
@@ -28,11 +31,32 @@ Your branch should take the following format: `YY-CY-FN-type-summary` where:
...
+This branch name can be conveniently generated directly from within a Linear ticket's "Properties" pane by clicking the `Copy git branch name` button.
+
+![button - Copy git branch name](../01-workflow/branch-name-gen.png)
+
+For example `johndoe/eng-3055-update-docs`.
+
+
+**Old/Stale Branches**
+
+Periodically old branches not being worked on will be deleted.
```
jaco-brink, line 52: Wonder if we should remove this considering we don't do this?

PR #7197, opened 2025-07-15
bug: Add default email greeting text when recipient missing

jaco-brink:


jaco-brink (commented):
No Account
`Hi ,` 

Existing Account
`Hi Jaco,`

Proposed change for No Account
`Hi there,`

Notes:
* The only usage of recipient?.firstName is for display purposes in the greeting, so there are no side effects.
* The rest of the component already handles missing recipient data gracefully (e.g., fallback avatar).
* This change will not break any existing functionality, as it only affects the display string when firstName is missing or empty.

Confirmed with storybook with `recipient.fullName == ""

## TeamInviteNoAccountEmail
image


## JourneySharedNoAccountEmail
image

## EmailVerifyEmail
image

edmonday (approved):

PR #7327, opened 2025-08-03
refactor: rename parent parents

mikeallisonJS:
## Summary by CodeRabbit

* **Refactor**
  * Updated the naming of the "parent" field to "parents" in video-related features for improved clarity. This change ensures consistent naming across video relations without impacting functionality.


apis/api-media/src/schema/video/video.ts
```diff
@@ -421,12 +421,12 @@ builder.prismaObjectField(Video, 'parents', (t) =>
         }
       }
       return {
-        parent: {
+        parents: {
           where: whereCondition
         }
       }
     },
-    resolve: (parent: any) => parent.parent
```
tataihono, line 429: this should have the correct type. I think its the child right? so it should be video.parents

tataihono (changes requested):

PR #7451, opened 2025-08-14
fix: reorder locale resolution priorities in middleware

lumberman:
## Summary by CodeRabbit

- **Bug Fixes**
  - Locale is now determined exclusively by the page URL; bookmarked/shared URLs always open in the URL-specified language.
  - Rewrites occur only when a non-default locale is present in the path; default-locale requests are unchanged.

- **Behavior Change**
  - Updating site language no longer sets a site-language cookie; audio/subtitle preferences still persist.

- **Tests / UI**
  - Tests updated to reflect path-only locale behavior; a site-language selector UI was removed from the language dialog and related tests adjusted.

### DIFF
--- a/apps/watch/middleware.ts
+++ b/apps/watch/middleware.ts
@@ -88,13 +88,13 @@ function getBrowserLanguage(req: NextRequest): string {
 }
 
 function getLocale(req: NextRequest): string {
-  // Priority 1: Cookie
-  const cookieLocale = req.cookies.get('NEXT_LOCALE')?.value?.split('---')[1]
-  if (cookieLocale != null) return cookieLocale
-
-  // Priority 2: URL Path
+  // Priority 1: URL Path
   const pathLocale = getLocaleFromPath(req.nextUrl.pathname)
   if (pathLocale != null) return pathLocale
+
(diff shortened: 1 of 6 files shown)


apps/watch/middleware.ts
```diff
@@ -1,57 +1,11 @@
...
-      ? languageCode
-      : DEFAULT_LOCALE
-}
+import { DEFAULT_LOCALE, LANGUAGE_MAPPINGS } from './src/libs/localeMapping'
 
 function getLocaleFromPath(pathname: string): string | undefined {
   const parts = pathname.split('/').filter(Boolean)
   const pathParts = parts.slice(1)
-  const lastPart = pathParts[pathParts.length - 1]
+  const lastPart = pathParts[pathParts.length - 1]?.split('?')[0]
```
csiyang, line 8: This lets slugs with more things at the end of the url be handled correctly.
eg `/watch/jesus.html/french.html?params=value`

lumberman:
@tataihono, we don't want to kill language switch completely.

tataihono:
The site language can't be changed with your changes so we removed the switcher.

PR #7585, opened 2025-09-03
chore: multi step form styling refactor

Ur-imazing:


apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -39,6 +39,7 @@ interface PageWrapperProps {
   user?: User
   initialState?: Partial<PageState>
   background?: string
+  backgroundOverride?: string
```
Kneesal, line 42: revert the changes to this file, you can just use the background prop to pass in the linear gradient/

apps/journeys-admin/pages/templates/[journeyId]/customize.tsx
```diff
@@ -25,14 +32,21 @@ function CustomizePage() {
...
-      <JourneyProvider
-        value={{
-          journey: data?.journey,
-          variant: 'default'
-        }}
+      <PageWrapper
+        user={user}
+        showMainHeader={false}
+        mainBodyPadding={false}
+        backgroundOverride="linear-gradient(to bottom, #1f2c430f, #2568994d)"
```
Comment (Kneesal) [inline apps/journeys-admin/pages/templates/[journeyId]/customize.tsx:39]: instead of passing the styles in, we can just nest
```suggestion
        background="linear-gradient(to bottom, #1f2c430f, #2568994d)"
```

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/MultiStepForm.tsx
```diff
@@ -49,11 +49,12 @@ export function MultiStepForm(): ReactElement {
 
   return (
     <Container
-      maxWidth="lg"
+      maxWidth="sm"
```
Kneesal, line 52:
```suggestion
      maxWidth="lg"
```

Kneesal (changes requested):

apps/journeys-admin/pages/templates/[journeyId]/customize.tsx
Comment (Ur-imazing) [reply apps/journeys-admin/pages/templates/[journeyId]/customize.tsx:39]: Using the background prop of PageWrapper only sets the wrapper's stack backgroundColor, which is incompatible with linear-gradients. If we want a gradient color, we need to set the stack's background prop instead.

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -102,7 +104,9 @@ export function PageWrapper({
             flexGrow={1}
             direction={{ xs: 'column', md: 'row' }}
             sx={{
-              backgroundColor: background ?? 'background.default',
+              ...(backgroundOverride != null
+                ? { background: backgroundOverride }
+                : { backgroundColor: background ?? 'background.default' }),
```
Kneesal, line 109:
```suggestion
              background: background,
              backgroundColor: backgroundColor ?? 'background.default'
```

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
```diff
@@ -58,7 +59,8 @@ export function PageWrapper({
   customSidePanel,
   user,
   initialState,
-  background
+  background,
+  backgroundOverride
```
Kneesal, line 63:
```suggestion
  backgroundColor
```

apps/journeys-admin/src/components/PageWrapper/PageWrapper.tsx
Kneesal: You will prolly need to update this prop for all the pages that use this

Kneesal (approved):

PR #7625, opened 2025-09-09
fix: language screen language filter

Kneesal:


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/MultiStepForm.tsx
```diff
@@ -52,9 +52,9 @@ export function MultiStepForm(): ReactElement {
       sx={{
         width: '100%',
         height: '100%',
-        maxHeight: { xs: '100%', sm: '852px' },
+        maxHeight: { xs: '100%', sm: '900px' },
```
Ur-imazing, line 55: Hannah specifically asked for the container height to grow to 852px. Is there a particular reason why it's been changed to 900px?

PR #7690, opened 2025-09-16
feat: add special case for done screen stepper node

Ur-imazing:


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/MultiStepForm.tsx
```diff
@@ -122,11 +122,12 @@ export function MultiStepForm(): ReactElement {
             {t('Edit Manually')}
           </Button>
         </NextLink>
-
-        <ProgressStepper
-          activeStepNumber={screens.indexOf(activeScreen)}
-          totalSteps={totalSteps}
-        />
+        {totalSteps > 3 && (
```
Kneesal, line 125: use screens assertion rather than total steps. We could add/remove screens in the future causing this to fail so assert the render logic by screens. you can get the `hasEditableText` and `hasCustomizableLinks` variables from `getCustomizeFlowConfig` function


```suggestion
        {hasEditableText && hasCustomizableLinks && (
```

PR #7700, opened 2025-09-17
feat: use apologist for suggestions

edmonday:
This PR makes an update to use apologist over gemini.

Fixed a few lint errors while I was at it.


jaco-brink (approved):
Nice, only one nit 🚀

PR #7704, opened 2025-09-18
fix: placement of ai chat button

edmonday:
This PR removes the ai chat button from the list of other chat buttons. 

As it is undecided as of yet where this button should live, this ai chat button will only appear if there are no chat widgets.


jaco-brink (approved):
Thanks for sorting all the missing showAssistants

PR #7726, opened 2025-09-19
feat: add pre loading of language and suggestions

jaco-brink:


jaco-brink (commented):
once this is in, we can deprecate the /suggestions endpoint.

libs/journeys/ui/src/libs/useJourneyAiContextGenerator/useJourneyAiContextGenerator.ts
```diff
@@ -0,0 +1,177 @@
...
+      )
+      const fullContext = createFullContext(blockContexts, contextResponse)
+      setAiContextData(fullContext)
+    } catch (err) {
+      setError(getErrorMessage(err))
+      console.error('Error processing AI context:', err)
+    } finally {
+      setIsLoading(false)
+    }
+  }, [treeBlocks])
```
edmonday, line 127: Is this refetch logic needed as well?

libs/journeys/ui/src/libs/JourneyAiContextProvider/JourneyAiContextProvider.spec.tsx
```diff
@@ -0,0 +1,254 @@
...
+    },
+    {
+      blockId: 'block-2',
+      contextText: 'Test context 2',
+      language: 'spanish',
+      suggestions: ['sugerencia 1', 'sugerencia 2']
+    }
+  ]
+
+  describe('JourneyAiContextProvider', () => {
```
edmonday, line 51: is there a reason why there is another describe block here? its also the same name

libs/journeys/ui/src/libs/useJourneyAiContextGenerator/useJourneyAiContextGenerator.ts
jaco-brink: haha, no, I thought I removed it already.
Will remove

libs/journeys/ui/src/libs/useJourneyAiContextGenerator/useJourneyAiContextGenerator.ts
```diff
@@ -0,0 +1,177 @@
...
+          setError(getErrorMessage(err))
+          console.error('Error processing AI context:', err)
+        } finally {
+          setIsLoading(false)
+        }
+      }
+
+      void processContext()
+    }
+  }, [treeBlocks])
```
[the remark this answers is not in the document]
jaco-brink: Okay, tested this out and it appears we do need to watch `treeBlocks`

The reason:

`treeBlocks` starts as an empty array and appears to be populated asynchronously. 
Without the watching `treeBlocks`, the hook only runs once with empty data (initial state) and never process the actual blocks when the async population completes.

I was able to remove the "hasChanged" logic as that was found redundant, useEffect effectively already does what it was trying to accomplish.

Kneesal (approved):

PR #7732, opened 2025-09-19
fix: easier language variant creator for publishers

Kneesal:


apps/journeys-admin/src/libs/useGetParentTemplateJourneyLanguages/useGetParentTemplateJourneyLanguages.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+              }
+            ]}
+          >
+            {children}
+          </MockedProvider>
+        )
+      }
+    )
+
+    await act(
```
Ur-imazing, line 41: Why are we using an act here, when we're not updating a component before assertion? Wouldn't the await waitFor() suffice?

PR #7789, opened 2025-09-25
fix: unify customization flow next step button styling

Ur-imazing:
## Summary by CodeRabbit

* **New Features**
  * Added a reusable Next/Done action button used across Links, Social, Text and Language screens and exported for reuse.

* **Style**
  * Standardized button appearance; adjusted typography variants and responsive heading displays.

* **Accessibility**
  * Added aria-label support and stable test identifiers for the new button.

* **API**
  * Several screens now accept an additional navigation handler prop.

* **Tests**
  * Added tests for the new button and updated screen tests to use the new control.

* **Chores**
  * Updated localization keys and removed an unused import.


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/TextScreen/TextScreen.tsx
```diff
@@ -323,7 +323,7 @@ export function TextScreen({
           >
             {t('Next Step')}
           </Typography>
-          <ArrowRightIcon />
+          <ArrowRightIcon sx={{ fontSize: { xs: '24px', sm: '16px' } }} />
```
Kneesal, line 326: Lets make a common compoennt for this button so that styling is always consistent and only has to be changed in one place. 

Let it take props: "label", that way we can use it for the Done button too in the file above (SocialScreen.tsx)

Ur-imazing: Done! Thanks for the suggestion bro.

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/LanguageScreen/LanguageScreen.tsx
```diff
@@ -226,34 +221,12 @@ export function LanguageScreen({
                   {t('Select a team')}
                 </Typography>
                 {isSignedIn && <JourneyCustomizeTeamSelect />}
-                <Button
-                  variant="contained"
-                  color="secondary"
-                  disabled={loading}
+                <CustomizeFlowNextButton
+                  label={t('Next')}
```
Kneesal, line 225:
```suggestion
```
default text is Next, you can remove this from everywhere that has this prop

Ur-imazing: But the default text is not wrapped in a t-function. Doesn't this mean that the button title won't be translated, if we rely on the default text?

Kneesal (approved):

Kneesal (dismissed):

Kneesal (approved):

PR #7868, opened 2025-09-29
fix: allow @example.com emails to verify with EXAMPLE_EMAIL_TOKEN

kiran-redhat:
- Add special handling for @example.com emails in validateEmail function
- Bypass job-based validation for test users with EXAMPLE_EMAIL_TOKEN
- Fixes email verification for additional Playwright test users (playwrightuser2-5@example.com)
- Resolves ENG-3253

## Summary by CodeRabbit

* **New Features**
  * Added test email validation support for @example.com domain using environment-based bypass mechanism.

* **Tests**
  * Extended end-to-end test infrastructure with additional test account credentials for broader coverage.
  * Improved email verification test scenarios with enhanced per-test environment isolation.


apis/api-users/src/schema/user/validateEmail.ts
```diff
@@ -9,6 +9,54 @@ export async function validateEmail(
   userId: string,
   token: string
 ): Promise<boolean> {
+  // Get user email to check if it's a test user
+  const user = await prisma.user.findUnique({
```
tataihono: don't call the db again. take in user email as an arg for this function.

.github/workflows/e2e-tests.yml
tataihono: these changes are not used anywhere in this PR is this a mistake?

csiyang: These are already being used here https://github.com/JesusFilm/core/blob/main/apps/journeys-admin-e2e/src/framework/helpers.ts

apps/journeys-admin/next-env.d.ts
tataihono: revert

tataihono (changes requested):

PR #8058, opened 2025-10-23
feat: stub launch darkly client

csiyang:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 1-second initialization timeout to prevent app hangs during feature-flag service startup.
  * Added a safe fallback that logs a warning and uses benign defaults when feature-flag initialization fails or times out.

* **Chores**
  * Upgraded the feature-flag SDK to a newer scoped release.

### DIFF
--- a/apis/api-journeys-modern/src/schema/action/blockUpdateAction.mutation.ts
+++ b/apis/api-journeys-modern/src/schema/action/blockUpdateAction.mutation.ts
@@ -13,7 +13,7 @@ import { canBlockHaveAction } from './canBlockHaveAction'
 import { BlockUpdateActionInput } from './inputs'
 
 const emailSchema = z.object({
-  email: z.string().email()
+  email: z.email()
 })
 
 const linkActionInputSchema = z.object({
@@ -24,7 +24,7 @@ const linkActionInputSchema = z.object({
 
 const emailActionInputSchema = z.object({
   gtmEventName: z.string().nullish(),
-  email: z.string().email()
+  email: z.email()
 })
 
 const navigateToBlockActionInputSchema = z.object({
--- a/apis/api-journeys-modern/src/schema/action/emailAction/blockUpdateEmailAction.mutation.ts
+++ b/apis/api-journeys-modern/src/schema/action/emailAction/blockUpdateEmailAction.mutation.ts
@@ -12,7 +12,7 @@ import { EmailActionRef } from './emailAction'
 import { EmailActionInput } from './inputs'
 
 const emailSchema = z.object({
-  email: z.string().email()
+  email: z.email()
 })
 
 builder.mutationField('blockUpdateEmailAction', (t) =>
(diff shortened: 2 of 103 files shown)


libs/shared/ui/src/libs/getLaunchDarklyClient/getLaunchDarklyClient.ts
```diff
@@ -2,18 +2,54 @@ import { LDClient, LDUser, init } from 'launchdarkly-node-server-sdk'
...
+}
+
 /**
  * This requires the LAUNCH_DARKLY_SDK_KEY environment variable to be set.
- * @returns a LaunchDarkly server-side client as a singleton.
+ * @returns a LaunchDarkly server-side client as a singleton, or a stub client if initialization fails or times out.
  */
 export async function getLaunchDarklyClient(user?: LDUser): Promise<LDClient> {
   if (launchDarklyClient == null) {
     launchDarklyClient = init(process.env.LAUNCH_DARKLY_SDK_KEY ?? '')
```
csiyang: This stub doesnt quite work, because this line seems to continue trying to connect even if the error state is entered below

libs/shared/ui/src/libs/getLaunchDarklyClient/getLaunchDarklyClient.ts
```diff
@@ -1,19 +1,47 @@
-import { LDClient, LDUser, init } from 'launchdarkly-node-server-sdk'
+import { LDClient, LDUser, init } from '@launchdarkly/node-server-sdk'
 
 let launchDarklyClient: LDClient
 
+const createStubClient = (): LDClient => {
```
tataihono: why does this need to be a function?

tataihono (changes requested):

libs/shared/ui/src/libs/getLaunchDarklyClient/getLaunchDarklyClient.ts
tataihono: its inefficient. like why make a function when you can just us and object.

libs/shared/ui/src/libs/getLaunchDarklyClient/getLaunchDarklyClient.ts
```diff
@@ -2,18 +2,54 @@ import { LDClient, LDUser, init } from 'launchdarkly-node-server-sdk'
...
+      new Promise((_, reject) =>
+        setTimeout(
+          () => reject(new Error('LaunchDarkly initialization timeout')),
+          1000
+        )
+      )
+    ])
 
-  if (user != null) launchDarklyClient.identify(user)
+    if (user != null) launchDarklyClient.identify(user)
```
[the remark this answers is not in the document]
tataihono: this is probably correct.

tataihono (changes requested):

PR #8422, opened 2025-11-27
feat: template site create

csiyang:
## Summary by CodeRabbit

* **New Features**
  * Track whether a journey template has an associated site.
  * Expanded analytics with 10 new capture event types (prayer requests, gospel start/complete, Christ decisions, RSVP, special video start/complete, and 3 custom captures).
  * Option to include unconnected cards when filtering journey events.
  * Option to disable creation of shared links when creating sites.

* **Tests**
  * Added coverage for template-site creation flows and shared-links behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


apis/api-journeys/src/app/modules/plausible/plausible.service.ts
```diff
@@ -204,6 +217,26 @@ export class PlausibleService implements OnModuleInit {
...
+    console.log('creating template sites...')
+    const chunkedTemplateIds = chunk(
+      (
+        await this.prismaService.journey.findMany({
+          where: { template: true },
+          select: { id: true }
+        })
+      ).map(({ id }) => id),
+      5
+    )
```
csiyang: @mikeallisonJS This PR still needs to be updated for your new changes, but Im not sure how how we should check when templateSiteCreate gets called for a journey.

- journeySite and and teamSite's are called based of if they have a token (which is added when a site is created),
- but in our case we dont want to create token (since it's purpose is to give public access to the dashboard for a site)

Do you think we should just a new prop thats a boolean to tell us if a templateSite has been created for the journey yet?

PR #8424, opened 2025-11-27
feat: hide global template-specific elements for local templates

Ur-imazing:


apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/MetadataTabPanel/MetadataTabPanel.tsx
```diff
@@ -65,36 +65,42 @@ export function MetadataTabPanel(): ReactElement {
...
-            hiddenLabel
-            variant="filled"
-            value={format(parseISO(journey?.publishedAt as string), 'P')}
-            InputProps={{
-              readOnly: true
-            }}
-          />
-        )}
-      </Stack>
+      {journey?.team?.id === 'jfp-team' && (
```
jaco-brink: Similar to @AboutTabPanel comment above we want the parent the orchestrate.
But in this case is more complex as we only want to show/hide a part of MetadataTabPanel opposed to complete AboutTabPanel.

Suggestion:
- Pass a `shouldShowFeatured` prop into this component and drive the partial show/hide using it. Use this variable name indicates that "child" is being told what do instead of decide for it self (e.g. if I'm in team jfp don't show featured)
- In the Parent > `const shouldShowFeature = isGlobalTemplate `

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -203,23 +204,24 @@ export function TemplateSettingsDialog({
...
-                />
-                <Tab label={t('About')} {...tabA11yProps('about', 2)} />
+                {isGlobalTemplate && (
+                  <Tab label={t('Categories')} {...tabA11yProps('categories', 1)} />
+                )}
+                <Tab label={t('About')} {...tabA11yProps('about', isGlobalTemplate ? 2 : 1)} />
               </Tabs>
               <TabPanel name="metadata" value={tab} index={0}>
                 <Stack sx={{ pt: 6 }} gap={5}>
                   <MetadataTabPanel />
```
jaco-brink: As per comment about, pass `shouldShowFeature` to orchestrate child behaviour from parent

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -203,23 +204,24 @@ export function TemplateSettingsDialog({
...
-              </TabPanel>
-              <TabPanel name="about" value={tab} index={2}>
+              {isGlobalTemplate && (
+                <TabPanel name="categories" value={tab} index={1}>
+                  <Stack sx={{ pt: 6 }} gap={5}>
+                    <CategoriesTabPanel />
+                  </Stack>
+                </TabPanel>
+              )}
+              <TabPanel name="about" value={tab} index={isGlobalTemplate ? 2 : 1}>
```
jaco-brink: As per comment above, conditionally render AboutTabPanel OR the TabPanel.

I think TabPanel is better to avoid accidental tab selection when it's a local template, however you might have to confirm if tab selection still works fine when doing this.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/AboutTabPanel/AboutTabPanel.tsx
```diff
@@ -45,29 +47,33 @@ export function AboutTabPanel(): ReactElement {
...
-        InputProps={{
-          endAdornment: <LinkAngled sx={{ color: 'secondary.light' }} />
-        }}
-      />
-      <StrategySection
-        strategySlug={values.strategySlug}
-        variant="placeholder"
-        isError={Boolean(errors?.strategySlug)}
-      />
+      {journey?.team?.id === 'jfp-team' && (
```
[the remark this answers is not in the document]
jaco-brink: As discussed - I over looked the fact this was not the entire component, rather a subsection that needed conditional rendering.
Instead we should conditionally render in AboutTabPanel using a prop (representing GlobalTemplate bool) passed in from TemplateSettingDialog representing

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
jaco-brink: As discussed, Ignore previous suggestion.
Instead treat similar to MetadataTabPanel comment (pass prop instead conditionally rendering)

PR #8465, opened 2025-12-05
refactor: feature plausible

mikeallisonJS:


apis/api-journeys-modern/src/schema/plausible/templateFamilyStatsBreakdown/templateFamilyStatsBreakdown.query.spec.ts
```diff
@@ -166,7 +166,7 @@ describe('templateFamilyStatsBreakdown', () => {
       expect.stringContaining('/api/v1/stats/breakdown'),
       expect.objectContaining({
         params: expect.objectContaining({
-          site_id: 'api-journeys-template-template-journey-id',
+          site_id: 'template-site',
```
csiyang: The original is correct, template-site is only used when I was testing out calls locally, i have a PR up fixing it. We should leave this tests unchanged

PR #8501, opened 2025-12-11
feat: add empty template tab text

jaco-brink:


jaco-brink (commented):
confirmed wrap behaviour with UX, aware of trade offs regarding slight impact to journey tab text wrap appearance

PR #8514, opened 2025-12-15
feat: hide copy to from journey menu for local templates

Ur-imazing:


jaco-brink (approved):
nit: we can add a test for this change

PR #8516, opened 2025-12-15
feat: add ux alterations for local templates

jaco-brink:


Ur-imazing (approved):
Looks good to me! Thanks so much Jaco :D

PR #8538, opened 2025-12-16
feat: add standard margin on menu divider

jaco-brink:
## What?
Looking at this area, all `<Dividers/>` look the same, yet one does not have the same margin as the others.
Note, the standard margin is (4px 0px), where as this one is (0px 0px)

## Why?
When the Divider's preceding sibling is a Box (not a MenuItem), it prevents MUI from automatically 
applying margins that it would apply when following a direct MenuItem.

e.g. `ShareItem` (precedes the divider) wraps its `MenuItem` in a `Box`, so this Divider's preceding sibling is a `Box` (not a `MenuItem`)

## Note
- This fixes the margin issue for both issues listed in the ticket


Ur-imazing (approved):
Absolutely ginormous piece of work, well done! 💪💪

PR #8610, opened 2026-01-19
feat: duplicating selected journey from archived

irenekim21:
## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate detection now includes archived journeys, improving cross-state duplicate identification.
  * Created duplicates now explicitly clear archived/trashed/deleted timestamps and apply a predictable "copy" suffix to slugs/titles.

* **Tests**
  * Added a test ensuring archived journeys duplicate correctly with the expected "copy" suffix and creation payload.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/journeys-admin/pages/templates/[journeyId].tsx
+++ b/apps/journeys-admin/pages/templates/[journeyId].tsx
@@ -31,7 +31,10 @@ function TemplateDetailsPage(): ReactElement {
   const user = useUser()
   const { data } = useJourneyQuery({
     id: router.query.journeyId as string,
-    idType: IdType.databaseId
+    idType: IdType.databaseId,
+    options: {
+      skipRoutingFilter: true
+    }
   })
   const { activeTeam, refetch, query } = useTeam()
 
--- a/apps/journeys-admin/pages/templates/[journeyId]/customize.tsx
+++ b/apps/journeys-admin/pages/templates/[journeyId]/customize.tsx
@@ -26,7 +26,10 @@ function CustomizePage() {
   const user = useUser()
   const { data } = useJourneyQuery({
     id: router.query.journeyId as string,
-    idType: IdType.databaseId
+    idType: IdType.databaseId,
+    options: {
+      skipRoutingFilter: true
+    }
   })
 
   return (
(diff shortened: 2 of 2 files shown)


apis/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -545,25 +544,31 @@ export class JourneyResolver {
       }
     }
 
-    const existingActiveDuplicateJourneys =
-      await this.prismaService.journey.findMany({
+    const baseTitleMatch = journey.title.match(/^(.+?)(?:\s+copy(?:\s+\d+)?)?$/)
+    const baseTitle = baseTitleMatch ? baseTitleMatch[1] : journey.title
+
+    const archiveFilter: Prisma.DateTimeNullableFilter | null =
+      journey.archivedAt == null ? null : { not: null }
```
edmonday: This is not needed. We already have title matching logic in functions above. 

See lines 435 and 425.

apis/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -545,25 +544,31 @@ export class JourneyResolver {
...
         }
-      })
+      }
+    )
     const duplicates = this.getJourneyDuplicateNumbers(
-      existingActiveDuplicateJourneys,
-      journey.title
+      existingDuplicateJourneys,
+      baseTitle
     )
```
edmonday: We can simply just remove `archivedAt` from the filter. This is because we don't actually care about wether or not a journey is archived or not.

It would then look something like this.

```suggestion
    const existingDuplicateJourneys = await this.prismaService.journey.findMany(
      {
where: {
title: {
            contains: baseTitle
},
trashedAt: null,
deletedAt: null,
template: false,
team: { id: teamId }
}
      }
    )
const duplicates = this.getJourneyDuplicateNumbers(
      existingDuplicateJourneys,
      baseTitle
)
```

apis/api-journeys/src/app/modules/journey/journey.resolver.ts
```diff
@@ -610,11 +610,14 @@ export class JourneyResolver {
                   'templateSite'
                 ]),
                 id: duplicateJourneyId,
-                slug,
+                slug: slug,
```
edmonday: change this back to just slug

PR #8672, opened 2026-02-02
feat: language code to app header language switcher

VincentHan12:


apps/journeys-admin/src/components/PageWrapper/AppHeader/AppHeader.tsx
```diff
@@ -42,38 +45,54 @@ export function AppHeader({
...
+              >
+                <LanguageIcon
+                  sx={{ fontSize: '12.36px', color: 'background.paper', mr: 1 }}
+                />
+                <Typography
+                  sx={{
+                    fontSize: '12px',
+                    fontWeight: 600,
+                    fontFamily: 'Montserrat'
+                  }}
```
edmonday: please use the variant prop on Typography. We have theming set for Typography. 

Take a look at: `libs/shared/ui/src/libs/themes/journeysAdmin/tokens/typography.ts`

edmonday (approved):
LGTM

PR #8703, opened 2026-02-09
fix: move blockcustomizationtoggle to top of image and video accordions

jianwei1:
Refactored the code to place the "Needs Customization" toggle to the top of the Image and Video accordions (previously it was at the bottom). Did this by:
- Image:
  - Remove the BlockCustomizationToggle out of Image and BackgroundMediaImage components, and placed it inside ImageSource instead. This now covers both cases.
- Video: 
  - Remove the BlockCustomizationToggle out of Video and BackgroundMediaVideo components, and placed it inside VideoBlockEditor instead. This now covers both cases

## Release Notes

* **Refactor**
  * Reorganized the placement of customization options within the editor interface to streamline the layout and consolidate related settings in dedicated editor panels.


jaco-brink (approved):
LGTM

PR #8708, opened 2026-02-10
fix: editing and saving share dialog triggers hover state

jaycnz:
## Summary by CodeRabbit

* **New Features**
  * Components can now optionally notify parents when share/copy/other dialogs open or close to coordinate UI state.
* **Bug Fixes**
  * Hover overlay no longer flickers or changes while a dialog is open.
  * Menu and dialog interactions consistently respect hover/overlay state when dialogs are active.
* **Tests**
  * Added tests verifying dialog open/close callbacks and hover behavior while dialogs are open.


apps/journeys-admin/src/components/Editor/Toolbar/Items/ShareItem/ShareItem.tsx
```diff
@@ -93,6 +101,10 @@ export function ShareItem({
         }/${journey?.slug}`
       : undefined
 
+  useEffect(() => {
+    setIsDialogOpen?.(Boolean(anchorEl))
+  }, [anchorEl, setIsDialogOpen])
```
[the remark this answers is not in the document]
jaycnz: this situation most likely cannot occur as you would need to close nested dialogs before closing the flat dialog that toggles setIsDialogOpen

apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCard.tsx
```diff
@@ -75,6 +75,7 @@ export function JourneyCard({
   const [isCardHovered, setIsCardHovered] = useState(false)
   const [isImageLoading, setIsImageLoading] = useState(true)
   const [breakdownDialogOpen, setBreakdownDialogOpen] = useState(false)
+  const [isDialogOpen, setIsDialogOpen] = useState(false)
```
csiyang: Just a minor rename so its a bit clearer how these should be used when passed to children
```suggestion
  const [dialogOpen, setDialogOpen] = useState(false)
```

apps/journeys-admin/src/components/Team/CopyToTeamMenuItem/CopyToTeamMenuItem.spec.tsx
```diff
@@ -885,4 +885,39 @@ describe('CopyToTeamMenuItem', () => {
     expect(handleCloseMenu).toHaveBeenCalled()
     expect(getByText('Journey Copied')).toBeInTheDocument()
   })
+
+  it('should call setIsDialogOpen when opening and closing copy to team dialog', async () => {
+    const setIsDialogOpen = jest.fn()
+
+    const { getByRole } = render(
```
csiyang: We've changed pattern for how we write these tests, can you update this, and other tests you touched to use this pattern
```suggestion
    render(
    	....
    )
   ... 
   screen.getByRole(...
   
```

csiyang (changes requested):

apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCardMenu/JourneyCardMenu.spec.tsx
```diff
@@ -300,4 +311,148 @@ describe('JourneyCardMenu', () => {
       expect(queryByText('Create Translated Copy')).not.toBeInTheDocument()
     )
   })
+
+  it('should call setHasOpenDialog when opening and closing edit details dialog', async () => {
+    const setHasOpenDialog = jest.fn()
+
+    const { getByRole, findByRole } = render(
```
csiyang: Make this change for every test you've touched
```suggestion
    render(
```

apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCard.tsx
```diff
@@ -88,6 +89,12 @@ export function JourneyCard({
     }
   }, [duplicatedJourneyId, journey])
 
+  const updateHoverState = (hovered: boolean) => {
+    if (!hasOpenDialog) {
+      setIsCardHovered(hovered)
+    }
+  }
```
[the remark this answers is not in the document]
jaycnz: isCardHovered is set to false when any menu item that opens a dialog is clicked. Instead of using useEffect, I've ensured that setHasOpenDialog(true) is called last, so handleCloseMenu() (which calls updateHoverState(false)) executes first without being blocked.

csiyang (approved):

PR #8715, opened 2026-02-11
feat: video editing section

jianwei1:


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
```diff
@@ -1,23 +1,171 @@
...
+      })
+        .then(() => {
+          showSnackbar(t('File uploaded successfully'), 'success')
+        })
+        .catch(() => {
+          showSnackbar(t('Upload failed. Please try again'), 'error')
+        })
+        .finally(() => {
+          setUpdating(false)
+        })
```
edmonday: I think we don't need to do catch and finally here as the `onUploadError` should catch the error.

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
```diff
@@ -1,23 +1,171 @@
...
+              id: journey.id,
+              idType: IdType.databaseId,
+              options: { skipRoutingFilter: true }
+            }
+          }
+        ]
+      })
+        .then(() => {
+          showSnackbar(t('File uploaded successfully'), 'success')
+        })
```
edmonday: Instead of chaining with `then` when can move this inside the `onUploadComplete`.

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
```diff
@@ -1,23 +1,171 @@
...
+        <Typography
+          variant="subtitle2"
+          fontSize={14}
+          sx={{ color: 'secondary.main' }}
+        >
+          {t('Upload file')}
+        </Typography>
+      </Button>
+    </Box>
+  )
```
edmonday: I would recommend defining this as a functional component at the top of the file instead. This will help with readability . This also means uploadButton is not created on every render of VideosSection.

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
```diff
@@ -1,23 +1,171 @@
...
+        >
+          <CircularProgress size={24} />
+        </Box>
+      ) : (
+        videoBlock != null && <VideoPreviewPlayer videoBlock={videoBlock} />
+      )}
+      {videoBlock != null &&
+        !loading &&
+        videoBlockDisplayTitle !== '' &&
+        videoTitle(videoBlockDisplayTitle)}
```
edmonday:
```suggestion
       <VideoTitle title={videoBlockDisplayTitle} />
```

This is more idiomatic to React.

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
jaco-brink: Thanks for the comment, bit of a complex flow here.

Why the catch and finally is here is to prevent a stuck loading state.
Basically we are watching the videoBlockUpdate mutation for faults, this runs after uploads succeed.

`onUploadError` only covers errors inside the upload hook and is unaware of the this mutation failures. 

If the mutation fails without catch, the error is silently swallowed and without  the finally clause, updating stays true and the UI gets stuck on a spinner forever.

To validate my understanding I did this:
- commented out the catch and finally
- moved the `setUpdating(false)` call into the `onUploadError`
- confirmed the test cases fails _shows error snackbar and clears loading when videoBlockUpdate rejects_

Hope this clarifies

edmonday (approved):

PR #8749, opened 2026-02-18
fix: use userId instead of id in findOrFetchUser Prisma where clause

up-tandem:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed user update logic so account records are correctly identified and updated, preventing failed or mis-applied updates for existing users (including cases where email verification is unset).
  * Updated tests to reflect and validate the corrected update behavior.


edmonday (approved):
LGTM

PR #8751, opened 2026-02-18
feat: add video background upload processing

jaco-brink:


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/TemplateVideoUploadProvider/useUploadTaskMap.ts
```diff
@@ -0,0 +1,91 @@
...
+  >
+  updateTask: (
+    videoBlockId: string,
+    updates: Partial<UploadTaskInternal>
+  ) => void
+  removeTask: (videoBlockId: string) => void
+  getUploadStatus: (videoBlockId: string) => VideoUploadState | null
+  hasActiveUploads: boolean
+  uploadInstancesRef: React.MutableRefObject<Map<string, { abort: () => void }>>
+  activeBlocksRef: React.MutableRefObject<Set<string>>
```
edmonday:
```suggestion
  uploadInstancesRef: RefObject<Map<string, { abort: () => void }>>
  activeBlocksRef: RefObject<Set<string>>
```

`MutableRefObject` has been deprecated and also import RefObject from react

apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/TemplateVideoUploadProvider/useMuxVideoProcessing.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+    act(() => {
+      result.current.startPolling('video-block-1', 'mux-video-id')
+    })
+
+    await waitFor(
+      () => {
+        expect(result.current.getUploadStatus('video-block-1')).toBeNull()
+      },
+      { timeout: 5000 }
```
edmonday: Instead of using a real `setTimeout` in this test file, for anywhere where you need to wait some time, can you use `jest.useFakeTimers()` so that you can advance the time.

We want to try to not make the CI runs any longer than it needs to be.

PR #8794, opened 2026-03-02
feat: add customizable field to journey 

jianwei1:
## Summary by CodeRabbit

* **New Features**
  * Added a `customizable` boolean on journeys (exposed for create/update) and image blocks; defaults to false.
  * `customizable` is preserved or set appropriately when duplicating journeys and honored across journey-related operations and tests.


apis/api-journeys/src/app/modules/journey/journey.resolver.spec.ts
```diff
@@ -2195,6 +2206,54 @@ describe('JourneyResolver', () => {
         resolver.journeyDuplicate(ability, 'journeyId', 'userId', 'teamId')
       ).rejects.toThrow('journey not found')
     })
+
+    it('carries customizable: true forward when duplicating a local template', async () => {
```
edmonday: NIT: do you think we need to add a test for global template?

edmonday (changes requested):

apis/api-journeys/src/app/modules/journey/journey.resolver.spec.ts
jianwei1: added

edmonday (dismissed):

PR #8833, opened 2026-03-11
feat: apply QA slug filter to Related Templates on template detail page

jaco-brink:
## Summary
- Exports `QA_ONLY_TEMPLATE_SLUG_PREFIX` from `TemplateSections` so it can be shared
- Applies the same QA slug filter to the **Related Templates** carousel in `TemplateView` (`/templates/[journeyId]`), matching the existing filter on the `/templates` gallery page
- Adds two tests in `TemplateView.spec.tsx` covering the filter: one asserting QA templates are hidden, one asserting normal templates are shown

## Test plan
- [ ] QA-prefixed templates do not appear in the Related Templates carousel on the template detail page
- [ ] Normal templates continue to appear in the Related Templates carousel
- [ ] Existing `TemplateSections` QA filter tests still pass
- [ ] All new `TemplateView` QA filter tests pass

## Summary by CodeRabbit

* **New Features**
  * QA-only test templates are now excluded from Related Templates recommendations so users see only production-ready journey templates.

* **Tests**
  * Added tests verifying QA templates are filtered out and regular templates remain visible to improve recommendation reliability.

* **Chores**
  * Made the QA template identifier accessible across components for consistent QA filtering and reuse.

### DIFF
--- a/libs/journeys/ui/src/components/TemplateSections/TemplateSections.tsx
+++ b/libs/journeys/ui/src/components/TemplateSections/TemplateSections.tsx
@@ -13,10 +13,6 @@ import { GetJourneys_journeys as Journey } from '../../libs/useJourneysQuery/__g
 import { ContentCarousel } from '../ContentCarousel'
 import { TemplateGalleryCard } from '../TemplateGalleryCard'
 
-// QA-only templates use this slug prefix. They are intentionally always hidden
-// from the public gallery so real users never see them. QA testers access them
-// directly via their template URL. Remove this constant and the filter below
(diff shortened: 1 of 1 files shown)


jaco-brink:
## Review feedback addressed (7ccdf06dd)

**Fixed:**
- Hardened the `should not show a QA template in Related Templates` test to also assert that the non-QA card (`relatedJourneyBase.title`) is still present in the same render. Switched from `waitFor(getByRole('Related Templates'))` to `await findByRole('Related Template')` so the assertion gates on the surviving non-QA card being rendered — a regressed filter that drops all cards would now correctly fail the test.

PR #8882, opened 2026-03-19
fix: team selection race condition in template customization (NES-1465)

up-tandem:
## Summary
- **Fixed two root causes** of journeys being duplicated to the wrong team during template customization:
  1. Formik's `enableReinitialize` silently reset the user's team selection whenever async data re-rendered the component
  2. `handleJourneyDuplication` ignored the form's `teamSelect` value and always used `lastActiveTeamId` from the server
- Added loading gate so the form only renders once both team and journey data are available
- Validates `lastActiveTeamId` against available teams (falls back to first team if stale/deleted)
- Persists selected team as `lastActiveTeamId` after successful duplication (fire-and-forget)
- Removed `setActiveTeam` side effect from team dropdown `onChange` to prevent mid-flow re-renders
- Added try/catch around duplication to prevent frozen loading state on network errors

## Testing
- 4 new tests: stale team fallback, `updateLastActiveTeamId` after duplication, network error handling, user-selected team honored over `lastActiveTeamId`
- Updated existing tests to account for loading gate (async data must resolve before form renders)
- All 15 LanguageScreen tests pass
- All 3 JourneyCustomizeTeamSelect tests pass

- **What to monitor/search**
  - Logs: `journeyDuplicate` mutation errors in Datadog/logging
  - Metrics: Journey duplication success rate, `journeyProfileUpdate` mutation call rate
- **Expected healthy behavior**
  - Users report journeys landing in the correct team
  - No increase in duplication errors
- **Failure signal(s) / rollback trigger**
  - Users report journeys still going to wrong team → revert this PR
  - Increase in "Failed to duplicate journey" snackbar errors → investigate
- **Validation window & owner**
  - Window: 48 hours post-deploy
  - Owner: Ekkasit (Up)

## Linear Issue
[NES-1465](https://linear.app/jesus-film-project/issue/NES-1465/email-template-bugs-with-email-field-and-team-selection)

---

🤖 Generated with Claude Opus 4.6 (1M context) via [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Shows a loading spinner while teams/journey load and displays "Failed to load teams. Please refresh the page and try again." on error.
  * Falls back to the first available team when your last-active team no longer exists.

* **Bug Fixes**
  * Duplication uses the team you explicitly select and updates your last-active team.
  * Clearer duplication error handling: shows failure message and re-enables Next after network failures.

* **Tests**
  * Expanded coverage for duplication flow, team-selection fallback, update-after-duplication, and network-error handling.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.spec.tsx
@@ -643,6 +643,9 @@ describe('Toolbar Menu', () => {
         screen.getByRole('menuitem', { name: 'Analytics 0 visitors' })
       ).toBeInTheDocument()
       expect(
+        screen.getByRole('menuitem', { name: 'Analytics 0 visitors' })
+      ).toHaveAttribute('href', '/journeys/journeyId/reports')
+      expect(
         screen.getByRole('menuitem', { name: 'Strategy' })
       ).toBeInTheDocument()
       expect(
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Menu/Menu.tsx
@@ -97,7 +97,9 @@ export function Menu({ user }: MenuProps): ReactElement {
         {!isLocalTemplate && (
           <AccessItem variant="menu-item" onClose={handleCloseMenu} />
         )}
-        {!mdUp && !isTemplate && <AnalyticsItem variant="menu-item" />}
+        {!mdUp && !isTemplate && (
+          <AnalyticsItem variant="menu-item" journeyId={journey?.id} />
+        )}
         {!sharedWithMeTeam && !isTemplate && (
           <>
             <CreateTemplateItem variant="menu-item" globalPublish={false} />


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/LanguageScreen/LanguageScreen.tsx
```diff
@@ -45,9 +56,12 @@ export function LanguageScreen({
   const { journey } = useJourney()
   const { query } = useTeam()
   const [journeyDuplicate] = useJourneyDuplicateMutation()
+  const updateLastActiveTeamId = useUpdateLastActiveTeamIdMutation()
   const { loadUser } = useCurrentUserLazyQuery()
   const [teamCreate] = useTeamCreateMutation()
   const [loading, setLoading] = useState(false)
+  const isDataReady = query?.data != null && journey != null
```
csiyang: Nit: Purely checking if the data is not null has the downside of showing loading state even if something errors out. We can consider using `query.loading` for some better checks.

jianwei1: Thanks for the suggestion. I kept the explicit  guard because it cleanly separates loading vs error states and avoids relying on loading semantics alone. The error branch now renders an accessible alert, while the loading branch keeps the spinner.

jianwei1: Clarification: I kept the explicit `hasTeamLoadError || !isDataReady` guard because it cleanly separates loading vs error states and avoids relying on loading semantics alone. The error branch now renders an accessible alert, while the loading branch keeps the spinner.

PR #8893, opened 2026-03-22
fix: reset YouTube URL state when switching cards in media screen (NES-1463)

jaco-brink:
## Summary
- **Bug:** Switching video cards in the media screen template customization caused the previously-typed YouTube URL to auto-apply to the newly selected card
- **Root cause:** `VideosSection` held a local `youtubeUrl` state that was never reset when `cardBlockId` changed, so the debounced `useEffect` re-fired with the stale URL against the new card's video block
- **Fix:** Added a `useEffect` keyed on `cardBlockId` that resets `youtubeUrl`/`youtubeUrlError` on card switch and hydrates from the existing YouTube source if one is configured

## Test plan
- [x] Existing tests updated (clear pre-populated URL before typing)
- [x] New test: resets YouTube URL when `cardBlockId` changes
- [x] New test: hydrates YouTube URL from existing video block source
- [x] New test: does not re-submit the hydrated URL (dedup map pre-populated)
- [x] `npx nx test journeys-admin --testPathPattern=VideosSection` — all pass
- [x] `npx nx lint journeys-admin` — 0 errors
- [x] `npx nx type-check journeys-admin` — clean

Closes NES-1463

## Summary by CodeRabbit

* **Bug Fixes**
  * YouTube URL input now resets when switching between video cards.
  * Existing YouTube videos are pre-filled with their canonical URLs for easier editing.
  * Debounce handling for the video input is more reliable, avoiding unintended validations.

* **Tests**
  * Updated tests to clear the input before typing and to verify reset, pre-fill, and debounce behaviors.

### DIFF
--- a/.cursor/skills/handle-pr-review/SKILL.md
+++ b/.cursor/skills/handle-pr-review/SKILL.md
@@ -1,40 +1,145 @@
 ---
 name: handle-pr-review
-description: Fetches PR review comments, applies fixes, commits, pushes, and posts a summary comment. Use when the user asks to check review feedback, address PR comments, fix review issues, or handle review feedback.
+description: Fetches PR review comments, triages them (fix, challenge, or skip), applies holistic fixes, resolves threads, and posts a summary. Use when the user asks to check review feedback, address PR comments, fix review issues, or handle review feedback.
 ---
 
 # Handle PR Review Feedback
 
 When the user asks to check or fix review feedback on a PR (e.g. on JesusFilm/core):
 
-## Steps
+## 1. Identify PR
 
-1. **Identify PR** — From context (branch, issue number) or ask. Prefer GitHub API lookup by head branch (`GET /repos/{owner}/{repo}/pulls?head=owner:branch`); fall back to branch-name inference or listing PRs if needed. Use GraphQL `pullRequest.reviewThreads` (includes `isResolved`) for thread resolution state; REST `get_review_comments` and `get_reviews` do not expose it.
+From context (branch, issue number) or ask. Prefer looking up via `gh pr view --json number,url` from the current branch. Fall back to `gh api` by head branch or listing PRs.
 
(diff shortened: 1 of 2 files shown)


apps/journeys-admin/src/components/TemplateCustomization/MultiStepForm/Screens/MediaScreen/Sections/VideosSection/VideosSection.tsx
```diff
@@ -139,6 +140,22 @@ export function VideosSection({
...
+      videoBlock != null &&
+      videoBlock.source === VideoBlockSource.youTube &&
+      videoBlock.videoId != null
+    ) {
+      const canonicalUrl = `https://www.youtube.com/watch?v=${videoBlock.videoId}`
+      setYoutubeUrl(canonicalUrl)
+      lastSubmittedRef.current.set(videoBlock.id, canonicalUrl)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cardBlockId])
```
jaco-brink: **Concern (minor):** The `eslint-disable` is the right call here — disagree with CodeRabbit's suggestion to add `videoBlock?.id/source/videoId` to the deps.

Adding those would cause this effect to re-fire after every successful `startYouTubeLink` mutation (the journey context updates with the resolved `videoId`), which would:
1. Flash the input value — user types `youtu.be/abc` → mutation succeeds → journey updates → effect re-runs → input changes to `youtube.com/watch?v=abc`
2. Potentially reset error state mid-flow if the journey updates for unrelated reasons

The dedup map already prevents stale-URL re-submission, so the current `[cardBlockId]`-only dep is correct and intentional.

**Suggestion:** Add a brief comment explaining *why* the lint is disabled so future readers don't "fix" it:

```tsx
// Only reset on card switch — adding videoBlock deps would re-fire
// after every YouTube mutation (journey context updates videoId).
// eslint-disable-next-line react-hooks/exhaustive-deps
```

PR #8958, opened 2026-04-07
fix: filter Google sync dropdown to only show current user's accounts (NES-1492)

jianwei1:
## Summary

**Bug (NES-1492):** In multi-manager teams, the Google Sheets sync dropdown showed ALL team integrations — including those created by other managers.

**Previous fix (#8938, reverted as #8957):** Filtered by `integration.user?.id === currentUserId`, but this comparison always failed because `integration.user.id` resolves to a Prisma UUID (via federation through `api-users`), while `useAuth().user.id` is the Firebase UID. Result: dropdown showed zero integrations.

**This fix:** Uses the `userId` scalar field exposed in #8967 to filter integrations by `integration.userId === currentUserId` — both are Firebase UIDs, so the comparison works correctly.

### Changes
- Added `userId` to the `GetIntegration` query (kept `user { id }` — used by `GoogleIntegrationDetails` for ownership checks)
- Updated filter in `GoogleSheetsSyncDialog` to compare `integration.userId` instead of `integration.user?.id`
- Added tests for user filtering, null userId, and unauthenticated user scenarios

### Dependencies
- **Requires #8967 to be deployed first** (exposes `userId` on `IntegrationGoogle` backend)

## Test plan
- [ ] Verify Google sync dropdown only shows the logged-in user's integrations
- [ ] Verify another manager's integrations do NOT appear in the dropdown
- [ ] Verify creating a new Google integration shows it in the dropdown
- [ ] Verify the Team Integrations page still shows all team integrations (no regression)

## Summary by CodeRabbit

* **Bug Fixes**
  * Google Sheets integration dropdown now only shows integrations belonging to the authenticated user, falls back to the placeholder when none match, and excludes unaffiliated accounts.
* **Tests**
  * Added tests validating dropdown behavior for matching, non-matching, null-owner, and unauthenticated scenarios.
* **Documentation**
  * Added a plan describing the update to ensure integration-owner matching works reliably.


mikeallisonJS (approved):

jianwei1:
## Review feedback addressed (ee4fc7a3e)

**Challenged:**
- `GoogleSheetsSyncDialog.spec.tsx:260-371` — CodeRabbit suggests wrapping new tests with provider chain (SnackbarProvider → MockedProvider → ThemeProvider). All 6 pre-existing tests in this file use the same `jest.mock()` pattern to mock hooks directly. The new tests are consistent with the file's established approach. Converting the whole file to providers would be a separate refactor. See inline reply.

**Skipped (optional/nit):**
- `docs/plans/...plan.md:15` — CodeRabbit notes plan text is stale. Plan documents are point-in-time historical artifacts; the implementation is the source of truth. Updating doc prose to match final code is low-value busywork.

PR #9036, opened 2026-04-17
feat(journeys): wire up LaunchDarkly flag provider

jaco-brink:
## Summary

- Wires LaunchDarkly flag provider into `apps/journeys` viewer, matching the established pattern from `apps/watch` and `apps/resources`
- Adds `getFlags()` server-side utility, `FlagsProvider` in `_app.tsx`, and `flags` prop through all 9 page routes via `getStaticProps`
- Adds live LD integration test (skips gracefully when `LAUNCH_DARKLY_SDK_KEY` is absent) and Conductor FlagsProvider unit tests

## Context

Resolves [NES-1563](https://linear.app/jesus-film-project/issue/NES-1563). Blocks NES-1554 (Slice 1 — Prototype Base).

The `apologistChat` flag will gate the Apologist AI chat widget for internal-only access, independent of the DB `showAssistant` field. The actual Conductor gating (`showPinnedChat` behind `apologistChat`) will land with NES-1554, which adds `LastCardChatBar` — this PR provides the flag infrastructure.

## Changes

**New files:**
- `apps/journeys/src/libs/getFlags/getFlags.tsx` — server-side flag fetcher (LD singleton + anonymous user)
- `apps/journeys/src/libs/getFlags/index.ts` — barrel export
- `apps/journeys/src/libs/getFlags/getFlags.spec.ts` — live LD test with env var skip guard

**Modified files:**
- `apps/journeys/pages/_app.tsx` — `FlagsProvider` as outermost wrapper
- All 9 page routes — `flags: await getFlags()` in `getStaticProps` success paths
- `apps/journeys/src/components/Conductor/Conductor.spec.tsx` — FlagsProvider integration tests
- `apps/journeys/AGENTS.md` — documented FlagsProvider in provider stack

## Pre-merge checklist

- [ ] Verify `LAUNCH_DARKLY_SDK_KEY` is in Doppler `stg_dev` config for the journeys app
- [ ] Ensure `apologistChat` flag exists in LD Stage environment with serve=true
- [ ] CI passes

## Test plan

- [x] `nx run journeys:lint` — pass (0 errors)
- [x] `nx run journeys:type-check` — pass
- [x] `nx run journeys:test` — 61/61 pass
- [ ] Live LD test passes in CI with real SDK key

## Summary by CodeRabbit

* **New Features**
  * Feature flags system integrated throughout the application, enabling dynamic feature control and toggling.
  * Flags are automatically fetched during static page generation and made accessible to all components.

* **Tests**
  * Added comprehensive test coverage for feature flags functionality and system integration.

* **Documentation**
  * Updated developer documentation to include feature flags setup and configuration guidance.


apps/journeys/src/libs/getFlags/getFlags.ts
```diff
@@ -0,0 +1,17 @@
...
+  const launchDarklyClient = await getLaunchDarklyClient()
+
+  const flags = (
+    await launchDarklyClient.allFlagsState({ key: uuidv4(), anonymous: true })
+  ).toJSON() as {
+    [key: string]: boolean | undefined
+  }
+
+  return flags
+}
```
jaco-brink, line 17: **Concern:** File extension should be `.ts` rather than `.tsx` — there is no JSX in this module. Minor, but `.tsx` signals to readers (and some tooling) that the file contains markup when it does not.

apps/journeys/src/libs/getFlags/getFlags.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { getFlags } from './getFlags'
+
+describe('getFlags', () => {
+  it('returns apologistChat as true for Stage environment', async () => {
+    if (!process.env.LAUNCH_DARKLY_SDK_KEY) {
+      // Skip gracefully in local dev where the SDK key is not configured.
+      // This test runs as a live check in CI with real env vars.
+      console.warn('Skipping live LD test: LAUNCH_DARKLY_SDK_KEY is not set')
```
jaco-brink: **Concern:** This test silently passes (via early `return`) when `LAUNCH_DARKLY_SDK_KEY` is absent, which is most local dev and CI environments. The suite reports green without actually testing anything.

Two options:
1. Use `it.skip()` instead of `return` so skipped tests are visible in output.
2. Add a separate unit test with a mocked `getLaunchDarklyClient` to verify the function calls `allFlagsState` and returns the JSON result — that test would always run.

apps/journeys/src/libs/getFlags/getFlags.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { getLaunchDarklyClient } from '@core/shared/ui/getLaunchDarklyClient'
+
+export async function getFlags(): Promise<{
+  [key: string]: boolean | undefined
+}> {
+  const launchDarklyClient = await getLaunchDarklyClient()
+
+  const flags = (
+    await launchDarklyClient.allFlagsState({ key: uuidv4(), anonymous: true })
```
jaco-brink, line 11: **Concern (awareness):** Each `getFlags()` call creates a fresh `uuidv4()` anonymous user. This is fine for boolean on/off flags like `apologistChat`, but means percentage-based rollouts would give inconsistent results across ISR regenerations (each page build rolls the dice independently). This matches the existing watch/resources pattern, so it is a known trade-off — just flagging for visibility as you add more flags in the future.

jaco-brink (commented):
## Review summary

**Critical:** 0  
**Concerns:** 3 items flagged (see inline)  
**Overall:** Clean implementation that follows the established watch/resources pattern. A few minor items worth discussing — nothing blocking.

apps/journeys/src/libs/getFlags/getFlags.ts
jaco-brink: Acknowledged — flagging here for visibility only. Sticking with the per-call `uuidv4()` pattern for consistency with the existing `watch`/`resources` implementations. Boolean flags (`apologistChat`) are unaffected; if we introduce percentage-based rollouts later we can revisit by keying on a stable identifier (user id, device id, or cached anonymous key).

jaco-brink:
## Review feedback addressed (5caa9dfe6)

**Fixed:**
- `apps/journeys/src/libs/getFlags/getFlags.tsx` → `getFlags.ts` — module contains no JSX, extension corrected.
- `apps/journeys/src/libs/getFlags/getFlags.spec.ts` — replaced env-gated live-LD test (which silently passed via early return when `LAUNCH_DARKLY_SDK_KEY` was absent) with a mocked unit test. Spec now always runs: mocks `getLaunchDarklyClient`, asserts `allFlagsState` is called with `{ key, anonymous: true }`, and verifies the JSON result is returned. Approach #2 from the suggestion.

**Acknowledged (no change):**
- Per-call `uuidv4()` for the anonymous user key — known trade-off, matches existing `watch`/`resources` pattern. Fine for boolean flags; noted for future percentage-rollout considerations.

**Follow-up:**
- Automated live LaunchDarkly integration test scope deferred to [NES-1560](https://linear.app/jesus-film-project/issue/NES-1560) — CI does not currently run Doppler (`fetch-secrets`), so `LAUNCH_DARKLY_SDK_KEY` is unavailable in the test job.

apps/journeys/pages/[hostname]/[journeySlug].tsx
```diff
@@ -104,6 +105,7 @@ export const getStaticProps: GetStaticProps<HostJourneyPageProps> = async (
     const { rtl, locale } = getJourneyRTL(data.journey)
     return {
       props: {
+        flags: await getFlags(),
         hostname: context.params?.hostname?.toString() ?? '',
         host: context.params?.host?.toString() ?? '',
```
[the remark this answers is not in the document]
Comment (jaco-brink) [reply apps/journeys/pages/[hostname]/[journeySlug].tsx:110]: Thanks for catching this — the bug is real (the `host` prop reads a non-existent route param), but it is pre-existing on `main`, not introduced by this PR. The diff here only adds `flags: await getFlags()` immediately above; the `host`/`hostname` lines are unchanged from before. The PR's scope is wiring up the LaunchDarkly flag provider, so I'd prefer to keep it focused and address this OpenGraph URL bug in a separate ticket where it can get its own review and (ideally) a regression test.

Follow-up filed: [NES-1567](https://linear.app/jesus-film-project/issue/NES-1567/fix-malformed-opengraph-url-on-hosted-journey-pages-host-reads-non).

apps/journeys/pages/[hostname]/embed/[journeySlug].tsx
```diff
@@ -110,6 +111,7 @@ export const getStaticProps: GetStaticProps<HostJourneyEmbedPageProps> = async (
     const { rtl, locale } = getJourneyRTL(data.journey)
     return {
       props: {
+        flags: await getFlags(),
         host: context.params?.host?.toString() ?? '',
```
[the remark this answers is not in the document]
Comment (jaco-brink) [reply apps/journeys/pages/[hostname]/embed/[journeySlug].tsx:115]: Same response as the [companion thread on `[journeySlug].tsx`](https://github.com/JesusFilm/core/pull/9036#discussion_r3107654759): the bug is real but pre-existing on `main` — this PR only added `flags: await getFlags()` immediately above. To keep the LaunchDarkly wiring focused, this is being tracked separately.

Follow-up filed: [NES-1567](https://linear.app/jesus-film-project/issue/NES-1567/fix-malformed-opengraph-url-on-hosted-journey-pages-host-reads-non) — covers both files.

jaco-brink:
## Review feedback addressed

**Challenged (out of scope) — tracked in [NES-1567](https://linear.app/jesus-film-project/issue/NES-1567/fix-malformed-opengraph-url-on-hosted-journey-pages-host-reads-non):**
- `apps/journeys/pages/[hostname]/[journeySlug].tsx:110` — CodeRabbit flagged that `host` reads a non-existent route param. Bug is real but pre-existing on `main` — this PR only adds `flags: await getFlags()` above it. Keeping PR scope focused on LaunchDarkly wiring. See inline reply.
- `apps/journeys/pages/[hostname]/embed/[journeySlug].tsx:115` — Same pre-existing bug, same response. See inline reply.

**Skipped (already resolved in prior review round):**
- `apps/journeys/src/libs/getFlags/getFlags.ts:17` — file extension nit.
- `apps/journeys/src/libs/getFlags/getFlags.spec.ts` — silently-passing test concern.
- `apps/journeys/src/libs/getFlags/getFlags.ts:11` — per-call `uuidv4()` awareness note.

**No code changes pushed in this round.**

csiyang (approved):

PR #9041, opened 2026-04-20
fix: limit guest user access to non-customisable templates

csiyang:
## Summary
- Removes `allowGuest: true` from the template detail page (`/templates/[journeyId]`) so anonymous users with a guest session from the customise flow are redirected to login on non-customisable templates
- Fresh visitors (no token) are unaffected — they already bypass `checkConditionalRedirect`

Fixes NES-1564

## Test plan
- [ ] Visit a customisable template as a guest (should still work, creates anonymous session)
- [ ] Navigate to a non-customisable template — should now prompt login
- [ ] Visit a non-customisable template directly with no session — should still render normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Sign-in detection now considers presence of an email address; footer, action buttons, and CTAs reflect signed-in state accordingly.

* **Tests**
  * Updated tests to match the revised authentication input and added a case for anonymous users verifying account-check UI appears and team-copy dialog does not render.


jianwei1 (changes requested):
## Review summary

**Request changes** — one missed spot re-introduces the same regression this PR is trying to fix.

### Critical — missed usage in `TemplateCardPreview.tsx:310`

```tsx
<TemplateActionButton signedIn={authUser?.id != null} />
```

This component is rendered by `TemplatePreviewTabs` on every template detail page (see `TemplateView.tsx:113`). Guests reach the page because `allowGuest: true` is kept in `initAndAuthApp` for `/templates/[journeyId]`, and the "{{count}} more cards" overlay (which contains this `TemplateActionButton`) renders whenever `steps.length > slidesToRender.length` — i.e. for most real templates. A guest clicking "Use This Template" here still gets `signedIn = true` (they have an `id`), goes straight into `CopyToTeamDialog`, and bypasses `AccountCheckDialog`. Same fix needed here: `authUser?.email != null`.

Can't leave this as an inline comment since the file isn't in the diff — but that's exactly the reason it's worth flagging.

### Concern — missing regression test

The two spec updates only patch existing tests to keep them green (adding `email: 'user@example.com'`). No test asserts the new behaviour: `authUser={ id: '123' }` (no email) + `nonCustomizableJourney` → clicking "Use This Template" opens `AccountCheckDialog` instead of `CopyToTeamDialog`. Without it, someone could revert the `email`/`id` swap and the suite would still pass — the fix has no coverage. Worth adding one spec case to `TemplateViewHeader.spec.tsx` (and ideally one to `TemplateCardPreview.spec.tsx` for the overlay button once that spot is fixed).

### Positive

Surgical three-line product change; aligns with the backend's `isAnonymous = email == null` convention in `apis/api-journeys-modern/src/schema/authScopes.ts:99`.

jianwei1 (approved):

PR #9064, opened 2026-04-23
fix: remove login tab sync

csiyang:
## Summary by CodeRabbit

* **Refactor**
  * Removed the `login` query parameter from sign-in redirects.
  * Sign-in tabs now initialize and update only via user interaction (no URL-driven preselection).

* **Bug Fix / Behavior**
  * Sign-in dialog handler simplified to a no-argument callback for both “Log in” and “Create account”.

* **Tests**
  * Updated tests to remove `login` query checks and assert the no-arg handler is invoked exactly once.

* **Documentation**
  * Added guidance describing the issue and the applied fix.

### DIFF
--- a/apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
+++ b/apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
@@ -18,4 +18,25 @@ describe('SignInTabs', () => {
     fireEvent.click(getByRole('tab', { name: 'Log In' }))
     expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
   })
+
+  it('should retain tab selection after re-render (regression: router effect reset)', () => {
+    const { getByRole, rerender } = render(<SignInTabs />)
+
+    fireEvent.click(getByRole('tab', { name: 'Log In' }))
+    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
+
+    rerender(<SignInTabs />)
+    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
+  })
+
+  it('should not reset tab selection on repeated re-renders (regression: shallow route update)', () => {
+    const { getByRole, rerender } = render(<SignInTabs />)
+
+    fireEvent.click(getByRole('tab', { name: 'Log In' }))
+    rerender(<SignInTabs />)
+    rerender(<SignInTabs />)
+    rerender(<SignInTabs />)
+
+    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
+  })
 })
--- a/apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
+++ b/apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
@@ -19,24 +19,5 @@ describe('SignInTabs', () => {
     expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
   })
 
-  it('should retain tab selection after re-render (regression: router effect reset)', () => {
-    const { getByRole, rerender } = render(<SignInTabs />)
 
-    fireEvent.click(getByRole('tab', { name: 'Log In' }))
-    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
-
-    rerender(<SignInTabs />)
-    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
-  })
-
-  it('should not reset tab selection on repeated re-renders (regression: shallow route update)', () => {
-    const { getByRole, rerender } = render(<SignInTabs />)
-
-    fireEvent.click(getByRole('tab', { name: 'Log In' }))
-    rerender(<SignInTabs />)
-    rerender(<SignInTabs />)
-    rerender(<SignInTabs />)
-
-    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
-  })
 })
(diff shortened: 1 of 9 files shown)


apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.tsx
```diff
@@ -1,28 +1,20 @@
...
-    setTabValue(newValue)
-  }
-
   return (
     <Tabs
       value={tabValue}
-      onChange={handleTabChange}
+      onChange={(_, newValue) => {
+        setTabValue(newValue)
+      }}
```
jianwei1: **Concern (convention):** `apps/journeys-admin/AGENTS.md` requires event handlers to be prefixed `handle` (e.g. `handleClick`, `handleKeyDown`). Previously this was a named `handleTabChange`; the inline anonymous arrow is a minor step back on convention compliance. Keeping a named `handleTabChange` costs nothing and is consistent with the rest of the admin app.

```tsx
function handleTabChange(_: unknown, newValue: number): void {
  setTabValue(newValue)
}
// ...
<Tabs value={tabValue} onChange={handleTabChange} ...>
```

---

**Concern (test coverage):** No regression test was added for the bug this PR fixes. `SignInTabs.spec.tsx` still only asserts default render + click behaviour — identical to before. If someone later reintroduces a `useEffect([router])` (or similar effect that writes `tabValue` from an unstable source), nothing catches it.

The solution doc you just wrote (`docs/solutions/ui-bugs/mui-tabs-snapping-to-first-tab-router-sync.md`) explicitly lists the two tests worth adding:

- Click tab 1 → cause a router reference change (or re-render) → assert `aria-selected` is still on tab 1.
- Simulate a shallow route update (different unrelated query param) → assert tab selection is unchanged.

Even one of these as a guard would lock in the fix.

docs/solutions/ui-bugs/mui-tabs-snapping-to-first-tab-router-sync.md
```diff
@@ -0,0 +1,214 @@
...
+category: ui-bugs
+tags:
+  - mui-tabs
+  - next-js
+  - router
+  - use-effect
+  - controlled-component
+  - sign-in
+  - state-management
+symptoms:
```
jianwei1, line 14: **Concern (doc accuracy):** These `components:` paths are wrong and will mislead future `grep`/`Read` lookups:

- `AccountCheckDialog` actually lives at `libs/journeys/ui/src/components/TemplateView/AccountCheckDialog/AccountCheckDialog.tsx`
- `CreateJourneyButton` at `libs/journeys/ui/src/components/TemplateView/CreateJourneyButton/CreateJourneyButton.tsx`
- `UseThisTemplateButton` at `libs/journeys/ui/src/components/TemplateView/UseThisTemplateButton/UseThisTemplateButton.tsx`
- `SignInTabs` is at `apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.tsx` (missing the `SignIn/` segment)

The same inaccuracy appears in the prose under **Investigation** (points 2 and 3 — `AccountCheckDialog.tsx`).

jianwei1 (commented):
## Review summary

Nice, focused fix — the `useEffect([router])` root cause is called out cleanly and the callback simplification propagates nicely through callers + tests. No blocking issues. Posting as a comment review; a few items to consider before merge (see inline).

**Concerns:**
- No regression test for the reported bug — the solution doc itself lists the two tests that would guard it.
- Solution doc frontmatter (`components:`) and prose reference wrong paths for `AccountCheckDialog` / `CreateJourneyButton` / `UseThisTemplateButton` (they live under `libs/journeys/ui/src/components/TemplateView/`, not `apps/journeys-admin/src/components/`).
- `onChange` handler went from named `handleTabChange` to an inline anonymous arrow — minor backslide vs `apps/journeys-admin/AGENTS.md` "event handlers prefix with `handle`".
- `AccountCheckDialog` "Login with my account" no longer pre-selects the Log In tab — intentional per the doc, but worth confirming with product/QA on NES-1599 so it isn't later flagged as a regression.

**FYI:** Playwright E2E (journeys) failed on this run — likely unrelated, worth a re-run before merge.

apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.tsx
csiyang: Fixed — extracted a named `handleTabChange` function per AGENTS.md convention.

docs/solutions/ui-bugs/mui-tabs-snapping-to-first-tab-router-sync.md
csiyang: Fixed — corrected all component paths in frontmatter and prose.

csiyang:
**Regression tests:** Added two tests in `SignInTabs.spec.tsx`: one verifying the tab stays selected after a re-render, and one verifying stability across repeated re-renders (simulating what a shallow route update would previously have triggered).

**Tab pre-selection change:** Confirmed intentional — documented in the solution doc as a deliberate tradeoff. The tabs don't change the underlying sign-in flow (real branching happens post-email-submit via `fetchSignInMethodsForEmail`), so the snap-back fix is the higher-priority UX win.

apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
```diff
@@ -18,4 +18,25 @@ describe('SignInTabs', () => {
...
+  it('should not reset tab selection on repeated re-renders (regression: shallow route update)', () => {
+    const { getByRole, rerender } = render(<SignInTabs />)
+
+    fireEvent.click(getByRole('tab', { name: 'Log In' }))
+    rerender(<SignInTabs />)
+    rerender(<SignInTabs />)
+    rerender(<SignInTabs />)
+
+    expect(getByRole('tab', { selected: true })).toHaveTextContent('Log In')
+  })
```
jianwei1:
```suggestion
```
On second thought, drop these regression tests. Now that SignInTabs doesn't import useRouter and has no router-reactive code, there's nothing for the re-render to actually test against.

jianwei1 (changes requested):

apps/journeys-admin/src/components/SignIn/SignInTabs/SignInTabs.spec.tsx
csiyang: Agreed — removed both. No router code left to regress against.

jianwei1 (approved):

PR #9075, opened 2026-04-24
fix: correct event label to Plausible goal mapping for capture events

csiyang:
## Summary

- Adds `BLOCK_EVENT_LABEL_TO_PLAUSIBLE_EVENT` mapping in `plausibleHelpers.ts` that correctly maps `BlockEventLabel` enum values to their registered Plausible goal names (e.g. `decisionForChrist` → `christDecisionCapture`)
- Fires the correctly-named capture event in `RadioQuestion`, `Button`, and `VideoEvents` instead of the raw enum value — this ensures Plausible goal conversion tracking works and `templateKey` values use the right event name
- Fixes capture event URL to use `stepId` instead of `blockId` so Plausible correctly attributes capture events to the journey step page
- Adds `target` to `templateKey` for RadioQuestion and Button capture events for consistency with primary events

Backward compatibility: `EVENT_TO_CAPTURE_MAP` in `transformBreakdownResults.ts` is kept to correctly handle historical `decisionForChrist` events already stored in Plausible.

Fixes QA-359.

## Test plan

- [x] `RadioQuestion.spec.tsx` — verifies `custom1Capture` (not `BlockEventLabel.custom1`) is fired with `stepId` in URL
- [x] `Button.spec.tsx` — verifies `custom1Capture`/`custom2Capture` are fired for click and chat events with `stepId` in URL
- [x] `VideoEvents.spec.tsx` — verifies `custom1Capture`/`custom2Capture` are fired for start/complete events with `stepId` in URL
- [x] `plausibleHelpers.spec.ts` — all existing tests pass

## Release Notes

* **Bug Fixes**
  * Fixed overcounting in analytics for specific engagement metrics caused by incorrect event attribution
  * Improved accuracy of event tracking across interactive components 
  * Corrected event-to-step attribution mapping to prevent duplicate counts

* **Documentation**
  * Added troubleshooting guide for analytics event tracking issues and resolution steps


edmonday (commented):
## Review summary

Fix itself looks right — exhaustive `Record<BlockEventLabel, ...>` is a nice guard, the stale-closure deps are fixed, and the shared `fireCaptureEvent` cleanly de-duplicates the three call sites. CI is green.

Two things to address before merge:

- **`todos/` directory at repo root (17 new files) looks unintended.** Not in `.gitignore`, not a pre-existing convention. Looks like AI self-review follow-up notes that should not land on `main`. The `docs/plans/` and `docs/solutions/` additions are fine — those directories already exist.
- **RadioQuestion `blockId` semantics changed** — see inline.

Out of scope: the `Object.defineProperty(process.env, 'NODE_ENV', …)` changes in `apis/api-journeys-modern/src/schema/event/utils.spec.ts` are unrelated to this fix — consider splitting or calling out in the PR description.

The pending `todos/*.md` items (dual-map sync test, PII guard, JSDoc, etc.) read as reasonable follow-ups; none block this PR in my view once the directory itself is removed and the RadioQuestion question is answered.

todos/001-complete-p1-videoevents-stale-closure-missing-deps.md
```diff
@@ -0,0 +1,102 @@
+---
```
edmonday: **Concern:** The entire `todos/` directory (17 files added in this PR) does not appear to be an established convention in this repo — it did not exist before this PR and is not in `.gitignore`. These look like internal self-review / follow-up notes that probably shouldn't be committed to `main`.

Suggest removing the `todos/` directory from this PR. If you want to keep the follow-ups tracked, convert the actionable ones to Linear issues (or move them under `docs/solutions/` if they're the kind of learning doc that belongs there). `docs/plans/2026-04-24-001-...` and `docs/solutions/logic-errors/plausible-analytics-event-name-mismatch-qa359.md` are fine to keep — those directories are already established.

libs/journeys/ui/src/components/RadioQuestion/RadioQuestion.tsx
```diff
@@ -149,31 +150,15 @@ export function RadioQuestion({
...
-                journeyId: journey?.id
-              })
-            }
-          })
-        }
+        fireCaptureEvent(plausible, radioOptionBlock.eventLabel, {
+          u: `${window.location.origin}/${journey.id}/${input.stepId}`,
+          input,
+          stepId: input.stepId ?? '',
+          blockId: radioOptionBlock.id,
```
edmonday, line 157: **Concern:** This is a semantic change in what Plausible receives for RadioQuestion capture events.

Inside `fireCaptureEvent`, `blockId` is spread *after* `...input`, so it overwrites `input.blockId`. Previously the capture event's `blockId` prop was `input.blockId` (the RadioQuestion block id, e.g. `RadioQuestion1`); now it becomes `radioOptionBlock.id` (e.g. `RadioOption1`). The spec was updated to match (`RadioQuestion.spec.tsx` line ~447).

Is this intentional? It lines `blockId` up with what the `key`/`simpleKey` already used, which is consistent, but it does change the shape of data flowing into Plausible — worth confirming it won't affect any downstream breakdown consumers. Button and VideoEvents are unaffected because `input.blockId` already matched the passed `blockId` at those call sites.

edmonday (approved):

Ur-imazing (approved):

PR #9082, opened 2026-04-24
chore: customdomaincreate modern

mikeallisonJS:
## Summary by CodeRabbit

* **New Features**
  * Add create and list custom domain support with domain validation, automated provisioning to the hosting provider, and automatic updates to team short-links; supports routing all team journeys and optional collection linking.
* **Tests**
  * Added comprehensive tests covering successful creation, input validation, authorization checks, duplicate-domain errors, and error paths.


apis/api-journeys-modern/src/schema/customDomain/customDomainCreate.mutation.ts
```diff
@@ -0,0 +1,90 @@
...
+  createVercelDomain,
+  deleteVercelDomain,
+  isDomainValid,
+  updateTeamShortLinks
+} from './service'
+
+const ERROR_PSQL_UNIQUE_CONSTRAINT_VIOLATED = 'P2002'
+
+builder.mutationField('customDomainCreate', (t) =>
+  t.withAuth({ isAuthenticated: true }).prismaField({
```
csiyang: The mutation only gates on `isAuthenticated: true` and then does the team-manager check post-hoc, after `createVercelDomain` has already run. That means any authenticated user can trigger a Vercel domain creation against your project for an arbitrary `input.name` before authorization fails — the cleanup `deleteVercelDomain` is best-effort, and if it returns a non-200/404 it throws `'vercel response not handled'`, which masks the original FORBIDDEN and leaves the domain registered in Vercel.

The codebase already has an `isTeamManager` scope in [`authScopes.ts`](https://github.com/JesusFilm/core/blob/main/apis/api-journeys-modern/src/schema/authScopes.ts) wired into the builder, and the equivalent pattern is used by `integrationGoogleCreate`:

```ts
.withAuth((_parent, args) => ({
  $all: {
    isAuthenticated: true,
    isInTeam: (args.input as typeof IntegrationGoogleCreateInput.$inferInput).teamId
  }
}))
```

Suggest the same shape here:

```ts
.withAuth((_parent, args) => ({
  $all: {
    isAuthenticated: true,
    isTeamManager: (args.input as typeof CustomDomainCreateInput.$inferInput).teamId
  }
}))
```

That gates the team-manager check before the resolver body runs, so the Vercel call never happens for unauthorized callers and the post-create ACL block + cleanup dance can be removed entirely.

apis/api-journeys-modern/src/schema/customDomain/service.ts
```diff
@@ -0,0 +1,177 @@
...
+    throw new GraphQLError('Journeys url not configured', {
+      extensions: { code: 'INTERNAL_SERVER_ERROR' }
+    })
+
+  const journey = await prisma.journey.findUniqueOrThrow({
+    where: { id: toJourneyId }
+  })
+
+  const base = `https://${customDomainName}`
+  const blockPath = toBlockId != null ? `/${toBlockId}` : ''
```
csiyang, line 143: Minor behavior regression vs legacy: the old `getTo` used `prisma.block.findUniqueOrThrow({ where: { journeyId: journey.id, id: toBlockId } })` which validated that the block exists and belongs to the target journey. The new code skips that and just concatenates `toBlockId` into the path.

The URL output is identical when the data is consistent (legacy used `block.id`, which equals the input `toBlockId`), so the only practical difference is the throw on stale data — a `qrCode.toBlockId` that's been deleted or moved to another journey will now silently produce a broken URL instead of failing loudly. Probably edge-case, but worth a conscious decision rather than an accidental drop.

csiyang (commented):
Overall the migration looks clean, and the `for...of await` replacement of the legacy `forEach(async ...)` anti-pattern is the right call.

One side effect of that fix worth flagging: short-link mutation failures now propagate up and roll back the whole transaction, where previously they were silently swallowed (the legacy `forEach` never awaited the inner promises and the surrounding try/catch never saw them). If any team has stale QR data — e.g. a `qrCode.toJourneyId` pointing at a journey that no longer exists — `customDomainCreate` will start failing for that team where it used to appear to succeed. Worth a quick check that no such stale rows exist before this rolls.

Left two inline comments — the auth-ordering one is the only thing I'd want addressed before merge.

csiyang (approved):
Auth ordering fix looks good — isTeamManager scope gates before Vercel call now.

PR #9085, opened 2026-04-25
chore: upgrade next i18next

mikeallisonJS:


tsconfig.base.json
```diff
@@ -4,12 +4,12 @@
     "rootDir": ".",
     "sourceMap": true,
     "declaration": false,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
csiyang: next-i18next v16 does need a module-resolution change so TypeScript can read the package's `exports` field and resolve the new `/pages` subpath — `node`/`node10` doesn't honor `exports`, but `node16`, `nodenext`, and `bundler` all do. The concern is the **scope** of the change.

Flipping the base means every project that extends `tsconfig.base.json` silently inherits `module: esnext` + `moduleResolution: bundler`, which is why this PR has to add ~12 compensating overrides re-pinning `module: commonjs` + `moduleResolution: node` across NestJS APIs, Prisma libs, tools, and several spec configs. That's also why every future NestJS service or shared backend lib added to the workspace will need to remember to override or it'll typecheck differently from how it runs.

This is a CJS-majority workspace (~6 NestJS APIs + 5 Prisma libs + tools + Strapi cms + Node-runtime shared libs vs ~7 Next.js apps). Inverting the base default means the majority pays the override tax for the minority.

Safer scoping: leave `tsconfig.base.json` alone and add `moduleResolution: bundler` + `module: esnext` to **just the ~5 Next.js app tsconfigs that actually import `next-i18next/pages`** (`apps/journeys-admin`, `apps/journeys`, `apps/resources`, `apps/watch`, `libs/journeys/ui`). Roughly 5 lines changed in 5 files, no compensating overrides needed anywhere else. (`apps/watch/tsconfig.json` already has these settings, fwiw.)

If you'd rather avoid touching `module` at all, `moduleResolution: "node16"` on those same apps also resolves `next-i18next/pages` and doesn't conflict with `module: commonjs` (TS5095 only fires for `bundler`).

apps/arclight/src/app/videoPlayerUrl/page.tsx
```diff
@@ -112,24 +111,18 @@ export default async function Page(props: {
 
   // Validate time parameters
   if (startTime != null && (isNaN(startTime) || startTime < 0)) {
-    return {
-      message: 'Invalid start time parameter',
-      status: 400
-    }
+    // eslint-disable-next-line i18next/no-literal-string
+    return <p>Invalid start time parameter</p>
```
csiyang: These three validation branches now return `<p>...</p>` directly, which means **HTTP 200 OK with naked HTML for invalid input** — previously the code returned `{ status: 400, ... }`. Anything downstream that branches on status code (cache rules, monitors, embedders, the calling client) will see invalid params as a successful response.

The other error paths in this same file (missing `refId`, missing `hlsUrl`) use `notFound()` and return a real 404. Consistency here would mean using `notFound()` for the time-validation failures too — that also lets you delete the three `eslint-disable-next-line i18next/no-literal-string` comments, which feel out-of-place in a PR titled "chore: upgrade next i18next".

Also — this hunk and the unrelated `GET_SHORT_LINK_QUERY` visibility change in `[keyword]/route.ts` look like a separate bug fix that drifted into the i18next migration. Splitting them out would make this PR a clean mechanical rename and let the arclight fix get reviewed on its own merits.

tsconfig.base.json
mikeallisonJS: Agreed — reverted tsconfig.base.json to commonjs/node and scoped `module: esnext` + `moduleResolution: bundler` to just the 4 Next.js app tsconfigs that need it (journeys-admin, journeys, resources, libs/journeys/ui). Watch already had them. Removed all compensating overrides from backend/lib tsconfigs. (49324da2e)

apps/arclight/src/app/videoPlayerUrl/page.tsx
mikeallisonJS: Restored the original `{ message, status }` return values — these may be consumed by embedding applications. The `GET_SHORT_LINK_QUERY` visibility change (`export` → `const`) is a minor cleanup that fell into this PR; happy to split it out if preferred. (49324da2e)

mikeallisonJS:
## Review feedback addressed (49324da2e)

**Fixed:**
- `tsconfig.base.json` — reverted to `commonjs`/`node`; scoped `module: esnext` + `moduleResolution: bundler` to only the Next.js apps that need it (journeys-admin, journeys, resources, libs/journeys/ui). Removed all ~20 compensating overrides from backend/lib tsconfigs.

**Challenged:**
- `apps/arclight/src/app/videoPlayerUrl/page.tsx` — restored original `{ message, status }` return values (may be consumed by embedding applications). See inline reply.

**Skipped:**
- Splitting arclight changes into a separate PR — valid process suggestion but splitting after the fact adds overhead.

csiyang (approved):
Re-reviewed after the latest changes. All four previously-flagged issues are resolved:

**1. ✅ `tsconfig.base.json` reverted** — diff vs `main` is empty. Still `module: commonjs`, `moduleResolution: node`.

**2. ✅ No compensating backend overrides** — `git diff --stat` shows zero tsconfig changes under `apps/api-*`, Prisma packages, or `tools/`. The `module: esnext` / `moduleResolution: bundler` overrides are scoped to:
- 4 Next.js apps: `journeys`, `journeys-admin`, `resources`, `videos-admin`
- 2 shared libs consumed by them: `libs/journeys/ui`, `libs/shared/ui` (necessary because they import `next-i18next/pages`)
- `arclight/tsconfig.json`'s only change is a reorder of `include[]` — no compiler-options change

This is a slightly wider scope than "just the 4 Next.js app tsconfigs" but the libs additions are correct: type-check would fail otherwise on the migrated `next-i18next/pages` imports inside `libs/journeys/ui` and `libs/shared/ui`.

**3. ✅ Arclight error handling reverted** — `git diff origin/main HEAD -- apps/arclight/src/app/videoPlayerUrl/page.tsx` returns empty. Commit `49324da2e` cleanly restored the `{ message, status }` returns. The pre-existing return-object-from-Server-Component pattern is unchanged from `main` and out of scope for this PR.

**4. ✅ i18next v15→v16 import migration is clean** — all `from 'next-i18next'` → `from 'next-i18next/pages'` and `from 'next-i18next/serverSideTranslations'` → `from 'next-i18next/pages/serverSideTranslations'`. Verified no stragglers via grep across `apps/` and `libs/`. Jest mocks (`jest.mock('next-i18next/pages', …)`), TypeScript types (`SSRConfig`, `TFunction`), and the JSDoc `UserConfig` annotation in `next-i18next.config.js` are all migrated consistently. `_app.tsx` and `next-env.d.ts` updated accordingly.

### Minor non-blocking observation

`apps/videos-admin/tsconfig.spec.json` had `moduleResolution: node` added to override the parent's bundler setting, but `journeys-admin`, `journeys`, and `resources` `tsconfig.spec.json` files did not get the same override — they keep `module: commonjs` while inheriting `moduleResolution: bundler` from the parent. TypeScript disallows that combination, so IDE/`tsc -b` against the spec config directly may error.

In practice this is unlikely to break CI because:
- The `type-check` targets run `tsc -b <app>/tsconfig.json` (not the spec config), and the app tsconfigs have no `references` to the spec configs
- Jest uses `babel-jest`, so spec-config compiler options don't affect runtime

Worth a quick follow-up to add `moduleResolution: node` to the three other spec configs for consistency, but not a merge blocker.

Approving — ship it.

PR #9094, opened 2026-04-28
feat(apologist): wire Langfuse tracing + prompt management on /api/chat (NES-1572)

jaco-brink:
## Summary

- Wire Langfuse tracing on the apologist `/api/chat` endpoint so every conversation produces a trace with one generation per request, tagged with `sessionId`, `journeyId`, language, IP country, provider, and model.
- Source the apologist system prompt from Langfuse prompt management (`apologist-world-cup-chat`, env-scoped `production` / `development` label) with a hard-coded fallback when Langfuse is unavailable or returns a non-text prompt.
- Mint a per-tab `sessionId` in `AiChat` (sessionStorage-backed UUID) and forward it plus `journeyId` to the API so Langfuse can group turns into sessions.
- Pin `langfuse@^3.38.20`.

## Behaviour notes

- `getLangfuse()` returns `null` when `LANGFUSE_PUBLIC_KEY`, `LANGFUSE_SECRET_KEY`, or `LANGFUSE_BASE_URL` is unset — the route then skips tracing/prompt fetch and uses the fallback system prompt, so absence of Langfuse env never breaks chat.
- Generation lifecycle is end-once: `onError`, `onFinish` (incl. `finishReason === 'error'`), the outer `try/catch`, and `pipeUIMessageStreamToResponse.onError` all funnel through `endGenerationIfPending` and `flushAsync`, so we don't lose traces on stream errors, sync throws, or pipe-step failures.
- Prompt label resolves to `production` on `VERCEL_ENV=production`, else `development` — matches the lowercase convention the Langfuse SDK expects.
- The `language` Mustache variable is omitted when the request has no language (rather than substituted with empty string), so the Langfuse template can use a section block (`{{#language}}…{{/language}}`). The fallback path likewise skips the language line when missing. The template variable receives a BCP-47 code (e.g. `fr`, `es-419`) sourced from `journey.language.bcp47` — the prompt template is authored to accept those codes directly.

## Codegen drift (autofix-ci)

`apps/resources/__generated__/{GetJourney,JourneyFields}.ts` and `apps/watch/__generated__/{GetJourney,JourneyFields}.ts` pick up a `showAssistant: boolean | null` field that was added to the journey GraphQL fragment in #8822. Only `apps/journeys/__generated__/*` and `apps/journeys-admin/__generated__/*` were regenerated at the time; the autofix-ci bot caught the resources/watch files up here. Unrelated to NES-1572 functionally but necessary catch-up.

## Test plan

- [ ] Verify chat works end-to-end on stage with Langfuse env vars set: trace appears in Langfuse with one generation, correct model/usage, session groups multiple turns from the same tab.
- [ ] Verify chat still works on stage with Langfuse env vars unset: fallback system prompt used, no errors, no Langfuse calls.
- [ ] Confirm prompt label resolution: stage uses `development`, prod uses `production`.
- [ ] Trigger a stream error (e.g. invalid model) and confirm the generation ends with `level: ERROR` and `statusMessage` populated.
- [ ] `npx jest --config apps/journeys/jest.config.ts --no-coverage 'apps/journeys/pages/api/chat/index.spec.ts'` passes.

## Summary by CodeRabbit

* **New Features**
  * Persistent per-tab session tracking and journey IDs included in chat requests.
  * Optional tracing integration to capture prompt selection, compilation, and generation lifecycle.

* **Bug Fixes**
  * More robust streaming error handling with single-generation completion guarantees and deterministic prompt fallbacks.
  * Improved provider resolution and clearer error responses when upstream keys are missing.

* **Tests**
  * Expanded end-to-end and unit tests covering tracing, prompt fallback, provider resolution, and streaming error scenarios.

* **Chores**
  * Added runtime dependency for tracing integration and updated default AI model fallback.

* **Documentation**
  * New end-to-end testing guide for chat tracing behavior.


apps/journeys/pages/api/chat/index.ts
```diff
@@ -174,7 +242,44 @@ export default async function handler(
...
+      undefined,
+      { label: getActivePromptLabel() }
+    )
+    if (promptClient.type !== 'text') {
+      console.warn(
+        `[langfuse] expected text prompt for ${APOLOGIST_PROMPT_NAME}, got ${promptClient.type} — using fallback`
+      )
+      return { system: fallback, promptClient: null }
+    }
+    const compiled = promptClient.compile({ language: language ?? '' })
```
jaco-brink: **Concern (language semantics):** the `language` body field is set by `AiChat.tsx:176` from `journey?.language?.bcp47` (e.g. `fr`, `es-419`, `zh-Hans-CN`), but the spec at `index.spec.ts:501` calls `compile({ language: 'French' })` — implying the prompt template variable is meant to receive a human-readable name. In production the LLM will see `Respond in the following language: fr` (or `zh-Hans-CN`), not `French`. Worth confirming the Langfuse `apologist-world-cup-chat` template is authored to accept BCP-47 codes, or mapping to a name in the handler before sending.

**Nit (empty-string handling):** `language ?? ''` here substitutes empty string into `{{language}}` whenever the field is missing, so the Langfuse-managed prompt may render `Respond in the following language: ` with a dangling colon. The fallback path at `buildFallbackSystemMessage` skips that sentence entirely when `length === 0`. Coalesce: either skip in both, or substitute a default (`'auto'` / `'the user's language'`) consistently.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -164,6 +227,11 @@ export default async function handler(
       name: err?.name,
       message: err?.message
     })
+    endGenerationIfPending({
```
jaco-brink, line 243: **Concern (terminal-path coverage):** the `catch`, `streamText.onError`, and `streamText.onFinish` paths all funnel through `endGenerationIfPending`, but `pipeUIMessageStreamToResponse(res, { onError })` at lines 210–221 (just above this hunk) is a fourth terminal path that doesn't. If the AI SDK pipe step itself fails (serialization, write to a closed `res`) after upstream succeeded, that branch fires and the Langfuse generation is left dangling. Cheap fix: route the pipe's onError through the same guard + `flushAsync` too.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -141,6 +177,33 @@ export default async function handler(
...
+                ? {
+                    input: usage.inputTokens,
+                    output: usage.outputTokens,
+                    unit: 'TOKENS'
+                  }
+                : undefined,
+            level: 'DEFAULT'
+          })
+        }
+        await langfuse?.flushAsync()
```
jaco-brink, line 208: **Concern:** flushing on every request defeats Langfuse's batching layer, which adds network latency on the success path before `res.end` (the AI SDK awaits `onFinish` before closing the stream). This is the right trade-off in serverless — the lambda freezes after the response stream closes, so without flush the trace is dropped — but worth knowing it's a per-request sync cost. If/when Vercel exposes per-request `waitUntil` on Pages Router for this app, that's the better home for this.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -122,17 +131,44 @@ export default async function handler(
...
+  const langfuse = getLangfuse()
+  const { system, promptClient } = await resolveSystemMessage({
+    language,
+    langfuse
+  })
   const modelMessages = await convertToModelMessages(messages)
 
   const { provider, modelId } = modelResult.resolved
+  const ipCountry =
+    (req.headers['x-vercel-ip-country'] as string | undefined) ?? undefined
```
jaco-brink: **Nit:** the trailing `?? undefined` is a no-op against a `string | undefined` cast — drop it. Header-as-`string[]` isn't a real concern for `x-vercel-ip-country` (single scalar).

apps/resources/__generated__/JourneyFields.ts
```diff
@@ -860,4 +860,5 @@ export interface JourneyFields {
    * used to display quick start label on customizable templates
    */
   customizable: boolean | null;
+  showAssistant: boolean | null;
```
jaco-brink, line 863: **Concern (unrelated drift):** this `showAssistant` field was added to the journey fragment in #8822 (`feat: customizable chat widget front end`) but `apps/resources/__generated__/*` and `apps/watch/__generated__/*` weren't regenerated at the time — `apps/journeys/__generated__/JourneyFields.ts` already has it on `main`. So the codegen change here is benign, but it's unrelated to NES-1572 (Langfuse) and adds 4 files of noise to the diff. Cleaner as a separate `chore: regenerate codegen for resources/watch` PR; if you'd rather keep them, call it out in the PR description so reviewers don't hunt for the connection.

package.json
```diff
@@ -180,6 +180,7 @@
     "isomorphic-fetch": "^3.0.0",
     "jiti": "^2.4.2",
     "keyv": "^5.3.2",
+    "langfuse": "^3.38.20",
```
jaco-brink, line 183: **Concern (FYI):** `langfuse-core@3.38.20` pulls in `mustache@4.2.0` as a runtime dep — new transitive for the workspace. Not a blocker, just flagging because it's introduced by this PR.

jaco-brink (commented):
## Review summary

Tracing lifecycle is solid — the three-terminal-path coverage (`onError`, `onFinish` incl. `finishReason==='error'`, sync `catch`) plus the `endGenerationIfPending` idempotency guard is the right shape, and the spec covers all four orderings + the no-langfuse path. Lazy/cached client + single-warn flag is clean.

**Concerns (5):** language BCP-47 vs prompt-template variable name, `pipeUIMessageStreamToResponse` onError doesn't end the generation, unrelated `showAssistant` drift in `apps/resources` + `apps/watch` codegen, per-request `flushAsync` defeats batching, new transitive dep `mustache` via `langfuse-core`. See inline.

**Nits (2):** redundant `?? undefined`, inconsistent empty-`language` handling between Langfuse + fallback paths.

**Critical:** none.

Draft, so this is a `COMMENT`. PR description test plan is also unchecked — worth running through it on stage with both env-set and env-unset Langfuse before flipping to ready.

apps/journeys/pages/api/chat/index.ts
jaco-brink: Acknowledged — keeping the per-request `flushAsync` since the Vercel serverless lambda freezes after the response stream closes; without it, batched traces drop. Worth migrating to per-request `waitUntil` once it's available on Pages Router for this app, but that's out of scope for NES-1572.

apps/resources/__generated__/JourneyFields.ts
jaco-brink: Verified — these four files (`apps/resources/__generated__/{GetJourney,JourneyFields}.ts` and `apps/watch/__generated__/{GetJourney,JourneyFields}.ts`) were missed when `showAssistant` was added to the journey fragment in #8822 (only `apps/journeys/__generated__/*` and `apps/journeys-admin/__generated__/*` got regenerated then). The autofix-ci bot regenerated them as part of this PR. Keeping them here rather than splitting since they're necessary catch-up; noting in the PR description so the connection is clear.

package.json
jaco-brink: Acknowledged — `mustache@4.2.0` is a runtime transitive of `langfuse-core@3.38.20` used for prompt template substitution. No action needed; flagging for future awareness.

jaco-brink:
## Review feedback addressed (d16237915)

**Fixed:**
- `apps/journeys/pages/api/chat/index.ts:230` — `pipeUIMessageStreamToResponse.onError` now funnels through `endGenerationIfPending` + `flushAsync`, closing the 4th terminal path so a pipe-step failure can't leave a generation dangling.
- `apps/journeys/pages/api/chat/index.ts:143` — dropped the no-op `?? undefined` after the `string | undefined` cast on `x-vercel-ip-country`.

**Fixed (adjusted):**
- `apps/journeys/pages/api/chat/index.ts:266` — coalesced both paths to **skip** the language variable when missing/empty: `compile` now receives `{}` instead of `{ language: '' }`, matching the fallback's existing skip. Spec updated to assert `{}`. The BCP-47 vs human-readable concern is a Langfuse template-authoring contract (the template is authored to accept BCP-47 codes); noted in the PR description rather than mapped in code, since the fallback path also passes BCP-47 directly.

**Acknowledged (no code change):**
- `apps/journeys/pages/api/chat/index.ts:206` — kept per-request `flushAsync` since the Vercel serverless lambda freezes after the response stream closes; `waitUntil` is a future improvement, out of scope for NES-1572. Resolved.
- `apps/resources/__generated__/JourneyFields.ts:863` — verified codegen drift is necessary catch-up from #8822 (which missed the resources/watch `__generated__` files when adding `showAssistant` to the journey fragment); the autofix-ci bot regenerated them. Called out in the PR description so reviewers don't hunt for the connection. Resolved.
- `package.json:183` — `mustache@4.2.0` is a runtime transitive of `langfuse-core` used for prompt template substitution. FYI only, no action. Resolved.

Verified: 30/30 tests pass, type-check clean, lint clean (only pre-existing warnings in unrelated files).

apps/journeys/pages/api/chat/index.ts
```diff
@@ -122,17 +131,43 @@ export default async function handler(
...
+  const trace = langfuse?.trace({
+    name: 'apologist-chat',
+    sessionId,
+    metadata: { journeyId, language, ipCountry, provider, modelId }
+  })
+  const generation = trace?.generation({
+    name: 'apologist-generation',
+    model: modelId,
+    input: modelMessages,
+    prompt: promptClient ?? undefined
```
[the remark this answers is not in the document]
jaco-brink: Challenging — capturing the full transcript as Langfuse `input` and the assistant reply as `output` is the **stated purpose** of NES-1572. Without those payloads the trace is just a request count and the integration delivers no value:

- We need full `input`/`output` to evaluate prompt template changes (`apologist-world-cup-chat`), debug hallucinations, and measure answer quality on real user questions.
- Langfuse is the org's adopted observability platform for LLM traffic; data residency and retention are governed at the platform/contract level, not per-trace.
- Per-request redaction would defeat the prompt-management workflow that this PR is wiring up — there is no useful 'minimized' version of an apologist conversation that still enables prompt iteration.

If/when a stricter consent model is introduced, it'll be enforced upstream of this handler (e.g. a flag controlling whether the chat surface is shown at all), not by sanitizing here.

jaco-brink:
## Review feedback addressed — round 2 (9b7f9033a)

**Fixed:**
- `libs/journeys/ui/src/components/AiChat/AiChat.tsx:184-191` — wrapped `sessionStorage` access in try/catch with an in-memory uuid fallback. `apps/journeys` is a public no-auth viewer that can render in restricted contexts (Safari private mode, embedded iframes, storage-disabled browsers); an uncaught throw inside `useState`'s initializer would tear down the chat render rather than just losing Langfuse session grouping.

**Challenged:**
- `apps/journeys/pages/api/chat/index.ts:149-153` (CodeRabbit major — sending full transcript to Langfuse) — by-design and the stated purpose of NES-1572. Langfuse is the org's adopted LLM observability platform; full `input`/`output` is needed to evaluate prompt template changes, debug hallucinations, and inform prompt iteration. Per-request redaction would defeat the integration's value. Data residency/retention is governed at the platform level, not per-trace. Detailed reasoning in inline reply.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -153,6 +215,11 @@ export default async function handler(
           name: err?.name,
           message: err?.message
         })
+        endGenerationIfPending({
+          level: 'ERROR',
```
edmonday: **Concern:** `void langfuse?.flushAsync()` here is fire-and-forget, while the other three terminal paths (`streamText.onError`, `streamText.onFinish`, outer `try/catch`) all `await` the flush.

In a Vercel serverless handler the function can be frozen as soon as the response stream closes, cutting off the in-flight HTTPS POST to Langfuse. So this specific path — a failure during pipe-to-response (e.g. write to a closed socket) — is the one most likely to lose its trace, even though it's exactly the kind of failure you want recorded.

The `pipeUIMessageStreamToResponse.onError` callback can be `async` (return a `Promise<string>`); the simple fix is to mark it async and `await langfuse?.flushAsync()` before returning, matching the other paths.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -53,7 +59,7 @@ function resolveChatModel():
       apiKey
     })
     const modelId =
-      process.env.OPENROUTER_MODEL ?? 'google/gemini-2.5-flash-lite'
+      process.env.OPENROUTER_MODEL ?? 'google/gemini-3-flash-preview'
```
edmonday, line 62: **Concern (optional):** default bumped from `google/gemini-2.5-flash-lite` to `google/gemini-3-flash-preview`. This default is hit any time `CHAT_PROVIDER=openrouter` is set without `OPENROUTER_MODEL`. Blast radius is narrow today (apologist is the prod default and flag-gated), but defaulting to a `-preview` SKU is the kind of footgun that's easy to set now and forget later. Worth confirming this is intentional vs. a leftover from a diagnostic run — if OpenRouter is dev-only, ignore.

edmonday (commented):
## Review summary

Overall in good shape — thorough self-review (6 author-filed threads already resolved), solid lifecycle test coverage, sessionStorage properly guarded.

**One worth fixing before merge:** the pipe-step `onError` flushes Langfuse fire-and-forget while every other terminal path awaits — see inline.

**Two optional discussion items:** OpenRouter default model bumped to a `-preview` SKU (inline), and CodeRabbit's still-open data-residency thread on transcript-to-Langfuse that's worth answering in the PR thread.

Event: `COMMENT` (no blocking findings).

edmonday (dismissed):

apps/journeys/pages/api/chat/index.ts
jaco-brink: **Update — earlier Option B refactor reverted; see commit `40ec02bb7`.**

You're right that fire-and-forget on the pipe `onError` was a real concern. The first attempt was to extend the handler's lifetime past response close and do a single awaited `flushAsync` at the end, so every code path went through the same shipping channel.

Two things came out of testing that on the preview deploy:

1. **The mid-stream client disconnect did not produce a Langfuse trace** even with the post-response flush. Lambda freeze on Vercel is more aggressive than the await-and-then-flush dance can handle on the pipe-error path.
2. **Stepping back: Langfuse is the wrong tool for that anyway.** It's an LLM observability platform — it's there for prompt iteration, response quality, token usage. Pipe-step failures (write to closed socket etc.) are infrastructure errors and they belong in Datadog, not Langfuse.

So the simplified shape now:

- `streamText.onError` / `streamText.onFinish`: still update Langfuse + `flushAsync` (these *are* LLM lifecycle events).
- `pipe.onError`: only logs (`console.error` placeholder) and returns a generic `'stream failed'` to the client. **No Langfuse update**, and we no longer leak the raw error message into the SSE response.
- Outer `catch` (sync throws): still ends the generation in Langfuse for lifecycle hygiene (the trace + generation were created before we got here), and the client-facing 500 body is genericised.

The structural change you and I both wanted is real — it's just that the right home is Datadog via pino + dd-trace, which `apps/journeys` doesn't have wired up yet (other apps in the repo do — `journeys-admin`, `arclight`, `api-journeys`). I've filed [NES-1615](https://linear.app/jesus-film-project/issue/NES-1615/wire-pino-dd-trace-server-logger-into-appsjourneys) in the **Cleanup & Tech Debt** milestone to do that properly, and left `TODO(NES-1615)` markers at every `console.error` site in the handler so the follow-up has obvious landing spots.

Verification: `apps/journeys/pages/api/chat/TESTING.md` is removed; the canonical happy + error scenarios still pass via the existing unit tests (30/30) and on the preview deploy. Pipe-error verification will move with NES-1615 to Datadog where it belongs.

apps/journeys/pages/api/chat/index.ts
jaco-brink: similar cost, with 3-flash having better outputs.
since this is mainly used of dev and qa not to concerned if preview stops work, we'll toggle to another model if that happens

jaco-brink:
## Review feedback addressed — round 4 (40ec02bb7)

Closing out three remaining threads. All conversations were already complete; no new code changes.

**Resolved (no code change):**
- `apps/journeys/pages/api/chat/index.ts:153` (CodeRabbit, sensitive transcripts to Langfuse) — challenged earlier with the NES-1572 design rationale; CodeRabbit acknowledged and dropped the concern. Resolved.
- `apps/journeys/pages/api/chat/index.ts` (Edmond, fire-and-forget `flushAsync` on `pipe.onError`) — addressed by the Option B revert + Langfuse scope-narrow + [NES-1615](https://linear.app/jesus-film-project/issue/NES-1615/wire-pino-dd-trace-server-logger-into-appsjourneys) follow-up. Thread is outdated; the code being commented on no longer exists. Resolved.
- `apps/journeys/pages/api/chat/index.ts:62` (Edmond, openrouter default model bump) — accepted the trade-off (similar cost, better outputs, dev/qa-only path; will toggle if preview model is deprecated). Resolved.

edmonday (approved):
Approved. All 11 review threads resolved. Tests for the new AI chat components are tracked separately under NES-813.

PR #9114, opened 2026-04-30
chore: stepvieweventcreate modern

mikeallisonJS:
## Summary by CodeRabbit

* **New Features**
  * Added step view event mutation to track when users view journey steps and automatically update duration metrics and timestamps.

* **Tests**
  * Added comprehensive test suite for the new step view event creation mutation.


mikeallisonJS:
@coderabbit review

csiyang:
Behaviour change vs legacy and sibling modern resolvers — the new `stepViewEventCreate.mutation.ts` (lines 32-41) doesn't persist `Event.blockId`. Legacy `step.resolver.ts` (lines 40-47) includes it via the `...input` spread (which passes through `eventService.save`, a pure `prisma.event.create({ data: input })` passthrough — no field filtering), and modern siblings `chatOpenEventCreate` (via `...input`) and `textResponseSubmissionEventCreate` (explicit `blockId: input.blockId`) both persist it. So modern-created `StepViewEvent` rows will have `blockId = null` while every other StepView row in the DB has it set.

Practical risk is low — most readers gate on submission-event typenames (Google Sheets sync, default visitor-export filters; dashboard transform reads from Plausible). But the visitor export at `journeyVisitor.ts:285-291` accepts arbitrary `filter.typenames` with a `blockId: { not: null }` filter, so any caller passing `StepViewEvent` would silently drop modern-created rows.

One-line fix — add `blockId: input.blockId` to the `prisma.event.create` data.

csiyang:
Error type inconsistency — `stepViewEventCreate.mutation.ts:21-23` throws `new Error('User not authenticated')` for the unauthenticated branch, but other modern event resolvers like `chatOpenEventCreate.mutation.ts:26-30` use `new GraphQLError('User not authenticated', { extensions: { code: 'UNAUTHENTICATED' } })`. Cosmetic but breaks any client that keys off the GraphQL error code instead of the message string.

mikeallisonJS:
## Review feedback addressed (d0728e798)

**Fixed:**
- `stepViewEventCreate.mutation.ts:32` — Added `blockId: input.blockId` to `prisma.event.create` data to match legacy and sibling modern resolvers (chatOpenEventCreate, textResponseSubmissionEventCreate). Prevents `blockId = null` on modern-created StepViewEvent rows.
- `stepViewEventCreate.mutation.ts:21-23` — Removed redundant user auth check; already handled by `withAuth` scope auth.

**Challenged:** None

**Skipped:** None

csiyang (approved):
LGTM — both comments addressed. Cleaner fix than suggested on the auth check.

PR #9120, opened 2026-04-30
fix: language autocomplete workaround

mikeallisonJS:
Prod db is returning multiple primary languages, this should work around it

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced language autocomplete search to match results across multiple language name formats, improving ability to find languages using various naming conventions.
  * Improved language name extraction to more reliably handle edge cases.

* **Tests**
  * Added comprehensive test coverage for language autocomplete search behavior and keyboard navigation.


csiyang (dismissed):
Looks good — fix is correct for the all-primary name edge case.

csiyang:
Nice fix! One thing to flag for a follow-up — `MultipleLanguageAutocomplete.tsx` (at `libs/shared/ui/src/components/MultipleLanguageAutocomplete/MultipleLanguageAutocomplete.tsx:48-49`) has the same `name.find(({ primary }) => primary)` extraction logic that this PR fixes in `LanguageAutocomplete`. If those affected language records also appear in the multi-select picker, they'll have the same broken display. Worth either replicating this fix there or extracting a shared `extractLanguageNames` util that both components use.

csiyang (approved):
Re-approving after shared util extraction — clean refactor, MultipleLanguageAutocomplete parity closed.

PR #9122, opened 2026-05-01
feat(journeys): apply Claude Design mobile UX to apologist chat (NES-1617)

jaco-brink:
## Summary

Implements the Claude Design mobile UX for the apologist chat pinned bar in `libs/journeys/ui` so Aaron and Sway can demo a working stage link during regional-leader presentation week, instead of reviewing a static mockup.

Closes NES-1617. Mobile only — desktop polish, accessibility audit, and content/prompt changes remain out of scope.

## What changed

**New modules** (under `libs/journeys/ui/src/components/AiChat/`):
- `tokens.ts` — chat-only design tokens. Greys map to `base/tokens/colors.ts` verbatim; `brandRed (#C52D3A)` is the journeysAdmin primary kept local since there's no shared journeyUi token for it.
- `ChatHeader/` — sparkle avatar (32px brand-red gradient + inset highlight + outer red glow), title, caption, divider, faint red header wash. Supports an optional close button for future takeover use.
- `DragHandle/` — extracted from `AiChat.tsx`. Tap (<4px) toggles, drag-down (>32px) collapses, thumb darkens while dragging, keyboard accessible (Enter/Space).

**Restyled in place** (token-driven, no purple anywhere):
- `Message.tsx` — user bubble `grey.700` / 16px all-corner radius; assistant bubble `grey.200` / 16-16-16-4 radius.
- `PromptInput.tsx` — 22px capsule, 44px min-height, `inputFill` background, 40×40 circular send button. Retains `font-size: 16px` (M1 iOS Safari zoom guard).
- `Conversation.tsx` — uses `useStickToBottom`'s `isAtBottom` + `scrollToBottom` to render a "scroll to latest" pill when the user scrolls up.
- `AiChat.tsx` — drops the inline drag-handle/close-X markup, mounts `DragHandle` + `ChatHeader` for the `panel` variant, exposes `onActiveChange` callback so the sheet can resize, bumps typewriter cadence to 280 cps.
- `PinnedChatBar.tsx` — 16px top corners, sheet top shadow, `auto`-height when idle and `80%` when active (driven by `onActiveChange`), `env(safe-area-inset-bottom)` only (no double-padding when the iOS toolbar is visible).

The existing `PinnedChatBar` `data-testid` contract is preserved, so the existing `Conductor.apologistChat.spec.tsx` mocks and assertions continue to apply.

## Test plan

- [ ] Open a journey with `apologistChat` flag on and `showAssistant=true`, on iOS Safari (real device) — confirm idle bar hugs the bottom, drag handle visible, header reads "Ask a question / Replies may not be perfect", input is the 22px capsule.
- [ ] Send a message — bar expands to 80% height, typewriter cadence feels right (~280 cps), typing dots show before stream starts, send button switches to stop while streaming.
- [ ] Scroll up mid-conversation — "scroll to latest" pill appears bottom-center; tapping it sticks back to bottom.
- [ ] Drag handle down >32px — bar collapses to idle, conversation hides, input remains visible.
- [ ] Tap drag handle — same collapse behaviour.
- [ ] Repeat on Android Chrome and Samsung Internet.
- [ ] Confirm no regression on the iOS Safari input-focus zoom (`font-size >= 16px` on the input).
- [ ] Confirm no regression on existing M1 prototype behaviour for users already in the chat.
- [ ] Stage URL works for Aaron and Sway.

## Out of scope (separate tickets)

- Desktop polish.
- Takeover direction (handoff includes both; product chose Pinned for this slice).
- Accessibility audit.
- Content / prompt changes (NES-1572 + v2 prompt).
- NES-1606 markdown styling (folds in on top of `Response` later).

## Notes

- Pre-existing on `main`: `next-i18next/pages` doesn't resolve in jest, so any spec that touches `Conductor` / `StepFooter` (including `Conductor.apologistChat.spec.tsx`) fails before this PR. Verified by stashing this branch's changes and running the baseline. Not introduced here.
- TypeScript is clean for the touched files. Remaining repo-wide TS errors (storybook 10 stories, `@xyflow/react` migration, the `next-i18next/pages` issue above) all pre-date this branch.

## Summary by CodeRabbit

* **New Features**
  * Sheet-style chat with drag-to-expand/collapse, overlay close control, and sheet lifecycle callbacks
  * Chat header with animated thinking indicator and assistant surface theming
  * Auto-scroll with bottom fade and "scroll to latest" pill

* **Improvements**
  * Multi-line auto-growing prompt input and refreshed submit/stop controls
  * Theme-aware message surfaces, improved typing indicator visuals and faster type animation
  * Centralized chat styling tokens and overlay-specific retry/error styling

* **Tests**
  * Added unit tests for header, drag handle, and conversation behaviors


libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -101,33 +115,30 @@ interface AssistantBubbleProps {
...
   animate,
   isStreaming,
-  plain = false
+  surface = 'light'
 }: AssistantBubbleProps): ReactElement {
   const { display, isComplete } = useTypewriter(text, animate, isStreaming)
   return (
     <>
-      <Message role="assistant" plain={plain}>
+      <Message role="assistant" plain surface={surface}>
```
jaco-brink, line 149: **[Blocker] Assistant bubble is never actually rendered in panel mode.**

`plain` is hardcoded on every assistant `Message` here (and on `TypingIndicator` / error message branches at lines 353 and 363). But `Message.tsx` gates the new bubble on `!isPlainAssistant`:

```tsx
bgcolor: isUser ? PRIMARY : isPlainAssistant ? 'transparent' : ASSISTANT_BG
borderRadius: isUser ? '16px' : isPlainAssistant ? 0 : '16px 16px 16px 4px'
```

So the `ASSISTANT_BG` (`grey.200`) bubble with `16-16-16-4` radius described in the PR description is dead code — assistant text always renders as plain prose in both panel and overlay. Either the description is wrong (panel = plain) or the call sites should drop `plain` for the panel surface. Please clarify and align.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -101,33 +115,30 @@ interface AssistantBubbleProps {
...
     </>
   )
 }
 
 function TypingIndicator(): ReactElement {
-  const { t } = useTranslation('libs-journeys-ui')
   return (
     <Box
-      aria-label={t('Assistant is typing')}
+      aria-label="Assistant is typing"
```
jaco-brink: **[Blocker] i18n regression on `TypingIndicator` aria-label.**

This was previously `aria-label={t('Assistant is typing')}`. The hardcoded English string ships to non-English locales. `useTranslation` was removed from this function presumably to cut a hook call — restore it (or thread `t` in from the caller).

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -183,15 +200,11 @@ export function AiChat({
...
-      window.sessionStorage.setItem('aiChat.sessionId', fresh)
-      return fresh
-    } catch {
-      return uuidv4()
-    }
+    const existing = window.sessionStorage.getItem('aiChat.sessionId')
+    if (existing != null && existing.length > 0) return existing
+    const fresh = uuidv4()
+    window.sessionStorage.setItem('aiChat.sessionId', fresh)
+    return fresh
```
jaco-brink: **[Blocker] `sessionStorage` access lost its `try/catch`.**

The previous code intentionally caught failures (Safari private mode, blocked third-party storage, quota errors) and fell back to a fresh UUID. As written, this `useState` initializer will throw out of render in those environments and crash the chat surface. Mobile Safari private tabs are a real demo risk; restore the `try { ... } catch { return uuidv4() }` wrapper.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -258,7 +271,24 @@ export function AiChat({
...
+  // Collapse wins over message presence so the user can dismiss the
+  // sheet from idle (empty chat) as well as from active.
+  const sheetState: AiChatSheetState = collapsed
+    ? 'collapsed'
+    : hasMessages
+      ? 'active'
+      : 'idle'
+  useEffect(() => {
+    onSheetStateChange?.(sheetState)
+  }, [sheetState, onSheetStateChange])
```
jaco-brink, line 313: **[Notable] One-frame lag between `AiChat`'s `sheetState` and `PinnedChatBar`'s `height`.**

`sheetState` is derived synchronously, but the parent is notified via this `useEffect`, so `PinnedChatBar`'s height update lags by a render. The CSS transition hides most of it, but the `sheetState === 'collapsed'` styling on the floating input below (lines 405, 424–426) ticks on the same render the parent height starts updating — can produce a subtle "input slides before sheet shrinks" beat on the first collapse. Consider lifting `sheetState` to the parent and letting `AiChat` accept it as a prop, with a setter callback for drag events.

libs/journeys/ui/src/components/Conversation/Conversation.tsx
```diff
@@ -1,52 +1,188 @@
...
+  // content. Content resizes happen when streaming text grows or
+  // images load (need to recheck because scrollHeight changed beneath
+  // the user). Container resizes happen when the parent sheet
+  // collapses/expands (clientHeight changes — without this the pill
+  // would keep rendering against a 0-height scrollable area).
+  useEffect(() => {
+    const scrollEl = scrollRef.current
+    const contentEl = contentRef.current
+    if (scrollEl == null || contentEl == null) return undefined
+    const observer = new ResizeObserver(() => evaluateAtBottom())
```
jaco-brink: **[Blocker] Streaming auto-scroll regression.**

Replacing `useStickToBottom` with manual tracking means content growth during a single streamed message no longer pins the user to the bottom. Concretely:

1. User is at the bottom while the assistant streams.
2. Text grows downward — they're no longer within the 24px `NEAR_BOTTOM_THRESHOLD_PX`.
3. The `ResizeObserver` here flips `isAtBottom` to `false` mid-message.
4. The "scroll to latest" pill appears during the same answer they were already reading, and they have to tap it to keep up.

That's the opposite of the goal. Two clean options:
- Restore `useStickToBottom`'s behaviour.
- Keep the manual tracker but, inside this `ResizeObserver` callback, call `scrollToBottom(false)` when `wasNearBottomRef.current` is true (then re-evaluate).

libs/journeys/ui/src/components/PromptInput/PromptInput.tsx
```diff
@@ -70,52 +90,54 @@ export function PromptInput({
...
         boxShadow: isFloating
           ? '0 10px 30px rgba(0, 0, 0, 0.25), 0 1px 3px rgba(0, 0, 0, 0.1)'
-          : 'none'
+          : '0 6px 18px rgba(38, 38, 46, 0.16), 0 1px 3px rgba(38, 38, 46, 0.10)'
       }}
     >
-      <TextField
-        inputRef={textareaRef}
+      <Box
+        component="input"
```
jaco-brink: **[Blocker] Multi-line input lost.**

Swapping MUI `TextField multiline maxRows={6}` for a plain `<input type="text">` means:
- Users can't compose multi-paragraph questions.
- `Shift+Enter` no longer adds a newline (single-line input — it just submits).
- Long pasted content scrolls horizontally instead of growing.

If the design genuinely calls for single-line, fine — but the previous behaviour was a deliberate UX call (apologist questions are often nuanced and multi-line). Worth confirming with product before locking it in. Easiest fix: keep the visual capsule (Box wrapper styles) but render a `<textarea>` inside with `rows=1` + auto-grow up to ~6 rows.

libs/journeys/ui/src/components/AiChat/ChatHeader/ChatHeader.tsx
```diff
@@ -0,0 +1,156 @@
...
+  SPARKLE_GRADIENT,
+  TEXT_SECONDARY
+} from '../tokens'
+
+interface ChatHeaderProps {
+  /** When provided renders a close button on the right (used by takeover surfaces). */
+  onClose?: () => void
+}
+
+export function ChatHeader({ onClose }: ChatHeaderProps): ReactElement {
```
jaco-brink: **[Notable] Unused `onClose` prop.**

This prop and the close-button branch (lines 139–155) ship without a caller — explicitly framed as "for future takeover use". Per project convention (no premature abstractions), drop it until takeover lands. Easier to add with the actual usage than to maintain dead surface area.

libs/journeys/ui/src/components/AiChat/ChatHeader/ChatHeader.tsx
```diff
@@ -0,0 +1,156 @@
...
+            // transform-box/origin keeps each path scaling around its own
+            // centre inside the SVG coordinate space.
+            '& .jfp-mark-shape': {
+              transformBox: 'fill-box',
+              transformOrigin: 'center'
+            },
+            '& .jfp-mark-top': {
+              animation: 'jfpMarkTop 1.6s ease-in-out infinite'
+            },
+            '& .jfp-mark-bottom': {
```
jaco-brink, line 80: **[Notable] Continuous SVG animation runs even at rest.**

The "two-beat thinking rhythm" runs every render — including when the bar is idle and the user isn't engaging the chat. On mid-tier mobile devices this is a continuous repaint of a transformed SVG inside a fixed-position bar. `prefers-reduced-motion` is honoured (good), but consider gating the animation on `isStreaming` / loading state rather than always-on. It will also feel more meaningful — a static avatar at rest, alive when thinking.

libs/journeys/ui/src/components/Response/Response.tsx
```diff
@@ -10,7 +10,79 @@ interface ResponseProps {
...
+        '& li': {
+          marginBottom: '2px',
+          '& > p': { marginBottom: '2px' }
+        },
+        '& strong': { fontWeight: 600 },
+        '& em': { fontStyle: 'italic' },
+        '& a': {
+          color: 'inherit',
+          textDecoration: 'underline',
+          textDecorationColor: 'rgba(38,38,46,0.35)',
```
jaco-brink: **[Notable] Markdown styles assume a light surface.**

Here and at lines 61, 69, 80 — `rgba(38,38,46,…)` is hard-coded. These will be invisible / illegible on the desktop overlay's dark backdrop. Either drive these from `currentColor` + opacity, or accept a `surface` prop the way `Message` now does. Out of scope for the mobile cut, but mark as known follow-up rather than letting overlay regress silently.

jaco-brink (commented):
## Code review — NES-1617 mobile UX

Structural refactor (token extraction, `ChatHeader`/`DragHandle` split, `PinnedChatBar` state machine) is well done and reads cleanly. Five blocking items concentrated in specific places, plus a handful of notable concerns and nits — see inline comments and the additional points below.

### Additional concerns not pinned to a single line

**No tests for the new components.** `DragHandle` and `ChatHeader` are new modules with non-trivial logic (drag thresholds, keyboard a11y, animation gates). The new "scroll to latest" pill in `Conversation` is also untested. Note: `Conductor.apologistChat.spec.tsx` is broken on `main` due to a pre-existing `next-i18next/pages` resolution issue — fair, but that doesn't preclude leaf tests for the new pieces. At minimum:
- `DragHandle.spec.tsx` — drag-down >32px calls `onCollapse`, drag-up >32px calls `onExpand`, sub-threshold does nothing, Enter/Space toggles.
- `ChatHeader.spec.tsx` — renders title + caption; close button only when `onClose` is supplied.
- `Conversation.spec.tsx` — pill appears when scrolled away from bottom and disappears when at bottom.

**`useEffect` for `onSheetStateChange` causes a one-frame lag.** `sheetState` is derived synchronously, but the parent is notified via `useEffect`, so `PinnedChatBar`'s `height` lags by one render. The CSS transition mostly hides this, but the `sheetState === 'collapsed'` styling on `AiChat` itself ticks on the same render the parent height starts updating, which can cause a subtle "input slides before sheet shrinks" beat on the first transition. Consider lifting `sheetState` to the parent and letting `AiChat` accept it as a prop, with a setter callback for drag events. Optional polish.

### Nits

- **Magic `72` in `bottomClearance={isPanel ? 72 : 0}`** at `AiChat.tsx:329` — `IDLE_HEIGHT_PX` and `COLLAPSED_HEIGHT_PX` have helpful comments; this one deserves the same (input pill height + safe-area + 8px breathing room?).
- **`brandRed` duplication** — fine as-is, but consider exporting it from the journeys theme later so a single change updates both surfaces.
- **`DragHandle` `cursor: 'grab'`** at `DragHandle.tsx:132` is meaningless on touch devices but harmless. Optional: gate behind a hover media query.
- **`ChatLoadingSkeleton`** at `PinnedChatBar.tsx:27` — nice touch, but it's now ~80 lines of structural duplication of `ChatHeader` / `DragHandle` / `PromptInput`. If the real shapes drift, the skeleton will look subtly wrong. Acceptable for the demo; consider compositing skeleton variants of the real components later.
- **`PromptInput` placeholder** changed from `Ask me anything` to `Ask anything…` — verify with product/copy that this is intentional.
- **`PromptInput` Enter handler** drops the `!e.shiftKey` guard — fine for `<input>`, but if you ever go back to a textarea this regression is silent.

### Test plan delta

- Verify in **Safari private browsing** (sessionStorage throws) — currently breaks per blocker #3.
- Verify the "scroll to latest" pill **does not appear during a single streamed response** when the user was at the bottom when the response started.
- Confirm assistant bubble vs plain prose matches design intent on a real device, given blocker #1.

### Summary

Resolve the five blocking items, restore the `try/catch`, add leaf tests for the two new components, and this is in good shape for the demo cut.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
jaco-brink: Skipping for this round — flagged on the original review as 'optional polish' and the CSS height transition masks it well enough for the demo cut. Will revisit if the staging walkthrough exposes the input/sheet timing seam.

libs/journeys/ui/src/components/Response/Response.tsx
jaco-brink: Skipping for this round — explicitly out-of-scope per the original review (overlay markdown styling sits behind the desktop polish ticket). Tracking as a follow-up rather than expanding the mobile-cut PR.

jaco-brink:
## Review feedback addressed (05dcc24d5)

**Fixed (blockers):**
- `AiChat.tsx:130` — drop hardcoded `plain` on assistant `Message` so the panel surface renders the spec'd `ASSISTANT_BG` / 16-16-16-4 bubble; overlay still gets plain prose.
- `AiChat.tsx:141` — restore `t('Assistant is typing')` aria-label.
- `AiChat.tsx:207` — restore `try/catch` around `sessionStorage` access; falls back to a fresh UUID in private mode / sandboxed iframes.
- `Conversation.tsx` — split the `ResizeObserver` so content growth pins the reader to the bottom while they were already there (`useStickToBottom`-equivalent), without stealing the screen from a reader who's scrolled up.
- `PromptInput.tsx` — restore textarea with auto-grow up to ~6 rows and `Shift+Enter` newline; iOS zoom guard kept.

**Fixed (notable):**
- `ChatHeader.tsx` — drop unused `onClose` prop + close button (will reintroduce with the takeover surface that needs it).
- `ChatHeader.tsx` — gate the sparkle-mark animation on a new `thinking` prop driven by `isLoading`, so the avatar is alive when the assistant works and at rest otherwise.

**Tests added:**
- `DragHandle.spec.tsx` — aria-label per state, drag-down >32px → onCollapse, drag-up >32px → onExpand, sub-threshold no-op, Enter/Space toggling.
- `ChatHeader.spec.tsx` — title/caption render, both `thinking` states. `next-i18next/pages` mocked virtually so the new specs run despite the pre-existing jest infra issue.

**Skipped (deferred, see inline replies):**
- `AiChat.tsx:283` — one-frame `useEffect` lag on `onSheetStateChange`; optional polish, the CSS transition masks it for the demo.
- `Response.tsx:52` — markdown styles assuming a light surface; overlay follow-up rather than blocking the mobile cut.

**Nits noted in the original review** (placeholder copy / brandRed duplication / cursor on touch / skeleton duplication): tracked as follow-ups; addressing the `bottomClearance={isPanel ? 72 : 0}` magic number with a comment.

jaco-brink:
Addressed the Response markdown styling review comment in fba3e9922. The markdown underline, code background, and blockquote border now use named currentColor-based constants, so those accents inherit the active message color and stay legible on both the light mobile panel and dark desktop overlay.

libs/journeys/ui/src/components/Conversation/Conversation.tsx
```diff
@@ -1,52 +1,222 @@
...
-        },
-        '&::-webkit-scrollbar-thumb:hover': {
-          backgroundColor: 'rgba(128, 128, 128, 0.7)'
-        }
-      }}
-    >
+    <Box sx={{ position: 'relative', flex: 1, minHeight: 0, display: 'flex' }}>
       <Box
-        ref={contentRef}
+        ref={scrollRef}
```
edmonday, line 145: **Concern:** 200+ new lines of behavioural code here ship without any spec — `scrollKey`-driven scroll, dual `ResizeObserver`s, near-bottom threshold, `clientHeight===0` short-circuit, mask gradient. `DragHandle` and `ChatHeader` got new specs; `Conversation` is the higher-risk delta and has none.

Worth at least one spec covering:
- pill renders only when not at bottom
- pill hidden when `clientHeight===0` (collapsed sheet)
- scroll on `scrollKey` change only when previously near bottom

Easy to regress otherwise.

libs/journeys/ui/src/components/PromptInput/PromptInput.tsx
```diff
@@ -70,52 +116,53 @@ export function PromptInput({
...
-        inputRef={textareaRef}
+      <Box
+        component="textarea"
+        ref={textareaRef}
+        rows={1}
         value={input}
-        onChange={(e) => onInputChange(e.target.value)}
-        placeholder={t('Ask me anything')}
+        onChange={handleInputChange}
+        onKeyDown={handleKeyDown}
```
edmonday, line 123: **Concern:** This replaces MUI `TextField` with a raw `Box component="textarea"`. The original auto-grow regression made the swap necessary, so I get the why — but it loses `TextField`'s a11y wiring (label association, helper-text plumbing, error states) and bypasses the project-wide "prefer MUI components over raw HTML" convention.

Would `OutlinedInput` (or `InputBase`) work here instead, with the same auto-grow effect applied? Then we keep the convention and the auto-grow fix.

libs/journeys/ui/src/components/PromptInput/PromptInput.tsx
```diff
@@ -141,18 +188,30 @@ export function PromptInput({
...
+            height: 32,
             flexShrink: 0,
             p: 0,
-            bgcolor: canSubmit ? '#6D28D9' : '#e0e0e0',
-            color: canSubmit ? '#ffffff' : '#999',
+            bgcolor: canSubmit
+              ? PRIMARY
+              : isFloating
+                ? OVERLAY_FILL_LOW
+                : DIVIDER,
```
edmonday, line 186: **Concern:** The `&:hover` branch and the `&.Mui-disabled` branch resolve to the same colours when `!canSubmit`, so hovering a disabled send button is a no-op visually. Probably the intent (don't tease an action you can't take), but worth confirming with design — usually disabled controls still get a subtle hover differentiation.

edmonday (commented):
## Review summary

No Critical findings. A few **Concerns** worth discussing before merge — most notably the lack of test coverage for the new `Conversation` behaviour. Token consolidation and follow-through on prior review threads is genuinely good. Posted as COMMENT.

**Concerns:** 5 inline (Conversation tests, raw textarea, disabled hover, drag-then-return edge case, color-mix support).
**Nit:** 1 inline (chatStyles theme-path vs literal split).
**Positive:** good resolution of all prior blocker threads; thoughtful trade-off comments throughout.

Out-of-scope items (desktop polish, a11y audit, content) honoured per PR description.

libs/journeys/ui/src/components/AiChat/DragHandle/DragHandle.tsx
```diff
@@ -0,0 +1,178 @@
...
+      startYRef.current = null
+      movedRef.current = 0
+      if (moved > DRAG_THRESHOLD_PX) {
+        suppressClickRef.current = true
+        onCollapse()
+      } else if (moved < -DRAG_THRESHOLD_PX) {
+        suppressClickRef.current = true
+        onExpand()
+      }
+    }
```
edmonday, line 75: **Concern (edge case):** `movedRef` tracks current offset, not max displacement. If a user drags down 100px then back up to ~0 before releasing, `moved` at `mouseup` is ~0, neither threshold fires, and `suppressClickRef` stays false — so the synthetic click then *toggles* the sheet. After a clearly intentional drag, a tap-toggle on release is surprising.

Low priority. Consider tracking max absolute displacement during the drag and suppressing the click if it ever crossed the threshold.

libs/journeys/ui/src/components/Response/Response.tsx
```diff
@@ -8,9 +8,90 @@ interface ResponseProps {
   content: string
 }
 
+const MARKDOWN_LINK_UNDERLINE =
+  'color-mix(in srgb, currentColor 35%, transparent)'
+const MARKDOWN_SUBTLE_BACKGROUND =
+  'color-mix(in srgb, currentColor 6%, transparent)'
+const MARKDOWN_BLOCKQUOTE_BORDER =
+  'color-mix(in srgb, currentColor 15%, transparent)'
+
```
edmonday, line 17: **Concern:** `color-mix(in srgb, …)` requires Chrome 111+ / Safari 16.2+ / Firefox 113+. Fine for the stage-demo audience, but no fallback colour is declared, so older browsers will get the unstyled inherit only. Worth confirming the demo device set is all on supported versions, or adding a fallback (e.g. a static rgba constant in `chatStyles.ts`) before this rolls beyond the demo.

libs/journeys/ui/src/components/AiChat/chatStyles.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export const MUTED_FG = 'text.secondary'
+
+// --- Brand ---
+
+/** Second stop of the sparkle-avatar gradient. Darker than journeysAdmin's primary.dark, tuned to the gradient. */
+export const BRAND_RED_DARK = '#A8202C'
+
+/** Sparkle-avatar fill — brand red shaded toward BRAND_RED_DARK. */
+export const SPARKLE_GRADIENT = `linear-gradient(135deg, ${brandRed} 0%, ${BRAND_RED_DARK} 100%)`
```
edmonday, line 47: **Nit:** This file mixes theme-path strings (`PRIMARY = 'grey.700'`) and concrete literals (`SHEET_TOP_SHADOW = '0 -8px …'`) under one umbrella. Consumers use both via the same `sx`-form import. Future contributors will easily inline `bgcolor: PRIMARY` the same way they would `boxShadow: SHEET_TOP_SHADOW` and not realise one resolves through `theme.palette` and the other doesn't.

Consider grouping into two named exports — e.g. `chatPalette` (theme paths) and `chatTokens` (literals) — so the distinction is visible at the import site. Not blocking.

libs/journeys/ui/src/components/Conversation/Conversation.tsx
jaco-brink: Addressed in db3ee6a2c3. Added `libs/journeys/ui/src/components/Conversation/Conversation.spec.tsx` covering exactly the three cases you flagged plus a couple of supporting ones:

- pill hidden when at bottom
- pill renders when scrolled away
- pill hidden when `clientHeight === 0` (sheet collapsed) — fired via captured ResizeObserver callback
- scrolls on `scrollKey` change when previously near bottom
- does **not** scroll on `scrollKey` change when reader has scrolled away
- pill click triggers smooth scroll-to-bottom

All 6 pass.

libs/journeys/ui/src/components/PromptInput/PromptInput.tsx
jaco-brink: Addressed in db3ee6a2c3. Swapped the raw `Box component="textarea"` for `<InputBase multiline maxRows={6}>`. The native auto-grow that `TextareaAutosize` provides covers the original regression, so the manual `useEffect` + `textareaRef` are gone. Net -13 lines. Preserved:

- 16px font-size (iOS Safari zoom guard from M1)
- `fontFamily: inherit` so the chat surface still tracks the journey typography
- 40px capsule min-height (via padding maths on the root)
- Enter-submits / Shift+Enter newline
- Disabled colour via `WebkitTextFillColor` so Safari respects it

A11y wiring (focus management, label/error semantics, `FormControl` integration if we ever need it) is back.

libs/journeys/ui/src/components/PromptInput/PromptInput.tsx
jaco-brink: Deferring to a follow-up milestone — captured under [NES-1641 — General UX improvements (apologist chat mobile)](https://linear.app/jesus-film-project/issue/NES-1641/general-ux-improvements-apologist-chat-mobile) in M3.

libs/journeys/ui/src/components/AiChat/DragHandle/DragHandle.tsx
jaco-brink: Deferring to a follow-up milestone — captured under [NES-1641 — General UX improvements (apologist chat mobile)](https://linear.app/jesus-film-project/issue/NES-1641/general-ux-improvements-apologist-chat-mobile) in M3.

libs/journeys/ui/src/components/Response/Response.tsx
jaco-brink: Deferring to a follow-up milestone — captured under [NES-1641 — General UX improvements (apologist chat mobile)](https://linear.app/jesus-film-project/issue/NES-1641/general-ux-improvements-apologist-chat-mobile) in M3. Worth adding the `rgba` fallback in `chatStyles.ts` before this rolls beyond the regional-leader stage demo — i.e. before M5 public launch.

libs/journeys/ui/src/components/AiChat/chatStyles.ts
jaco-brink: Deferring to a follow-up milestone — captured under [NES-1641 — General UX improvements (apologist chat mobile)](https://linear.app/jesus-film-project/issue/NES-1641/general-ux-improvements-apologist-chat-mobile) in M3.

libs/journeys/ui/src/components/Conversation/Conversation.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+    if (originalScrollTo == null) {
+      delete (Element.prototype as { scrollTo?: typeof Element.prototype.scrollTo }).scrollTo
+    } else {
+      Element.prototype.scrollTo = originalScrollTo
+    }
+  })
```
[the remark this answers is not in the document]
jaco-brink: Valid quick win — addressed in b5f7f95143. Stored `originalResizeObserver` in `beforeEach` and restore it in `afterEach`, mirroring the existing pattern for `Element.prototype.scrollTo`. All 6 Conversation specs still pass.

edmonday (dismissed):
Approved.

Both Concerns I asked to address in this PR are fixed with clean, MUI-conventional code:

- **`Conversation` tests** (`db3ee6a2c3` + `b5f7f9514`) — 171-line spec covers pill visibility at-bottom / scrolled / `clientHeight===0` collapsed, smooth-scroll on `scrollKey` change only when near bottom, and click-to-scroll. Mock teardown is symmetric (ResizeObserver + scrollTo both restored).
- **`PromptInput` → `InputBase multiline maxRows={6}`** (`db3ee6a2c3`) — drops the raw `Box component="textarea"`, removes the manual `useEffect`/ref auto-grow, preserves the 16px iOS-Safari zoom guard. Net −13 lines.

The remaining 3 Concerns + 1 Nit deferred to NES-1641 (M3) are all reasonable for that scope — none block the demo. Nice work on the self-correcting follow-up commit for the ResizeObserver cleanup.

libs/journeys/ui/src/components/Conversation/Conversation.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  it('does not render the scroll-to-bottom pill while the reader is at the bottom', () => {
+    mockScrollMetrics({ scrollHeight: 400, scrollTop: 0, clientHeight: 400 })
+    const { queryByTestId } = render(
+      <Conversation scrollKey={1}>
+        <div>message</div>
+      </Conversation>
+    )
+    expect(queryByTestId('ScrollToBottomPill')).not.toBeInTheDocument()
+  })
```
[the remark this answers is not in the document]
edmonday: @jaco-brink this might be worth adding real quick.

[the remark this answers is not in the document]
jaco-brink: Good catch — the test passed only because `isAtBottom` defaults to `true` in component state, so the mocked metrics were never read. Fired `fireEvent.scroll` on the scroll container so `evaluateAtBottom` actually runs against the mocked metrics. Fixed in 8b88d882a2.

jaco-brink:
## Review feedback addressed (8b88d882a2)

**Fixed:**
- `libs/journeys/ui/src/components/Conversation/Conversation.spec.tsx:92` — fired `fireEvent.scroll` on the scroll container in the at-bottom test so the mocked metrics actually drive `evaluateAtBottom`. Without this, the assertion only confirmed the `isAtBottom=true` initial state and would have passed even if the evaluation logic were broken (per @coderabbitai / @edmonday).

edmonday (dismissed):

edmonday (approved):
LGTM

PR #9126, opened 2026-05-01
fix: mobile journey xyflow click

mikeallisonJS:
## Bug Fixes

* Fixed an issue where clicking step menu buttons would unintentionally select the step block, allowing users to interact with step controls without triggering unwanted selection.


mikeallisonJS:
## Review feedback addressed (fe9264a13)

**Fixed:**
- `JourneyFlow.tsx:438` — Replaced shallow `parentNode` / `id` checks in `isMenuElement` with `target.closest('#StepBlockNodeMenuIcon, #edit-step')` so nested SVG children inside the menu FAB are correctly caught. Also corrected the backtick syntax error in CodeRabbit's suggested selector.

PR #9128, opened 2026-05-01
fix: update journeyAiTranslateCreateSubscription to allow anonymous u…

mikeallisonJS:
…sers


csiyang (approved):
LGTM — auth scope change matches established pattern, ACL still gates at resolver level.

PR #8988, opened 2026-04-08
chore: next 16

mikeallisonJS:
- Removed deprecated inputs from nx.json and added @nx/vitest:test configuration
- Updated package dependencies to latest versions, including @emotion/react, @next packages, and @nx packages
- Excluded story files from TypeScript configurations in multiple apps
- Refactored middleware and next.config.js files to streamline configurations and remove unnecessary comments
- Updated next-env.d.ts references for various apps to point to the correct paths
- Removed unused middleware files from journeys and resources apps

## Summary by CodeRabbit

* **Dependencies**
  * Upgraded Next.js, Storybook, Nx tooling and other build/test toolchain packages.

* **Bug Fixes**
  * Video player routes now yield proper “not found” behavior for invalid or missing parameters.

* **Chores**
  * Harmonized Next.js build configs and removed CI lint overrides.
  * Renamed middleware entrypoints to proxy and updated tests accordingly.
  * Expanded TypeScript/ESLint exclusions for story files and Storybook libs; added image asset typings.

PR #9148, opened 2026-05-04
chore: remove unused next-env.d.ts files and update .gitignore to exc…

mikeallisonJS:
…lude next-env.d.ts from version control

## Summary by CodeRabbit

* **Chores**
  * Updated build and TypeScript configuration files across applications for improved consistency and build optimization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,8 @@ terraform.tfstate.backup
 
 # next.js
 .next
+next-env.d.ts
+!libs/**/next-env.d.ts
 
 # playwright
 **/test-results/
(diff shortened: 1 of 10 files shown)


csiyang (approved):
LGTM — clean and correctly scoped gitignore + delete pass. Verified the libs/ negation preserves the one committed next-env.d.ts that has custom content, CI type-check and lint both green, and apps' tsconfig **/*.ts include patterns will pick up regenerated files. Two non-blocking follow-up notes for awareness: (1) the per-app index.d.ts image-module asymmetry from #8988 is unchanged here — every apps/*/index.d.ts still declares only *.svg, so a typecheck before next build could miss png/jpg type resolution; (2) the 9+6+3 redundant 'next-env.d.ts' entries across tsconfig.json / tsconfig.eslint.json files could come out in a follow-up since **/*.ts already matches.

PR #9152, opened 2026-05-05
fix: lint generated next

mikeallisonJS:


csiyang:
Reviewed against the failing autofix.ci run ([25351353715](https://github.com/JesusFilm/core/actions/runs/25351353715)). The TS2307 cascade was caused by losing `/// <reference types="next/image-types/global" />` when #9148 deleted the per-app `next-env.d.ts` files — this fix correctly addresses the root cause by adding the missing ambient `*.png` / `*.jpg` (and for `apps/player`, `*.svg`) declarations to each app's `index.d.ts`, without recreating gitignored generated files. All green on the latest run. ✅

Some asymmetry / consistency nits worth knowing about (none are blockers — happy for these to be a follow-up):

1. **Two coexisting styles for the same declaration in this PR.** Pick one before merging or in a follow-up:
   - Inline (used in `journeys-admin`, `player`, `watch`):
     ```ts
     const content: import('next/image').StaticImageData
     ```
   - Top-level import (used in `resources`, `videos-admin`):
     ```ts
     import { StaticImageData } from 'next/image'
     const content: StaticImageData
     ```

2. **`apps/player/index.d.ts` declares `*.svg` as `StaticImageData`** ([apps/player/index.d.ts:1-4](https://github.com/JesusFilm/core/blob/26-00-MA-fix-lint-generated-next/apps/player/index.d.ts#L1-L4)) while every other Next app declares `*.svg` as `any` with a `ReactComponent` export (the legacy SVGR shape). Functionally fine because player only uses default imports of SVGs (verified — `SharedPlaylistBanner.tsx`, `TopNavBar.tsx`), but worth aligning so SVGR-style usage doesn't surprise someone later.

3. **Unused `*.jpg` declarations added to `journeys-admin` and `watch`.** Neither app currently imports any `.jpg` (`journeys-admin`: 4 png, 0 jpg; `watch`: 1 png, 0 jpg). Harmless ambient noise. Either drop them or accept it as forward-compat.

4. **PR body is empty** and the title (`fix: lint generated next`) doesn't disambiguate that this also bundles the `@nx/enforce-module-boundaries` eslint-disable for the 6 `i18next.config.js` files (added in #9080, never had a disable). That second fix is unrelated to next-env.d.ts. Worth a one-liner in the PR description so the archaeologists in 6 months know what happened.

5. **Re the previously-flagged image-module asymmetry on #8988**: post-fix, only the 5 apps that actually import images get the declarations (`arclight`/`docs`/`short-links`/`watch-modern` are correctly skipped — 0 image imports in each). `libs/journeys/ui/next-env.d.ts` is preserved by the `!libs/**/next-env.d.ts` carve-out in the gitignore, so the ~11 `.jpg` imports in that lib still resolve. So scope-wise, it's correct.

Bigger-picture: a single shared `image-modules.d.ts` referenced from `tsconfig.base.json` would let us delete the duplication across 5 apps and the libs `next-env.d.ts`, but that's a refactor for another PR — not a reason to hold this fix.

Verdict: **approve**, the comments are nits.

csiyang (approved):
LGTM — targeted fix for the #9148 lint fallout. Restores next/image-types/global declarations in each affected app's index.d.ts (5 apps + player) without reintroducing the committed next-env.d.ts pattern. Detailed comment with notes on per-app declaration-style asymmetry and a follow-up opportunity for a shared image-modules.d.ts already posted on the PR.

PR #9164, opened 2026-05-06
feat(journeys): per-card showAssistant + expandChatByDefault [NES-1622]

jaco-brink:
## Summary

Frontend half of the per-card chat-assistant rollout (Linear NES-1622). The backend (NES-1556) is already on `main`; this PR consumes the new `CardBlock.showAssistant` and `CardBlock.expandChatByDefault` fields and uses them to:

1. Decide per-card whether the AI chat affordance shows. The chat is now purely card-level — the deprecated journey-level `Journey.showAssistant` is no longer consulted at runtime (field removal tracked in NES-1624).
2. Auto-open the chat overlay once per `(journeyId, cardId)` per browser tab when a card opts in.
3. Lift overlay open/close state out of `AiChatButton` / `PinnedChatBar` into a single `ChatOverlayProvider` so the affordance, the auto-open trigger, and the pinned chat bar all share one source of truth.

Plan: [`docs/plans/2026-05-04-001-feat-per-card-showassistant-frontend-plan.md`](../tree/jacobusbrink/nes-1622-frontend-per-card-showassistant-expandchatbydefault/docs/plans/2026-05-04-001-feat-per-card-showassistant-frontend-plan.md).

## Files for human review (logical changes)

These contain real behaviour changes — please read carefully:

### Fragments / type plumbing
- `libs/journeys/ui/src/components/Card/cardFields.ts` — adds `showAssistant` + `expandChatByDefault` to the shared `CardFields` fragment.
- `libs/journeys/ui/src/libs/JourneyProvider/journeyFields.tsx` — marks `Journey.showAssistant` as transitional (removal tracked in NES-1624).
- `libs/journeys/ui/src/libs/block/getCardChild.ts` (new) — centralises the `children.find(b => b.__typename === 'CardBlock')` guard used by Conductor, StepFooter, and OverlayContent.
- `libs/journeys/ui/src/libs/block/getCardChild.spec.ts` (new)
- `libs/journeys/ui/src/libs/block/index.ts` — exports the helper.

### Lifted overlay state
- `libs/journeys/ui/src/libs/ChatOverlayProvider/ChatOverlayProvider.tsx` (new) — context provider exposing `{ open, setOpen, shouldAutoOpen, markAutoOpened }`. Auto-open dedup is `sessionStorage`-backed and namespaced by `journeyId`, so it survives StrictMode double-invokes, per-card subtree remounts, and in-tab navigation but resets on tab close / full reload. Falls back to a no-op context when used outside a provider so legacy admin previews keep working without wrapping.
- `libs/journeys/ui/src/libs/ChatOverlayProvider/ChatOverlayProvider.spec.tsx` (new)
- `libs/journeys/ui/src/libs/ChatOverlayProvider/index.ts` (new)
- `apps/journeys/src/components/JourneyPageWrapper/JourneyPageWrapper.tsx` — wraps the player tree with `ChatOverlayProvider` (passes `journeyId` for namespacing).
- `libs/journeys/ui/src/components/AiChatButton/AiChatButton.tsx` — replaces local `useState` with `useChatOverlay()` so the button, the pinned bar, and the auto-open trigger all share state.
- `libs/journeys/ui/src/components/AiChatButton/AiChatButton.spec.tsx` (new)

### Per-card opt-in/out + auto-open
- `libs/journeys/ui/src/components/Card/utils/getFooterElements/getFooterElements.ts` — `hasAiChatButton` / `getFooterMobileSpacing` / `getFooterMobileHeight` now accept an optional `card` and key off `card.showAssistant === true`. The deprecated `Journey.showAssistant` fallback was intentionally dropped in 7248535 after stage testing surfaced legacy journeys forcing the chat on against creator intent. (Closes NES-1603.)
- `libs/journeys/ui/src/components/Card/Card.tsx` — captures the active card props into `cardForFooter` and threads it down.
- `libs/journeys/ui/src/components/Card/{Contained,Expanded,Website}Cover/*.tsx` — accept and forward the `card` prop.
- `libs/journeys/ui/src/components/Card/OverlayContent/OverlayContent.tsx` — uses `card` for footer-spacing and pinned-chat decisions.
- `libs/journeys/ui/src/components/StepFooter/StepFooter.tsx` — derives the active card via `getCardChild` and passes it through.
- `libs/journeys/ui/src/components/StepFooter/StepFooter.apologistChat.spec.tsx` — extended for per-card opt-in / opt-out / fallback.
- `apps/journeys/src/components/Conductor/Conductor.tsx` — card-aware `showPinnedChat` plus the auto-open `useEffect` keyed on `activeCard?.id`. Living at the navigation chokepoint means SwipeNavigation, NavigationButton, HotkeyNavigation, and initial deep-link mounts all fire it, but prefetched neighbours rendered off-screen by `DynamicCardList` do not. The effect intentionally does not depend on the `journey` reference (`hasAiChatButton` is now purely card-level), so Apollo cache writes for unrelated fields no longer cause re-fires.
- `apps/journeys/src/components/Conductor/Conductor.apologistChat.spec.tsx` — extended for per-card flag + auto-open + dedup + StrictMode double-invoke.

### Test-utility tweak
- `libs/journeys/ui/src/components/Card/utils/getFormInitialValues/getFormInitialValues.spec.ts` — adds two optional fields to the local `MockBlock` test type to keep the existing `as unknown as TreeBlock` casts type-safe.

### Out-of-scope (please flag if you'd rather it not ride along)
- `apps/journeys/tsconfig.json`
- `apps/journeys-admin/tsconfig.json`

  Both now exclude `**/*.stories.{ts,tsx,js,jsx}` from `tsc`. Storybook is not exercised in our system anymore (no CI step builds it), but the `type-check` Nx target — invoked by `.github/workflows/autofix.ci.yml` — still ran `tsc -b <project>/tsconfig.json` over story files via the `**/*.tsx` include. The new `CardBlock` fields surfaced ~30 type errors on dead code. Excluding stories from these two projects only keeps `tsc -b` green without modifying any story file. **This is unrelated to NES-1622** and should ideally be ratified separately; happy to split it out if reviewers prefer.

## Files NOT needing review (mechanical)

- All `__generated__/*.ts` — codegen output for the new fragment fields.
- All other modified `*.spec.tsx` / `*.mock.ts` / `data.ts` files in `apps/journeys-admin`, `apps/journeys`, and `libs/journeys/ui` — these were updated by a script that adds `showAssistant: null, expandChatByDefault: null` to every `CardBlock` literal so existing fixtures keep type-checking against the new fragment. No behaviour changes.

## Test plan

- [x] `nx run journeys-ui:test` — 1264 passed
- [x] `nx run journeys:test` — 104 passed
- [x] `nx run journeys-admin:test` — 3151 passed
- [x] `nx run journeys-ui:lint` / `journeys:lint` / `journeys-admin:lint` — clean
- [x] `nx run journeys:type-check` / `nx run journeys-admin:type-check` — clean (post-stories-exclusion)
- [ ] Manual smoke in stage:
    - [ ] Card with `showAssistant=true`, `expandChatByDefault=false` — pinned chat appears, overlay does not auto-open.
    - [ ] Card with `showAssistant=true`, `expandChatByDefault=true` — overlay auto-opens once on first visit, doesn't reopen on revisit within the same tab.
    - [ ] Card with `showAssistant=false`, `expandChatByDefault=true` — silent (no overlay, no pinned bar).
    - [ ] Card with `showAssistant=null` (legacy, no per-card opt-in) — chat does **not** render, even on journeys with `Journey.showAssistant=true`. The journey-level fallback has been dropped.
    - [ ] `variant=admin` and `variant=embed` — neither pinned bar nor auto-open ever fires.
    - [ ] Swipe / arrow-key / button navigation across cards — auto-open fires per (journeyId, cardId) once.
    - [ ] Refresh page — auto-open fires again (sessionStorage tied to tab, full reload resets).
    - [ ] **Mobile asymmetry (known gap):** `expandChatByDefault` only seeds the `PinnedChatBar`'s sheet state on the **first** card a visitor lands on. Navigating from a card with `expandChatByDefault=true` to one with `false` (or vice versa) preserves whatever sheet state the user / previous card produced. Mobile per-card re-seeding is tracked as follow-up and is intentionally out of scope here; verify the desktop overlay still flips per-card.

## Linear

- Closes [NES-1622](https://linear.app/jesus-film-project/issue/NES-1622)
- Closes [NES-1603](https://linear.app/jesus-film-project/issue/NES-1603) (`hasAiChatButton` now respects the active card)
- Follow-up tracked in [NES-1624](https://linear.app/jesus-film-project/issue/NES-1624) for removing the `Journey.showAssistant` fallback once verification SQL confirms zero null `Block.showAssistant` rows.

## Summary by CodeRabbit

* **New Features**
  * Per-card AI chat visibility now drives whether chat UI appears; page-level provider manages shared overlay state and per-card auto-open with session-based deduplication.
  * Pinned chat and AI chat controls support configurable initial expanded/collapsed state.

* **Tests**
  * Expanded test coverage for per-card gating, auto-open rules, session deduplication, provider wiring, and mobile/desktop behaviors.

* **Documentation**
  * Delivery plan updated with rollout status and follow-up items.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9164)


apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -73,12 +79,58 @@ export function Conductor({ blocks }: ConductorProps): ReactElement {
...
+  const initialChatExpanded = activeCard?.expandChatByDefault === true
+
+  const { setOpen, shouldAutoOpen, markAutoOpened } = useChatOverlay()
+
+  // Desktop overlay auto-open on `expandChatByDefault`. Lives at the
+  // navigation chokepoint so prefetched neighbours mounted off-screen by
+  // DynamicCardList do not trigger it. Mobile has no equivalent — the
+  // pinned bar starts in 'idle' (input visible, ready) which already
+  // satisfies the "expanded" semantics; 'active' is reserved for sheets
+  // that contain a real conversation.
```
jaco-brink: **Concern:** This block contradicts the comment six lines above (lines 90–94) and the actual behaviour. After the user-driven mobile fix, the pinned bar starts in `'collapsed'` (drag handle only) when `card.expandChatByDefault` is `false` / `null`, and `'idle'` only when `true`. The mobile equivalent of the desktop auto-open lives in `initialChatExpanded` → `<PinnedChatBar initialExpanded=...>` (declared just above as `initialChatExpanded`), not in the absence of one.

Suggest replacing this paragraph with something like:

> Mobile: handled separately via `initialChatExpanded` → `PinnedChatBar.initialExpanded` (declared above). `setOpen` here only drives the desktop overlay; mobile sheet state is owned by `PinnedChatBar`'s local state machine after mount.

Leaving the stale claim will mislead the next reader.

apps/journeys/src/components/JourneyPageWrapper/JourneyPageWrapper.tsx
```diff
@@ -57,14 +58,16 @@ export function JourneyPageWrapper({
...
       <JourneyProvider value={{ journey, variant: variant ?? 'default' }}>
-        <ThemeProvider
-          {...journeyTheme}
-          rtl={rtl}
-          locale={locale}
-          fontFamilies={fontFamilies}
-        >
-          {children}
-        </ThemeProvider>
+        <ChatOverlayProvider journeyId={journey.id}>
```
jaco-brink: **Nit:** `journeyId` is already extracted on line 28 (and used to build `journeyDomain`). Passing `journey.id` here instead is a minor inconsistency — and `journey.id` (no optional chain) would also throw if `journey` were ever null, whereas the existing `journeyId = journey?.id` is defensive. Suggest `<ChatOverlayProvider journeyId={journeyId}>`.

apps/journeys/src/components/JourneyPageWrapper/JourneyPageWrapper.spec.tsx
```diff
@@ -1,18 +1,32 @@
...
+import userEvent from '@testing-library/user-event'
+
+import { useChatOverlay } from '@core/journeys/ui/ChatOverlayProvider'
 
 import { GetJourney_journey as Journey } from '../../../__generated__/GetJourney'
 import { ThemeMode, ThemeName } from '../../../__generated__/globalTypes'
 
 import { JourneyPageWrapper } from './JourneyPageWrapper'
 
+function ChatOverlayProbe(): JSX.Element {
```
jaco-brink: **Nit:** `apps/journeys/AGENTS.md` prescribes `ReactElement` from `react` as the component return type. Same convention applies to test probes; this returns `JSX.Element`.

jaco-brink (commented):
## Review summary

23 functional files, ~1.5k lines. The gen-vs-functional PR split via #9178 worked beautifully — the diff is clean.

**Concerns:** 3 (one stale comment, one PR-description routing fix, one missing test from the plan).
**Nits:** 2 (variable consistency, `JSX.Element` vs `ReactElement`).
**Critical:** none.

### Out-of-band concerns (no obvious inline anchor)

- **PR description routing.** Description claims `Conductor.apologistChat.spec.tsx` was extended for per-card / auto-open / dedup / StrictMode. The file actually still contains only the four pre-existing flag-gating tests. The new coverage lives in a separate file, **`Conductor.perCardChat.spec.tsx`** (13 cases, including the LD-flag-flip stale-closure regression), which the description doesn't mention. Worth fixing the description so reviewers find the right file.
- **StrictMode double-invoke test missing.** The plan in `docs/plans/2026-05-04-001-feat-per-card-showassistant-frontend-plan.md` explicitly listed `StrictMode-on test: assert setOpen is called exactly once on the double-invoke` as an acceptance criterion. The `sessionStorage` dedup in `ChatOverlayProvider` makes the double-fire safe in theory, but there's no regression-fence for it. Either add the test or call out the deferral in the plan.

### Positives

- `ChatOverlayProvider` design is solid: `'use client'`, no-op `FALLBACK_CONTEXT` for legacy callers (so admin previews keep working without wrapping), `sessionStorage` dedup with in-memory mirror for Safari private mode, `[journeyId]` reset effect.
- `JourneyPageWrapper.spec.tsx` introduces a probe component that flips `data-open` via `setOpen(true)` — directly catches a silent regression to the no-op fallback. This is exactly the kind of fence that would have caught the desktop auto-open bug we hit during iteration.
- Auto-open `useEffect` in `Conductor.tsx` is well-defended: `apologistChatEnabled` in deps explicitly catches the late-LD-flag-load race (and the docstring explains the regression scenario). `shouldAutoOpen` / `markAutoOpened` handle dedup so the extra deps are inert re-fires.
- `getCardChild` centralises a duplicated type guard cleanly.
- Per-card opt-in / fallback / opt-out covered in both `StepFooter.apologistChat.spec.tsx` and `Conductor.perCardChat.spec.tsx`.

jaco-brink:
## Self-review feedback addressed (60351f7fda)

**Fixed:**
- `apps/journeys/src/components/Conductor/Conductor.tsx` — replaced the stale "Mobile has no equivalent" paragraph (lines 99–104) with a pointer to `initialChatExpanded` → `PinnedChatBar.initialExpanded`, so the desktop-overlay block stops contradicting the mobile-bar comment six lines above.
- `apps/journeys/src/components/JourneyPageWrapper/JourneyPageWrapper.tsx:61` — passed the already-extracted `journeyId` to `ChatOverlayProvider` instead of `journey.id` (defensive + consistent with line 28).
- `apps/journeys/src/components/JourneyPageWrapper/JourneyPageWrapper.spec.tsx:11` — switched `ChatOverlayProbe` return type from `JSX.Element` to `ReactElement` per `apps/journeys/AGENTS.md:38`.

Verified locally: `JourneyPageWrapper.spec.tsx` (2 tests) + `Conductor.spec.tsx` (5 tests) all pass.

**Challenged:** none.
**Skipped:** none.

libs/journeys/ui/src/components/Card/utils/getFooterElements/getFooterElements.ts
```diff
@@ -45,11 +53,13 @@ export function hasChatWidget({
...
-  journey,
-  variant = 'default'
+  variant = 'default',
+  card
 }: JourneyInfoProps): boolean {
   if (variant === 'admin' || variant === 'embed') return false
-  return journey?.showAssistant === true
+  // Card-level only. The deprecated `Journey.showAssistant` is intentionally
+  // ignored here — the field itself is removed in NES-1624.
+  return card?.showAssistant === true
```
jianwei1, line 62: **Critical:** The `Journey.showAssistant` runtime fallback has been removed here (and across `getFooterMobileSpacing` / `getFooterMobileHeight`), but the PR description still claims `card.showAssistant ?? journey.showAssistant` and the test plan includes a "Card with no per-card values (legacy) — falls back to `journey.showAssistant`" case that this code cannot satisfy. `journeyFields.tsx:122-125` also flipped from "Transitional … kept as the read-time fallback" to "No longer consulted at runtime."

This folds the NES-1624 follow-up into NES-1622. The PR's own follow-up note says NES-1624 should wait for "verification SQL confirms zero null `Block.showAssistant` rows." If any production `CardBlock` row still has `showAssistant = NULL` on a journey with `Journey.showAssistant = true`, that card will silently lose its chat button, pinned bar, and overlay the moment this lands.

Please either:
1. Restore the read-time fallback until NES-1624 lands cleanly:
   ```ts
   return (card?.showAssistant ?? journey?.showAssistant) === true
   ```
   …and keep this PR scoped to "per-card override of the journey-level value."
2. Or attach the verification SQL output (zero NULLs across stage + prod) to this PR, explicitly fold NES-1624 into the scope, and update the description + test plan so QA isn't testing a behaviour that no longer exists.

---

libs/journeys/ui/src/components/Card/Card.tsx
```diff
@@ -91,8 +91,18 @@ export function Card({
...
 }: CardProps): ReactElement {
+  const cardForFooter = {
+    id,
+    children,
+    backgroundColor,
+    backdropBlur,
+    coverBlockId,
+    fullscreen,
+    ...rest
+  } as TreeBlock<CardFields>
```
jianwei1: **Concern:** `cardForFooter` is rebuilt from a destructured-and-spread shape with `as TreeBlock<CardFields>` masking the cast. Two things make this fragile:

1. A new `CardFields` field added later will be silently captured only because it falls into `...rest` — there's no compile-time signal if `Card` is missed when the fragment grows.
2. The cast hides whether `__typename`, `parentOrder`, and friends are actually present. They are, because `CardProps extends TreeBlock<CardFields>` — but a reader has to verify that to trust the cast.

Cleaner options:
- Change `Card`'s signature to receive the whole card block once and forward it: `function Card({ card, wrappers }: { card: TreeBlock<CardFields>; wrappers?: WrappersProps })` and let the body destructure off `card`. The cover children then take `card={card}` directly.
- Or, if you'd rather keep the current API, keep `...rest` but drop the cast and build the object as `{ ...rest, id, children, backgroundColor, backdropBlur, coverBlockId, fullscreen } satisfies TreeBlock<CardFields>` — `satisfies` gives you the type guarantee the cast is currently silencing.

---

libs/journeys/ui/src/components/Card/OverlayContent/OverlayContent.tsx
```diff
@@ -71,12 +84,12 @@ export function OverlayContent({
...
 
-  const footerMobileSpacing = getFooterMobileSpacing({ journey, variant })
+  const footerMobileSpacing = getFooterMobileSpacing({ journey, variant, card })
 
   const flags = useFlags()
   const pinnedChatActive =
     flags.apologistChat === true &&
-    journey?.showAssistant === true &&
+    hasAiChatButton({ journey, variant, card }) &&
     variant !== 'admin' &&
```
jianwei1: **Concern (nit-pattern):** `hasAiChatButton(...)` already returns `false` for `variant === 'admin'` and `variant === 'embed'` (see `getFooterElements.ts:59`). The trailing `&& variant !== 'admin' && variant !== 'embed'` here is dead code. Same pattern at `apps/journeys/src/components/Conductor/Conductor.tsx:84-88`.

Either drop the redundant guards at both callsites, or drop the internal early-return and centralise the check at the callsites — but pick one. Right now the duplication reads as defensive, which masks the real invariant.

---

apps/journeys/src/components/Conductor/Conductor.tsx
```diff
@@ -73,12 +79,58 @@ export function Conductor({ blocks }: ConductorProps): ReactElement {
...
+    setOpen(true)
+  }, [
+    activeBlock?.id,
+    apologistChatEnabled,
+    journey,
+    variant,
+    setOpen,
+    shouldAutoOpen,
+    markAutoOpened
+  ])
```
jianwei1, line 138: **Concern:** Depending on the whole `journey` reference here re-fires the effect on every Apollo cache write that touches the journey object (analytics mutations, language updates, etc.). The `shouldAutoOpen` / `sessionStorage` dedup makes this safe, but it's noisy and obscures what the effect actually depends on.

The body of the effect reads `journey` only via `hasAiChatButton({ journey, variant, card })` — and since the chat affordance is now card-only, `hasAiChatButton` doesn't actually consult `journey` at all. You can narrow the deps to `[activeBlock?.id, apologistChatEnabled, variant, setOpen, shouldAutoOpen, markAutoOpened]`. If/when the journey-level fallback comes back (see the critical comment on `getFooterElements.ts`), `journey?.showAssistant` is the only field that needs to be in deps — not the whole reference.

Also worth reusing `activeCard` (declared on line 82) instead of recomputing `getCardChild(activeBlock)` inside the effect body — cosmetic, but it keeps the two readings of "the active card" in lockstep.

---

libs/journeys/ui/src/components/PinnedChatBar/PinnedChatBar.tsx
```diff
@@ -150,9 +159,14 @@ const COLLAPSED_HEIGHT_PX = 32
...
-export function PinnedChatBar({ sx }: PinnedChatBarProps): ReactElement | null {
+export function PinnedChatBar({
+  sx,
+  initialExpanded = true
+}: PinnedChatBarProps): ReactElement | null {
   const { variant } = useJourney()
-  const [sheetState, setSheetState] = useState<AiChatSheetState>('idle')
+  const [sheetState, setSheetState] = useState<AiChatSheetState>(
+    initialExpanded ? 'idle' : 'collapsed'
+  )
```
jianwei1, line 169: **Concern:** `initialExpanded` is read only at mount, and `PinnedChatBar` isn't keyed by `activeCard?.id` upstream, so per-card `expandChatByDefault` on mobile only seeds the **first** card a user lands on. Navigating from card A (`expandChatByDefault: true`) to card B (`expandChatByDefault: false`) leaves the sheet in whatever state card A produced.

CodeRabbit already flagged this on the Conductor side as "first-card-only until PR #9122 lands." Two options:

1. Land the mobile parity now: pass `key={activeCard?.id}` to `PinnedChatBar` (or accept a card id and remount internally) so each card re-seeds `sheetState` from its own `expandChatByDefault`. Be aware this also discards any user-initiated drag state mid-journey — which is arguably what creators expect when they set the flag per-card.
2. Land the asymmetry consciously: call this gap out in the PR description's manual smoke list ("Mobile: per-card `expandChatByDefault` applies on first card only; subsequent cards inherit user state"), so QA isn't trying to verify a behaviour that doesn't yet exist on mobile.

As-is, the field name on `CardBlock` implies a per-card behaviour that holds on desktop but breaks silently on mobile after the first card.

---

jianwei1 (changes requested):
## Review summary

Nice shape overall — `ChatOverlayProvider` is well-designed (sessionStorage with in-memory mirror, namespacing by `journeyId`, no-op fallback context, proper StrictMode handling), centralising `getCardChild` is a clean cleanup, and the auto-open's placement at the navigation chokepoint correctly avoids prefetched off-screen card mounts in `DynamicCardList`. Test coverage on auto-open dedup + StrictMode double-invoke + admin/embed exclusion is comprehensive.

Requesting changes on one critical item:

**Critical**
- `getFooterElements.ts` — the `Journey.showAssistant` runtime fallback has been removed but the PR description and test plan still describe it. This silently folds the NES-1624 follow-up into NES-1622 and risks production journeys losing the chat affordance on any `CardBlock` row still carrying `showAssistant: NULL`. Either restore the fallback or attach verification SQL and rescope the PR.

**Concerns** (none blocking on their own, all worth discussing)
- `Card.tsx` — `cardForFooter` reconstruction via `...rest` + cast is fragile; `satisfies` or a single `card` prop is cleaner.
- `OverlayContent.tsx` + `Conductor.tsx` — redundant `variant !== 'admin' && variant !== 'embed'` after `hasAiChatButton` (which already excludes them).
- `Conductor.tsx` auto-open `useEffect` — dep on full `journey` reference re-fires on every cache write; narrow to what's actually read. Also reuse `activeCard` instead of recomputing.
- `PinnedChatBar.tsx` — `initialExpanded` only seeds mount state and the bar isn't keyed by `activeCard?.id`, so per-card `expandChatByDefault` is first-card-only on mobile. CodeRabbit raised the same. Either key on `activeCard?.id` now, or document the asymmetry in the manual smoke list.

**Skipped / acknowledged**
- Pre-existing review threads from @jaco-brink on `Conductor.tsx`, `JourneyPageWrapper.tsx`, and `JourneyPageWrapper.spec.tsx` are already resolved.
- The PR description references files (e.g. `cardFields.ts`, fixture sweeps, `tsconfig.json` exclusions) that aren't actually in this PR's diff — they appear to already be on `main` from NES-1556. Minor staleness, no review impact.

---

jaco-brink:
## Review feedback addressed (7cf2384724)

Thanks for the thorough pass, @jianwei1 and @coderabbitai. All six unresolved threads addressed in [db4a210754](https://github.com/JesusFilm/core/commit/db4a210754) and PR description refreshed.

**Fixed:**

- `apps/journeys/src/components/Conductor/Conductor.tsx` ([#discussion_r3215605163](https://github.com/JesusFilm/core/pull/9164#discussion_r3215605163)) — Narrowed the auto-open `useEffect` deps. `journey` removed (`hasAiChatButton` is card-only now, so cache writes elsewhere no longer re-fire the effect). Reused the outer `activeCard` instead of re-deriving `getCardChild(activeBlock)` inside the body. Comment block updated to explain the `journey` omission.
- `apps/journeys/src/components/Conductor/Conductor.tsx` ([#discussion_r3206225790](https://github.com/JesusFilm/core/pull/9164#discussion_r3206225790)) — Extended the `initialChatExpanded` comment to call out that mobile `expandChatByDefault` only seeds the first card on mount; subsequent navigations stay sticky. (#9122 has landed; per-card re-seeding on mobile is still its own follow-up.)
- `libs/journeys/ui/src/components/Card/Card.tsx` ([#discussion_r3215605061](https://github.com/JesusFilm/core/pull/9164#discussion_r3215605061)) — Replaced the destructure-and-rebuild + `as TreeBlock<CardFields>` cast with `{ wrappers, ...cardForFooter }: CardProps`. `cardForFooter` is now `TreeBlock<CardFields>` directly. No cast, no `...rest` quietly swallowing new fragment fields.
- `libs/journeys/ui/src/components/Card/OverlayContent/OverlayContent.tsx` + `apps/journeys/src/components/Conductor/Conductor.tsx` ([#discussion_r3215605112](https://github.com/JesusFilm/core/pull/9164#discussion_r3215605112)) — Dropped the redundant `&& variant !== 'admin' && variant !== 'embed'` guards from `pinnedChatActive` and `showPinnedChat`. The admin/embed exclusion now lives in exactly one place: inside `hasAiChatButton`.

**Fixed (PR description + test plan, code stays):**

- `libs/journeys/ui/src/components/Card/utils/getFooterElements/getFooterElements.ts` ([#discussion_r3215604908](https://github.com/JesusFilm/core/pull/9164#discussion_r3215604908)) — Critical catch. The code is correct as written — the `Journey.showAssistant` fallback was intentionally dropped in [7248535](https://github.com/JesusFilm/core/commit/7248535) after stage testing surfaced legacy journeys forcing the chat on against creator intent. The PR description and test plan were stale: they still claimed `card.showAssistant ?? journey.showAssistant` and included a "legacy fallback" smoke case that no longer applies. Both updated in this push — the legacy smoke case is inverted (chat does **not** render for `showAssistant=null`), and the description now reflects the card-only behaviour with a reference to the drop commit. The NES-1624 follow-up still tracks the schema-level field removal, but at the runtime level we are already card-only.

**Fixed (adjusted — documented instead of remounting):**

- `libs/journeys/ui/src/components/PinnedChatBar/PinnedChatBar.tsx` ([#discussion_r3215605269](https://github.com/JesusFilm/core/pull/9164#discussion_r3215605269)) — Took option 2 (call out the gap). Adding `key={activeCard?.id}` to `PinnedChatBar` would discard the user's manually-set sheet state on every card navigation, which is a UX regression for visitors who collapse the bar mid-journey. That's a deliberate trade-off worth its own ticket, not a silent fold-in. The inline comment in `Conductor.tsx` now states the asymmetry explicitly, and the PR description's manual smoke list has a new item so QA isn't testing a behaviour that doesn't yet exist on mobile.

**Skipped:** none.

CI was green on the previous head; pushed on top of a fresh main merge. Re-requesting review at your convenience.

jianwei1 (approved):
## Review summary

All five items from the previous review (1 Critical + 4 Concerns) addressed:

- **Critical — `Journey.showAssistant` fallback** → resolved via PR-description rewrite. NES-1624 consciously folded in; description, code comments, and test plan now consistently say `showAssistant=null` cards do not render chat. Rationale (stage testing found legacy journeys forcing chat on against creator intent) captured in the body.
- **`cardForFooter` reconstruction** → `Card.tsx` now `function Card({ wrappers, ...cardForFooter }: CardProps)` with local fields destructured off `cardForFooter`. No cast; type flows from `CardProps` naturally.
- **Redundant `variant !== 'admin' && variant !== 'embed'` guards** → removed at both callsites (`Conductor.tsx`, `OverlayContent.tsx`); `hasAiChatButton` is the single source of truth.
- **`useEffect` deps over-broad** → `journey` dropped, `activeCard` reused inside the body, scoped `eslint-disable-next-line` with a clear inline rationale.
- **Mobile asymmetry (CodeRabbit)** → documented in code comment + dedicated "Mobile asymmetry (known gap)" item in the manual smoke list.

CI green across CLA / CodeRabbit / Danger. No new concerns introduced by the revision. Nice clean turnaround — `Card.tsx`'s rest-then-pick pattern is genuinely simpler than what was there before.

LGTM.

---

PR #9167, opened 2026-05-06
fix: ensure StepBlock nodes are correctly added to th…

mikeallisonJS:
…e nodes array and update internal node state on step change

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected element ordering in the journey flow editor to ensure step blocks display correctly before associated actions.
  * Improved editor responsiveness when actions are added or modified within a step.

* **Tests**
  * Updated test suite to validate node ordering and positioning across various flow editor scenarios.


csiyang (approved):
LGTM — both fixes target real causes: parent-before-child ordering matches xyflow's documented contract, and the useEffect for updateNodeInternals removes the render-phase side effect anti-pattern. Together they explain the click-makes-it-work symptom on QA-458.

PR #9172, opened 2026-05-06
feat: add teamPlausibleStatsBreakdown query for aggregated journey stats

mikeallisonJS:
- Introduced a new GraphQL query `teamPlausibleStatsBreakdown` to fetch aggregated Plausible breakdown stats for journeys within a team.
- Updated schema definitions and generated types accordingly.
- Deprecated `showAssistant` field in the Journey type, directing users to use `CardBlock.showAssistant` instead.


mikeallisonJS:
not needed, going straight to plausible

PR #9173, opened 2026-05-07
feat(journeys-admin): nes-1608 use template deep link flow on admin root

csiyang:
## Summary

Admin-side receiver for the public template gallery's "Use" deep link (`https://admin.nextstep.is/?useTemplate=<journeyId>` — emitted by [#9144](https://github.com/JesusFilm/core/pull/9144)). When the param is present on the dashboard, the app fetches the public template, opens a `CopyToTeamDialog` defaulted to the active team, calls the existing `journeyDuplicate` mutation on submit, navigates to `/?type=journeys&refresh=true`, and strips the param via shallow `router.replace` on close. The `?onboarding=true` popover is suppressed while `useTemplate` is present so the two popovers don't compete for attention. Auth-redirect already preserves the param via `redirectToLogin(ctx)` → `ctx.resolvedUrl` — no change needed.

Pure frontend; reuses existing `useJourneyDuplicateMutation`, public `useJourneyQuery({ idType: databaseId, options: { skipRoutingFilter: true } })`, and `TeamProvider`. No new GraphQL operations, no schema changes.

> **Base branch:** stacked on [`siyangcao/nes-1552-public-gallery-page-renderer-template-galleryslug`](https://github.com/JesusFilm/core/pull/9144). Merge that first, then rebase onto `main`.

Plan: [docs/plans/2026-05-07-001-feat-nes-1608-use-template-deep-link-flow-on-admin-root-route-plan.md](https://github.com/JesusFilm/core/blob/siyangcao/nes-1608-use-template-deep-link-flow-on-admin-root-route/docs/plans/2026-05-07-001-feat-nes-1608-use-template-deep-link-flow-on-admin-root-route-plan.md).

## What's in the change

| Area | Path |
| --- | --- |
| Page wiring | `apps/journeys-admin/pages/index.tsx` (mount the receiver, suppress onboarding popover when `useTemplate` is present) |
| Receiver | `apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx` |
| Tests | `apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.spec.tsx` (7/7 passing) |
| Plan | `docs/plans/2026-05-07-001-feat-nes-1608-use-template-deep-link-flow-on-admin-root-route-plan.md` |

## Behaviour

- Param coercion: `string` → use; `string[]` → first element wins; `''`/missing → render nothing.
- Dialog opens immediately on detection; submit blocks until the team list resolves (existing dialog validation).
- Submit calls `journeyDuplicate({ id, teamId, forceNonTemplate: true })` — same shape as `CreateJourneyButton`.
- Success → `enqueueSnackbar('Journey Copied')` → `router.push('/?type=journeys&refresh=true')` → dialog auto-closes → param stripped via shallow `router.replace`.
- Failure → `enqueueSnackbar('Journey duplication failed')` → submit re-throws so the dialog stays open for retry on a different team.
- Translation toggle is supported (mirrors `CreateJourneyButton`'s flow): kicks off `useJourneyAiTranslateSubscription`, holds the dialog open with progress, navigates on completion.

## Testing

- `npx jest --config apps/journeys-admin/jest.config.ts --no-coverage 'apps/journeys-admin/src/components/UseTemplateDeepLink'` — **7/7 passing** (param absent / empty / array / present, submit happy path, submit failure, close-without-submit).
- `npx tsc --noEmit -p apps/journeys-admin/tsconfig.json` — clean.
- `npx nx lint journeys-admin` — clean (0 new errors; pre-existing warnings only).

### Manual smoke (TODO before un-drafting)

- [ ] Logged-in visitor lands on `/?useTemplate=<known-published-template-id>` → modal opens with active team preselected → submit copies journey → lands on `/?type=journeys&refresh=true` with new copy in list.
- [ ] Logged-out visitor → redirected to sign-in, returns to `/?useTemplate=<id>`, modal opens.
- [ ] `/?useTemplate=<id>&onboarding=true` → modal opens, onboarding popover does not.
- [ ] Close modal without submitting → URL becomes `/`, no journey created.
- [ ] Translation toggle path → progress UI shows, navigates on completion.

## Out of scope

- "Use all templates from a gallery" multi-template flow (future).
- `/templates/<id>` route changes (existing in-app entry point untouched).
- Custom-domain (`[hostname]/?useTemplate=...`) handling — public galleries are intentionally only served from the root domain.

## Linear

[NES-1608 — Use Template deep link flow on admin root route](https://linear.app/jesus-film-project/issue/NES-1608)

Roadmap: [NES-1607 — Template Gallery Pages](https://linear.app/jesus-film-project/issue/NES-1607).

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.7, 1M context, extended thinking).

## Summary by CodeRabbit

* **New Features**
  * Added deep link support enabling direct template duplication from the public template gallery
  * Template gallery cards now display "Use" and "Preview" action buttons for improved discoverability
  * Template duplication flow integrates team selection and optional translation capabilities

* **Documentation**
  * Added implementation plan documenting template deep link feature specifications and behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9173?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


edmonday (commented):
## Review summary

**Critical:** 1 — PR title fails commitlint (Danger is red on this).
**Concerns:** 3 inline — missing i18n extraction, `pages/index.tsx` reaching into the feature's activation rule, and a request to file a follow-up ticket for the deferred hook extraction.
**Positive:** outer/inner mount-gating split is the right shape; test coverage on the subtle translation-while-loading + navigatedAwayRef paths is solid; `target="_blank" rel="noopener noreferrer"` + `encodeURIComponent` on the deep link are correct.

Local review only — no blocking ask beyond fixing the title and extracting translations before merge.

apps/journeys-admin/pages/index.tsx
```diff
@@ -13,6 +13,10 @@ import { PageWrapper } from '../src/components/PageWrapper'
 import { SidePanelTitle } from '../src/components/SidePanelTitle/SidePanelTitle'
 import { TeamMenu } from '../src/components/Team/TeamMenu'
 import { TeamSelect } from '../src/components/Team/TeamSelect'
+import {
+  UseTemplateDeepLink,
+  getJourneyIdParam
+} from '../src/components/UseTemplateDeepLink'
```
edmonday, line 19: **Critical:** PR title fails commitlint (Danger check is failing on this) — `feat(journeys-admin): NES-1608 Use Template deep link flow on admin root` has a sentence-case subject. Lowercase it (e.g. `feat(journeys-admin): nes-1608 use template deep link flow on admin root`) so Danger goes green. This is the only reason CI is red.

*(Anchored here because GitHub requires inline comments on a diff line — title-level concerns don't have a natural anchor.)*

apps/journeys-admin/pages/index.tsx
```diff
@@ -37,6 +41,10 @@ export default function IndexPage(): ReactElement {
     (router?.query?.type as 'journeys' | 'templates' | 'collections') ??
     'journeys'
   const showSidePanel = currentContentType === 'journeys'
+  const useTemplateDeepLinkActive =
+    getJourneyIdParam(router.query.useTemplate) != null
+  const showOnboardingPopover =
+    router.query.onboarding === 'true' && !useTemplateDeepLinkActive
```
edmonday: **Concern:** `pages/index.tsx` now has to know two things about the deep link — where to mount the receiver and how to detect when it's active (the `getJourneyIdParam` check duplicated here for onboarding suppression). The page should only need to mount it. Consider exposing the active-state through a hook (`useUseTemplateDeepLinkActive()`) or React context so the activation rule lives in one place. If the rule changes (e.g. also gated on auth state), only the feature file should need to change.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
```diff
@@ -0,0 +1,262 @@
...
+      const wantsTranslation =
+        showTranslation === true && selectedLanguage != null
+
+      // Translation needs the source journey's language metadata. Throwing
+      // (instead of a plain return) halts CopyToTeamDialog's submit pipeline
+      // BEFORE updateTeamState + resetForm run, so the user's selections
+      // survive the retry. The throw fires before `setLoading(true)` and
+      // outside the try block below — no "duplication failed" snackbar leaks.
+      if (wantsTranslation && journey == null) {
+        enqueueSnackbar(t('Loading template — please retry'), {
```
edmonday, line 143: **Concern:** i18n extraction not run for the new keys. This file introduces `Loading template — please retry`, `Journey duplication failed`, `Journey Copied`, `Journey Translated`, `Use This Template`, and `Add`; `GalleryTemplateCard.tsx` adds `Use` and `Preview`. Neither `libs/locales/en/apps-journeys-admin.json` nor `apps-journeys.json` is touched in this diff. Please run `pnpm dlx nx run journeys-admin:extract-translations` and `pnpm dlx nx run journeys:extract-translations` before merge.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * NOTE: shares duplication-and-translate orchestration semantics with
+ * libs/journeys/ui/src/components/TemplateView/CreateJourneyButton/CreateJourneyButton.tsx.
+ * If you change the mutation, subscription, error handling, or wait-for-load
+ * behaviour here, consider whether the same change applies there. Hook
+ * extraction was considered (NES-1608 review 2026-05-13) and deferred to
+ * reduce shared-lib blast radius.
+ */
+export function UseTemplateDeepLink(): ReactElement | null {
+  const router = useRouter()
```
edmonday: **Concern:** This NOTE and the mirror in `libs/journeys/ui/.../CreateJourneyButton.tsx` ask future editors to keep the duplication-and-translate orchestration in sync by hand. The deferral itself is fine — limits shared-lib blast radius from this PR — but two-way "remember to also edit X" comments are the weakest form of coupling and will rot. Could you file a follow-up NES ticket for the hook extraction and link it from both NOTE comments so there's a tracked home for the work? No follow-up ticket exists under NES-1608 today (checked Linear).

apps/journeys-admin/pages/index.tsx
csiyang: Title fixed via `gh pr edit` — lowercased to `feat(journeys-admin): NES-1608 use template deep link flow on admin root` so commitlint's subject-case rule passes. Danger should go green on the next run.

apps/journeys-admin/pages/index.tsx
csiyang: Done — extracted `useTemplateDeepLinkActive` (boolean) and `useTemplateDeepLinkJourneyId` (`string | null`) hooks into a new `useTemplateDeepLinkActive.ts` module co-located with the component. Both call `useRouter` so they're real hooks, not pure functions in disguise.

`pages/index.tsx` now uses the boolean variant for the onboarding-suppression check; `UseTemplateDeepLink` uses the journeyId variant for its mount-gate. The activation rule (array "first wins", empty string is absent) lives in one place — the page no longer has to know two things about the deep link. `getJourneyIdParam` is now file-private to the hook module.

Spec covers both hooks (9 cases via `renderHook` — absent / empty string / non-empty / array-with-value / empty-array). Two-hook layering rather than one + caller-coerces because Siyang wanted the hook "to actually just be the hook" with hook-style naming at each call site.

Commit: 494c41028.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
csiyang: Added the 4 actually-missing keys: `Loading template — please retry` and `Use This Template` to `apps-journeys-admin.json`, and `Use` + `Preview` to `apps-journeys.json`.

Worth flagging: `nx run journeys-admin:extract-translations` and `nx run journeys:extract-translations` from the workspace root both report "No files were updated" — i18next-cli's input path resolution doesn't pick up the source files from that CWD (running from the app dir picks them up but writes to the wrong output location). I checked the other keys you listed (`Journey Copied`, `Journey Translated`, `Journey duplication failed`, `Add`) — those were already in the file from earlier extractions, so this PR's actual new-key delta is just the 4 above, added manually with the matching JSON shape. Happy to file a separate ticket for the extraction-from-workspace-root regression if you'd like.

Commit: 494c41028.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
csiyang: Filed: **NES-1680** — *Extract shared duplicate-and-translate orchestration into a hook (UseTemplateDeepLink + CreateJourneyButton)*.
Linear: https://linear.app/jesus-film-project/issue/NES-1680

Both NOTE comments are now `TODO(NES-1680): …` with the Linear URL and a one-line "if editing this orchestration, also edit <other file>" pointer. Same fix mirrored in the plan doc. The cross-reference is now time-limited rather than a permanent ask of future editors.

Commit: 494c41028.

csiyang: Update: cancelled NES-1680 and dropped the cross-reference NOTE comments from both files (commit 02ca8c1ba).

On re-examination most of the "duplication" between `UseTemplateDeepLink` and `CreateJourneyButton` is **GraphQL-contract-mandated boilerplate** rather than true logic duplication:

- The `journeyDuplicate` mutation variables (`{ id, teamId, forceNonTemplate: true }`) match because that's what the mutation requires — you couldn't write them differently.
- The `useJourneyAiTranslateSubscription` `variables/skip/onComplete/onError` structure is what the subscription hook expects — also not arbitrary.
- The success/error snackbar copy happens to match, but that's a coincidence of UX choice, not a shared contract. Either component could change its copy independently without breaking the other.

The only genuinely non-obvious shared bit is the translation-variables construction (~10 lines, building the 7-field request from the source journey's language metadata). That doesn't justify a shared-lib hook with cross-call-site callback contracts today. If a third "use template" entry point ever appears, we'll revisit then — but at that point a pure helper function for the variable construction would likely be the right shape, not a full orchestration hook.

You were right that bare "remember to also edit X" comments rot. Removing them entirely rather than tracking the deferral is the cleaner accept-the-duplication choice.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
```diff
@@ -0,0 +1,241 @@
...
+        const { data } = await journeyDuplicate({
+          variables: { id: journeyId, teamId, forceNonTemplate: true }
+        })
+
+        if (data?.journeyDuplicate?.id == null) {
+          throw new Error('Journey duplication failed')
+        }
+
+        if (!wantsTranslation || journey == null) {
+          setLoading(false)
```
edmonday, line 161: **Concern:** Dead disjunct — the `|| journey == null` branch is unreachable.

The earlier guard (`if (wantsTranslation && journey == null) throw`) establishes the invariant `wantsTranslation === true ⇒ journey != null` for the rest of the function. So here:

- `wantsTranslation === false` → first disjunct true → enter copy-only branch ✓
- `wantsTranslation === true` → `journey != null` (from guard) → second disjunct false → fall through to translation path ✓
- `wantsTranslation === true && journey == null` → impossible, guard already threw

The extra clause reads as belt-and-braces defense but actually obscures the invariant — a reader sees it and reasonably asks whether `journey` can be null on the translation path below (it can't). Suggest:

```ts
if (!wantsTranslation) {
  setLoading(false)
  enqueueSnackbar(t('Journey Copied'), { variant: 'success', preventDuplicate: true })
  navigateToJourneyList()
  return
}
```

Then the translation path below reads cleanly: `wantsTranslation` is true ⇒ `journey` is non-null, and the `journey.title` / `journey.language` accesses don't need a reader to reconstruct why they're safe.

edmonday (commented):
Single inline concern on `handleSubmit` control flow — see comment. No criticals; CI not yet finished.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
csiyang: Good catch — applied in 2e8898997.

You're right that the disjunct was load-bearing in an earlier revision (when the guard above `return`-ed instead of throwing, it routed translation-mode-without-journey to the copy-only branch as a fallback). After we changed the early return to a throw for the form-preservation fix, the clause lost its job but stayed — exactly the obscured-invariant smell you described.

One wrinkle: TypeScript's control-flow narrower can't carry the transitive invariant (`wantsTranslation === true ⇒ journey != null` across the `!wantsTranslation` early-return) across branches, so dropping the disjunct outright surfaced three `journey is possibly undefined` errors on the translation path below. Added a local `if (journey == null) throw new Error('unreachable: ...')` re-assertion right before the translation path, with a comment explaining why it exists. The throw is unreachable at runtime; the assertion is purely there to inform the type narrower (and the next reader).

edmonday (commented):
## Review summary

Two concerns flagged inline — both minor consistency/comment fixes. Otherwise the implementation is in good shape.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const journey = journeyData?.journey
+
+  const [journeyDuplicate] = useJourneyDuplicateMutation()
+
+  const stripParamFromUrl = useCallback((): void => {
+    const { useTemplate, ...rest } = router.query
+    void router.replace({ pathname: router.pathname, query: rest }, undefined, {
+      shallow: true
+    })
+  }, [router])
```
edmonday, line 89: **Concern:** `stripParamFromUrl` uses `router.pathname` while `navigateToJourneyList` a few lines down hardcodes `pathname: '/'`. Both end up at `/` because this receiver is only mounted on the index page, but using `router.pathname` in one and `'/'` in the other makes the code *look* portable to other routes when the rest of the component (translation guard, `defaultToActiveTeam`, `/?type=journeys&refresh=true` nav) all assume index. Pick one — preferably hardcode `'/'` in both — so the next reader doesn't waste time wondering whether this works elsewhere.

apps/journeys/src/components/TemplateGalleryView/TemplateGalleryGrid/GalleryTemplateCard.tsx
```diff
@@ -19,6 +32,14 @@ export function GalleryTemplateCard({
   template,
   priority = false
 }: GalleryTemplateCardProps): ReactElement {
+  const { t } = useTranslation('apps-journeys')
+
+  // Read inside the component so test specs can override
+  // NEXT_PUBLIC_JOURNEYS_ADMIN_URL via vi.stubEnv / process.env without
+  // having to reset module state. Fallback matches the prod admin host.
```
edmonday: **Concern:** This comment references `vi.stubEnv` (a Vitest API), but the spec for this file uses Jest — `process.env = { ...originalEnv }` in `beforeEach`. The comment will mislead a future reader (or a Vitest migration) about how this is actually tested. Drop the `vi.stubEnv` mention and just say something like *"Read inside the component so specs can override `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` via `process.env` without module-state resets."*

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
csiyang: Done in f956adb68 — `router.pathname` → `'/'` in `stripParamFromUrl`.

apps/journeys/src/components/TemplateGalleryView/TemplateGalleryGrid/GalleryTemplateCard.tsx
csiyang: Done in f956adb68 — comment removed entirely.

edmonday (commented):
## Review summary

Second-pass review (dry-run).

**Critical:** none
**Concerns:** none new — the two unresolved threads from the prior pass are addressed in head (f956adb): `stripParamFromUrl` now hardcodes `pathname: '/'` to match `navigateToJourneyList`, and the stale `vi.stubEnv` comment is gone.
**Nits:** none worth flagging.

**Positive notes**
- Outer/inner split keeps `GET_JOURNEY` from firing on ordinary admin loads and gives each deep-link session a fresh state slate (no `navigatedAwayRef` leakage).
- `navigateToJourneyList` + `navigatedAwayRef` correctly serializes the success-path nav against `CopyToTeamDialog`'s post-submit `onClose`.
- Translation-before-journey-resolved throws (not returns) so Formik's `resetForm` doesn't wipe the user's selections — nice regression test for that.
- Shared `useTemplateDeepLinkJourneyId` / `useTemplateDeepLinkActive` keeps `pages/index.tsx` and the receiver in lockstep on what "active" means.
- `forceNonTemplate: true` + `IdType.databaseId` + `skipRoutingFilter: true` match existing precedents — no new attack surface.
- 20-case spec coverage across the component and the hook; covers param shapes, success/error, translation success/error, the loading-on-translate regression, strip-vs-nav, env override, and URL encoding.

LGTM.

edmonday (approved):
LGTM — prior review threads addressed, CI green.

PR #9181, opened 2026-05-08
chore(api-users): migrate test suite from Jest to Vitest

mikeallisonJS:
## Summary

Mirrors the api-analytics (`bc0da938d`) and api-media (#9179) migrations: swap Jest for Vitest on `apis/api-users`. Replaces `jest.config.ts` with `vitest.config.mts`, switches the Nx executor from `@nx/jest:jest` to `@nx/vitest:test`, ports `test/prismaMock.ts` to `vitest-mock-extended` with the async `vi.importActual` factory, and rewrites all 9 spec files from `jest.*` to `vi.*`.

Adds one matching runtime line to `apis/api-users/src/yoga.ts`:

```ts
maskedErrors: process.env.NODE_ENV === 'test' ? false : undefined
```

This is the same line shipped in api-media — required so GraphQL Yoga surfaces `GraphQLError` extensions (e.g., `Email already in use`) in test responses instead of masking them as "Unexpected error". Production behavior is unchanged (the gate is `NODE_ENV === 'test'`).

## Plan

`docs/plans/2026-05-08-001-chore-api-users-vitest-migration-plan.md` (all 7 R-IDs satisfied).

## Test plan

- [x] `nx test api-users --skip-nx-cache` — 74/74 passing
- [x] `nx lint api-users` — 0 errors, 2 pre-existing warnings (no regression)
- [x] `grep -rE "\bjest\b|jest-mock-extended|@jest/globals|ts-jest|jest.preset" apis/api-users/` — zero hits
- [x] Code review: 7 reviewers dispatched, all clean (1 false-positive dropped by validator, 1 P3 advisory about `resolve.dedupe: ['graphql']` worth investigating later)
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Migrated the api-users test suite and tooling from Jest to Vitest, updating configs, mocks, and spec files.
  * Retained existing test coverage and assertions; no application behavior changes.
* **Documentation**
  * Added a migration plan describing the Jest→Vitest transition and verification steps.
* **Minor**
  * Adjusted GraphQL server error masking behavior for test environments only.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9181)

PR #9179, opened 2026-05-07
chore(api-media): migrate test suite from Jest to Vitest

mikeallisonJS:
## Summary

Migrates the `apis/api-media` test suite from Jest to Vitest, mirroring the recipe established by [PR #9121 / commit `bc0da938d`](https://github.com/JesusFilm/core/commit/bc0da938d) for `api-analytics`.

- Replaces `jest.config.ts` with `vitest.config.mts`; switches the Nx `test` target executor from `@nx/jest:jest` to `@nx/vitest:test`
- Rewrites all 63 spec files mechanically (`jest.fn` → `vi.fn`, `jest.mock` → `vi.mock`, etc.) plus three test setup files (`prismaMock.ts`, `crowdinMock.ts`, `bullmqMock.ts`)
- Adapts `tsconfig.spec.json` to `vitest/globals` and drops the `jest` types from `tsconfig.app.json`

A few non-mechanical adaptations were needed:

- **`src/yoga.ts`** — disables Yoga's error masking when `NODE_ENV=test` so resolver-thrown `GraphQLError` messages reach test clients. Vitest's module-realm isolation defeats Yoga's `instanceof GraphQLError` check, masking otherwise-passable errors as "Unexpected error.".
- **`src/schema/mux/video/video.spec.ts`** — `jest.requireMock` calls become `await vi.importMock<any>(...)` and the surrounding callbacks are made async.
- **`src/schema/videoVariant/videoVariant.spec.ts`** — top-level `jest.requireMock` destructuring replaced with namespace imports + `vi.mocked()`; the `jest.requireActual` factory rewritten as `vi.mock(path, async () => ({ ...(await vi.importActual(path)) }))`.
- **`src/schema/shortLink/shortLinkDomain/shortLinkDomain.service.spec.ts`** — `Response` now imported from `node-fetch`; mock factory spreads `vi.importActual` to preserve the real Response class while replacing the default export.
- **`src/workers/blocklist/service/service.spec.ts`** — rewritten to drop `moxios` (which OOMed under Vitest) and mock `axios` directly with `vi.mock` + `vi.mocked(axios.get).mockResolvedValueOnce(...)`.
- **`src/workers/dataExport/service/service.spec.ts`** — per-test timeout third-arg `it(name, fn, 10000)` replaces the in-test `vi.setConfig`; `vi.mocked(S3Client)` with a shape cast replaces the prior `jest.requireMock`.

## Verification

`nx test api-media --skip-nx-cache` passes:

```
[...]

PR #9121, opened 2026-05-01
chore: vitest analytics

mikeallisonJS:
## Summary by CodeRabbit

* **Chores**
  * Upgraded internal testing infrastructure to enhance system performance and maintainability standards.

* **Tests**
  * Refactored test configurations and utilities to provide more consistent and reliable test execution across the project.

PR #9182, opened 2026-05-08
chore: languages vitest

mikeallisonJS:


csiyang (approved):
Clean mechanical Jest→Vitest migration for api-languages, consistent with the merged sibling recipe (#9121, #9132, #9179, #9181).

Verified locally: ran the suite via vitest, 9 files / 35 tests pass. No source changes outside specs — appropriately omits the `yoga.ts` `maskedErrors` tweak (no test in this app asserts on GraphQLError extensions) and the `dedupe: ['graphql']` / `NODE_ENV: 'test'` config bits (not needed here, matches api-analytics' minimal config). Coverage config preserved.

The few non-mechanical touches (chalk/fs mocks wrapped for ESM default-import resolution, defensive `closeHandler?.[1]` null-guards in two specs) are correct and benign.

PR #9193, opened 2026-05-09
feat: scribe agent

mikeallisonJS:
Proof of concept for journeys agent. Not for production use


csiyang:
### Problem
Build an interactive Claude-driven CLI ("scribe") that signs into dev/stage/prod via a browser flow, caches a Firebase ID token, and exposes journey CRUD as MCP tools to a Claude agent — including a superadmin `/impersonate` mode for acting as another user.

### Solution
New `apps/scribe` (Ink REPL + Claude Agent SDK + `tool()`-backed GraphQL ops for journey create/fetch/diff/validate/update). New `apps/journeys-admin/pages/users/cli-auth.tsx` brokers the browser → CLI handoff by POSTing the user's ID token to a one-shot loopback server. Impersonation goes via existing `userImpersonate` GraphQL mutation, exchanged for an ID token through Firebase REST using a Doppler-fetched API key.

### Findings

**P1 — Token harvesting via cli-auth page** (`apps/journeys-admin/pages/users/cli-auth.tsx:55-94`)
The page auto-POSTs the signed-in user's Firebase ID token to any `http://(localhost|127.0.0.1):*/callback` URL passed via `?callback=`, with no user consent click, no PKCE-style proof of possession, and an attacker-chosen `state` (any 16–64 hex chars passes the regex). Threat: attacker runs a process on victim's loopback (browser ext, malicious dev tool, dropped binary) and social-engineers victim into clicking `https://admin.nextstep.is/users/cli-auth?callback=http://127.0.0.1:NNNN/callback&state=...`. Victim is signed in → page auto-fires → attacker gets a ~1-hr ID token tied to the victim's full account. The README at line 180 sells "validates the callback origin" as the protection — that's necessary but not sufficient. Fix: render an explicit "Authorize scribe to sign in" button showing the callback host:port + environment, and require a click before the POST. Optionally have scribe send a short verification code that the user must paste/confirm in-browser (PKCE-style).

**P1 — Agent can write to prod without per-call user confirmation** (`apps/scribe/src/repl/App.tsx:664`, `apps/scribe/src/tools/journey/index.ts:161-182`)
`permissionMode: 'default'` + SDK MCP tools = `create_journey` and `update_journey` execute as soon as the LLM emits the tool_use block. The only guardrail is text in `systemPrompt.ts:74` ("Never call update_journey without explicit user approval after a diff"). A prompt-injected, confused, or jailbroken agent (or one fed a malicious journey title or description it then echoes back) will happily mutate production. Given the README states staging/prod are real, this is the wrong default. Fix: pass a `canUseTool` callback (or equivalent in the agent SDK version pinned) that intercepts `mcp__scribe__update_journey` and `mcp__scribe__create_journey`, surfaces a yes/no confirmation in the Ink UI showing the diff/payload, and only proceeds on Enter.

**P2 — Stale closure in `setActiveTeam` persists team to wrong profile** (`apps/scribe/src/repl/App.tsx:235-269`)
`useCallback` deps are `[appendSystemMessage, state.session]` but the body reads `getEffectiveSession(state)`, which depends on `state.impersonating`. After `/impersonate`, `state.session` is unchanged so the memoized callback retains a closure where `impersonating: null`, then writes the new team to the operator's `JourneyProfile.lastActiveTeamId` instead of the impersonated user's. Fix: add `state.impersonating` (or the whole `state`) to deps, or compute `effective` from the latest state via a ref.

**P2 — `replaceSession` doesn't clear impersonation on `/env` switch** (`apps/scribe/src/repl/App.tsx:154-182`)
`/env stage → /impersonate x → /env prod` leaves `impersonating` populated with a stage-Firebase-issued ID token. Prod gateway will UNAUTHENTICATED-reject every call, but the system prompt at `systemPrompt.ts:107` still says "IMPERSONATION ACTIVE — operating as x". Confusing failure mode and a real risk of operators believing they're impersonating in prod when they aren't. Fix: reset `impersonating: null` in `replaceSession`.

**P2 — Card reorder produces empty diff while changing entry card** (`apps/scribe/src/tools/journey/diffJourney.ts:83-121`, `validateJourney.ts:83`)
`diffJourney` indexes by `id`, so swapping `cards[0]` and `cards[1]` shows zero diff. But `validateJourney` treats `cards[0]?.id` as the entry card (and presumably the server does too). The system-prompted workflow has the user approve the diff before update_journey — they can approve "no changes" and have the entry card silently swap on apply. Fix: include an ordered ID list comparison in `JourneyDiff` (e.g., `entryCardChanged: {before, after}` and `reordered: string[]`).

**P2 — Zero test coverage** (entire `apps/scribe/`, plus `apps/journeys-admin/pages/users/cli-auth.tsx`)
5,243 LOC, 0 spec files. At minimum before un-drafting: `validateJourney` (error/warning code matrix), `diffJourney` (add/remove/change/reorder), `credentials.{save,load,clear}` (perms + corruption recovery), `isAllowedCallback` (URL parsing edge cases), and a render test on `CliAuthPage` asserting `notFound` on bad state/callback.

**Medium — CORS wildcard on loopback callback** (`apps/scribe/src/auth/browserLogin.ts:152-158`)
`Access-Control-Allow-Origin: '*'` is broader than needed; only the journeys-admin origin should be permitted to POST. Defense in depth.

**Medium — README out of sync with code** (`apps/scribe/README.md`)
References `src/repl/runRepl.ts` at lines 161, 189 (doesn't exist — entry is `App.tsx`/`Root.tsx`). Slash-command table at lines 82-89 omits `/team`, `/journey`, `/model`, `/impersonate`, `/stop-impersonate`, all of which are implemented. Layout tree is wrong.

**Medium — `environment` query param unvalidated against host** (`apps/journeys-admin/pages/users/cli-auth.tsx:165-166`)
Page accepts any `?environment=` string and displays it to the user. Cross-env confusion: `admin-stage.nextstep.is/users/cli-auth?environment=prod&...` will render "Environment: prod" while signing into stage. Fix: derive `environment` from the host server-side, or assert it matches.

**Low — Windows `openBrowser` uses `shell: true`** (`apps/scribe/src/auth/openBrowser.ts:14-19`)
URL is app-built so currently safe, but `shell: true` is fragile. Prefer `spawn('cmd', ['/c', 'start', '', url], { shell: false })`.

**Low — Doppler shell-out for `NEXT_PUBLIC_FIREBASE_API_KEY`** (`apps/scribe/src/config/environments.ts:65-97`)
The `NEXT_PUBLIC_` prefix means this is already client-bundled in journeys-admin; it isn't a secret. Requiring `doppler login` to use `/impersonate` is friction without a security gain.

**Low — SDK pin `^0.1.0`** (root `package.json:43`)
Caret on 0.x permits semver-breaking minor bumps. Pin `~0.1.0` or `0.1.x` until 1.0.

### Recommendation
**Comment** (PR is draft). The two P1s are design conversations, not line-edits — they should be settled before this comes out of draft and before any of it runs against prod.

PR #9202, opened 2026-05-11
feat(journeys-admin): nest Upload file button in dropzone [NES-1647]

edmonday:
## Summary

- Move the "Upload file" button inside the dropzone for both image (Cloudflare) and Mux video uploads. Drag-and-drop and click-to-browse both still work.
- Mux video only: during upload/waiting/processing the button is replaced by an inline LinearProgress bar; on mobile the status label overlays the bar.
- Add a two-line default copy ("Drop a video/image here" + "or click to browse your files").
- Below `sm`, drop the dropzone chrome and render the full-width Upload button on its own (per design).
- Extract the duplicated dropzone Box styling into a new shared `UploadDropZoneShell` component used by both consumers.
- Preserve react-dropzone pipeline, accepted MIME types, size caps, and existing loading / success / error states.

## Verification

- ImageUpload.spec, AddByFile.spec, UploadDropZoneShell.spec: 26/26 pass
- `npx tsc --noEmit -p apps/journeys-admin/tsconfig.json` clean

## Linear

[NES-1647](https://linear.app/jesus-film-project/issue/NES-1647/move-upload-file-button-into-image-properties-custom-dropzone)

## Summary by CodeRabbit

* **New Features**
  * Consolidated upload drop-zone UI into a unified shell for images and videos, simplifying layout and removing the duplicate bottom upload button.

* **Bug Fixes**
  * Refined error-state logic to treat explicit errors distinctly from error codes in upload flows.

* **Tests**
  * Added and expanded tests for drop-zone rendering, props forwarding, interaction handlers, and upload-state behaviors.

* **Localization**
  * Added English copy for browse hint and uploading progress display.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9202)


csiyang:
Nice work on this — the `UploadDropZoneShell` extraction is the right semantic carve-out, and the mobile progress overlay is a clean UX touch. Few things below.

**One behavior change worth confirming**

`ImageUpload.tsx:139` passes `isActive={loading === true}` to the shell, which means the dashed border + tinted background no longer disappear when upload succeeds. Before this PR, `noBorder = loading || uploadSuccess || hasError`, so the success state was borderless. The PR description says state behavior is preserved, so I think this is unintentional. Suggest:

```tsx
isActive={loading === true || uploadSuccess}
```

In `AddByFile.tsx:167` the `isActive` now also fires on `waiting` (originally `noBorder` didn't include waiting). Probably intentional since the button is replaced by the indeterminate bar — just flagging.

**Nits**

- `libs/locales/en/apps-journeys-admin.json` doesn't yet have `'or click to browse your files'` or `'Uploading... {{progress}}%'`. If extraction isn't automatic, please add them.
- `AddByFile.tsx:172` and `:252` still inline `error != null || errorType != null` — you have `hasError` at line 115 now, may as well reuse it.
- `UploadDropZoneShell.spec.tsx:55` — the "merges caller sx" test only asserts the shell defaults survive, not that the caller's sx actually applies. Consider adding an assertion on a caller-passed property.
- `UploadDropZoneShell` `children: ReactNode` is already in `BoxProps`, so the explicit prop is redundant (harmless).

**Liked**
- Clean dedup of dropzone chrome with a sensible prop API (`isDragAccept` / `isActive` / `hasError`).
- `it.each` over Mux phases is a nice coverage pattern.
- Tightening `hasError` from `error || errorCode` to explicit boolean is a small correctness win.

edmonday:
## Review feedback addressed (36745cf58)

Thanks @csiyang — all four points actioned.

**Fixed:**
- `ImageUpload.tsx:139` — `isActive={loading === true || uploadSuccess}` restores the borderless success state. Good catch, regression from the carve-out.
- `AddByFile.tsx` — replaced three inline `error != null || errorType != null` with the existing `hasError` helper (icon condition, outer color, AlertTriangle display).
- `UploadDropZoneShell.spec.tsx` — sx-merge test now asserts caller's `mt: 10` actually applies (80px) in addition to the shell defaults.
- `UploadDropZoneShell.tsx` — removed redundant `children: ReactNode`; inherited from `BoxProps` now.

**Locale strings:** added `or click to browse your files` and `Uploading... {{progress}}%` to `libs/locales/en/apps-journeys-admin.json`. Note: `extract-translations` runs clean but didn't pick these two up — looks like an i18next-cli AST quirk (other strings in the same files extract fine). Added them manually.

**Acknowledged (no change):**
- `AddByFile.tsx` `isActive` firing on `waiting` is intentional — the button is replaced by the indeterminate bar so the active styling is desirable there.

csiyang (dismissed):
Approved — all the feedback from earlier is addressed cleanly, tests still green.

Two minor things, neither blocking:

1. `ImageUpload` — restoring `noBorder` on `uploadSuccess` also means the shell's darker bg tint (`rgba(239,239,239,0.9)`) now applies for ~4s before the success timer fires. Previously success kept the lighter idle bg. Visually subtle and arguably fine, just flagging.

2. Pre-existing (not introduced here): `UploadDropZoneShell` merges caller `sx` via object spread. Works for object-form `sx` but would break for function-form or array-form. No caller does this today; MUI's idiom is `sx={[defaults, ...(Array.isArray(sx) ? sx : [sx])]}`. Worth knowing if you touch this lib again.

csiyang (approved):

PR #9203, opened 2026-05-12
feat(NES-1659): tailscale dev-server access — setup docs + hostname fixes

csiyang:
## Summary

Adds Tailscale dev-server access so devs can hit local journeys/journeys-admin from their phone (or any Tailnet device) without per-developer env-var setup. Implements all three parts of [NES-1659](https://linear.app/jesus-film-project/issue/NES-1659):

- **Part 1** — `docs/development/tailscale-dev-access.md`: setup runbook (Tailscale install, MagicDNS, hostname convention `tailscale-<initials>-<machine>`, troubleshooting, screenshots TBD)
- **Part 2** — codebase audit for hardcoded dev hostnames (results in `docs/plans/nes-1659-tailscale-dev-server-access.md`)
- **Part 3** — code changes to make Tailscale access zero-config

## Code changes (7 commits, 5 areas)

| Commit | Area | Change |
|---|---|---|
| `ecd708d3c` | `apps/journeys/middleware.ts` | Accept `host.startsWith('tailscale-')` in dev |
| `236866937` | `apps/journeys-admin/.../apolloClient.ts` | Derive gateway URL from `window.location` in dev when `NEXT_PUBLIC_GATEWAY_URL` unset (DI'd `location` param for testability; prod byte-identical) |
| `8ad8c39b7` | `apps/journeys-admin/.../getAuthTokens.ts` | Allow `tailscale-*` hosts in redirect allow-list |
| `1901ff986` | both `next.config.js` | `allowedDevOrigins: ['tailscale-*']` (Next 16 wildcard syntax; regex not supported) |
| `f7be67ddc` | `QrCodeDialog.tsx` | Swap hardcoded `localhost:4100` → `window.location.hostname` in dev under tailscale host |
| `de9da3e18` | `docs/` | Runbook + plan doc |

## Audit findings (Part 2)

Confirmed in scope and applied: `journeys`, `journeys-admin`, `QrCodeDialog` (was hardcoded `:4100`). The earlier `apis/api-gateway/gateway.prod.config.ts` change (commit `655f31128`) was reverted in `069b490c` — dev `gateway.config.ts` has no CORS config, Yoga's default echoes any origin, so no gateway-side change was needed.

Verified out of scope (no journeys/admin cross-traffic during dev flow): `watch`, `videos-admin`, `arclight`, `resources`, `cms` — deferred to docs-only with `NEXT_PUBLIC_GATEWAY_URL` workaround documented.

Verified no-op: `apis/api-gateway/` already binds to `0.0.0.0` via Hive Gateway default (no `host` config set).

Verified deferred: HTTPS / Tailscale Funnel — HTTP tailnet-only is sufficient for dev.

## Test plan

- [x] Jest specs pass — `middleware.spec.ts` (9/9), `apolloClient.test.ts` (14/14), `getAuthTokens.spec.ts` (12/12)
- [x] Typecheck clean on touched files
- [x] Lint clean on touched files
- [ ] Manual: localhost dev still works (golden path)
- [ ] Manual: Tailscale dev — load journeys-admin on phone via `http://tailscale-<host>:4200`, sign in, edit a journey, view QR code
- [ ] Manual: prod-build smoke (`NODE_ENV=production` derivation guard fires correctly)
- [ ] Stage smoke after merge

## Summary by CodeRabbit

* **New Features**
  * Support for using configured developer hostnames so dev servers and gateway resolve correctly from non-local devices.
  * QR code generation and gateway URL resolution now respect developer hostnames for smoother local-to-remote testing.
  * Authentication redirect validation now allows safe developer hostnames when configured.

* **Documentation**
  * Added a detailed Tailscale/MagicDNS setup and troubleshooting guide for accessing services remotely.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9203)


apps/journeys-admin/src/libs/auth/getAuthTokens.ts
```diff
@@ -62,6 +63,21 @@ const ALLOWED_REDIRECT_HOSTS = [
...
+ * dev config) need to round-trip through the sign-in redirect flow. The
+ * secret is only set in dev's Doppler config, so `isDevHost` returns false
+ * everywhere else — absence of the secret IS the gate. The host may carry
+ * an optional `:port` suffix.
+ * See docs/development/tailscale-dev-access.md.
+ */
+export function isAllowedRedirectHost(host: string): boolean {
+  if (allowedHost(host, ALLOWED_REDIRECT_HOSTS)) return true
+  const hostWithoutPort = host.split(':')[0]
+  if (isDevHost(hostWithoutPort)) return true
```
edmonday, line 77: **Concern:** This path widens the prod redirect allow-list, which I think is unintentional given the plan docs PR description states production behavior is byte-identical.

`isDevHost` consults `getDevHosts()`, which in `libs/shared/dev-hosts/src/devHosts/devHosts.ts` unconditionally includes a `LOCAL_DEV_HOSTS = [localhost, 127.0.0.1]` baseline with no `NODE_ENV` gate. Combined with the `host.split(:)[0]` port strip here, the practical effect in **production** is:

- Pre-PR: `admin.nextstep.is` only accepted `?redirect=` targets whose host exactly matched `localhost:4200`, `admin.nextstep.is`, or `admin-stage.nextstep.is`.
- Post-PR: it also accepts `http://localhost/...`, `http://localhost:<anything>/...`, `http://127.0.0.1/...`, etc.

The `dec8cd17` commit message calls the port-stripping a modest, intentional widening for dev — agreed on the dev ergonomics, but the same code path runs in prod since the baseline isnt env-gated. Exploitability is weak (you can only bounce a user to their own machine), but its a wider redirect surface than the PR description claims.

Cheapest fix that preserves the dev ergonomics: gate this fallback on `NODE_ENV !== production` here, so prod stays on the strict exact-match list:

```ts
export function isAllowedRedirectHost(host: string): boolean {
  if (allowedHost(host, ALLOWED_REDIRECT_HOSTS)) return true
  if (process.env.NODE_ENV === production) return false
  const hostWithoutPort = host.split(:)[0]
  if (isDevHost(hostWithoutPort)) return true
  return false
}
```

Alternative: drop `localhost` / `127.0.0.1` from the `LOCAL_DEV_HOSTS` baseline and rely on Doppler `NEXT_PUBLIC_DEV_HOSTS` being unset in prod as the gate (matches the pattern the rest of the PR uses).

edmonday (commented):
Single review concern — no Criticals; CI is green.

See inline on `isAllowedRedirectHost`.

apps/journeys-admin/src/libs/auth/getAuthTokens.ts
csiyang: Good catch on the prod surface widening. Looked at it with Siyang — going with the cleanest fix that keeps the "absence of secret IS the gate" principle pure: dropped the `LOCAL_DEV_HOSTS` baseline entirely. `isAllowedRedirectHost` now does the strict allow-list check first (pre-PR behavior, includes `localhost:4200` exact match), then falls back to `isDevHost()` which is Doppler-only. In prod, Doppler `NEXT_PUBLIC_DEV_HOSTS` / `DEV_HOSTS` are unset → empty array → only strict list applies, identical to pre-PR. Pushed in `bf82a5033`.

edmonday: Confirmed `bf82a5033` restores pre-PR prod behaviour for this path — strict allow-list first, then Doppler-only `isDevHost`. Happy to resolve this thread. Two follow-ons surfaced while verifying:

**1. Proxy port-handling bug — `apps/journeys/proxy.ts:43`.** The other three call sites strip port before `isDevHost` (here via `host.split(':')[0]`; apolloClient/QrCodeDialog via `window.location.hostname`). `proxy.ts` passes the raw `Host` header, which browsers send as `tailscale-dev-<n>.taila2a609.ts.net:4100`. The Doppler list stores bare FQDNs, so `isDevHost` returns `false`, the `/home` rewrite is skipped, and the request falls through to `/${hostname}${path}` = `/tailscale-dev-<n>.taila2a609.ts.net:4100/...` — broken. Spec doesn't catch it (`proxy.spec.ts:71` uses a portless fixture). Fix: `isDevHost(hostname.split(':')[0])` + a `:4100` fixture.

Prod safety check: this fix doesn't re-expose the widening I flagged earlier. The earlier risk was the always-on `LOCAL_DEV_HOSTS` baseline making `getDevHosts()` non-empty in prod. That baseline is gone — `getDevHosts()` is Doppler-only, returns `[]` in prod, so `isDevHost(<anything>)` is `false` whether or not we strip the port.

**2. Runbook doc — gateway row is stale.** `069b490c` reverted the `DEV_HOSTS` consumer in `gateway.prod.config.ts`, and dev gateway runs `gateway.config.ts` with Hive's permissive default. So the `api-gateway` / `DEV_HOSTS` row in the 3-secret table in [03-projects-via-tailscale.md](apps/docs/docs/02-getting-started/04-running-projects/03-projects-via-tailscale.md) and the "every GraphQL request will be CORS-rejected by the gateway" troubleshooting line are now wrong — `DEV_HOSTS` in api-gateway Doppler is unused. Worth scrubbing the stale `655f13128` row from the PR description's commit table too.

csiyang: Thanks for the second look. New commit `580eb0a24` ships both:

1. Proxy port-strip — `apps/journeys/proxy.ts:43` now does `isDevHost(host.split(':')[0])` with a `:4100` Tailscale Host header fixture in the spec. Confirms the rewrite goes to `/home/...` and not the broken catch-all.

2. Runbook doc — `03-projects-via-tailscale.md` rewritten: the 3-secret table is now a single-source `core.dev.DEV_HOSTS` model, and the gateway-CORS troubleshooting entry is gone (dev `gateway.config.ts` is permissive by Yoga default, no config needed). PR description's stale `655f31128` commit row scrubbed too.

apps/docs/docs/02-getting-started/04-running-projects/03-projects-via-tailscale.md
```diff
@@ -0,0 +1,177 @@
...
+  "bob": "tailscale-dev-bob.tailXXXXX.ts.net",
+  "you": "tailscale-dev-<your-name>.tailXXXXX.ts.net"
+}
+```
+
+Multiple teammates merge into the same JSON — when you add yourself, copy the existing value, add your entry, save back. Coordinate in Slack if two of you are editing simultaneously.
+
+The shared helper that reads this lives at `libs/shared/dev-hosts/` (importable as `@core/shared/dev-hosts`). Missing var, empty string, malformed JSON, or a non-object payload all yield `[]` → no relaxation. Fail-closed by default.
+
+The api-gateway doesn't appear in the diagram — its dev config (`apis/api-gateway/gateway.config.ts`) sets no CORS allow-list, so Yoga's default kicks in and echoes any origin. The gateway will accept GraphQL requests from your Tailscale-served apps with no extra setup.
```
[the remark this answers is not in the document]
csiyang: Verified against HEAD `580eb0a24`:

- `apis/api-gateway/gateway.config.ts` (dev) — `grep -n "cors\|origin\|tailscale\|isDevHost\|getDevHosts" gateway.config.ts` returns zero matches. The dev gateway has no CORS configuration, so Yoga's permissive default applies (echoes any origin). The doc reflects this.
- `apis/api-gateway/gateway.prod.config.ts` (prod) — `git diff origin/main..HEAD -- gateway.prod.config.ts` is empty. The file is byte-identical to main after `069b490c` reverted the earlier CORS regex commit (`655f31128`) within this PR.

The doc describes the final state of the PR. The earlier gateway-CORS regex was reverted, so no allow-list configuration is needed — the doc isn't out of sync.

edmonday (commented):
## Review summary

Nicely engineered. The fail-closed Doppler-as-gate design is consistent across all five consumers, the parsing helper handles the awkward edge cases (`Object.values("string")` footgun, `null`, malformed JSON, non-object payloads), and the port-strip in `proxy.ts` pre-empts the `Host: tailscale-dev-x:4100` mismatch I flagged earlier (r3231489594) with a regression test.

A few things I particularly liked:

- Dedicated `libs/shared/dev-hosts` nx lib rather than a one-off helper — good call given five call sites.
- Acknowledged duplication of the parse logic in both `next.config.js` files with a clear lockstep instruction and rationale (CJS vs TS lib). Acceptable trade-off.
- Strong negative-path test coverage across `devHosts.spec.ts`, `getAuthTokens.spec.ts`, and `proxy.spec.ts` — spoofed hosts, prefix/suffix bypass attempts, case-sensitivity all explicitly tested.
- Clear, consistent inline comments on each consumer site explaining "absence of secret IS the gate" and pointing back to the runbook.
- Both of my earlier concerns on the `getAuthTokens.ts:77` thread (LOCAL_DEV_HOSTS baseline widening prod, then the proxy port-strip bug + stale runbook gateway-CORS rows) are addressed at HEAD in `bf82a5033` + `580eb0a24` — verified the helper is Doppler-only and prod returns `[]`.

**Follow-up suggestion (non-blocking):** could you spin up a follow-up ticket to fold the runbook content into the public JesusFilm developer documentation so future onboards find this without having to search the repo? The internal doc under `apps/docs/` is great for now; just want to make sure it surfaces in the right place long-term.

No inline comments — nothing else from me. 👍

edmonday (approved):

PR #9160, opened 2026-05-05
refactor: remove journeys-admin application and related configuration…

mikeallisonJS:
…s from workflows and infrastructure files

## Summary by CodeRabbit

* **Chores**
  * Updated deployment infrastructure for the `journeys-admin` application, transitioning from ECS-based deployment to Vercel deployment pipelines.
  * Removed legacy ECS workflow configurations and infrastructure-as-code definitions for `journeys-admin`.

PR #9204, opened 2026-05-12
refactor: remove DOPPLER_JOURNEYS_ADMIN tokens and related configurat…

mikeallisonJS:
…ions from workflows and infrastructure

## Chores
* Removed Journeys Admin service from production and staging deployments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9204)


csiyang (approved):
Approving — this is exactly the cleanup I asked for on #9160. Minimal and surgical: module callers, Doppler SSM data sources, and dead workflow env vars all removed cleanly; the `DOPPLER_JOURNEYS_ADMIN_TOKEN` use in `main.yml` for `fetch-secrets` is correctly preserved.

Two things to keep in mind before you apply (no need to paste anything here, just flagging):

1. The current Atlantis plan in this PR failed with a transient `openpgp: key expired` error while downloading terraform 1.14.8, not anything in the diff. Worth re-running once to get a clean destroy-only plan to eyeball before apply — should be just the journeys-admin ECS service, ALB target group, listener rule, task definition, CloudWatch log group, and IAM role attachments.

2. Confirm `admin.nextstep.is` / `admin-stage.nextstep.is` are currently serving from Vercel before you apply — once apply runs, the AWS fallback is gone. Almost certainly fine since #9160 merged before this, just worth eyeballing.

Stragglers worth a follow-up cleanup at some point (not this PR's job): the SSM parameters `/terraform/prd/DOPPLER_JOURNEYS_ADMIN_{PROD,STAGE}_TOKEN` in AWS Parameter Store stay around (they were `data` sources, not `resource` blocks), and the repo-level GitHub secret is retained because main.yml still uses it. No urgency on either.

PR #9209, opened 2026-05-14
chore(journeys): migrate tests from Jest to Vitest and update configu…

mikeallisonJS:
…rations

- Removed Jest configuration and related files.
- Updated ESLint and TypeScript configurations to support Vitest.
- Modified test files to replace Jest mocks with Vitest equivalents.
- Updated README to reflect the change in testing framework.
- Adjusted project settings to use Vitest for testing execution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue in video player controls where pending single-click actions could execute after the component was unmounted.

* **Chores**
  * Migrated testing framework from Jest to Vitest across the library.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9209?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mikeallisonJS:
## Review feedback addressed (fc42e1972)

All five CodeRabbit threads were evaluated against the current code and fixed. Verified: `SearchBar` directory suite (9 files, 72 tests) green; the four changed files are ESLint-clean.

**Fixed:**
- `setupTests.ts:68,87` — replaced `.catch(() => {})` with `.catch(() => undefined)`. Confirmed locally these were real `@typescript-eslint/no-empty-function` **errors**.
- `SearchBar.spec.tsx:20` — captured `originalFetch` and restored it in `afterAll` (inside the describe, per the `jest/require-top-level-describe` rule) so the process-wide `global.fetch` override no longer leaks across files. Mirrors the `Conductor.apologistChat.spec.tsx` pattern.
- `SearchBar.spec.tsx:45` — added `refine.mockClear()` in `beforeEach`. Worth flagging: I probed the call count and `toHaveBeenCalledTimes(1)` was passing on a **leaked call from the previous test**, not the debounced call it claims to verify. It now tests the real debounce behaviour.
- `RefinementGroup.spec.tsx:211` — restore `window.ResizeObserver` and the `HTMLElement.prototype` getter spies after the tooltip test (`vi.clearAllMocks()` only clears call data; it doesn't un-install a direct assignment or `spyOn`).

**Fixed (adjusted):**
- `Button.spec.tsx:1357` — CodeRabbit asked to reword the stale "Jest v30" comment. I went further and tested re-enabling the skipped test under Vitest — it still fails on a genuine `toHaveStyle` mismatch (default outline styles not applied), so it stays `it.skip` with a comment describing the real Vitest-era reason and the condition to re-enable.

**Note (pre-existing, out of scope):** the `lint` / `lint-work` checks were already failing before this commit due to repo-wide Prettier drift (~22 files) and ESLint warnings unrelated to these review comments — the three touched spec files were already in the failing Prettier list. My additions are Prettier-clean (all flagged lines are pre-existing). Happy to address the Prettier drift in a separate commit if wanted.

csiyang (approved):

PR #9212, opened 2026-05-14
feat: add analytics tracking for image upload events [NES-1662]

edmonday:
## Summary

- Adds GTM `dataLayer` pushes for image upload success/failure via `sendImageUploadEvent` helper, called from the image upload flow in `journeys-admin`.
- Adds analytics documentation at `apps/docs/docs/03-basics/06-analytics.md` covering the end-to-end GTM → GA4 setup (Data Layer Variables, Custom Event triggers, GA4 Event tags, preview/publish, replication to stage/prod) with screenshots and naming conventions.

## Dependencies

Depends on #9202 ([NES-1647](https://github.com/JesusFilm/core/pull/9202)) — that PR must merge first. This PR targets `main`, so once #9202 is merged this branch can be merged directly without re-targeting.

## Linear

[NES-1662](https://linear.app/jfp/issue/NES-1662)

## Test plan

- [ ] `pnpm run build` passes locally
- [ ] Trigger a successful image upload in `journeys-admin` and confirm `image_upload_success` appears in GTM Preview + GA4 DebugView with `fileName`, `fileSize`, `fileType`
- [ ] Trigger a failed image upload and confirm `image_upload_failure` appears with `errorCode`
- [ ] Verify the new docs page renders correctly via `npx nx serve docs`

## Release Notes

* **New Features**
  * Image upload component now reports upload success and failure events with file metrics.

* **Documentation**
  * Added comprehensive guide for configuring analytics via Google Tag Manager and Google Analytics 4, including event flow, setup instructions, troubleshooting, and naming conventions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9212)


apps/journeys-admin/src/libs/sendImageUploadEvent/sendImageUploadEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function sendImageUploadSuccessEvent({
+  fileName,
+  fileSize,
+  fileType
+}: ImageUploadSuccessEvent): void {
+  sendGTMEvent({
+    event: 'image_upload_success',
+    fileName,
```
jaco-brink: **Concern:** `fileName` is sent verbatim to the GTM `dataLayer` → GA4, and content creators can upload files with arbitrary names. Realistic examples in this admin would include `john-doe-baptism.jpg`, `[real-name]-passport.png`, etc. Google's GA4 terms prohibit PII in event parameters, and the admin auth gate reduces but doesn't eliminate the risk — Google's side doesn't know about our gate.

Worth a deliberate decision before this lights up in prod:
- Strip to just the extension (`png`)?
- Truncate / hash?
- Keep as-is and accept the risk?

If the team has already weighed this, ignore — but I couldn't see it in the PR description.

apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/ImageBlockEditor/CustomImage/ImageUpload/ImageUpload.tsx
```diff
@@ -84,8 +102,18 @@ export function ImageUpload({
...
         setTimeout(() => setSuccess(undefined), 4000)
         setUploading?.(undefined)
+        sendImageUploadSuccessEvent({
+          fileName: file.name,
+          fileSize: file.size,
+          fileType: file.type
+        })
       } catch {
         setSuccess(false)
+        sendImageUploadFailureEvent({
```
jaco-brink, line 109: **Concern:** This catch-block fires `image_upload_failure` with `errorCode: undefined`, while the rejected-file path sends `'file-too-large'` and the Cloudflare path sends `String(cloudflareError)`. In GA4 that means any `fetch` / JSON-parse / unexpected exception lands in an unattributable bucket — the most interesting failure class becomes the least diagnosable.

Two small follow-ons that would make this signal useful:

1. Pass a constant — e.g. `errorCode: 'upload-exception'` — so the failure mode is identifiable in GA4.
2. Capture the caught error. Today the `catch {}` swallows it silently (pre-existing behaviour, not introduced here, but the analytics event makes the loss more visible). At minimum `catch (e) { console.error(e); ... }` so it surfaces in browser tooling / error reporting.

Also: this path has no test. The three new tests in `ImageUpload.spec.tsx` cover success, rejected-file, and Cloudflare error response — none forces `fetch` to throw, so this new `sendImageUploadFailureEvent` call isn't asserted. Easy to add via `mockFetch.mockRejectedValueOnce(new Error('network'))`.

While we're here — out of scope but flagging: the `catch` still doesn't call `setUploading?.(false)`, so a fetch exception leaves the upload spinner stuck. Pre-existing, not yours to fix in this PR.

apps/journeys-admin/src/libs/sendImageUploadEvent/sendImageUploadEvent.ts
```diff
@@ -0,0 +1,41 @@
+import { sendGTMEvent } from '@next/third-parties/google'
+
+interface ImageUploadEventBase {
+  fileName: string
+  fileSize: number
+  fileType: string
+}
+
+interface ImageUploadSuccessEvent extends ImageUploadEventBase {}
```
jaco-brink: **Nit:** `interface ImageUploadSuccessEvent extends ImageUploadEventBase {}` is a no-op. Either drop it and use `ImageUploadEventBase` directly as the param type, or use `type ImageUploadSuccessEvent = ImageUploadEventBase`. Lint isn't catching it here so it's not blocking — just reads a bit awkwardly. (Same shape works in the failure interface because it actually adds `errorCode?`.)

jaco-brink (commented):
## Review summary

Nice, focused PR. The helper + docs combo is the right shape — a typed wrapper for `sendGTMEvent` and a real operational guide for the GTM→GA4 setup that the team can reuse next time. CI is fully green and `<GoogleTagManager>` is wired up in `journeys-admin/pages/_app.tsx`, so events have a real destination.

**Concerns (3 inline):**

- PII risk in `fileName` going to GA4 — user-supplied file names can contain identifying info. Admin-gated route reduces exposure but Google's GA4 ToS still applies.
- Catch-block failure path: no `errorCode`, no logging, no test coverage. The new `image_upload_failure` event in `catch {}` becomes an unattributable bucket in GA4.
- Tiny: empty `interface … extends X {}` no-op in the helper.

No blockers — happy to see this go in once the PII question has a deliberate answer.

edmonday:
## Review feedback addressed (768170d60)

Thanks @jaco-brink for the thorough review.

**Fixed (adjusted):**
- `sendImageUploadEvent.ts` — fileName PII concern: dropped `fileName` from the GTM payload entirely rather than stripping/hashing. `fileType` (MIME like `image/png`) already conveys the equivalent extension signal, so fileName carried no unique analytic value once the PII risk was removed. Updated the helper signature, the four callsites, and the related spec assertions. Will also drop the `dlv - fileName` Data Layer Variable on the GTM side.

**Fixed:**
- `ImageUpload.tsx:112` — catch block now passes `errorCode: 'upload-exception'` so fetch/JSON-parse exceptions are identifiable in GA4 instead of landing in an unattributed `undefined` bucket. Added a test using `mockFetch.mockRejectedValueOnce(new Error('network'))` to cover the catch branch.
  - Skipped the `console.error(e)` suggestion — without a browser error reporter wired to `console.error`, it just adds devtools noise nobody reads in prod. The new GA4 event + UI failure state already provide the diagnostic signal.
  - Left the pre-existing `setUploading?.(false)` issue alone per your own out-of-scope flag.
- `sendImageUploadEvent.ts:9` — dropped the no-op `interface ImageUploadSuccessEvent extends ImageUploadEventBase {}` and used `ImageUploadEventBase` directly as the param type for `sendImageUploadSuccessEvent`.

All 17 tests in `ImageUpload.spec.tsx` + `sendImageUploadEvent.spec.ts` pass.

jaco-brink (approved):
🤖nice🚀

PR #9217, opened 2026-05-15
fix(NES-1677): invalidate api-journeys-modern response cache on gallery mutations

csiyang:
## Summary

Backend split-out from NES-1644 (PR #9174). Closes the `useResponseCache` null-branch invalidation gap that lets a stale `notFound` survive a publish → unpublish → republish cycle on the public template-gallery page.

- Plumbs `cache: Cache` through `BaseContext` so resolvers can call `cache.invalidate(...)`.
- Adds `await context.cache.invalidate([{ typename: 'TemplateGalleryPage' }])` after the `prisma.$transaction` commit in all 6 templateGalleryPage mutations (publish, unpublish, delete, assignJourney, reorderTemplate, update). Gated on `didMutate` so idempotent re-runs and race-lost callers don't fire — the global response cache only churns when state actually changed.
- 60 s TTL on `Query.templateGalleryPageBySlug` preserved as defense-in-depth against the cache-poisoning vector that originally motivated it (NES-1547).
- Two integration regression specs covering the lifecycle end-to-end with the real plugin and real cache:
  - `responseCacheLifecycle.spec.ts` — stub schema canary, 3 tests with positive resolver-call-count assertions so the over-invalidation and back-to-back tests can't pass when invalidate is a no-op.
  - `responseCacheLifecycleReal.spec.ts` — production Pothos schema + real federation, 9 tests: plugin-installed sanity check, canonical NES-1644 lifecycle, 6-mutation parametrized eviction, and the `cache.invalidate` rejection contract.

Background: `useResponseCache`'s automatic entity-ID invalidation can't evict cached `null` responses — they carry no entity ID to match on. The library's null-inference branch (`@envelop/response-cache` plugin.js:212-224) does record bare typenames on cache set, so explicit typename-only invalidation reaches them. Trade-off (documented in `yoga.ts`): typename-only invalidation is intentionally over-eager — clears all cached `TemplateGalleryPage` entries on any mutation. Slug-keyed `buildResponseCacheKey` would be finer-grained; deferred.

Plan: `docs/plans/2026-05-14-001-fix-template-gallery-revalidate-cache-invalidation-plan.md` (moved with the work).

Parent: NES-1644 (frontend-only PR #9174).

## What the post-/ce:review commits added

1. **`fix(NES-1677): gate cache.invalidate on actual state change`** — `invalidate` only fires when the resolver actually mutated state. Idempotent re-publishes, race-lost callers (`updateMany count=0`), no-op unassigns, and reorder no-ops (`sourceIndex === newIndex`) all skip the eviction. Closes the authenticated-caller cache-flush DOS vector flagged in the review. Also moves `invalidate(...)` out of the publish/unpublish `try/catch` (was being coupled to the Prisma P2025 remap).

2. **`test(NES-1677): tighten cache-lifecycle specs`** — parametrized real-schema spec over all 6 mutations (was only publish/unpublish); added positive plugin-installed assertion (catches a silent NODE_ENV ordering regression); added cache-rejection test pinning the "fail loudly post-commit" contract; added `prismaTxSpy.invocationCallOrder < invalidateSpy.invocationCallOrder` assertions in 6 per-mutation specs so a regression that moves `invalidate` inside the transaction is caught; fixed the two vacuous canary tests with resolver-call-count assertions.

3. **`chore(NES-1677): plug per-mutation test gaps`** — added `expect(invalidateSpy).not.toHaveBeenCalled()` to the FORBIDDEN error-path tests across assignJourney + update (CodeRabbit nitpicks); added positive invalidate assertions across the remaining reorder happy-path branches; updated three stale "five mutations" references in the plan doc.

## Test plan

- [x] `responseCacheLifecycle.spec.ts` passes (3 tests, stub-schema canary)
- [x] `responseCacheLifecycleReal.spec.ts` passes (9 tests: plugin-installed sanity + canonical lifecycle + 6 mutations parametrized + cache-rejection contract)
- [x] All 6 mutation specs pass with new `invalidateSpy` assertions, FORBIDDEN-path negative assertions, and per-mutation call-order assertions
- [x] All 112 tests across 14 files in `apis/api-journeys-modern/src/schema/templateGalleryPage/` pass
- [x] `nx lint api-journeys-modern` clean (0 errors)
- [x] `nx type-check api-journeys-modern` clean
- [x] Full `api-journeys-modern` test suite: 909 passed, 1 skipped, 0 failed
- [x] Manually verified end-to-end against `nf start` inside `core_devcontainer-app-1`: publish → unpublish → 60s wait → republish → public page returns 200 (Siyang)
- [ ] CI green
- [ ] Stage QA: same publish → unpublish → republish cycle on stage (Sharon)

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved cache inconsistency so public reads return fresh template gallery pages after publish, unpublish, delete, assign, reorder, or update actions—preventing stale "null" responses.

* **Tests**
  * Added integration and lifecycle tests validating response-cache eviction across publish/unpublish/update flows and ensuring mutations trigger appropriate cache invalidation.

* **Documentation**
  * Added a mitigation plan for stale-cache scenarios and a workflow guide for verifying infra findings before escalation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9217?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


csiyang:
Pushed three commits addressing the /ce:review findings on top of the original PR:

- **`076dffb3e` fix(NES-1677): gate cache.invalidate on actual state change.** Moves `invalidate(...)` out of the publish/unpublish try/catch (was being coupled to the Prisma P2025 remap), and gates it on `didMutate` across publish/unpublish/assignJourney. Idempotent re-publishes, race-lost callers (`updateMany count=0`), and no-op unassigns no longer evict — closes the authenticated-caller cache-flush DOS vector. Tightens the `yoga.ts` comment block to describe what the fix does without speculative caveats.

- **`79437133e` test(NES-1677): tighten cache-lifecycle specs.** Parametrized `responseCacheLifecycleReal.spec.ts` over all 6 mutations (was only publish/unpublish). Added a plugin-installed sanity assertion at the top (catches a silent NODE_ENV ordering regression). Added a cache-rejection test that pins the "fail loudly post-commit" contract. Added `prismaTxSpy.invocationCallOrder < invalidateSpy.invocationCallOrder` assertions in 6 per-mutation specs — so a regression that moves invalidate inside the transaction is caught. Fixed the two vacuous canary tests with positive resolver-call-count assertions. Also fixes the pre-existing TS errors that the type-checker was flagging on the real spec.

- **`60f6baf53` chore(NES-1677): plug per-mutation test gaps; skip reorder no-op invalidate; fix stale plan-doc copy.** Extends the no-op skip pattern to reorder (`sourceIndex === newIndex`). Adds `expect(invalidateSpy).not.toHaveBeenCalled()` to the FORBIDDEN tests in assignJourney (target-team + unassign source-team) and update. Adds positive invalidate assertions to the remaining reorder happy-path branches. Updates three stale "five mutations" references in the plan doc to "six".

Verification: full `api-journeys-modern` test suite passes (909 passed, 1 skipped, 0 failed). `nx lint` 0 errors. `nx type-check` clean.

csiyang:
CodeRabbit cleanup pushed in `95e25679f`:

- **NODE_ENV restoration** in `responseCacheLifecycleReal.spec.ts`: the `vi.hoisted` block now captures the original value and an `afterAll` restores it (or deletes the env var when originally undefined). Closes the cross-spec leakage risk you flagged at `:36-40` and `:101-113`.
- **Assertion-order swap** in `templateGalleryPageAssignJourney.mutation.spec.ts` "throws NOT_FOUND when target page does not exist": the `expect(invalidateSpy).not.toHaveBeenCalled()` assertion now runs after the result-shape assertion (matches the ordering used across the other negative-path tests).

The other nitpicks (FORBIDDEN-on-assign target-team `:429-450`, FORBIDDEN-on-unassign source-team `:452-477`, FORBIDDEN on update `:133-156`) already had `expect(invalidateSpy).not.toHaveBeenCalled()` assertions from earlier passes — verified, no change needed.

All 112 templateGalleryPage tests pass at the new head.

@coderabbitai review

PR #10, opened 2021-08-29
Integrate storybook to journeys and review process

storyworks:
Feel free to merge block-renderer-conductor into main without waiting for this branch.

PR #9221, opened 2026-05-15
feat(journeys-admin): add Template Info side panel to Team Templates [NES-1538]

csiyang:
## Summary

Adds a static educational side panel to the **Team Templates** tab in `journeys-admin` that teaches template creators what templates are, how to make them, how tracking works, and how to share them.

The panel is gated by the existing `teamTemplateCollection` LaunchDarkly flag — same gate as the Collections panel. No new LD flag. The panel only makes sense on the local-template surface that flag already controls, so riding the existing gate avoids unnecessary flag management overhead.

On `md+` viewports the panel renders as a 375px right-anchored column beside the templates grid, narrowing the grid to fit. On `xs`/`sm` it's hidden by default and an inline info icon next to the **"Collections"** heading opens a swipeable bottom drawer (auto-height, max 80vh, drag-handle pill, X close).

## What the panel contains

Always-visible header — **"What templates are about:"** — followed by **four collapsible accordion sections** matching Figma frames `39657-66822`, `39653-66495`, `39657-66677`, `39657-66751`:

1. **Template Types** — Quick-Start (RECOMMENDED first, per Figma) and Regular
2. **How to create** — top-level steps, Links/Images/Video sub-section, Text sub-section with `{{date: May, 8}}` variable example
3. **Tracking and Analytics** — incl. inline Tracking-tab icon next to the "Tracking" label
4. **Sharing and Publishing** — Publish + Use in a team flows

Single-expand UX (opening one section collapses the previous, matching Figma's expanded-state frames). `TemplateInfoPanel` is fully presentational — no Apollo, no `JourneyProvider`, no `useEditor` — so the planned NES-1642 editor floating helper can reuse it as-is.

## Scope adjustments during PR life

This PR went through scope refinements in flight:

1. **`a6fc7908b`** — Removed the entire "Embedding Canva or Google Slides" section + the CollectionDialog Popover affordance (originally absorbed from NES-1660). That helper was educating about the Strategy section feature, which the LTL QA bundle (#9220) is removing entirely from journeys + journeys-admin. NES-1660 stays Canceled (the contextual help intent is gone with the underlying feature).
2. **`029ed6286`** — Dropped the proposed `templateInfoSidePanel` LD flag in favour of riding on the existing `teamTemplateCollection` flag. The panel only makes sense alongside Collections, so a second flag added management overhead without independent rollout value.
3. **[NES-1686] mobile trigger relocation** — Replaced the floating top-right fixed-position IconButton with an inline `text.secondary`-colored info icon placed next to the "Collections" heading on `xs`/`sm` viewports. Removes a misplaced UI element that QA flagged on mobile.
4. **Review fixes (Mike)** — `aria-labelledby` moved off `PaperProps` (where MUI's `role="presentation"` wrapper orphans it) onto the inner Stack with `role="dialog"`; hardcoded `#26262E` + duplicated Open Sans typography in `HowToCreateSection` replaced with `<Typography variant="body1">` + `text.primary`; `subHeadingSx` `#6D6D7D` swapped for `text.secondary`; `infoPanelActive` prop docblock extended to explain the three-level threading; `TemplateInfoPanel.spec` no longer hardcodes a four-entry buttons array.

## Placeholder assets

Seven required screenshots/GIFs ship as clearly-labelled placeholder PNG/GIF assets at Figma dimensions (333px wide). Generated via `apps/journeys-admin/scripts/generate-template-info-placeholders.mjs` for reproducibility. To be swapped for the real Figma exports before broader rollout (or in a follow-up PR).

## Feature flag

- Gated by the existing **`teamTemplateCollection`** flag. No new flag.

## Test plan

- [x] Jest specs: 48 tests passing across `TemplateInfoPanel/*`, `JourneyList.spec.tsx`, `TemplateGalleryPageList.spec.tsx`
- [x] TypeScript clean (`tsc -b apps/journeys-admin/tsconfig.json`)
- [x] ESLint clean (0 errors), Prettier clean
- [x] **Live-tested on Vercel preview** — verified the consolidated `teamTemplateCollection` gate renders the panel correctly at `/?type=templates`
- [ ] **Manual smoke (reviewer)**: with `teamTemplateCollection` on, navigate to the Team Templates tab. Verify the right-anchored 375px column appears with the header + 4 collapsed accordions; expand each in turn and confirm copy + image paths match Figma intent. Resize to mobile — confirm the inline info icon next to the "Collections" heading opens the bottom drawer, swipe-to-dismiss works, X close works.

## Notable implementation notes

- **Desktop panel uses `permanent` MUI `Drawer`** rather than an in-flow column. Deliberate deviation from the plan doc — kept because it matches the journeys-tab `SidePanel` chrome exactly (borders, radii, mt/mr) without re-implementing the layout. Side effect: `infoPanelActive` has to be threaded through `JourneyListView → TeamMode` so the 3-dots overflow menu can avoid the panel's left edge — see the `infoPanelActive` docblock.
- **Mobile drawer**: `SwipeableDrawer anchor="bottom"`, auto-height + 80vh cap, swipe-to-dismiss + X close fallback.
- **Figma typos corrected**: `RECOMENDED → RECOMMENDED`, `your need → you need`, `there dots menu → three dots menu`.
- **Template Types renders Quick-Start before Regular** (Figma `39657-66822` order), even though the ticket body listed Regular first.

## Summary by CodeRabbit

* **New Features**
  * Template Info side panel for Team Templates: desktop right-side permanent panel and mobile Info button opening a bottom drawer with four single-expand sections (Template Types, How to Create, Tracking & Analytics, Sharing & Publishing).
* **Content & Localization**
  * Added localized copy and illustrative media (GIFs/screenshots) for each section.
* **Assets**
  * Added placeholder image assets for the Template Info panel.
* **Tests**
  * New/updated test coverage for panel behavior, accordions, mobile trigger and drawer.
* **Documentation**
  * Added implementation plan and requirements for the Template Info panel.


mikeallisonJS (changes requested):
## Code Review: feat(journeys-admin): template info side panel [NES-1538]

### Summary

The component is well-architected — purely presentational, no Apollo/provider coupling, and a thorough test suite. The accordion single-expand logic is correctly implemented, accessibility foundations are solid, and the placeholder asset strategy is clean. Two issues need resolution before merge; the rest are warnings and suggestions.

---

### 🔴 Critical

#### 1. Feature flag gate contradicts the committed plan document

**File:** `JourneyList.tsx` (~line 55–60) + `docs/plans/2026-05-14-001-feat-nes-1538-template-info-side-panel-plan.md`

The implementation gates the panel on `teamTemplateCollection`:

```tsx
const showTemplateInfoPanel =
  teamTemplateCollection === true && currentContentType === 'templates'
```

The plan document committed in this PR (Requirements R1, U1, Key Technical Decisions) still requires a **separate** `templateInfoSidePanel` flag. That's a direct contradiction in the committed codebase.

Using the existing flag is a defensible call — but the plan must reflect that decision. Either:
- **(a)** Update the plan to document the rationale for consolidation and remove the conflicting requirement language, or
- **(b)** Implement the separate `templateInfoSidePanel` flag as the plan specifies.

Shipping with contradictory documentation is the concrete risk here.

---

#### 2. PR title promises Canva URL help — Section 5 is missing

**File:** `TemplateInfoPanel.tsx`

The PR title says **"with Canva URL help"** and the plan specifies five accordion sections. The implementation ships four — `EmbeddingCanvaOrGoogleSlidesSection` (Section 5) is absent from `TemplateInfoPanel.tsx`, missing from the `TemplateInfoSectionId` union type, and has no i18n keys.

This is the most operationally valuable section (directly addresses the "refused to connect" support burden). If it's intentionally deferred, please update the PR title/description with an explicit follow-up ticket reference.

---

### 🟡 Warnings

#### 3. Desktop panel uses `permanent` Drawer — contradicts plan, introduces fragile layout

**File:** `JourneyList.tsx` (~line 195–230)

The plan's Key Technical Decisions explicitly states: *"Render the panel as a static right-side column in `JourneyList.tsx` rather than as a MUI `Drawer`. Rationale: Figma `39640-65061` shows the panel docked inside the content area, not as an overlay."*

The `permanent` Drawer renders into a MUI Portal at body root with `position: fixed`. Consequences:
- The panel won't scroll with the content area
- `mt: 15.5, mr: 5` are magic numbers that break silently if the header height ever changes
- It's also *why* `infoPanelActive` had to be threaded through `JourneyListView → TeamMode` to fix the 3-dots menu alignment — that prop wouldn't exist if the panel were in document flow

If keeping the Drawer approach, please add a code comment explaining the deliberate deviation from the plan and acknowledge the `mt`/`mr` fragility.

#### 4. `aria-labelledby` on `PaperProps` won't be read by screen readers

**File:** `JourneyList.tsx` (~line 252)

```tsx
PaperProps={{
  'aria-labelledby': 'template-info-drawer-title'
}}
```

MUI's `SwipeableDrawer` wraps Paper in a `<div role="presentation">`. Screen readers look for `aria-labelledby` on the element with an interactive role — placing it on the inner Paper means it's orphaned. For correct association, apply `role="dialog"` and `aria-labelledby` directly on the inner `Stack` (`data-testid="TemplateInfoPanelMobileDrawer"`).

#### 5. `color: '#26262E'` in `HowToCreateSection` breaks the shared token pattern

**File:** `HowToCreateSection.tsx` (~line 22)

Every other section and `styles.ts` uses `'text.primary'` for body text. The hardcoded hex deviates from the theme and won't adapt to dark mode. Replace with `color: 'text.primary'`.

#### 6. `next-i18next/pages` import — verify consistency

**File:** `JourneyList.tsx` (~line 8)

`next-i18next/pages` is the Pages Router–specific export. Confirm this matches the established convention in the rest of `journeys-admin`.

#### 7. `<code>` elements with proportional font lose monospace semantics

**File:** `HowToCreateSection.tsx` (~lines 28–31)

```tsx
const inlineCodeSx = {
  fontFamily: 'Open Sans, sans-serif',
  fontStyle: 'italic'
}
```

Variable syntax examples like `{{date: May, 8}}` rendered in a proportional font may confuse users who try to type them verbatim. If Figma requires the italic treatment, keep it — but add a comment explaining the intentional deviation, or use `<Box component="em">` if monospace semantics aren't intended.

---

### 🔵 Suggestions

#### 8. `#6D6D7D` in `styles.ts` was flagged as unresolved in the plan

The plan's open questions note: *"Whether `#6D6D7D` should map to `text.secondary` — verify in `libs/shared/ui/themes/`."* This was left as a hardcoded hex. Check if `text.secondary` resolves to `#6D6D7D`; if so, replace. If not, add a comment: `// #6D6D7D = Editor/Secondary/Light; no theme token match at time of authoring`.

#### 9. Flat i18n keys risk collisions

`"Regular"`, `"Text"`, and `"RECOMMENDED"` as top-level keys in `apps-journeys-admin.json` could collide with existing uses of those strings elsewhere in the admin. Check for existing key conflicts.

#### 10. `getByRole('presentation')` in tests is fragile

**File:** `JourneyList.spec.tsx` (~lines 314–335)

Any other MUI component using `role="presentation"` in the render tree will cause a false positive. Use `screen.getByTestId('TemplateInfoPanelMobileDrawer')` as the open-state assertion instead.

#### 11. `generate-template-info-placeholders.mjs` double-encodes PNG

The script runs `sharp()` twice (buffer → re-decode → file). Simplify to:
```js
await sharp(Buffer.from(svg))
  .png({ compressionLevel: 9 })
  .toFile(resolve(OUT_DIR, slot.file))
```

#### 12. `infoPanelActive` prop threading is tech debt worth documenting

The prop exists solely because of the Drawer's fixed positioning (see warning #3). Add a comment linking this back to the layout choice so future maintainers understand why it's there.

#### 13. `getAllByRole('button')` order test hardcodes four sections

Will need updating when Section 5 lands. Consider querying individual buttons by accessible name instead.

---

### What Looks Good

- **Fully presentational architecture** — no Apollo/provider coupling means it will slot into NES-1642 as-is ✓
- **Single-expand accordion logic** — `makeHandleChange` + fully-controlled `TemplateInfoAccordion` is the right pattern ✓
- **`useEffect` closing the drawer when the flag turns off** — a thoughtful edge case that's easy to miss ✓
- **`loading="lazy"` on every image** — combined with single-expand, users only load what they open ✓
- **Feature flag gate computed once** at the top of `JourneyList` — no scattered `useFlags()` calls in leaf components ✓
- **Test coverage** — testing accordion cross-interaction, Figma typo corrections, and image `src`/`alt` pairs are all meaningful regression guards ✓
- **Placeholder asset strategy** — production filenames at correct dimensions, generated script for reproducibility, swap-in is a one-file replace ✓

csiyang:
Thanks @mikeallisonJS — pushed `06e72e9` (review fixes) and `f243dd2` (merge `main`). Recap by your numbering:

### 🔴 Critical
- **#1 (plan doc vs `teamTemplateCollection` flag):** intentionally deferred. The plan doc is going stale on purpose now that the solution lives in the code + Linear; ce/me will keep the solution doc current and let the plan rot.
- **#2 (PR title promises "Canva URL help"):** ✅ retitled to `feat(journeys-admin): add Template Info side panel to Team Templates [NES-1538]` and rewrote the PR body to reflect the actual shipped scope + the in-flight scope refinements (Canva help removal, flag consolidation, NES-1686 mobile relocation, review fixes).

### 🟡 Warnings
- **#3 (desktop `Drawer permanent` deviates from plan):** keeping as-is — confirmed this is the approach we actually want. The fragility you flagged (`mt: 15.5, mr: 5` magic numbers + the existence of `infoPanelActive` prop) is now documented inline via #12.
- **#4 (`aria-labelledby` orphan on `PaperProps`):** ✅ moved off `PaperProps` onto the inner `Stack` with `role="dialog"` so the label is on the element that actually carries the dialog role.
- **#5 (`color: '#26262E'` in `HowToCreateSection`):** ✅ replaced the entire `calloutBoxSx` typography (hardcoded family/weight/size + the hex) with `<Typography variant="body1" color="text.primary">` — chrome `sx` now only carries border/radius/padding.
- **#6 (`next-i18next/pages` import):** verified — 357 files in `journeys-admin` use this path, 0 use bare `next-i18next`. Established convention, no change.
- **#7 (proportional-font `<code>` for `{{date: May, 8}}`):** ✅ replaced `inlineCodeSx` with `<Typography component="code" variant="body1" sx={{ fontStyle: 'italic' }}>`. Per Figma the italic treatment is intentional (it's a placeholder marker, not literal code), but it now inherits the theme variant instead of duplicating the Open Sans block. Happy to swap to a true monospace font if you'd rather the syntax be copy-pasted verbatim.

### 🔵 Suggestions
- **#8 (`#6D6D7D` in `styles.ts`):** ✅ swapped for `'text.secondary'`. Confirmed `journeysAdmin` palette → `text.secondary` = `palette[700]` = `#6D6D7D` exactly.
- **#9 (flat i18n keys `Regular` / `Text` / `RECOMMENDED`):** deferred. Grepped the locale file — no current collision. Will add `t('…', { context: … })` discriminators if/when a real conflict shows up.
- **#10 (`getByRole('presentation')` in tests):** **tried, reverted.** `SwipeableDrawer` keeps its inner content `Stack` mounted even with `disableSwipeToOpen` (for the bottom-edge swipe gesture). Only the modal portal wrapper (`role="presentation"`) actually mounts/unmounts with `open`, so the testid-based assertion you suggested can't signal close — `queryByTestId('TemplateInfoPanelMobileDrawer')` returns the element in both states. Kept the `role="presentation"` assertion + added a comment in the spec explaining the constraint and noting the spec stubs `TemplateGalleryPageList` so no other `presentation`-role element appears in this render tree.
- **#11 (`sharp` double-encodes the placeholder PNGs):** skipped — placeholders are being replaced with real Figma exports, not worth touching the generator now.
- **#12 (`infoPanelActive` prop threading is tech debt):** ✅ extended the prop's docblock in `JourneyListView.tsx` to explain the three-level threading + tie it explicitly to the `Drawer permanent` choice (so removing the Drawer would let this prop go).
- **#13 (`getAllByRole('button')` hardcodes four sections):** ✅ rewrote both order/collapsed tests in `TemplateInfoPanel.spec.tsx` to look up each section by accessible name + assert positional ordering, so adding a future section doesn't require rewriting either test.

### Status
50/50 Jest, 0 lint errors, Prettier clean, TS clean. CLA bot is green (the two original commits are now authored as `Siyang Cao` after a rebase). The new CodeRabbit pass on `06e72e9` only flagged the plan-doc MD040 markdown-lint warnings, which fall under #1 → deferred.

---

mikeallisonJS (approved):
## Follow-up Review — Round 2

Thanks for the thorough pass on these. Checking each item from my original review, plus flagging two new issues introduced by the fixes.

---

### Original findings — status

| # | Finding | Status |
|---|---------|--------|
| 1 | Flag gate contradicts plan doc | ✅ Fixed — plan doc + code comment now document the deliberate consolidation |
| 2 | Section 5 / Canva URL help missing from PR title | ✅ Fixed — title updated to match the four implemented sections |
| 3 | Desktop uses `permanent` Drawer vs plan's `Box` | ✅ Accepted + documented — PR description and `infoPanelActive` docblock now explain the trade-off |
| 4 | `aria-labelledby` on `PaperProps` orphaned | ✅ Fixed — `role="dialog"` + `aria-labelledby` moved to inner `Stack` |
| 5 | `color: '#26262E'` hardcoded hex | ✅ Fixed — `color="text.primary"` on `Typography` |
| 6 | `next-i18next/pages` import consistency | ➖ Not addressed — if it's the app-wide convention this is fine; flagging for your awareness |
| 7 | `<code>` elements using proportional font | ➖ Intentional per Figma (Open Sans Italic 16/24) — noted, not a blocker |
| 8 | Mobile trigger floating-button placement | ✅ Fixed — inline icon next to "Collections" heading (NES-1686) |
| 12 | `#6D6D7D` not checked against theme token | ✅ Fixed — `subHeadingSx` now uses `color: 'text.secondary'` |
| 13 | `getByRole('presentation')` fragile | ✅ Addressed — `TemplateGalleryPageList` stubbed to isolate the element; comment explains the constraint |
| 14 | `infoPanelActive` prop threading undocumented | ✅ Fixed — detailed JSDoc added to both `JourneyListViewProps` and `TeamModeProps` |
| 16 | `getAllByRole('button')` hardcodes four sections | ✅ Fixed — test now queries by accessible name |

---

### Two new issues introduced by the fixes

#### 🟡 A — `sx={{ width: 1, height: 1 }}` resolves to 100%, not 1px

**File:** `JourneyList.tsx` — the visually-hidden `aria-labelledby` target

```tsx
<Box
  id="template-info-drawer-title"
  sx={{
    position: 'absolute',
    width: 1,      // ← resolves to width: 100% in MUI sx
    height: 1,     // ← resolves to height: 100% in MUI sx
    overflow: 'hidden',
    clip: 'rect(0 0 0 0)'
  }}
>
```

In MUI's `sx`, a numeric `1` for `width`/`height` uses the sizing scale and resolves to `100%`, not `1px`. The visually-hidden pattern requires explicit string values: `width: '1px', height: '1px'`. Today the element stays invisible because `clip: rect(0 0 0 0)` suppresses rendering — but `clip` is deprecated CSS2, and if it's ever cleaned up, this element will become a 100%-wide invisible overlay that intercepts pointer events inside the drawer.

Fix: `width: '1px', height: '1px'` (strings, not numbers), or import `visuallyHidden` from `@mui/utils`:
```tsx
import { visuallyHidden } from '@mui/utils'
// ...
<Box id="template-info-drawer-title" sx={visuallyHidden}>
  {t('Template info')}
</Box>
```

#### 🔵 B — Missing `aria-modal="true"` on the drawer `Stack`

**File:** `JourneyList.tsx` — inner `Stack` with `role="dialog"`

```tsx
<Stack
  role="dialog"
  aria-labelledby="template-info-drawer-title"
  // ← missing aria-modal="true"
>
```

`role="dialog"` without `aria-modal="true"` means VoiceOver/NVDA may continue reading content behind the open drawer on some browsers. MUI's Modal does set `aria-hidden` on the rest of the document internally, which partially mitigates this — but the explicit `role="dialog"` without the companion `aria-modal` is non-standard.

Fix: add `aria-modal="true"` to the Stack.

---

### Other notes

- The `useEffect` that auto-closes the drawer has `templateInfoDrawerOpen` in its dependency array, which causes an unnecessary extra run on every drawer-open/close cycle when the panel is visible. Moving to `[showTemplateInfoPanel]` only would be cleaner, but the `if` guard makes the current form correct. Low-priority cleanup.
- i18n flat keys (`"Regular"`, `"Text"`) are a future translator risk, but not a today problem. Fine to leave.

---

### Verdict

**Approve** — the two new issues above are the only things I'd want fixed before this merges. Issue A is the more important one (silent `clip` dependency); Issue B is straightforward to add. Everything else from round 1 is cleanly resolved. Great work on the mobile trigger relocation and the a11y wiring.

PR #9222, opened 2026-05-18
feat(ci): add firewall-rate-limit-check workflow (NES-1581)

jaco-brink:
## Summary

Adds a manual GitHub Actions workflow that probes `/api/chat` from GHA-owned IPs while a parallel sentinel verifies legit traffic still serves — the detection harness for the Vercel Firewall rule being rolled out in [NES-1581](https://linear.app/jesus-film-project/issue/NES-1581/safety-r4-vercel-firewall-rules-on-apichat).

## Why a workflow, not app code

The Vercel Firewall sits *in front of* the Next.js serverless function — edge protection, by design, lives outside the function. There is nothing meaningful to add inside `apps/journeys` for this risk; the rule is configured in the Vercel dashboard. This PR is the **test harness** that proves the rule fires (probe) without dragging real traffic down (sentinel). It is the "rules detection check" called for in section 4 of the Linear ticket.

## How to run

1. Owner configures the Vercel Custom Rule per the rollout plan (Log first, never straight to Deny).
2. GitHub → Actions → **Firewall Rate-Limit Check (NES-1581)** → **Run workflow**.
3. Pick `target_env` (defaults to `stage` — production is opt-in). Defaults of `probe_count=100`, `probe_rps=2`, `sentinel_path=/` are sane starting points.
4. Read the run summary: probe table (2xx / 429 / 5xx counts) + sentinel table (success rate, p50, p95). The probe job intentionally does **not** fail on 429 — 429 is the success signal once the rule is past Log mode.
5. Promote Log → Challenge → Deny as each stage settles (per the rollout plan); re-run the workflow at each stage.

## Vercel-side context

This PR is the test harness only. The actual firewall rule lives in the Vercel dashboard and is rolled out per the plan in NES-1581:

| Stage | Action | What we verify |
| --- | --- | --- |
| 1 | Log | Threshold fires correctly; zero user impact; firewall logs reflect probe hits. |
| 2 | Challenge | Bots blocked; real browsers solve transparently; sentinel green. |
| 3 | Deny | Hard block at the edge; sentinel still green. |

Owner allowlists their home IP at the top of the rule chain *before* anything in this PR runs — see section 1 of the Linear ticket.

## Notes for reviewers

- Stage host literal `your-stage.nextstep.is` (and prod `your.nextstep.is`) is hardcoded with a comment pointing at the repo references it was copied from. If `vars.JOURNEYS_STAGE_HOST` / `vars.JOURNEYS_PROD_HOST` get added later, swap to `${{ vars.* }}`.
- No `environment:` block — repo has no `.github/environments` config. A `TODO` comment in the workflow flags this for the owner to wire up later if a `production` environment with required reviewers gets added.
- Runner is `blacksmith-2vcpu-ubuntu-2204` to match repo convention (every other workflow uses it).

## Test plan

- [ ] Lint / required CI checks green on this PR
- [ ] Workflow appears under the **Actions** tab as "Firewall Rate-Limit Check (NES-1581)"
- [ ] Manual run against `stage` with the rule in **Log mode** completes: probe shows ~100% 2xx, sentinel shows 100% 2xx with p95 < 3s
- [ ] After rule promoted to **Challenge** / **Deny**: probe shows non-zero 429 count, sentinel still 100% 2xx

## Summary by CodeRabbit

* **Chores**
  * Added a manually-triggered workflow to validate firewall rate-limit behavior against the /api/chat endpoint for stage and production targets.

* **Documentation**
  * Added documentation describing how to run the validation, pass/fail rules for probes and sentinel checks, and configuration options (request counts, RPS, sentinel path, duration).

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9222?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/firewall-rate-limit-check.yml
```diff
@@ -0,0 +1,295 @@
...
+          {
+            echo "## Run configuration"
+            echo ""
+            echo "| Setting | Value |"
+            echo "| --- | --- |"
+            echo "| target_env | \`${TARGET_ENV}\` |"
+            echo "| host | \`${host}\` |"
+            echo "| probe_count | ${count} |"
+            echo "| probe_rps | ${rps} |"
+            echo "| sentinel_path | \`${{ inputs.sentinel_path }}\` |"
```
jaco-brink: **Concern (hardening):** `${{ inputs.sentinel_path }}` is interpolated directly into this bash heredoc. Every other input in this same step (`TARGET_ENV`, `PROBE_COUNT`, `PROBE_RPS`) is passed via the `env:` map at line 66–69 and dereferenced as `${SENTINEL_PATH}` — the sentinel job at line 207 already does this. Workflow_dispatch requires repo write access so blast radius is contained, but this is the documented GHA script-injection hardening pattern (https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable) and the inconsistency is the only place in this file that diverges.

Suggested fix: add `SENTINEL_PATH: ${{ inputs.sentinel_path }}` to the step `env:` block and dereference it as `${SENTINEL_PATH}` here, the same way the sentinel job at lines 205–208 already does.

.github/workflows/firewall-rate-limit-check.yml
```diff
@@ -0,0 +1,295 @@
...
+            echo "| sentinel duration | ~${duration}s |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  probe:
```
jaco-brink, line 90: **Concern:** No `timeout-minutes` on this job. With `--max-time 10` per curl and an unbounded `probe_count` input, an unreachable host can keep the probe job running for `probe_count × 10s` (~16 min at the default 100; arbitrary at higher values, capped only by GHA's 6-hour default). Recommend `timeout-minutes: 15` (or whatever the realistic upper bound is for the configured defaults plus headroom). Same applies to the sentinel job at line 200.

.github/workflows/firewall-rate-limit-check.yml
```diff
@@ -0,0 +1,295 @@
...
+          if [ "${count_5xx}" -gt 0 ]; then
+            err5_pct=$(awk -v e="${count_5xx}" -v t="${total}" 'BEGIN { printf "%.2f", (e*100)/t }')
+            if awk -v p="${err5_pct}" 'BEGIN { exit !(p > 50.0) }'; then
+              echo "::error::5xx rate ${err5_pct}% (> 50%) — origin under stress"
+              fail=1
+            fi
+          fi
+          exit "${fail}"
+
+  sentinel:
```
jaco-brink, line 115: **Concern (same as probe job):** Add `timeout-minutes:` to bound the sentinel — its `while date +%s < end` loop will keep going as long as `DURATION_SECONDS` says, and `DURATION_SECONDS` is derived from `probe_count / probe_rps`, both user-controlled. A hard ceiling here defends against a malformed (or just generous) input pair.

jaco-brink (commented):
## Review summary

Clean, well-scoped detection harness — the 429-as-success semantics, sentinel/probe coupling via derived duration, and per-env `concurrency` group are all well thought through. Three concerns + one nit below; nothing blocking for a draft.

**Concerns**

- **Missing `permissions:` block** — endorsing CodeQL's three inline notes (lines 106, 200, 295). No job needs write; add `permissions: contents: read` at the workflow root. Matches repo convention (`codeql-analysis.yml`, `danger.yml`, `ai-build-spike.yml` all set explicit permissions).
- **Script-injection hardening on `sentinel_path`** — see inline at line 102.
- **No `timeout-minutes` on probe/sentinel jobs** — see inline at lines 106 / 200.

**Nit**

- `probe_count` / `probe_rps` could use `type: number` instead of `type: string` + regex, so the dispatch form validates before the run starts.

**Positives**

- 429-as-success documented and asserted in code (won't rot).
- Sentinel duration derived from `ceil(count/rps) + 15s` — probe and sentinel stay coupled without manual sync.
- p95 calc correctly drops `time=0` rows from errored curls.
- Host literals carry a comment pointing at their in-repo references and the `vars.JOURNEYS_*_HOST` swap path.

CI is green except for Danger's "assign someone to merge" — not a code issue.

jaco-brink:
## Review feedback addressed (8d25aeb59)

**Fixed:**
- `firewall-rate-limit-check.yml` (top-level) — added `permissions: contents: read` block. Closes the three CodeQL findings (workflow permissions) in one shot since the block applies to all jobs.
- `firewall-rate-limit-check.yml:66-69` — added `SENTINEL_PATH: ${{ inputs.sentinel_path }}` to the prepare step's env map; line 102 now dereferences `${SENTINEL_PATH}` instead of interpolating `${{ inputs.sentinel_path }}` directly. Matches the pattern already used by the sentinel job and by every other input in the prepare step.
- `firewall-rate-limit-check.yml` probe + sentinel jobs — added `timeout-minutes: 15` to both. Bounds the worst case from user-controlled `probe_count` / `probe_rps` inputs.

**Skipped / not in my scope:**
- The `Danger` check failure is a pre-existing 'Please assign someone to merge this PR' procedural ask, not a code issue. Will assign a reviewer when ready to come out of draft.

YAML re-validated locally with PyYAML; CI will run on push.

jaco-brink:
## Review feedback addressed (2e0294f0c)

**Fixed:**
- `firewall-rate-limit-check.yml:92-94` (CodeRabbit) — added explicit `awk` positive-value check after the regex, rejects `probe_rps=0` / `0.0` before they hit the duration division (and the probe-step `1.0/r` sleep). Matches the exit-status idiom this file already uses for the 10%/50% threshold checks.

jaco-brink:
## Refactor: rewrite firewall-rate-limit-check logic in TypeScript

**What + why:** Replaced ~250 lines of dense inline bash (awk percentiles, exit-code juggling, two `mktemp` scratch dirs) in `.github/workflows/firewall-rate-limit-check.yml` with typed tsx scripts so the workflow is junior-readable. The workflow is now orchestration-only (~140 lines: inputs + permissions + concurrency + 3x checkout/pnpm/node preamble + 3x one-line `tsx` invocation).

### Files

- **Modified:** `.github/workflows/firewall-rate-limit-check.yml` (now orchestration-only)
- **New:** `tools/scripts/firewall-rate-limit-check/`
  - `prepare.ts` — input validation + host resolution + sentinel-duration math
  - `probe.ts` — POSTs to `/api/chat` at the requested rps via native `fetch` + `AbortSignal.timeout(10s)`
  - `sentinel.ts` — GETs sentinel path at ~1 req/sec, computes p50/p95 latency
  - `types.ts` — shared `RequestOutcome` / `ProbeSummary` / `SentinelSummary`
  - `README.md` — junior-readable purpose + when-to-run + pass/fail guide

### Behaviour parity

Confirmed unchanged across the board:
- Host literals (`your-stage.nextstep.is` / `your.nextstep.is`) and the `Unknown target_env` rejection.
- Input validation: `probe_count` positive integer, `probe_rps` positive number, both `0`/`0.0` rejected.
- Duration formula: `ceil(count / rps) + 15`.
- Probe headers, request body shape (`{messages, sessionId: "firewall-probe-<RUN_ID>"}`), User-Agent string, 10s timeout, ~`1/rps` pacing, 429-is-not-failure policy, and the two failure thresholds (network errors > 10%, 5xx > 50%).
- Sentinel path normalisation (exactly one leading slash), 15s timeout, ~1 req/sec, percentile indexing replicated (`floor((N-1)*pct) + 1`, clamped, computed over successful requests only), pass criteria 100% 2xx AND p95 < 3.0s.
- Markdown summary tables and the "29-in-Log-mode-is-expected" note retained verbatim.
- `permissions: contents: read`, `concurrency` group, `timeout-minutes: 15`, and `prepare → {probe, sentinel}` parallel topology preserved.

No bugs found in the bash worth flagging — the prior version was correct, just dense.

### No new dependencies

Node 22 built-ins only: native `fetch`, `AbortSignal.timeout`, `performance.now()`, `node:fs/promises`. `tsx` is already a dev dep used by `app-deploy.yml`.

### Validation

- TypeScript type-check (`tsc --noEmit --strict` against the four new `.ts` files): clean.
- Smoke-tested `prepare.ts` with valid inputs (stage/10/2 → `duration_seconds=20`; production/100/2 → `duration_seconds=65`) and verified non-zero exit on `probe_count=0`, `probe_rps=0`, and unknown `target_env`.
- Smoke-tested `probe.ts` and `sentinel.ts` against an unresolvable host (`nonexistent.invalid`); both handle the network error gracefully, write the expected summary table, and exit 1 on the 100% error condition.
- YAML re-parsed cleanly: `python3 -c "import yaml; yaml.safe_load(open(...))"` → OK.
- ESLint skipped: monorepo's flat eslint config isn't usable from the worktree without a full `pnpm install`. TS type-check + smoke tests cover the new files.

Pushed as `ea4145241` on top of `2e0294f0c`. The workflow remains manual (`workflow_dispatch` only) and was not triggered.

.github/workflows/firewall-rate-limit-check.yml
```diff
@@ -0,0 +1,137 @@
...
+# (`stage`) plus the explicit choice prompt to keep production opt-in.
+
+concurrency:
+  group: firewall-rate-limit-check-${{ inputs.target_env }}
+  cancel-in-progress: false
+
+jobs:
+  prepare:
+    name: Resolve host
+    runs-on: blacksmith-2vcpu-ubuntu-2204
```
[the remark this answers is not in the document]
jaco-brink: The `blacksmith-2vcpu-ubuntu-2204` runner is the established convention across this repo (18 other workflows use it). Adding `.github/actionlint.yml` to register it is a repo-wide tooling concern outside the scope of this PR.

tools/scripts/firewall-rate-limit-check/probe.ts
```diff
@@ -0,0 +1,214 @@
...
+  for (let i = 1; i <= probeCount; i += 1) {
+    const outcome = await sendProbe({ url, body, userAgent })
+    outcomes.push(outcome)
+    if (outcome.error != null) errorMessages.push(outcome.error)
+    if (i % 10 === 0) {
+      console.log(`  sent ${i}/${probeCount}`)
+    }
+    if (i < probeCount) {
+      await sleep(sleepMs)
+    }
```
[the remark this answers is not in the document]
jaco-brink: Thanks — the concern carries over from the old bash version, but I'm going to keep the current pacing. Two reasons:

1. The workflow input describes `probe_rps` as "approximate" and explicitly "Keep gentle — this is a detection check, not a load test." Smooth, strictly-under-configured pacing is the goal.
2. The proposed catch-up scheduling (`nextTickAt` with `Math.max(0, ...)`) will fire back-to-back requests after any fetch-latency spike. Bursts can produce intermittent firewall-rule triggers that distort the detection signal, where a smooth probe would be steady. At the default 2 rps + ~100 ms fetch latency the under-shoot is ~17% — negligible for "did the rule fire" detection.

Marking outdated/won't-fix.

jaco-brink:
## Review feedback addressed (c9b603167)

No code changes this round — both remaining unresolved threads challenged inline.

**Challenged:**
- `.github/workflows/firewall-rate-limit-check.yml:65` (CodeRabbit, actionlint `blacksmith-2vcpu-ubuntu-2204`) — repo-wide convention (18 workflows already use this runner); registering it in actionlint config is outside this PR's scope. See inline reply.
- `tools/scripts/firewall-rate-limit-check/probe.ts` (CodeRabbit, probe pacing under-shoots RPS) — workflow is a gentle "detection check, not a load test" with explicitly "approximate" RPS; proposed catch-up scheduling produces bursts on latency spikes that distort detection signal. See inline reply.

**Already resolved (prior rounds):**
- CodeQL workflow-permissions warnings (×3) — addressed by `permissions:` block in 8d25aeb59.
- `SENTINEL_PATH` script-injection hardening — addressed by env-pass in 8d25aeb59.
- `timeout-minutes` on probe + sentinel jobs — addressed in 8d25aeb59.
- `probe_rps=0` division by zero — addressed in 2e0294f0c.

jaco-brink:
## Review feedback addressed (5caf3293a)

**Fixed:**
- `tools/scripts/firewall-rate-limit-check/probe.ts:21-26` — Added validation for `PROBE_COUNT` (positive integer) and `PROBE_RPS` (positive finite number) immediately after parsing. Mirrors the existing pattern in `sentinel.ts:21-24`. Without these guards, an invalid `PROBE_COUNT` would parse as `NaN`, the loop would never run, and the script would exit 0 — a silent false positive when run standalone outside the workflow.
- `tools/scripts/firewall-rate-limit-check/probe.ts:61-64` — Added zero-request guard after the loop, matching `sentinel.ts:52-55`. Defends against any future regression that would skip the loop body.
- `tools/scripts/firewall-rate-limit-check/probe.ts:108-109` — Dropped `total === 0 ? 0 :` defensive guards on `errorPct` / `fiveXxPct` since the new `results.length === 0` exit makes them provably unreachable.

**Challenged / skipped:** None — the CodeRabbit suggestion was correct and aligned with the existing `sentinel.ts` pattern.

PR #9232, opened 2026-05-19
feat(tools): k6 load-test scaffold for /api/chat [NES-1581]

jaco-brink:
## Summary

Adds `tools/load-test/`, a k6-based load tester for Journeys API endpoints driven by YAML scenario files. Built for NES-1581 to drive the Vercel Firewall rate-limit rule on `/api/chat` from a controllable local generator, with each run producing a stable, share-friendly JSON result file. Adding a new endpoint or load profile is a one-file copy.

## Layout

```
tools/load-test/
  run-chat.ts             # single-arg wrapper: pnpm exec tsx run-chat.ts <yaml>
  scenarios/              # YAML load profiles
    smoke.yaml
    sustained-single.yaml
    concurrent-clients.yaml
    firewall-trip.yaml
  targets/chat.js         # k6 entry — /api/chat payload (ai-sdk v6 UIMessage shape)
  lib/scenario.js         # shared runner: per-HTTP-status counters + latency + handleSummary
  results/                # per-run JSON (committed as evidence; manual retention rule lives in the README)
```

Adds the `yaml` package (`^2.9.0`) for scenario parsing; otherwise uses Node 22 builtins and `tsx` (already in the workspace). Requires `brew install k6`.

## Verification — the firewall rule fires, and we proved it

End-to-end verified against `stage` with the Vercel Firewall rule on `/api/chat` in both **Challenge/Deny** and **Log** modes. The same scenario (`sustained-single`: 1 VU at 1 RPS for 60s from one source IP) was run before and after switching the rule to Log mode:

| # | Scenario             | Rule mode      | 2xx           | 429   | p50    | p95   | p99   | max    |
| - | -------------------- | -------------- | ------------- | ----- | ------ | ----- | ----- | ------ |
| 1 | sustained-single     | Challenge/Deny | 55/57 (96.5%) | **2** | 1.26s  | 3.37s | 7.76s | 13.14s |
| 2 | sustained-single     | Log            | 59/59 (100%)  | 0     | 1.27s  | 1.76s | 2.22s | 2.41s  |
| 3 | concurrent-clients   | Log            | 141/141 (100%)| 0     | 1.34s  | 3.67s | 4.60s | 8.24s  |

### What this proves

1. **The firewall rule actively blocks `/api/chat` traffic from a single source IP at ~1 RPS sustained.** With the rule in Challenge/Deny mode, two of 57 requests came back with 429 in a 60-second window. Empirical per-IP threshold sits around 60 RPM.
2. **Log mode is non-blocking.** Same load, same payload — zero 429s, identical p50. Confirms the rule can be soaked in Log mode without affecting real users.
3. **The long-tail latency in run `#1` was the firewall, not the endpoint.** Compare runs 1 vs 2: p50 is identical (~1.27s) but p99 drops from 7.76s → 2.22s and max from 13.14s → 2.41s when the rule moves to Log. The endpoint itself is consistent; the slow requests in run 1 were the rate-limited ones.
4. **Concurrent load from one IP handles cleanly.** Run `#3` (5 VUs, 5 RPS aggregate) returns 141/141 of 200s with no app-layer failures. p95 widens ~2× under concurrency (1.76s → 3.67s) but p50 holds — expected queuing shape.

Raw JSONs are committed under `tools/load-test/results/` as evidence and follow a stable schema (see README) — `by_status` for exact code distribution, `buckets` for the categorical roll-up, `latency_ms` for p50/p95/p99/max/avg, `config.durationMs` for the actual run length. The README documents a manual retention convention (rename one-off baselines descriptively; prune older timestamped runs before pushing) — guidance for future operators rather than an enforced policy.

## Identity model — note for reviewers

The tool deliberately does NOT simulate Firebase auth. Reading the handler at `apps/journeys/pages/api/chat/index.ts`, `/api/chat` doesn't authenticate today — only the LaunchDarkly flag gate and request bounds. The Vercel Firewall rule keys on source IP, so multi-VU runs from one machine cannot model multi-IP load. The README's **Identity model** section spells this out so future contributors don't add a token-sending flow before it's needed.

When the per-Firebase-uid gate lands on `api-journeys` (NES-1580 / NES-1584 with server-side token verification), that's the moment to add an authenticated target (`targets/chat-auth.js`) — not before.

## How to run

```sh
brew install k6
pnpm exec tsx tools/load-test/run-chat.ts tools/load-test/scenarios/smoke.yaml
pnpm exec tsx tools/load-test/run-chat.ts tools/load-test/scenarios/sustained-single.yaml
pnpm exec tsx tools/load-test/run-chat.ts tools/load-test/scenarios/concurrent-clients.yaml
```

Adding a scenario = copy a `*.yaml` and edit the values. Adding a target = copy `targets/chat.js`, swap the request body, swap `TARGET_PATH` in `run-chat.ts` (or copy that too as `run-<name>.ts`).

## Test plan

- [x] Sanity smoke against stage with rule in Log mode → all 200s
- [x] `sustained-single` against stage with rule in Challenge/Deny → expected 429s observed (2/57)
- [x] `sustained-single` against stage with rule in Log mode → no 429s, same p50, narrow tail
- [x] `concurrent-clients` against stage with rule in Log mode → all 200s, latency tail widens as expected under concurrency
- [x] Lint / required CI checks green on this PR
- [ ] `firewall-trip.yaml` run against stage with rule re-promoted to Challenge / Deny — separate exercise, do once the rule progresses past Log

## Summary by CodeRabbit

* **New Features**
  * Added a load-testing framework: documentation, CLI tooling, reusable test scenarios, and example run result artifacts.

* **Chores**
  * Added the `yaml` package to devDependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9232?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tools/load-test/results/sustained-single-20260519T225718.json
```diff
@@ -0,0 +1,40 @@
...
+    "maxIterations": null,
+    "startedAt": "2026-05-19T22:57:18.676Z",
+    "finishedAt": "2026-05-19T22:58:18.678Z",
+    "durationMs": 60001
+  },
+  "totals": {
+    "sent": 57,
+    "by_status": {
+      "200": 55,
+      "429": 2
```
jaco-brink, line 22: 2 x 429 returned as I configured the vercel rate limit rule output the error code

tools/load-test/results/sustained-single-20260519T230340.json
```diff
@@ -0,0 +1,39 @@
...
+    "duration": "60s",
+    "maxIterations": null,
+    "startedAt": "2026-05-19T23:03:40.096Z",
+    "finishedAt": "2026-05-19T23:04:41.294Z",
+    "durationMs": 61197
+  },
+  "totals": {
+    "sent": 59,
+    "by_status": {
+      "200": 59
```
jaco-brink, line 21: no 429's returned as I configured the vercel rate limit rule to log only

csiyang:
Nice work — the code is clean, well-typed, and the verification evidence in the description is genuinely thorough. Two real blockers around documentation drift, plus a couple of correctness niggles worth a look before merge.

## Should fix before merge

**1. PR description + README contradict the committed diff** (P1)

Three stale claims that a reader can't reconcile without commit archaeology:

- PR body says `results/  # per-run JSON (gitignored)` — `results/` is not in `.gitignore` and 4 JSONs are committed.
- PR body says `No new npm deps — Node 22 builtins + tsx + a built-in flat YAML reader` — `package.json` adds `yaml: ^2.9.0`, `run-chat.ts:16` imports it.
- `README.md:19` and `:86` still say `(gitignored)`. `README.md:27` still says `No YAML library is added — the scenario parser is built in (flat key: value only)`. Then `:127` says "Result files are committed to git as evidence" and `:156` says "YAML is parsed by the `yaml` package". Both versions of the story live in the same file.

Pick the real story and update PR body + the three stale README lines.

**2. `results/` retention is documented but unenforced** (P2)

README:127-146 directs operators to "prune to the last 2 files per scenario manually" with a copy-paste bash loop. Nothing in CI / pre-commit / `.gitignore` enforces this — it will drift. Either:

- gitignore `results/` and rely on the PR-body table as the canonical evidence, or
- gitignore `results/*-[0-9]*.json` and check in only the named-baseline files the README already calls out (`*-challenge-deny-baseline.json` etc.).

## Worth a look

**3. `max_iterations: 0` from YAML throws instead of meaning "no cap"** (P2)

`lib/scenario.js:197` defaults `MAX_ITERATIONS` to `0` (treated as "no cap" at line 226). But if someone explicitly writes `max_iterations: 0` in YAML, `run-chat.ts:162` sets the env to `"0"`, then `intEnv`'s `value <= 0` check throws. No bundled scenario hits this today; next contributor copying a template might. Either treat `0` the same as missing inside `intEnv`, or document that `max_iterations` must be omitted (not set to 0) to disable.

**4. Two independent runId fallbacks will diverge** (P2)

`targets/chat.js:21` (`__ENV.RUN_ID || ` `` `local-${Date.now()}` `` ``) and `lib/scenario.js:199` (`stringEnv('RUN_ID', ` `` `${name}-${isoNowSafe()}` `` `)`) each generate their own fallback. With `run-chat.ts` they agree (env is set). The README's "Or invoke k6 directly" path (line 163-165) skips that — User-Agent and result filename will then report different IDs computed milliseconds apart. Pick one source: either make `targets/chat.js` require `RUN_ID`, or expose the resolved runId from `lib/scenario.js` for the target to consume.

## Nits (your call)

- `run-chat.ts:101` returns `fields as ScenarioConfig`, discarding the URL-required guarantee enforced at `:93-94`. A `Required<Pick<…,'url'>> & Partial<Omit<…,'url'>>` shape would carry the invariant, but reasonable as-is for a CLI script.
- No unit tests for `loadScenario`, `intEnv`, `recordStatus`, `buildStatusBreakdown`. They're pure and trivial to test. Lower bar for tooling, but the XOR check in `loadScenario` is exactly the kind of thing that breaks silently on a refactor.
- Test-plan checkbox `Lint / required CI checks green on this PR` still unchecked.

jaco-brink:
## Review feedback addressed (700a2167b)

Thanks @csiyang — landed three of the four substantive items, with a thought-out push-back on the retention enforcement.

**Fixed**

- **Docs drift (P1)** — `README.md:19`/`:27`/`:86` and the PR description are now internally consistent with the committed state: `results/` is described as **committed evidence** (matching the four JSON files already in the diff), and the "no YAML library" claim is replaced with an honest mention of the `yaml` package (`^2.9.0`) and the `SCENARIO_KEYS` allowlist that gates it. PR body's layout block, npm-deps blurb, and verification footer all updated to match.
- **`max_iterations: 0` no longer throws (P2)** — added `nonNegativeIntEnv` in `lib/scenario.js` and use it only for `MAX_ITERATIONS`. `intEnv` stays strict (rate and concurrency knobs should not accept `0`). Explicit `0` in YAML now means "no cap," matching the line-226 semantics and the README phrasing.
- **`runId` unified across User-Agent and result filename (P2)** — `buildScenario` resolves `runId` once and passes it into `buildRequest({ runId })`. `targets/chat.js` drops the duplicate `local-${Date.now()}` fallback so a bare `k6 run` invocation can no longer produce a User-Agent and a result filename that disagree by milliseconds. Docstring on `buildScenario` updated to reflect the new contract.

**CI checkbox** — checked. All required checks were green before this push too; the prior version of the test plan was just stale.

**Pushing back on `results/` retention enforcement (#2)**

The four committed JSONs aren't accidental — committing them is the explicit intent (the PR's value as evidence depends on those files being inspectable from the diff). The README rule is **guidance for future operators**, not a policy that should be enforced by tooling. This tool is expected to be run infrequently (firewall-rule promotion checkpoints), so the directory growth is bounded by hand without much risk of drift. Adding a gitignore + ignore-exceptions for baseline names would push complexity for a problem the team hasn't actually encountered yet. Happy to revisit if the directory ever gets noisy.

**Nits left as-is (with reasoning)**

- `run-chat.ts:101 fields as ScenarioConfig` — you already noted "reasonable as-is for a CLI script"; agreed, leaving it.
- No unit tests for `loadScenario`/`intEnv`/`recordStatus`/`buildStatusBreakdown` — fair point in principle, especially for the XOR check, but adding a Jest project under `tools/` purely for this tool feels heavier than what the current scope warrants. Will reach for it the next time one of these helpers gets a real refactor.

**Self-annotation threads** — the two inline notes I left on the result JSONs (about expected 429 counts under Challenge/Deny vs Log) were observational, resolved.

jaco-brink:
## Review feedback addressed (9741e6cd2) — CodeRabbit pass

Three findings from CodeRabbit, all valid edge cases. Fixed in one commit.

**Fixed**

- **`lib/scenario.js` — harden `intEnv` / `nonNegativeIntEnv` (Major).** `Number.parseInt('10ms', 10)` returns `10`, so a malformed YAML value like `rps: "10ms"` would silently produce a 10-RPS run instead of erroring. Added a `/^\d+$/` digits-only check before parsing in both helpers — non-digit input now throws `'${key} must be a positive/non-negative integer (got: ${raw})'`.
- **`run-chat.ts` — normalize blank `run_id` (Minor).** `??` only fell back on null/undefined, so `run_id: ""` (or whitespace) would slip through and produce a broken result filename like `results/.json`. Now we trim and treat empty as missing, falling back to `${scenarioName}-${isoTimestamp}` like an unset value would.
- **`README.md` — label fenced code blocks (Minor).** Two unlabeled fences (line 9 layout tree, line 132 rename example) now use ```text to satisfy markdownlint `MD040`.

All three threads resolved.

csiyang (approved):

PR #9236, opened 2026-05-20
fix(journeys): tag langfuse traces with deployment environment [NES-1688]

jaco-brink:
## Summary

* Pass `environment` to the Langfuse SDK constructor so traces stop bucketing as `default`. Today **every** trace from `/api/chat` (prod, stage, preview, local) shows `environment: default` in the Langfuse UI, which breaks per-env filtering needed for World-Cup launch observability.
* Derive the value from `VERCEL_ENV` (and `VERCEL_GIT_COMMIT_REF` for the `stage` branch), with `LANGFUSE_TRACING_ENVIRONMENT` as a manual override.

## Mapping

| `VERCEL_ENV` | `VERCEL_GIT_COMMIT_REF` | Langfuse env  |
|--------------|-------------------------|---------------|
| `production` | —                       | `production`  |
| `preview`    | `stage`                 | `stage`       |
| `preview`    | other                   | `preview`     |
| unset        | —                       | `development` |

## Why this is a one-line fix in two places

Langfuse's SDK already supports `environment` on the constructor (`langfuse-core@3.38.20`, `lib/index.cjs.js:890`); we just weren't passing it. The fallback `getEnv("LANGFUSE_TRACING_ENVIRONMENT")` isn't set in Doppler either. Either knob alone fixes the bug; this PR wires both so we can override per environment without a redeploy if needed.

## Test plan

- [x] `npx vitest run --config apps/journeys/vitest.config.mts 'apps/journeys/src/libs/langfuse/client.spec.ts' --coverage=false` — 17 pass (was 11; added 6).
- [ ] After merge to preview: open Langfuse → confirm traces from the PR preview deployment appear under `preview` (not `default`).
- [ ] After merge to `stage` branch: traces appear under `stage`.
- [ ] After prod release: traces appear under `production`.

## Out of scope

* Per-trace `environment` overrides at the call site in `pages/api/chat/index.ts`. Constructor-level is sufficient.
* Folding `getActivePromptLabel()` into the new env helper — they answer different questions (prompt label vs Langfuse env). Tracked as a "could-do" if it ever bites.

## Linear

[NES-1688](https://linear.app/jesus-film-project/issue/NES-1688/fix-langfuse-pass-environment-to-constructor-so-prodstagedev-traces)

### DIFF
--- a/apps/journeys/src/libs/langfuse/client.spec.ts
+++ b/apps/journeys/src/libs/langfuse/client.spec.ts
@@ -19,6 +19,8 @@ describe('langfuse client', () => {
     delete process.env.LANGFUSE_SECRET_KEY
     delete process.env.LANGFUSE_BASE_URL
     delete process.env.VERCEL_ENV
+    delete process.env.VERCEL_GIT_COMMIT_REF
+    delete process.env.LANGFUSE_TRACING_ENVIRONMENT
   })
 
   afterEach(() => {
@@ -108,7 +110,8 @@ describe('langfuse client', () => {
       expect(mockLangfuseConstructor).toHaveBeenCalledWith({
         publicKey: 'pk-test',
         secretKey: 'sk-test',
-        baseUrl: 'https://lf.test'
+        baseUrl: 'https://lf.test',
+        environment: 'development'
       })
     })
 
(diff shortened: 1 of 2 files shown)


edmonday (commented):
## Review summary

**Critical:** none
**Concerns:** none
**Nits:** none

Tight, well-scoped fix. `resolveTracingEnvironment()` reads top-to-bottom in priority order with early returns, the empty-string override case is handled, and the test suite covers the full truth table (production / preview-on-stage / preview-other / unset / override / empty-override). Inline comment explains the *why* (Langfuse bucketing as `default`) rather than the what. Out-of-scope notes about `getActivePromptLabel()` and per-trace overrides are reasonable.

Nothing to flag.

edmonday (approved):

PR #9247, opened 2026-05-24
refactor(journeys-admin): unify collection create/edit/publish flow

csiyang:
## Summary

- **Create Collection** no longer opens a dialog — it creates a collection instantly with an auto-incremented name (`Collection 1`, `Collection 2`, …).
- **One state-aware card menu item** in place of separate Edit / Publish / Unpublish entries: drafts show **Publish**, published show **Edit**. Both open the same `CollectionDialog`.
- **Publish and Unpublish moved inside the dialog footer.** Drafts get `Cancel | Save Draft | Publish`; published get `Cancel | Unpublish | Save`. Empty / custom-domain gating moved from the card menu to the in-dialog Publish button.
- **Slug field** now visible in edit *and* publish modes (was edit-only), styled with a section header like the other fields, and reordered to sit directly under Page Description.
- **Collection card restyle** — sits on page grey with a 1px `grey.400` border, more rounded corners, and a soft drop shadow. All Templates grid drops its inset padding so cards line up with the collection-card edges above.
- **QA-459**: in the editor's mobile dropdown menu, the **TEMPLATE chip stacks below the journey title** instead of getting squeezed beside it. Desktop unchanged.

## Test plan

- [ ] Click **Create Collection** with no collections present — verify it appears named `Collection 1` with no modal
- [ ] Click **Create Collection** again with `Collection 1` and `Collection 3` present — verify the new one is named `Collection 2` (smallest unused number)
- [ ] On a draft collection card, open the menu — verify only **Publish** (no separate Edit) is shown; clicking it opens the dialog with `Cancel | Save Draft | Publish`
- [ ] On a published collection card, open the menu — verify only **Edit** (no separate Publish/Unpublish) is shown; clicking it opens the dialog with `Cancel | Unpublish | Save`
- [ ] In Publish dialog with no templates selected, verify Publish button is disabled with the "Add at least one template" tooltip; Save Draft still works
- [ ] In Publish dialog, hit Save Draft → collection remains draft with updated fields; hit Publish → success dialog opens with the live URL
- [ ] In Edit dialog on a published collection, hit Unpublish → collection flips back to draft, dialog closes, success snackbar appears
- [ ] On a custom-domain team, verify the Publish menu item is disabled with the gate-reason tooltip
- [ ] Visually verify collection cards have the new grey background + grey border + rounded corners + drop shadow, and that template cards in **All Templates** line up with the collection card edges
- [ ] Verify Slug field appears in both Edit and Publish dialogs, sits under Page Description, and renders with the "Slug" section header
- [ ] On mobile (~375px), open a template journey in the editor → open the 3-dots dropdown → verify the **TEMPLATE** chip sits below the journey title (stacked, not beside)

## Summary by CodeRabbit

* **New Features**
  * Create‑and‑publish flow with a dedicated publish dialog and publish success handling.
  * Publish/unpublish actions available from dialog footer; slug editing normalized outside create mode.

* **UI/UX Improvements**
  * Collection card styling refreshed (background, border, shadow).
  * Journey title stacks vertically on mobile for clearer chip display.
  * Card menu entries reorganized for draft vs. published states.

* **Behavior & Tests**
  * "Publish" can be invoked for empty drafts (dialog enforces gating); tests updated for menu, dialog, and footer behaviors.

* **Documentation**
  * Added spacing and Formik multi‑action pattern guidance and a UI-bug note on mobile stacking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9247?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialog.tsx
+++ b/apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialog.tsx
@@ -5,6 +5,7 @@ import ButtonBase from '@mui/material/ButtonBase'
 import Collapse from '@mui/material/Collapse'
 import Stack from '@mui/material/Stack'
 import TextField from '@mui/material/TextField'
+import Tooltip from '@mui/material/Tooltip'
 import Typography from '@mui/material/Typography'
 import { Form, Formik } from 'formik'
 import { useTranslation } from 'next-i18next/pages'
@@ -25,7 +26,7 @@ import { useCollectionForm } from './useCollectionForm'
 
 export interface CollectionDialogProps {
   open: boolean
-  mode: 'create' | 'edit'
+  mode: 'create' | 'edit' | 'publish'
   teamId: string
   collection?: TemplateGalleryPage
   availableJourneys: readonly Journey[]
@@ -44,6 +45,8 @@ export interface CollectionDialogProps {
   /** Tooltip copy for the disabled preview button. */
   publishBlockedReason?: string | null
   onClose: () => void
+  /** Forwarded to useCollectionForm in publish mode. */
+  onPublished?: (collection: TemplateGalleryPage) => void
 }
 
 // Matches the Figma "Editor / Subtitle/2" type token on section headers.
@@ -64,12 +67,27 @@ export function CollectionDialog({
   parentBusy = false,
   canPublish = true,
   publishBlockedReason = null,
-  onClose
+  onClose,
+  onPublished
 }: CollectionDialogProps): ReactElement {
   const { t } = useTranslation('apps-journeys-admin')
 
-  const { initialValues, schema, isPublished, handleSubmit } =
-    useCollectionForm({ mode, teamId, collection, parentBusy, onClose })
+  const {
+    initialValues,
+    schema,
+    isPublished,
+    handleSubmit,
+    setSubmitIntent,
+    handleUnpublishAction,
+    isMutating
+  } = useCollectionForm({
+    mode,
+    teamId,
+    collection,
+    parentBusy,
+    onClose,
+    onPublished
+  })
 
   // O(1) lookup so the preview pane can resolve the user-pick-ordered
   // journey list per Formik render without an O(M) scan.
@@ -130,23 +148,124 @@ export function CollectionDialog({
               open={open}
               onClose={guardedClose}
               maxWidth="md"
-              // Disable the submit + close buttons while the mutation is in
-              // flight. Formik's submitForm has no internal double-click guard
-              // — without this, a fast second click before React re-renders
-              // would fire a second create/update mutation.
-              loading={isSubmitting}
+              // Disable the submit + close buttons while a mutation is in
+              // flight. Covers Formik's submitForm (isSubmitting) and
+              // the out-of-band unpublish action (isMutating). Without
+              // this, a fast second click before React re-renders would
+              // fire a duplicate mutation.
+              loading={isSubmitting || isMutating}
               dialogTitle={{
                 title:
                   mode === 'create'
                     ? t('New Collection')
-                    : t('Edit Collection'),
+                    : mode === 'publish'
+                      ? t('Publish Collection')
+                      : t('Edit Collection'),
                 closeButton: true
               }}
-              dialogAction={{
-                onSubmit: handleSubmit,
-                closeLabel: t('Cancel'),
-                submitLabel: mode === 'create' ? t('Create') : t('Save')
-              }}
+              dialogAction={
+                // Custom footer in publish mode (Save Draft + Publish)
+                // and in edit-while-published mode (Unpublish + Save).
+                // Create + draft-edit fall through to the default
+                // Cancel + Save footer.
+                mode === 'publish' || (mode === 'edit' && isPublished)
+                  ? undefined
+                  : {
+                      onSubmit: handleSubmit,
+                      closeLabel: t('Cancel'),
+                      submitLabel:
+                        mode === 'create' ? t('Create') : t('Save')
+                    }
+              }
+              dialogActionChildren={(() => {
+                const busy = isSubmitting || isMutating
+                if (mode === 'publish') {
+                  // Publish is gated by (a) custom-domain teams that
+                  // can't host gallery pages and (b) an empty selection
+                  // — same checks that previously lived on the card
+                  // menu, moved here so the menu item itself can
+                  // always open the dialog (the user may want to fill
+                  // in metadata before adding templates).
+                  const publishBlockedHere =
+                    !canPublish || values.journeyIds.length === 0
+                  const publishTooltip = !canPublish
+                    ? (publishBlockedReason ?? '')
+                    : values.journeyIds.length === 0
+                      ? t('Add at least one template before publishing')
+                      : ''
+                  return (
+                    <>
+                      <Button onClick={guardedClose} disabled={busy}>
+                        {t('Cancel')}
+                      </Button>
+                      <Button
+                        onClick={() => {
+                          setSubmitIntent('draft')
+                          handleSubmit()
+                        }}
+                        disabled={busy}
+                      >
+                        {t('Save Draft')}
+                      </Button>
+                      <Tooltip
+                        title={publishTooltip}
+                        placement="top"
+                        disableHoverListener={!publishBlockedHere}
+                        disableFocusListener={!publishBlockedHere}
+                      >
+                        <span>
+                          <Button
+                            variant="contained"
+                            color="primary"
+                            onClick={() => {
+                              setSubmitIntent('publish')
+                              handleSubmit()
+                            }}
+                            loading={isSubmitting}
+                            disabled={publishBlockedHere || busy}
+                          >
+                            {t('Publish')}
+                          </Button>
+                        </span>
+                      </Tooltip>
+                    </>
+                  )
+                }
+                if (mode === 'edit' && isPublished) {
+                  // Unpublish lives inside the dialog so the card menu
+                  // has a single state-aware entry point. Skips
+                  // Formik on purpose — any pending field edits are
+                  // discarded, because "unpublish" is a deliberate
+                  // status change, not a save path.
+                  return (
+                    <>
+                      <Button onClick={guardedClose} disabled={busy}>
+                        {t('Cancel')}
+                      </Button>
+                      <Button
+                        color="error"
+                        onClick={() => {
+                          void handleUnpublishAction()
+                        }}
+                        loading={isMutating}
+                        disabled={busy}
+                      >
+                        {t('Unpublish')}
+                      </Button>
+                      <Button
+                        variant="contained"
+                        color="primary"
+                        onClick={() => handleSubmit()}
+                        loading={isSubmitting}
+                        disabled={busy}
+                      >
+                        {t('Save')}
+                      </Button>
+                    </>
+                  )
+                }
+                return undefined
+              })()}
               testId="CollectionDialog"
               sx={{
                 '& .MuiDialogContent-root': {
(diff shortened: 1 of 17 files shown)


apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialog.tsx
```diff
@@ -130,23 +148,123 @@ export function CollectionDialog({
...
+                        placement="top"
+                        disableHoverListener={!publishBlockedHere}
+                        disableFocusListener={!publishBlockedHere}
+                      >
+                        <span>
+                          <Button
+                            variant="contained"
+                            color="primary"
+                            onClick={() => {
+                              setSubmitIntent('publish')
```
edmonday: **Concern:** `dialogActionChildren={(() => { ... })()}` is a ~90-line multi-branch IIFE embedded in a JSX prop expression, stacking two unusual idioms (self-invoked arrow + JSX-as-prop-value). Three issues:

1. Readability — by the time you reach `testId="CollectionDialog"` you’ve lost track of which prop you’re configuring.
2. The closure is recreated and re-called every render — harmless at this size, but it noises up profiling traces.
3. The `return undefined` fall-through silently relies on `<Dialog>` treating `undefined` as “render nothing”, which the next reader has to verify against the shared Dialog source.

Suggest extracting to a local function returning `ReactNode` (or splitting the two branches into small `PublishFooter` / `EditPublishedFooter` components) and passing the result here:

```tsx
function renderFooter(): ReactNode {
  const busy = isSubmitting || isMutating
  if (mode === "publish") return (<>…</>)
  if (mode === "edit" && isPublished) return (<>…</>)
  return null
}
// …
dialogActionChildren={renderFooter()}
```

Logic itself reads correctly — this is purely a structural cleanup.

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
```diff
@@ -337,23 +333,39 @@ export function TemplateGalleryPageList({
...
-    const own = editTarget.templates
+    const own = target.templates
       .filter((tpl) => !seen.has(tpl.id))
       .map((tpl) => journeyById.get(tpl.id))
       .filter((j): j is Journey => j != null)
     return [...unsectioned, ...own]
-  }, [unsectioned, editTarget, journeyById])
+  }
+  const editAvailableJourneys = useMemo<readonly Journey[]>(
+    () => buildAvailableJourneys(editTarget),
```
edmonday, line 369: **Concern:** Both `useMemo`s disable `react-hooks/exhaustive-deps` and manually list the values that `buildAvailableJourneys` closes over. That manual list is now a contract the linter can’t enforce — if someone later adds a closure-captured value to `buildAvailableJourneys` (e.g. a new filter prop), the dep array stays stale and the memoized list silently desyncs. That’s the exact failure mode `exhaustive-deps` exists to prevent.

Three options, in increasing order of refactor:

1. **Inline the body in each memo** — the linter sees the real deps and the disable disappears. Slight duplication, but shallow and obvious.
2. **`useCallback` the helper** and add it to each memo’s deps.
3. **Hoist to module scope**, take `unsectioned` and `journeyById` as explicit args.

I’d take (1) — least machinery for what it buys.

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/useCollectionForm/useCollectionForm.ts
```diff
@@ -242,5 +339,61 @@ export function useCollectionForm({
...
+    }
+  }
+
+  return {
+    initialValues,
+    schema,
+    isPublished,
+    handleSubmit,
+    setSubmitIntent,
+    handleUnpublishAction,
```
edmonday, line 397: **Concern:** `isMutating` is returned as if it’s the total busy flag, but it only tracks `handleUnpublishAction` — the Formik submit path is still tracked separately via Formik’s `isSubmitting`. The dialog OR’s them together at every call site (`disabled={isSubmitting || isMutating}`), which works today but is an unenforced invariant — the next consumer of this hook will read `isMutating` and assume “anything mutating”, miss the Formik path, and silently get a button enabled during a Save.

Name it for what it tracks:

```ts
return { …, isUnpublishing: unpublishing }
```

Then the dialog reads `disabled={isSubmitting || isUnpublishing}` and the asymmetry is in the name, not hidden behind a generic one.

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
```diff
@@ -362,18 +374,68 @@ export function TemplateGalleryPageList({
...
+      if (mountedRef.current) {
+        enqueueSnackbar(t('Collection created'), {
+          variant: 'success',
+          preventDuplicate: true
+        })
+      }
+    } catch (error) {
+      if (mountedRef.current) {
+        enqueueSnackbar(
+          error instanceof Error
```
edmonday, line 429: **Concern:** Double-click race on the new instant-create flow.

`handleCreate` guards on `createLoading`, but `createLoading` only flips after Apollo dispatches the mutation. Two synchronous clicks before the first re-render both pass the guard:

- Click 1 → `handleCreate` runs to the `await`, mutation invoked, Promise pending. Apollo hasn’t flipped `loading` or re-rendered yet.
- Click 2 (within the same tick) → `createLoading` still `false`, `disabled` in the DOM still `false`, guard passes, second mutation fires.
- Net: a single intended click creates two collections (`Collection 1` + `Collection 2` — auto-name doesn’t protect because the second invocation reads the Apollo cache before the first lands).

Before this PR the button opened a dialog, so this race didn’t exist — it’s introduced by the instant-create.

Fix is the same synchronous-ref-latch pattern `submittingRef` in `useCollectionForm` already uses:

```ts
const creatingRef = useRef(false)

async function handleCreate(): Promise<void> {
  if (creatingRef.current || teamId == null) return
  creatingRef.current = true
  try {
    await templateGalleryPageCreate({ … })
    // …
  } finally {
    creatingRef.current = false
  }
}
```

The ref flips before any `await`, so the second click sees `true` and returns.

edmonday (commented):
## Review summary

4 concerns, none blocking — all maintainability / robustness items. CI green, behaviour looks right.

**Concerns:**
- IIFE-in-JSX `dialogActionChildren` is hard to read; extract to a `renderFooter()` local.
- `eslint-disable-next-line react-hooks/exhaustive-deps` on the two `buildAvailableJourneys` memos is a footgun — the linter can't catch dep drift if the helper changes.
- `isMutating` only tracks unpublish, not Formik submits — rename to `isUnpublishing` so the asymmetry is named, not hidden.
- Double-click race on the new instant Create Collection: `createLoading` flips async, so two synchronous clicks both pass the guard and create two collections. Add a `useRef` latch matching the existing `submittingRef` pattern.

**Positive:** strong inline rationale throughout, null-result trap on publish/unpublish, snackbar-suppression when publish follows update, three substantial `docs/solutions/` learnings captured.

_Triaged via `/review-pr` (dry run → posted on request)._

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
csiyang: Done in 2fbd95e. Added `creatingRef` matching the existing `submittingRef` / `dragInFlightRef` pattern — flips synchronously inside the click handler before any await, so the second click sees `true` and returns. Slightly embarrassing miss given I used the same pattern in `useCollectionForm` — should have followed my own template the first time.

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialog.tsx
csiyang: Done in f2e907af0. Extracted to a new `CollectionDialogFooter` component in its own folder (with tests + index). The dialog now reads top-to-bottom without losing track of which prop is being configured, and the four footer configurations (create / edit-draft / edit-published / publish) all live in one place inside the component. The `return undefined` fall-through is gone too — `dialogActionChildren` is now always a real element, the component handles the branching internally.

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
csiyang: Skipping for now. The only benefit of inlining is silencing the lint disable — the helper is used in exactly 2 call sites and doesn't close over much. If `buildAvailableJourneys` grows new closure deps later, we'll inline (your option 1) — agreed that's the right move at that point, just not paying the duplication cost preemptively. Happy to revisit if you'd rather have it now.

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/useCollectionForm/useCollectionForm.ts
csiyang: Done in f2e907af0. Renamed in `useCollectionForm`'s return + every consumer + the type comment. Bundled with CodeRabbit's catch on `guardedClose` (it was also only checking `isSubmitting`, so mid-unpublish the user could dismiss and orphan the post-await `onClose`/snackbar) — both fixed together since they're the same `isUnpublishing` asymmetry. The new `CollectionDialogFooter` takes `isSubmitting` and `isUnpublishing` as separate props for the same reason.

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialogFooter/CollectionDialogFooter.tsx
```diff
@@ -0,0 +1,191 @@
...
+    // gallery pages and (b) an empty selection — moved here from the
+    // card menu so the menu item can always open the dialog (the user
+    // may want to fill in metadata before adding templates). Save
+    // Draft stays enabled either way.
+    const publishBlocked = !canPublish || journeyCount === 0
+    const publishTooltip = !canPublish
+      ? (publishBlockedReason ?? '')
+      : journeyCount === 0
+        ? t('Add at least one template before publishing')
+        : ''
```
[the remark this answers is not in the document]
csiyang: Addressed in 214ef0ceb — but going with option A (fix the docstring to match the existing code) rather than option B (change the fallback string).

The reasoning: `useCanPublishCollection` always returns a non-null reason when `canPublish` is false today, so the null-with-canPublish-false branch isn't actually reachable in production. The docstring was a fabrication on my part (`Null falls back to the empty-selection copy` was wishful thinking, not actual behaviour). The existing pattern in `CollectionCard` also does `publishBlockedReason ?? ''` for the same reason, so I'd rather match the established pattern than introduce a fallback string that's semantically wrong (the empty-selection copy refers to a different gate).

Also addressed the `<span>` → `<Box component="span">` nitpick from the same review in the same commit — matches the pattern already used in `CollectionPublishSuccessDialog`.

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
```diff
@@ -337,23 +342,39 @@ export function TemplateGalleryPageList({
...
+  const editAvailableJourneys = useMemo<readonly Journey[]>(
+    () => buildAvailableJourneys(editTarget),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [unsectioned, editTarget, journeyById]
+  )
+  const publishAvailableJourneys = useMemo<readonly Journey[]>(
+    () => buildAvailableJourneys(publishTarget),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [unsectioned, publishTarget, journeyById]
+  )
```
edmonday, line 377: **Still open from previous review (Concern 2).**

Both memos still disable `react-hooks/exhaustive-deps` and manually mirror the deps of `buildAvailableJourneys`. The disable is a footgun — if a future edit adds a new closure-captured value to `buildAvailableJourneys`, the dep arrays stay stale and the memoized list silently desyncs. That's the exact class of bug the lint rule exists to catch.

Cleanest fix is to inline the helper body into each memo. The duplication is shallow and the deps then become correct by construction:

```ts
const editAvailableJourneys = useMemo<readonly Journey[]>(() => {
  if (editTarget == null) return unsectioned
  const seen = new Set(unsectioned.map((j) => j.id))
  const own = editTarget.templates
    .filter((tpl) => !seen.has(tpl.id))
    .map((tpl) => journeyById.get(tpl.id))
    .filter((j): j is Journey => j != null)
  return [...unsectioned, ...own]
}, [unsectioned, editTarget, journeyById])
```

Alternatives: `useCallback` the helper and list it as a dep, or hoist to module scope taking `unsectioned`/`journeyById` as args. Non-blocking — flagging again so it doesn't slip through.

edmonday (commented):
## Re-review on `214ef0ce`

Thanks for the turnaround. Three of the four concerns are resolved:

- ✅ Footer extracted to `CollectionDialogFooter` (with spec) — much easier to read, and the four-branch layout is now self-documenting.
- ✅ `isMutating` → `isUnpublishing` with a docstring that makes the OR-with-`isSubmitting` contract explicit.
- ✅ `creatingRef` latch closes the double-click race; flipped before any `await`, matches the existing `submittingRef` pattern.

One concern remains — the `eslint-disable react-hooks/exhaustive-deps` on the two `buildAvailableJourneys` memos. Inline comment below. Non-blocking on my end.

apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
edmonday: Agreed — fair call. The duplication-vs-disable trade tips the other way once the helper actually grows; flagging again then makes more sense than refactoring preemptively. Resolving both threads.

edmonday (approved):

PR #9250, opened 2026-05-25
feat(journeys): restyle public gallery page + share it with admin preview (NES-1694)

csiyang:
## What

Two things for the public gallery (template-collection) page:

1. **Restyle** to the **Sections** layout — a sectioned, animated single-page design (intro / 2 featured / the rest) with the journeys-admin red accent, full-height `svh` sections, scroll-reveal animations, and a gap-closing parallax. Mobile-tuned (smaller title, 2-up portrait cards, lighter parallax).
2. **Extract** the page into a shared `@core/journeys/ui/PublicGalleryPage` component consumed by **both** the journeys app and the journeys-admin collection dialog, so the live page and its admin preview render from one source of truth and stop drifting.

Ref: [NES-1694](https://linear.app/jesus-film-project/issue/NES-1694/public-page-restyle)

## Shared component

`PublicGalleryPage` takes a neutral `PublicGalleryPageData` view-model and renders one of two variants behind a single entry point:

- **`journey`** — the real full-screen sectioned page (`apps/journeys`).
- **`admin`** — the compact, decorative recreation shown in the collection dialog (`apps/journeys-admin`), replacing the previously hand-built mock that had started to drift from the real page.

Each app maps its own generated GraphQL types (`GetTemplateGalleryPage` / `GetAdminJourneys`) into the shared model.

## Background

This started as a prototype that explored several restyle directions behind an on-page "Style preview" toggle (Showcase, Spotlight, Landing, Bento, Feature rows, Sections). We picked **Sections**; the prototype toggle and the other presets have been removed.

## Changes

- New `libs/journeys/ui/.../PublicGalleryPage` (journey + admin variants, shared data model, stories, specs).
- `apps/journeys` `TemplateGalleryView` is now a thin mapper onto the shared component; the prototype style system and the migrated layout files were removed.
- `apps/journeys-admin` `CollectionPreviewPane` keeps its URL/open-in-new-tab chrome and renders the shared `admin` variant.
- i18n keys moved to the `libs-journeys-ui` namespace.

## Notes

- No backend/schema changes.
- `prefers-reduced-motion` disables the journey-view animations; the admin variant has none.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated gallery UI into a shared PublicGalleryPage; collection previews now render the admin variant for consistent admin/public display.

* **New Features**
  * Public Journey view with parallax, scroll-reveal, featured/More sections, responsive card layouts and “Use”/“Preview” actions.
  * Compact Admin preview view plus Storybook stories and mock data.

* **Documentation**
  * Added cross-app shared view and scrollspy navigation pattern docs.

* **Localization**
  * Added English strings for collection UI and empty-state messaging.

* **Tests**
  * Added shared PublicGalleryPage tests; removed several legacy template-gallery tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mikeallisonJS (changes requested):
## Summary

Solid refactor — `PublicGalleryPageData` is the right neutral view-model, both callsites shrink to thin `toData()` mappers, and security hygiene is intact (`rel="noopener noreferrer"`, `encodeURIComponent`, no `dangerouslySetInnerHTML`/`eval`/dead imports). `prefers-reduced-motion` is honored consistently across `useParallax`, `ScrollReveal`, and `scrollToSection`. No critical issues.

Requesting changes for two items I'd like landed before merge; the rest are non-blocking and listed as follow-ups.

## Please address before merge

1. **Trailing-slash brittleness in admin URL construction** — `libs/journeys/ui/src/components/PublicGalleryPage/JourneyView/JourneyViewCardActions.tsx`:
   ```ts
   const adminUrl = process.env.NEXT_PUBLIC_JOURNEYS_ADMIN_URL ?? 'https://admin.nextstep.is'
   const useTemplateHref = `${adminUrl}/?useTemplate=${encodeURIComponent(itemId)}`
   ```
   If the env var ends with a `/`, this produces `https://example.com//?useTemplate=...`. Browsers tolerate it, but it's ugly and brittle. Suggest `new URL('/', adminUrl)` + `searchParams.set('useTemplate', itemId)`, or strip with `adminUrl.replace(/\/$/, '')`.

2. **Duplicated featured-count carve-out** — `AdminView.tsx` (~L43) and `JourneyView.tsx` (~L191) both contain:
   ```ts
   const featuredCount = data.items.length === 3 ? 3 : FEATURED_COUNT
   ```
   Extract a shared `splitFeatured(items)` helper in `PublicGalleryPage.tsx` so the two views can't drift if the rule changes.

## Non-blocking suggestions (follow-up OK)

3. **`next/image` allowlist audit in journeys-admin.** The admin variant now renders gallery item thumbnails via `next/image`. The journeys app's `next.config.js` covers Cloudflare Images, Cloudflare Stream, Mux, arclight CloudFront, etc. — please confirm `apps/journeys-admin/next.config.js` `images.remotePatterns` covers every host `primaryImageBlock.src` can return, or the preview will 500 on uncovered hosts.

4. **i18n namespace migration cleanup.** Keys moved from `apps-journeys` / `apps-journeys-admin` → `libs-journeys-ui` (`Untitled collection`, `Creator name`, `A short description of your collection will appear here.`, `This gallery has no templates yet.`). Confirm the old keys in the app namespaces are removed in a follow-up so translators don't keep updating dead strings.

5. **MUI `slotProps` on `Drawer`.** `JourneyViewNav.tsx` uses `slotProps={{ paper, backdrop }}` — this is the v6+ API on `Drawer`. If the repo is still on v5 you may need `componentsProps`. Quick `package.json` check.

6. **`aria-controls` should be unconditional.** `JourneyViewNav.tsx` mobile hamburger:
   ```tsx
   aria-controls={drawerOpen ? 'gallery-nav-drawer' : undefined}
   ```
   The target id should always be set since the `Drawer` element exists regardless of open state. Just `aria-controls="gallery-nav-drawer"`.

7. **Seed `activeId` for first paint.** `useState<string | null>(null)` means no link is highlighted until the `IntersectionObserver` fires. Consider `useState(sections[0]?.id ?? null)`.

8. **Redundant `String(createdAt)` casts.** `createdAt` is already typed `string | null` in both mappers (`CollectionPreviewPane.tsx`, `TemplateGalleryView.tsx`) and in `metaLine`'s `parseISO(String(item.createdAt))`. Drop it for clarity.

9. **`bandBackground` inline structural type.** `JourneyView.tsx`:
   ```ts
   const bandBackground = (theme: { palette: { mode: string } }): string => ...
   ```
   Use `(theme: Theme)` from `@mui/material/styles` to match the rest of the file and stay grep-friendly.

10. **`ScrollReveal` empty deps array.** Functionally correct (the captured props aren't used inside the effect), but `react-hooks/exhaustive-deps` will likely complain. Either disable inline with a comment explaining why, or restructure so the lint stays clean.

11. **`Use` action accessibility.** On a page with many cards, several buttons all labelled "Use" are hard to disambiguate via voice control / screen reader. Consider `aria-label={t('Use {{title}}', { title: item.title })}` on the `JourneyViewCardActions` Use button.

12. **Barrel re-exports.** `GALLERY_ACCENT` and `GALLERY_CARD_RADIUS` are exported from `PublicGalleryPage.tsx` but not re-exported from `index.ts`. If they're truly internal, drop the `export`; if external consumers might want them, add them to the barrel.

## What looks good

- Abstraction boundary is the right one — neutral view-model, two view variants behind a single entry point.
- Security: `target="_blank" rel="noopener noreferrer"` on every external `<a>`, `'noopener,noreferrer'` on `window.open`, `encodeURIComponent` on both `itemId` and `slug`.
- `aria-hidden="true"` on the duplicate `AdminView` preview frame is correctly placed and well-commented.
- Scroll handlers are `passive: true` and RAF-throttled; observers and listeners are torn down in effect cleanup. No leak surface.
- `testid` continuity preserved (`TemplateGalleryHeader`, `TemplateGalleryMedia`, `TemplateGalleryEmptyState`, `GalleryTemplateCard`) — keeps any e2e/analytics selectors working.
- `CollectionPreviewPane.tsx` is ~185 lines smaller post-refactor and both surviving icon imports (`CopyRightIcon`, `Play3Icon`) are still in use in the URL row above.
- No `dangerouslySetInnerHTML`, `eval`, `console.log`, hardcoded secrets, or `any` casts introduced.
- Reasonable spec coverage for variant branching (3-vs-4 carve-out, empty state, conditional nav sections, decorative actions). Skipping scroll/motion tests in jsdom is the right call.

## Notes

- Danger flagged "some issues" on the latest run — worth checking its comment for any test/changelog requirements.
- CodeRabbit returned a clean status.
- No prior review comments on the PR.

— Review generated by Claude.

csiyang:
Thanks for the thorough review @mikeallisonJS — both blockers and most of the follow-ups are landed.

## Addressed

**Blockers** — [`6c8e3aad7`](https://github.com/JesusFilm/core/pull/9250/commits/6c8e3aad7)
1. **Trailing-slash URL.** `JourneyViewCardActions` now builds the deep link with `new URL('/', adminUrl)` + `searchParams.set('useTemplate', itemId)`, so a trailing slash on `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` no longer doubles the path (and the URL searchParams handle encoding).
2. **Duplicated `featuredCount`.** Extracted a shared `splitFeatured(items)` helper (with `FEATURED_COUNT` and the all-three carve-out) into `PublicGalleryPage.tsx`, called from both `JourneyView` and `AdminView`. Side-note: the prevention rule we just added to [\`shared-cross-app-view-component\`](https://github.com/JesusFilm/core/blob/siyangcao/nes-1694-public-page-restyle/docs/solutions/architecture-patterns/shared-cross-app-view-component-2026-05-26.md) anticipated exactly this — sharing the data model isn't enough; derived logic needs a shared home too.

**Follow-ups** — [\`ef2b1d57d\`](https://github.com/JesusFilm/core/pull/9250/commits/ef2b1d57d)
- **#6** \`aria-controls=\"gallery-nav-drawer\"\` is now unconditional.
- **#7** \`activeId\` is lazy-seeded to \`sections[0]?.id\` so the first link highlights on first paint.
- **#8** Dropped the redundant \`String(createdAt)\` casts in both \`toData()\` mappers and inside \`metaLine\`.
- **#9** \`bandBackground\` now uses \`Theme\` from \`@mui/material/styles\` instead of the inline structural type.
- **#11** \`JourneyViewCardActions\` threads an \`itemTitle\` prop and renders \`aria-label={t('Use {{title}}', { title })}\` (and the matching \`Preview\` aria-label). Ran \`nx run journeys-ui:extract-translations\` for the new keys.

## Not changed — reasoning

- **#3 image-host audit.** \`apps/journeys-admin/next.config.js\` \`images.remotePatterns\` already covers every host that \`primaryImageBlock.src\` can return (cloudfront/arclight, \`imagedelivery.net\`, \`customer-*.cloudflarestream.com\`, \`image.mux.com\`, \`i.ytimg.com\`, unsplash, imageshack). Matches \`apps/journeys\` for the gallery-relevant hosts.
- **#5 \`slotProps\`.** \`@mui/material\` is on v7.1.1 here — \`slotProps.paper\` / \`slotProps.backdrop\` are the correct v6+/v7 API.
- **#10 \`ScrollReveal\` empty deps.** The effect captures only \`setShown\` (stable React setter) and the ref; no missing deps, and lint stays clean for the file.

## Deferred follow-ups (filed mentally, not in this PR)

- **#4 i18n namespace migration cleanup.** Removing the legacy \`Untitled collection\` / \`Creator name\` / etc. keys from the old \`apps-journeys\` and \`apps-journeys-admin\` namespaces. Worth a tiny separate PR so it doesn't ride with the gallery diff.
- **#12 barrel re-exports.** Deciding whether \`GALLERY_ACCENT\` / \`GALLERY_CARD_RADIUS\` belong in \`index.ts\` or should drop their \`export\`. Same — best as a separate touch-up.
- **CodeRabbit's edge-case tests for \`TemplateGalleryView\`.** Reasonable suggestion; happy to add if you want them on this PR, otherwise will follow up.

Verification on these changes: \`tsc\` clean (journeys-ui, journeys, journeys-admin), Vitest passes \`PublicGalleryPage.spec.tsx\` (19) and \`TemplateGalleryView.spec.tsx\` (2), Jest passes \`CollectionPreviewPane.spec.tsx\` (4), and lint is clean for the three touched projects.

mikeallisonJS:
Solid refactor — the shared view-model + variant split is a clean fix for the admin/public drift, and the ADR is unusually thorough. CI is fully green. One real correctness risk, a couple of test-coverage gaps that should be ported from the deleted specs, and a handful of polish items. No security findings.

Leaning approve once #1 and #2 are addressed.

---

### High

**1. `toData` doesn't string-coerce `createdAt`** — `apps/journeys-admin/.../CollectionPreviewPane.tsx` (and the journey-app mapper in `TemplateGalleryView.tsx`).

The mapper writes `createdAt: journey.createdAt` straight through, but downstream `JourneyViewCard` calls `parseISO(item.createdAt)`. The deleted `GalleryTemplateCard.tsx` defensively did `parseISO(String(template.createdAt))`, and the companion design doc (`shared-cross-app-view-component-2026-05-26.md`) also specifies `createdAt != null ? String(createdAt) : null`. If Apollo ever hands back anything but a plain ISO string (custom scalar, `Date`, etc.), `parseISO` silently returns Invalid Date and the meta line drops the date without warning. One-line fix in both mappers:

```ts
createdAt: journey.createdAt != null ? String(journey.createdAt) : null,
```

**2. URL-encoding assertion lost in the test migration** — `libs/journeys/ui/.../PublicGalleryPage.spec.tsx`.

The deleted `GalleryTemplateCard.spec.tsx` had a test pinning `'tmpl/with spaces&chars'` → `'tmpl%2Fwith%20spaces%26chars'`. The new shared component uses `URL.searchParams.set('useTemplate', itemId)` which encodes spaces as `+`, not `%20` — so behavior actually changed under us. Port the assertion (pinned to whatever the new encoding is) so future tweaks don't silently drift again. If anything downstream is strict about `+` vs `%20`, flag that too.

### Medium

**3. Four `window` scroll listeners per page** — `JourneyView.tsx` instantiates `useParallax` three times (intro / featured / rest), and `JourneyViewNav` adds a fourth for the border fade. All are `{ passive: true }` and rAF-throttled so per-frame cost is bounded, but you're firing four rAFs per scroll batch. Consider consolidating to a single scroll handler that fans out via context.

**4. `priority={index < 2}` on featured rows preloads off-screen images** — `JourneyView.tsx`. The featured rows sit below the full-viewport cover and are never the LCP. Either set `priority={false}` for all of them, or move the hint to the cover image if/when one is added.

**5. Scrollspy never clears `activeId`** — `JourneyViewNav.tsx`. The "keep last active section" behaviour is intentional per the inline comment, but when the user scrolls past the last section into footer/whitespace the last nav link stays highlighted. Minor UX papercut — worth deciding whether to clear when the visible set is empty.

**6. `prefers-reduced-motion` read once, never re-evaluated** — `useParallax`, `ScrollReveal`, `scrollToSection`. Standard pattern, but if the user toggles the OS setting mid-session animations don't update. Add a `mediaQuery.addEventListener('change', …)` if you want responsiveness to live changes.

### Low / Nit

**7. `toData()` allocated in render without `useMemo`** — both mappers. New object reference every render. Nothing memoizes on it today so it's harmless, but pre-emptive `useMemo` would future-proof against `React.memo`-ing children.

**8. `GALLERY_ACCENT = '#C52D3A'`** — better than the old per-file `#26262E` duplication (centralized + exported, good), but still a hardcoded brand-red literal. Consider exposing `palette.accent` on the journeys-app theme so the brand color can shift in one place.

**9. Missing stories** — `PublicGalleryPage.stories.tsx` only has the happy path for both variants. Add empty-state, 3-template featured-all, missing-image, and dark-mode stories (the `bandBackground` branches on theme mode) — Storybook is a design-review surface.

**10. Other test coverage gaps from the deleted specs** (besides #2):
- `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` env override
- `https://admin.nextstep.is` fallback when env unset
- Empty `language.name` array → no language in meta
- Missing image → placeholder icon renders
- Missing `createdAt` → meta line omitted
- `whiteSpace: 'pre-wrap'` newline preservation in the header

**11. i18n cleanup** — namespace move from `apps-journeys` → `libs-journeys-ui` looks clean on the new-keys side. Verify the old strings (`Use`, `Preview`, `This gallery has no templates yet.`) are also removed from `apps-journeys.json` if not used elsewhere — `apps-journeys.json` isn't in the diff, so dead keys may be left behind.

---

### Security
No XSS surface, no `dangerouslySetInnerHTML`, no untrusted iframe construction. All `target="_blank"` links carry `rel="noopener noreferrer"`. `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` is build-time injected (not runtime user input). Image URLs flow through `next/image`'s remote-patterns allowlist. Deep-link construction uses `URL.searchParams.set` which encodes correctly. 👍

### What looks good
- **Neutral view-model + `variant` prop** is the right abstraction — directly addresses the NES-1694 drift, and the ADR explicitly calls out the parent↔child circular-import gotcha and the render-time invariant. Genuinely thoughtful engineering doc.
- **`splitFeatured` extracted** with parametric specs covering both variants at 3 and 4 items — exactly the kind of shared invariant that should be one helper, not two implementations.
- **Per-card `aria-label={t('Use {{title}}', { title: itemTitle })}`** replaces the previous duplicate generic "Use" labels — meaningful a11y win on a multi-card page.
- **IntersectionObserver scrollspy** uses correct enter+leave tracking (set of visible IDs, topmost-in-order resolution) instead of the naive last-entry-wins pattern, with an inline comment explaining why.
- **Drawer transparent-backdrop fix** (`backdrop.sx.backgroundColor: 'transparent'` + `backgroundImage: 'none'` on paper) is the right non-obvious MUI escape hatch.

csiyang:
Thanks again @mikeallisonJS — round 2 in [`28befd311`](https://github.com/JesusFilm/core/pull/9250/commits/28befd311). All High and Medium items addressed, plus the Low ones that don't expand scope. Main has been merged in too (so we're on the new journeys-admin Vitest config).

## High (both blockers)

1. **`String(createdAt)` restored** in both `toData()` mappers and inside `JourneyViewCard.metaLine`. Comment in each call explains the defensive intent (custom DateTime scalar / Date instance → `parseISO` Invalid Date → silently dropped meta date).
2. **URL-encoding assertion ported** as `'encodes special characters in the template id via URLSearchParams'` — input `'tmpl/with spaces&chars'` pinned to `tmpl%2Fwith+spaces%26chars` (spaces are `+` under `application/x-www-form-urlencoded`, not `%20`). Also added trailing-slash, env-override, and env-fallback tests so the admin URL contract is fully nailed down.

## Medium

- **#3** New `ScrollProvider` / `useScrollSubscription` (`libs/.../JourneyView/scrollContext.tsx`). The three `useParallax` refs + the nav border-fade now share one rAF-throttled scroll/resize listener instead of four. `JourneyView` wraps its body in the provider.
- **#4** Dropped `priority` from featured-row `<Image>`s — below the cover, never the LCP.
- **#5** `activeId` now clears when no observed section intersects the centre band, so the last link stops sticking when you scroll past the last section.
- **#6** New `usePrefersReducedMotion` hook subscribes to `matchMedia('change')`; `useParallax` consumes it, so toggling the OS setting mid-session stops/starts the drift live (and resets any leftover transform on the way out).

## Low

- **#7** `toData()` wrapped in `useMemo` in both mappers, so a future `React.memo` on children gets the stable reference for free.
- **#8** `GALLERY_ACCENT` now reuses the existing `brandRed` token from `@core/shared/ui/themes/base/tokens/colors` (same token AiChat already uses). Not adding a palette slot — token is the source of truth.
- **#10** Test gaps from the deleted spec landed in `PublicGalleryPage.spec.tsx`: URL encoding (#2), env override, env fallback, trailing-slash env value, empty `languageName`, null `createdAt`, missing image (placeholder icon), and `pre-wrap` newline preservation in the description.

## Skipped — reasoning

- **#9 Storybook variants** — we're not actively maintaining stories in general;
- **#11 Orphaned `Use`/`Preview`/empty-state keys in `apps-journeys.json`** — the extract-translations workflow is authoritative for what should be in those files, so we leave it to the next normal extract to clean up rather than manually editing locale JSONs.

### Verification

- `tsc` clean on `journeys-ui`, `journeys`, `journeys-admin`.
- Vitest: `PublicGalleryPage.spec.tsx` (27, up from 19 with 8 new), `TemplateGalleryView.spec.tsx` (2), `CollectionPreviewPane.spec.tsx` (4, now on the new journeys-admin Vitest config from main).
- Lint clean on all three projects.

mikeallisonJS:
## Code Review

Solid refactor — neutral view-model + variant entry component + shared `splitFeatured` is the right shape and directly addresses the drift NES-1694 was about. CI green so far (unit, build, CodeQL, Danger, deploy previews); Playwright + lint-work still running at time of review. A few things to address before merge.

### High-priority

1. **Parent↔child import cycle in `PublicGalleryPage.tsx`.** The entry module exports `GALLERY_ACCENT`, `GALLERY_CARD_RADIUS`, `PublicGalleryPageData`, `PublicGalleryPageItem`, `splitFeatured` **and** imports `JourneyView` + `AdminView`, which import those exports back. The PR's own ADR (`shared-cross-app-view-component-2026-05-26.md`) flags this as safe only because `import/no-cycle` isn't enforced and references are render-time. One stray module-top-level `const x = GALLERY_ACCENT` in a child will silently get `undefined`. Extract tokens/types/`splitFeatured` into a leaf `galleryTokens.ts` — the ADR includes the recipe.

2. **Unescaped slug used as href in `JourneyViewCardActions.tsx`.**
   ```ts
   const previewHref = `/${itemSlug}`
   ```
   `itemSlug` flows from GraphQL `template.slug` / `journey.slug`. If a slug ever contained `//`, `..`, or URL-encoded segment separators, the link can navigate cross-origin or escape route scope. Backend probably validates, but `encodeURIComponent(itemSlug)` (or a slug-shape assertion) is free defense.

3. **Four i18n keys unrelated to this PR in `libs/locales/en/libs-journeys-ui.json`:**
   ```json
   "Ask anything…": "Ask anything…",
   "Scroll to latest message": "Scroll to latest message",
   "Ask a question": "Ask a question",
   "Replies may not be perfect": "Replies may not be perfect"
   ```
   These look like an accidental rebase/merge from a chat-UI feature — remove before merge.

4. **Admin URL encoding semantics changed.** `?useTemplate=` construction switched from template-string + `encodeURIComponent` (spaces → `%20`) to `URL` + `URLSearchParams` (spaces → `+`). The new spec pins the `+`. The admin route's receiver will be fine if it parses via `URLSearchParams.get()` (decodes `+` → space), but any code that reads `location.search` and pipes through `decodeURIComponent` directly will now break. Worth confirming on the admin side.

5. **Orphan i18n in `apps-journeys` namespace.** Empty-state and card-action strings moved namespace `apps-journeys` → `libs-journeys-ui`. The old keys in `libs/locales/en/apps-journeys.json` are likely still present — audit and remove so they don't linger.

### Medium / suggestions

- **`useParallax` writes `transform` every rAF tick** (`JourneyView.tsx`):
  ```ts
  element.style.transform = `translateY(${(-clamped * strength * viewport).toFixed(1)}px)`
  ```
  Cache last value, early-return on no-change, and skip when `Math.abs(progress) > 1` (off-screen). ~60 string allocs/sec/ref matters on low-end Android with 3 refs on a long page.
- **Reduced-motion flush.** If `prefers-reduced-motion` flips ON while page is idle (no scroll firing), parallaxed elements stay at the last drift offset until the next scroll. `usePrefersReducedMotion` is reactive — wire a one-shot transform-reset when it changes.
- **`ScrollReveal.tsx`** doesn't gate on `prefers-reduced-motion`. Observer keeps running until intersection even when motion is disabled. Reuse `usePrefersReducedMotion` for symmetry with `useParallax`.
- **`AdminView` default `aria-hidden`.** PR description says the admin tree is meant to sit behind `aria-hidden`. `CollectionPreviewPane` does this — but if anyone else mounts `AdminView` directly they won't know. Apply `aria-hidden="true"` at the component root by default, or hoist this into a louder JSDoc.
- **`scrollContext.tsx` subscriber error isolation.** `subscribers.current.forEach(cb => cb())` runs callbacks synchronously; one throwing subscriber skips the rest. Wrap each call in try/catch.
- **`JourneyView.tsx` is 475 lines** with `useParallax`, `SectionLabel`, `FeaturedRow`, `JourneyView`, `JourneyViewBody`, `SECTION_IDS`, `bandBackground`, `fullHeightSection` all colocated. `FeaturedRow` (~95 lines) and `useParallax` (~30 lines) deserve their own files for parity with `ScrollReveal.tsx`.
- **`JourneyViewNav.tsx` drawer id.** `slotProps.paper.id="gallery-nav-drawer"` should land on the actual `<aside>` MUI renders — worth a quick DOM check to make sure `aria-controls` resolves.
- **`galleryFixture.ts` likely partially dead.** Most card/grid specs that consumed it are gone; trim it to what `TemplateGalleryView.spec.tsx` actually uses.
- **Test coverage gaps** (not blockers): no unit tests for `useParallax`, `useScrollSubscription`, `usePrefersReducedMotion` reactive flip, or `ScrollReveal` reduced-motion immediate-show path.

### What looks good

- IntersectionObserver pattern in `JourneyViewNav.tsx` is exemplary — Set-based active tracking, picks topmost in scroll order, jsdom-guarded, disconnected on cleanup.
- `rel="noopener noreferrer"` is correct on every `target="_blank"`.
- `usePrefersReducedMotion` covers the SSR→hydrate gap with a post-mount re-sync.
- Both `toData` mappers wrap in `useMemo` with proper deps; `React.memo` on the admin consumer.
- New spec (~426 lines) covers admin URL env-override, trailing-slash handling, special-char encoding, the 3-vs-4-item featured split, empty state, decorative-vs-real actions, pre-wrap newline preservation.
- Legacy `apps/journeys` files properly renamed (not duplicated) into the shared lib — clean git history.
- Two architecture docs accompany the change.

### Verdict

**Request changes** — the list is short, but the import cycle, unrelated i18n keys, and slug encoding are worth resolving, and the URL encoding semantics change on `?useTemplate=` deserves explicit confirmation rather than a guess.

csiyang:
Round 3 in [`177295128`](https://github.com/JesusFilm/core/pull/9250/commits/177295128). All High + all Medium (including the hook unit tests you flagged "not blockers"), with replies for the items that aren't code changes.

## High

1. **Parent↔child import cycle broken.** New leaf `galleryTokens.ts` carries the tokens, view-model types, `FEATURED_COUNT`, `splitFeatured`, and `SplitFeaturedItems` — and imports *nothing* from the component tree. Children (`JourneyView/*`, `AdminView`, the mock) import the leaf directly; `PublicGalleryPage.tsx` does `export * from './galleryTokens'` so external consumers (`TemplateGalleryView`, `CollectionPreviewPane`) are unaffected. Exactly the recipe the ADR describes.
2. **`itemSlug` encoded in `previewHref`** via `encodeURIComponent` — free defence against `/`, `..`, or encoded segment separators escaping the route.

## Medium (all)

- **`useParallax` perf**: caches the last-applied transform string and short-circuits on no-change (no per-rAF allocation), plus an off-screen short-circuit when the element is well above/below the viewport.
- **Reduced-motion flush**: separate `useEffect` keyed on `reduceMotion` so toggling ON while the user is *idle* clears any lingering transform — the scroll callback alone can't help in that case.
- **`ScrollReveal` now consumes `usePrefersReducedMotion`** for symmetry with useParallax; the effect re-runs on OS toggles and skips attaching the IntersectionObserver when motion is disabled.
- **`AdminView` default `aria-hidden="true"`** at the component root — a consumer that forgets to wrap it still won't expose the duplicate subtree to AT.
- **`ScrollProvider` subscriber error isolation**: each callback wrapped in try/catch, one throw doesn't skip the rest in the same tick.
- **`Drawer keepMounted`**: keeps Paper in the DOM so the hamburger's unconditional `aria-controls="gallery-nav-drawer"` always resolves (MUI Drawer's default unmounts the Modal subtree). Drawer id verification.
- **JourneyView size**: now ~200 lines (down from ~475) by extracting `useParallax` and `FeaturedRow` to their own files.

## Hook unit tests (your "not blockers" — landed anyway)

- `usePrefersReducedMotion.spec.ts` (3) — initial true/false; reacts to `matchMedia` `change` mid-session.
- `scrollContext.spec.tsx` (3) — broadcasts on scroll inside provider, no-op outside, isolates subscriber errors.
- `useParallax.spec.tsx` (3) — writes `translateY` on overlap; skips when reduce-motion is on; skips when well off-screen.
- `ScrollReveal.spec.tsx` (2) — no IntersectionObserver attached under reduce-motion; observer attached when motion allowed.

Suite is now 5 files / 38 tests in the gallery area.

## Replies (not code changes)

- **#3 (chat-UI keys)** — `Ask anything…`, `Scroll to latest message`, `Ask a question`, `Replies may not be perfect` are **real `t()` calls** in `libs/journeys/ui/AiChat/ChatHeader`, `PromptInput`, `Conversation`, plus consumers in `apps/watch` / `apps/resources`. The extract-translations workflow is authoritative; removing them manually would be reversed by the next extract.
- **#4 (URL encoding semantics)** — confirmed safe. `apps/journeys-admin` reads the param via `router.query.useTemplate` (Next.js router → URLSearchParams under the hood, decodes `+` back to space). Template ids are journey UUIDs in practice, no spaces in real data.
- **#5 (apps-journeys orphans `Use`/`Preview`)** — deferred per the project's `removeUnusedKeys: false` convention (`i18next.config.base.js` explicitly preserves keys not in source; comment: *"Match old i18next-parser behavior: preserve keys not found in source"*). Not editing locale JSONs by hand.
- **`galleryFixture.ts` trim** — already minimal: every field on `mockTemplate` is required by the generated `GalleryTemplate` GraphQL type. Dropping any would fail tsc.

### Verification

- tsc clean on journeys-ui, journeys, journeys-admin.
- Vitest: 38/38 in the gallery area (27 `PublicGalleryPage.spec.tsx` + 3 + 3 + 3 + 2 new hook/component specs); 2/2 `TemplateGalleryView.spec.tsx`; 4/4 `CollectionPreviewPane.spec.tsx`.
- Lint clean on all three projects.

mikeallisonJS:
## Code review

Reviewed the diff for security, performance, correctness, and maintainability. Architecturally this is a clean win — one source of truth for the gallery, a well-tested scroll primitive, and a solid a11y posture. Two issues worth fixing before merge; the rest can ship as follow-ups.

### High-severity

1. **LCP regression — no `priority` on the new `<Image>`s.** `JourneyViewCard` accepts an optional `priority?` prop but no caller passes it (`JourneyView.tsx`, `FeaturedRow.tsx`, `AdminView.tsx`). The old `TemplateGalleryGrid` set `priority={index < 3}` for above-the-fold cards. The first featured image on the new design is likely the LCP element and is currently lazy-loaded.

2. **`new URL('/', adminUrl)` in `JourneyViewCardActions.tsx` can throw at render.** If `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` is ever set to a schemeless value (e.g. `admin.staging.local`), `new URL` throws and crashes the page. The old template-string approach degraded silently. Wrap in `try/catch` with a fallback href + console warn, or validate at build time.

### Suggestions

- **Mobile + desktop grids both render**, toggled via `display: { xs/md }`. Doubles `<JourneyViewCard>` instantiation per item (2× IntersectionObserver, 2× action buttons, potentially 2× image requests). Render once and swap variants via a media-query hook.
- **`willChange: 'transform'`** set permanently on three large containers in `JourneyView.tsx` — keeps them composited the entire session. Worth measuring on low-end Android.
- **`ScrollReveal` + `prefers-reduced-motion`.** Under reduced motion the component shows immediately but the CSS `transition: 700ms` still applies, so the "instant" reveal animates over 700ms. Set `transition: 'none'` when `reduceMotion`.
- **`JourneyViewNav` IntersectionObserver** depends on the memoised `sections`, which includes `t`. Any language change tears the observer down and rebuilds it. Use `[sections.map(s => s.id).join(',')]` instead.
- **Unrelated i18n keys** in `libs/locales/en/libs-journeys-ui.json`: `Ask anything…`, `Scroll to latest message`, `Ask a question`, `Replies may not be perfect` look like they belong to an assistant/chat UI, not the gallery. Possibly slipped in from a sibling branch — worth confirming. Also: only `en` is in the diff, verify other-locale sync.
- **`ScrollProvider` subscriber error logging.** A throwing subscriber will `console.error` on every rAF tick while scrolling (~60/s). Unsubscribe on first throw or rate-limit.
- **Test coverage gaps from removed specs.** `PublicGalleryPage.spec.tsx` (+426) replaces most of the deleted specs, but a few assertions weren't carried over:
  - explicit `rel="noopener noreferrer"` check (was in `GalleryTemplateCard.spec.tsx`)
  - Preview-button labelling assertion
  - `JourneyViewMedia` placeholder variant forwarding
  - `JourneyViewHeader` avatar `alt` fallback
- **Hook-order tripwire in `JourneyView.tsx`.** The file is fine today, but each `useParallax(...)` runs several hooks internally. A future early-return added before any of them would break the rules of hooks. A `// no early returns above this line` comment would help future-proof.
- **Add a spec for slug containing `/`** (e.g. `foo/bar` → `/foo%2Fbar`) to lock in the `encodeURIComponent` defence.
- **`metaLine`** is exported from `JourneyViewCard.tsx` and consumed by `FeaturedRow.tsx`. Move to `galleryTokens.ts` to keep the import graph clean.

### What looks good

- **Security.** No `dangerouslySetInnerHTML`, all `target="_blank"` carry `rel="noopener noreferrer"`, slugs are `encodeURIComponent`-ed, URLs built with `new URL` + `URLSearchParams`, encoding pinned by spec.
- **Scroll architecture.** `ScrollProvider` consolidates 4 listeners into one rAF-throttled listener with `{ passive: true }`. Cleanup, error isolation, and priming are all tested.
- **`prefers-reduced-motion`** is genuinely implemented and reactive via `matchMedia.addEventListener('change')`. `useParallax` flushes a lingering transform when the user toggles mid-session.
- **`useParallax` short-circuits** for reduce-motion, off-screen, and same-string-as-cached — all tested.
- **View-model boundary.** `PublicGalleryPageData` cleanly decouples GraphQL types and Formik values from the rendering library. Both mappers defensively `String(template.createdAt)`.
- **Shared `splitFeatured`** ensures admin preview and live page can't drift on the 3-vs-4-item layout decision.
- **Spec coverage** of new behaviour is strong (`PublicGalleryPage.spec.tsx` is 426 lines).

### Verdict

Request changes — minor. The two High items are quick fixes. Everything else is suggestion-level.

csiyang:
Round 4 in [\`57f27255c\`](https://github.com/JesusFilm/core/pull/9250/commits/57f27255c). Both High items addressed, all adopted suggestions landed (with new tests), and a few framing replies on the rest.

## High

1. **`new URL` schemeless guard** — `JourneyViewCardActions` wraps the deep-link construction in try/catch with a sanitised template-string fallback (`adminUrl.replace(/\/$/, '') + \`/?useTemplate=\${encodeURIComponent(itemId)}\``). A schemeless `NEXT_PUBLIC_JOURNEYS_ADMIN_URL` no longer throws and crashes render. Test covers the fallback path.

2. **`priority` on the first featured `<Image>`** — `FeaturedRow` now accepts an optional `priority` prop; `JourneyView` passes `priority={index === 0}` for the first row only. ⚠️ Note: this *reverses* your round-2 #4 reasoning. The cover is text-only (no image), so the actual LCP element is the **title h1**, not the first featured image. The featured rows sit below the full 100svh cover and are below the fold on initial paint. Adding `priority` is cheap insurance, but if you're sure the LCP is the title, this can come back off. Happy to defer to your call.

## Adopted

- **`ScrollReveal` under reduced motion**: `transition: 'none'` + `transitionDelay: '0ms'` + `willChange: 'auto'` — the "instant" reveal no longer animates over 700ms.
- **`JourneyViewNav` scrollspy dep**: keyed on the joined id string instead of the memoised `sections` array, so a language change doesn't tear the observer down and re-attach for the same id set.
- **`ScrollProvider` error handling**: drop a throwing subscriber from the set after its first failure — console no longer floods with the same error on every rAF tick.
- **Permanent `willChange: 'transform'` removed** from the three parallax containers in `JourneyView`. Letting the browser optimise on demand beats permanently composited surfaces.
- **`metaLine` moved to `galleryTokens.ts`** alongside `splitFeatured`, so every derivation of `PublicGalleryPageItem` lives in the leaf. `FeaturedRow` no longer crosses back through `JourneyViewCard`.
- **Hook-order tripwire** comment before the three `useParallax()` calls in `JourneyViewBody`.

## New tests

- `falls back to a sanitised template string when the env value is schemeless` — covers the new URL guard.
- `carry rel="noopener noreferrer" and target="_blank" on Use and Preview` — locks in the security hygiene.
- `encodes a slug containing "/" so it never escapes the route` — pins `encodeURIComponent(itemSlug)` on the preview href.
- `gives Preview buttons an aria-label keyed on the item title` — proves the `t('Preview {{title}}', { title })` call site is wired. (Stubbed `t()` doesn't interpolate, so we assert the `{{title}}` shape rather than an interpolated value.)

Spec is now 42 tests in PublicGalleryPage alone (53 across the 5 files in the gallery area).

## Replies (no code change)

- **Chat-UI keys** (third time flagged) — same reply as round 3. They're legitimate `t()` calls in `AiChat/ChatHeader`, `PromptInput`, `Conversation`, plus consumers in `apps/watch` / `apps/resources`. The extract-translations workflow is authoritative; removing them manually would be reversed by the next extract on any future PR that touches those keys.
- **Mobile + desktop grid both rendered** — trade-off for SSR-safe layout without a hydration mismatch (a `useMediaQuery`-driven swap would be either client-only or hydrate-flicker). Worth revisiting with real perf numbers, but out of scope for this PR.
- **`JourneyViewMedia` placeholder variant + `JourneyViewHeader` avatar alt fallback** — both behaviours are exercised end-to-end via the rendering tests; happy to add explicit assertions if they're load-bearing.

### Verification

- tsc clean on journeys-ui, journeys, journeys-admin.
- Vitest: `PublicGalleryPage.spec.tsx` 31/31, hook specs 11/11 (42 total in the lib), `TemplateGalleryView.spec.tsx` 2/2, `CollectionPreviewPane.spec.tsx` 4/4.
- Lint clean on all three projects.

mikeallisonJS:
## Code Review

CI is green across all three checks. Architecture is sound, accessibility/i18n/SSR/reduced-motion are handled deliberately, and test coverage is thorough.

### Critical Issues

None. No XSS sinks or `dangerouslySetInnerHTML`. Every `target="_blank"` carries `rel="noopener noreferrer"` (asserted in spec), slugs are `encodeURIComponent`'d, and `buildUseTemplateHref` `try/catch`es schemeless env values. SSR-sensitive globals (`window`, `matchMedia`, `IntersectionObserver`, `document.getElementById`) are inside `useEffect` or behind `typeof` guards — correct for the Next.js host.

### Suggestions

**Correctness**
- `scrollContext` priming misses late subscribers (`JourneyView/scrollContext.tsx` ~L87–96). `ScrollProvider`'s one-shot priming `update()` fires before subscribers registered on later renders. Visible impact is small because `useParallax` short-circuits off-screen, but consider invoking the new subscriber once inside `subscribe()` so late-mounters get an initial tick.
- `usePrefersReducedMotion` uses only modern `addEventListener('change')` (L23–32). Safari <14 needs `addListener`. Probably fine for the 2026 baseline; an `'addEventListener' in mq` fallback would harden it.
- `aria-current="true"` on scrollspy nav (`JourneyViewNav.tsx`) — for in-page section links, `aria-current="location"` is the canonical value. `aria-haspopup="true"` on the mobile menu trigger should be `"menu"` or `"dialog"` (the boolean form is deprecated).

**Performance**
- `willChange: 'opacity, transform'` persists after the one-shot reveal (`ScrollReveal.tsx` L80). The composited layer stays alive for the element's lifetime — clear it after reveal to free GPU memory.
- `next/image` `sizes` slightly over-fetches on mobile (`JourneyViewCard.tsx` overlay variant, L62): `60vw` at xs but cards render ~50vw in the 2-col grid.
- Shared `ScrollProvider` is a real win — three `useParallax` instances + the nav border-fade share one rAF-throttled passive listener with isolated per-subscriber error handling. Worth modelling for other scroll-heavy pages.

**Maintainability**
- `JourneyView.tsx` (327 lines) repeats section-wrapper sx four times (L168–321). Extract a `<GallerySection id label backgroundTint topAlign>` helper — halves the file and makes the band-tint rule legible in one place.
- Duplicate `SectionLabel` in `JourneyView.tsx` (L56–73) and `AdminView.tsx` (L29–43) differ only in spacing. Promote to `galleryTokens.ts` with a prop.
- `toData` mapper is duplicated in `apps/journeys/.../TemplateGalleryView.tsx` and `apps/journeys-admin/.../CollectionPreviewPane.tsx`. Same defensive `String(createdAt)`, same image mapping — extract before a third caller arrives.
- Old `Template*` `data-testid`s linger inside the new `PublicGalleryPage` tree (`TemplateGalleryHeader`, `TemplateGallerySections`, `GalleryTemplateCard`, `GalleryTemplateCardUseButton`, `GalleryTemplateCardPreviewButton`, `TemplateGalleryMedia`, `TemplateGalleryEmptyState`). Likely deliberate churn-avoidance, but worth a follow-up cleanup ticket.
- Magic constants scattered — `GALLERY_NAV_HEIGHT = 56`, `ACTION_SIZE = 40`, `FEATURED_COUNT = 2` — co-locate in `galleryTokens.ts`.

### What Looks Good

- Leaf-module architecture (`galleryTokens.ts` owns the view-model + `splitFeatured` + `metaLine`) breaks the parent↔child cycle the architecture doc warns against.
- Single source of truth for the featured/rest split (`splitFeatured`), spec-covered for the 3-item edge case.
- Defensive `buildUseTemplateHref` with `try/catch` on schemeless env values, plus slug encoding — both pinned by tests.
- `prefers-reduced-motion` honored across `ScrollReveal`, `useParallax`, and `scrollToSection`, each with its own spec.
- Decorative admin variant correctly drops link semantics (`aria-hidden`, no `role="link"`).
- `PublicGalleryPage.spec.tsx` (494 lines) covers variants, env override + fallback, slug escape, `target`/`rel`, aria-labels, empty/missing fields, media presence, pre-wrap preservation.
- i18n strings consolidated into `libs/locales/en/libs-journeys-ui.json` with disambiguated per-card aria-labels.

### Verdict

**Approve with non-blocking nits.** No security or correctness defects. The maintainability items (decompose `JourneyView`, dedupe `SectionLabel`, extract `toData`, rename stale `Template*` testids) and minor a11y polish are good follow-up tickets but shouldn't block merge.

mikeallisonJS:
## Code Review — restyle public gallery page + share with admin preview

🤖 Automated review (high-effort, recall-biased). One genuine correctness regression, plus a cluster of cleanup/altitude findings — several of which the PR's own architecture doc flags as the drift this refactor was meant to eliminate.

### 🔴 Correctness

**1. Empty state and media now render together when a gallery has no templates but has a `mediaUrl`** — `libs/journeys/ui/src/components/PublicGalleryPage/JourneyView/JourneyView.tsx:262`

The old `TemplateGalleryView` guarded the empty state with `!hasTemplates && !hasMedia`. The new `JourneyView` renders them on independent conditions:

```tsx
{!hasTemplates && ( ...<JourneyViewEmptyState /> )}        // line 262
{hasMedia && ( ...<JourneyViewMedia mediaUrl={data.mediaUrl} /> )}  // line 268
```

A published collection with zero templates but a configured Strategy embed now shows *"This gallery has no templates yet."* directly above a fully-populated media section. Restore the `!hasMedia` term, or make the empty state mutually exclusive with media.

**2. No test covers the no-templates-with-media case** — `libs/journeys/ui/src/components/PublicGalleryPage/PublicGalleryPage.spec.tsx`

Every `items: []` test uses mock data whose `mediaUrl` defaults to `null`, so finding #1 is invisible to the suite and will regress again silently. Add a spec with `items: []` + non-null `mediaUrl` asserting the empty state is suppressed.

### 🟡 Altitude / drift (the PR's stated goal)

**3. The two `toData` GraphQL→view-model mappers are duplicated across both apps** — `apps/journeys/src/components/TemplateGalleryView/TemplateGalleryView.tsx:15` and `CollectionPreviewPane.tsx`

Identical item-mapping logic (the `String(createdAt)` coercion, `primaryImageBlock` guard, field list) lives in two files. Add a field to `PublicGalleryPageItem` and miss one copy → the live page and admin preview render different content from the same collection — the exact drift this PR fixed at the *render* layer but left open at the *projection* layer. The PR's own doc (`shared-cross-app-view-component-2026-05-26.md`) documents a real drift bug from this pattern (the `FEATURED_COUNT` lone-card bug). Consider a shared `mapItem()` in `galleryTokens`.

**4. The shared "neutral" leaf hardcodes journeys-app URL concerns** — `JourneyViewCardActions.tsx:124`

`process.env.NEXT_PUBLIC_JOURNEYS_ADMIN_URL` and the `/${slug}` viewer route are baked into the shared component. The admin variant only works because it renders these as *decorative* look-alikes; any surface wanting real links must fork or thread config. URL/route construction is an app concern and belongs in the view-model or a prop.

### 🟡 Reuse / efficiency / cleanup

**5. Parallax reads `getBoundingClientRect()` on every rAF tick across 3 instances → layout thrash** — `useParallax.ts:59`. Each of the 3 subscribers reads layout then writes `element.style.transform` in the same tick; the interleaved read→write→read forces synchronous reflow and risks dropped frames on low-end devices. Cache `offsetTop` on mount/resize and read `scrollY` once per tick in the provider.

**6. `onDark` prop is fully dead** — `JourneyViewCardActions.tsx:30/71/136`. Declared, defaulted to `false`, and branched, but no caller ever passes `onDark`. Delete the prop and inline the accent branch.

**7. `usePrefersReducedMotion` re-implements MUI's `useMediaQuery`** — `usePrefersReducedMotion.ts`. MUI's `useMediaQuery('(prefers-reduced-motion: reduce)')` (already a repo dependency) provides reactive, SSR-aware matching with the same hydration handling this hook hand-rolls.

**8. `metaLine()` recomputes `parseISO` + `intlFormat` per card render and duplicates the date/language derivation already in `TemplateGalleryCard.tsx`** — `galleryTokens.ts:136`. The two copies already differ (`month: 'long'` vs `'short'` + `isThisYear`) and will drift further. Precompute the meta string in `toData` or `useMemo` it.

**9. `ScrollReveal` flashes hidden state for reduce-motion users** — `ScrollReveal.tsx:47`. Initial render is `reduceMotion=false` (SSR-matching), so a reduce-motion client paints content at `opacity:0`/translated for one frame before the effect flips `shown`. The old grid showed everything immediately. Minor UX regression for exactly the users who asked for less motion.

**10. `GALLERY_CARD_RADIUS` token bypassed by inline `borderRadius: 3`** — `JourneyViewCard.tsx:97` (panel variant) and `AdminView.tsx:144`. The token governs some corners but not these; bumping it leaves the panel and admin placeholder square. Same drift class the tokens exist to prevent.

---

The strongest action item is **#1** — a user-visible regression with no test coverage (**#2**). Everything below the line is quality/altitude; **#3** and **#4** are worth taking seriously because they re-open the drift surface this PR was built to close.

mikeallisonJS:
## Re-review — after `307342b` + `6f71ded`

🤖 Re-reviewed the two follow-up commits against my earlier findings. **All correctness issues resolved; most cleanup items addressed.** Verified the lib specs pass locally (`ScrollReveal.spec` + `PublicGalleryPage.spec`: 32 passing, including the new empty-state-with-media regression guard).

### Resolved ✅
- **#1 Empty-state/media double-render** — `JourneyViewEmptyState` removed; sections now gate per-data, so a template-less gallery with a `mediaUrl` renders media only. The contradictory copy is gone.
- **#2 Missing test** — added *"renders the media section but no Explore/More when items are empty and mediaUrl is set"* — passes.
- **#6 Dead `onDark` prop** — removed entirely (along with the whole light/dark `useButtonSx` branch).
- **#7 `usePrefersReducedMotion` reinvents `useMediaQuery`** — swapped to MUI's `useMediaQuery('(prefers-reduced-motion: reduce)')` in both `ScrollReveal` and `useParallax`; hook + spec deleted.
- **#9 Reduce-motion reveal flash** — improved: `useMediaQuery` (via `useSyncExternalStore`) resolves the client value at hydration without a mismatch, and the spec now asserts no observer attaches under reduce-motion. (Content still starts at `opacity:0` and reveals via the post-mount effect — inherent to SSR scroll-reveal, acceptable.)
- **#10 `GALLERY_CARD_RADIUS` token bypass** — both `JourneyViewCard` and `AdminView` now use the token.

### Not addressed (all non-blocking nits) ⚪
- **#3** duplicated `toData` mappers across the two apps
- **#4** admin-URL coupling in the shared leaf (reduced in scope now that the `/slug` Preview link was removed)
- **#5** parallax `getBoundingClientRect()` per rAF tick
- **#8** `metaLine` not memoized / duplicated with `TemplateGalleryCard`

### Note
The Preview (`/slug`) icon-button was removed entirely along with the `itemSlug` prop — a scope change beyond my findings (review nitpick), not a bug. `slug` remains in `PublicGalleryPageItem` but is now unused by the view; harmless and easy to reinstate.

Couldn't execute `TemplateGalleryView.spec.tsx` locally — it hits the pre-existing `setupTests.ts` path quirk that also fails untouched journeys specs (e.g. `Conductor.spec`), unrelated to this PR.

**LGTM.** Remaining items are optional cleanup.

mikeallisonJS (approved):

PR #9258, opened 2026-05-27
feat(journeys-admin): flag journey tracking metrics editing [QA-485]

csiyang:
## Summary

- Shows the tracking-metric editor (the `EventLabel` control) on **regular, non-template journeys** when the `editJourneyTrackingMetrics` flag is enabled. Template behaviour is unchanged.
- Gate applied identically in the **Button**, **RadioOption**, **Video**, and **Card** property panels: `journey?.template` → `journey?.template || editJourneyTrackingMetrics`.

## Why (QA-485)

The template breakdown's "Decision for Christ" count came out far higher than the editor's button-click count. Root cause: the decision metric was attached to **both** the card (fires on *view*, via `Step.tsx`) **and** the "Yes" button (fires on *click*), and both land in the same `christDecisionCapture` column — a double count.

This PR is **temporary relief**: it exposes the metric editor on journeys so an editor can open the affected journey and clear the duplicate label (keep it on the button, remove it from the card). The deeper fixes — `Step.tsx` emitting the raw event name instead of routing through `fireCaptureEvent`, and `transformBreakdownResults` summing per-row unique counts instead of deduping like the page-view path does — are tracked separately.

## Notes

- `useFlags()` returns `{}` with no provider, so the flag is simply falsy when absent — no behaviour change when off.
- The control's helper text still reads "…every project created from your template", which is template-specific wording now also shown on journeys. Left as-is for this relief PR — can adjust if desired.

## Test plan

- [x] Existing `Button` / `RadioOption` / `Video` / `Card` specs pass (51) + 2 new `Card` flag tests (non-template shows with flag on, hidden with flag off).
- [ ] Flag OFF, non-template journey → no "Tracking" section on button/card/radio/video.
- [ ] Flag ON, non-template journey → "Tracking" section appears; event label can be changed/cleared.
- [ ] Templates still show the tracking editor regardless of the flag.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XavX9DG6XZam6RkBCw4yYV)_

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Button/Button.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Button/Button.tsx
@@ -6,6 +6,7 @@ import { ReactElement, useEffect } from 'react'
 import type { TreeBlock } from '@core/journeys/ui/block'
 import { useEditor } from '@core/journeys/ui/EditorProvider'
 import { useJourney } from '@core/journeys/ui/JourneyProvider'
+import { useFlags } from '@core/shared/ui/FlagsProvider'
 import ActivityIcon from '@core/shared/ui/icons/Activity'
 import AlertCircleIcon from '@core/shared/ui/icons/AlertCircle'
 import AlignLeft from '@core/shared/ui/icons/AlignLeft'
@@ -48,6 +49,7 @@ export function Button({
 }: TreeBlock<ButtonBlock>): ReactElement {
   const { dispatch } = useEditor()
   const { journey } = useJourney()
+  const { editJourneyTrackingMetrics } = useFlags()
   const { t } = useTranslation('apps-journeys-admin')
 
   const startIcon = children.find(
@@ -70,7 +72,7 @@ export function Button({
 
   return (
     <Box data-testid="ButtonProperties">
-      {journey?.template && (
+      {(journey?.template || editJourneyTrackingMetrics) && (
         <Accordion
           icon={<ActivityIcon />}
           id={`${id}-event-label`}
--- a/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/Card.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/Card.spec.tsx
@@ -130,6 +130,29 @@ describe('Card', () => {
       })
     })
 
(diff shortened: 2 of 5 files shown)


apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Button/Button.tsx
```diff
@@ -70,7 +72,7 @@ export function Button({
 
   return (
     <Box data-testid="ButtonProperties">
-      {journey?.template && (
+      {(journey?.template || editJourneyTrackingMetrics) && (
```
edmonday, line 75: **Concern (optional):** Button, RadioOption, and Video all get this identical `journey?.template || editJourneyTrackingMetrics` gate, but only Card has flag-on/flag-off tests. The loaded testing rule does call for a test when a flag drives branching logic — and since "keep it on the button" is the actual relief target here, Button is the one most worth covering. The gate is byte-identical across all four though, so Card is a fair representative; up to you whether one more test earns its keep.

edmonday (commented):
## Review summary

Tight, well-scoped relief PR. Verified the four edited panels (Button, Card, RadioOption, Video) are the **complete** set of `journey?.template`-gated `EventLabel` accordions — no block missed (the `BackgroundMediaVideo` `EventLabel` reference is just a default-value assignment, not a gate). The underlying mutations are standard block updates that work on regular journeys, so editors can actually clear the duplicate label. Fallback semantics are correct: `useFlags() → {}` keeps behaviour unchanged when the flag is off.

**Concern — misleading helper text shown on journeys:** `EventLabel.tsx:305-307` reads _"Tracking covers user actions in every project created from your template."_ With the flag on, editors see this on non-template journeys where it's wrong. You flagged this in the description and chose to leave it — reasonable for relief — but since the `label`/`showHelperText` props already exist, a one-line conditional or a fast-follow would avoid confusing the internal editors using this. Judgment call; not blocking. (File isn't in the diff, so noted here rather than inline.)

**Operational note (not a code finding):** The relief is inert until `editJourneyTrackingMetrics` is created and enabled in the **LaunchDarkly dashboard** — worth confirming that's tracked, since it lives outside this diff.

**Positive:**
- Correct fallback semantics — absent flag is falsy, zero behaviour change when off.
- Consistent application across exactly the right four panels; no block missed.
- New tests exercise both branches through a real `FlagsProvider`, not a stub.
- Honest, root-cause-level description that scopes this as temporary relief and points to the deeper `Step.tsx` / `transformBreakdownResults` fixes.

**Overall:** Good to merge. No critical issues. One optional copy tweak and a flag-branch test gap noted inline.

edmonday (approved):

PR #9267, opened 2026-06-02
fix(journeys-ui): make ChatOverlay backdrop fully opaque [NES-1654]

jaco-brink:
Closes [NES-1654](https://linear.app/jesus-film-project/issue/NES-1654/apologist-chat-ux-polish-from-lucindas-feedback-solid-background-input).

## What

Makes the desktop `ChatOverlay` backdrop **fully opaque** so the underlying journey card is no longer visible behind the chat.

```diff
- bgcolor: (theme) => alpha(theme.palette.grey[900], 0.88),
- backdropFilter: 'blur(6px)',
- WebkitBackdropFilter: 'blur(6px)'
+ bgcolor: 'grey.900'
```

3-line change to `libs/journeys/ui/src/components/ChatOverlay/ChatOverlay.tsx` + drop unused `alpha` import.

## Why

Per Lucinda's UX feedback (captured in NES-1654), the previous backdrop (`alpha(grey[900], 0.88)` + 6px blur) let the underlying journey card show through. Coloured shapes and blurred imagery behind the chat hurt legibility and created "is this part of the page or a separate surface?" ambiguity. Solid backdrop closes that.

Side win: removing `backdrop-filter: blur` drops a GPU-expensive effect and sidesteps Safari's backdrop-filter rendering quirks.

## Scope guard

- **Mobile `PinnedChatBar` is untouched.** It already renders with a fully opaque surface (`SURFACE = 'common.white'`) and continues to do so. Mobile UX was approved in QA for NES-1622; this PR adds no risk to that surface.
- Only `ChatOverlay` (rendered at `sm+` per `Conductor.tsx:284–319`) is affected.

## Deferred from this ticket

The related "input resting position / reflow feels off vs ChatGPT/Claude" feedback is **deferred** with rationale documented in NES-1654:

- The desktop chat already mirrors the ChatGPT macOS native pattern (input bottom-anchored).
- The remaining gap (ChatGPT-web-style empty-state-centred input that drops on first message) has a poor cost/benefit ratio for our overlay shape (`max-height: 80vh, max-width: 48rem, align-items: flex-end`).
- The opaque-backdrop fix may itself reduce the perceived "off" feel — we re-review after this lands.
- The ticket's acceptance criterion explicitly accepts a written deferral note in lieu of a behaviour change.

If the next preview review shows the M2 bar still feels off, a follow-up ticket with proper design exploration will be spawned.

## Acceptance criteria

- [x] Desktop `ChatOverlay` renders with a fully opaque dark backdrop — journey-card content behind the overlay is no longer visible.
- [x] Mobile `PinnedChatBar` behaviour is unchanged.
- [x] Input position/behaviour is unchanged in this ticket; deferral note documented in NES-1654.

## Testing

- **Vercel preview (this PR):** open a journey with `showAssistant: true` on the final card on a desktop browser at `sm+` width, open the chat, confirm the backdrop fully obscures the card behind.
- **Stage:** merge to `stage` for cross-device sanity check. Mobile (`xs`) should look identical to current main — that's the regression guard.

## Why no unit test

Per the frontend testing guideline (don't add tests for static style/config edits — they duplicate the diff and have no regression value), no `ChatOverlay.spec.tsx` is added. Existing `AiChat` and `Conductor` tests cover the behavioural surface this change does not touch.

## Summary by CodeRabbit

* **New Features**
  * Added hero element in chat overlay's empty state with animated wave effect
  * Enhanced overlay layout and spacing for optimal message positioning

* **Style**
  * Improved overlay background opacity and removed blur effect
  * Added styling token for overlay text color

* **Tests**
  * Added test suite for overlay empty-state hero visibility across various chat states

### DIFF
--- a/libs/journeys/ui/src/components/AiChat/AiChat.tsx
+++ b/libs/journeys/ui/src/components/AiChat/AiChat.tsx
@@ -406,7 +406,10 @@ export function AiChat({
   // Burmese, CJK) the split returns a single chunk and the whole
   // greeting wave-lifts as one unit — still on-tone.
   const heroWords = useMemo(
-    () => t('Ask me anything').split(/\s+/).filter((w) => w.length > 0),
+    () =>
+      t('Ask me anything')
+        .split(/\s+/)
+        .filter((w) => w.length > 0),
     [t]
   )
   // We keep header/conversation/input mounted in every state and rely on
--- a/libs/journeys/ui/src/components/AiChat/AiChat.tsx
+++ b/libs/journeys/ui/src/components/AiChat/AiChat.tsx
@@ -1,10 +1,9 @@
 'use client'
 
 import { useChat } from '@ai-sdk/react'
-import CloseRoundedIcon from '@mui/icons-material/CloseRounded'
 import Box from '@mui/material/Box'
 import Button from '@mui/material/Button'
-import IconButton from '@mui/material/IconButton'
+import Typography from '@mui/material/Typography'
 import { DefaultChatTransport, UIMessage } from 'ai'
 import { useTranslation } from 'next-i18next/pages'
 import {
@@ -29,9 +28,8 @@ import { ChatHeader } from './ChatHeader'
 import {
   HEADER_WASH,
   MUTED_FG,
-  OVERLAY_CLOSE_BG,
-  OVERLAY_CLOSE_BG_HOVER,
   OVERLAY_FG_RETRY,
+  OVERLAY_HERO_FG,
   SHEET_BOTTOM_FADE
 } from './chatStyles'
 import { DragHandle } from './DragHandle'
@@ -71,12 +69,6 @@ interface AiChatProps {
    * are ignored, drag interactions own the state from there.
    */
   initialCollapsed?: boolean
-  /**
-   * Optional close action for parent-owned overlay chrome. Rendered as a
-   * sibling of the floating input so it stays discoverable without covering
-   * typed text.
-   */
-  onClose?: () => void
 }
 
 export type AiChatSheetState = 'idle' | 'active' | 'collapsed'
@@ -230,8 +222,7 @@ export function AiChat({
   collapsible = true,
   variant = 'panel',
   onSheetStateChange,
-  initialCollapsed = false,
-  onClose
+  initialCollapsed = false
 }: AiChatProps): ReactElement {
   const isOverlay = variant === 'overlay'
   const isPanel = !isOverlay
@@ -386,7 +377,12 @@ export function AiChat({
 
   const showDragHandle = isPanel && collapsible
   const showHeader = isPanel
-  const showOverlayClose = isOverlay && onClose != null
+  // Empty-state hero is overlay-only: gives the user a clear "this is
+  // the chat" signal when the overlay auto-opens with no messages yet
+  // (NES-1654). Hidden once a message exists, is being sent, or there's
+  // an error so it doesn't compete with conversation content.
+  const showOverlayHero =
+    isOverlay && messages.length === 0 && !isLoading && error == null
   // We keep header/conversation/input mounted in every state and rely on
   // the parent sheet's height transition + overflow:hidden to clip them
   // as the sheet collapses. Hiding via display:none would short-circuit
@@ -418,6 +414,9 @@ export function AiChat({
 
       <Box
         sx={{
+          // `relative` anchors the overlay hero's absolute positioning
+          // to the conversation area rather than the viewport.
+          position: 'relative',
           display: 'flex',
           flex: 1,
           flexDirection: 'column',
@@ -427,6 +426,32 @@ export function AiChat({
           mx: 'auto'
         }}
       >
+        {showOverlayHero && (
+          <Box
+            aria-hidden
+            sx={{
+              position: 'absolute',
+              inset: 0,
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'center',
+              px: 4,
+              pointerEvents: 'none'
+            }}
+          >
+            <Typography
+              sx={{
+                fontSize: { xs: 22, sm: 26, md: 28 },
+                fontWeight: 500,
+                color: OVERLAY_HERO_FG,
+                textAlign: 'center',
+                lineHeight: 1.3
+              }}
+            >
+              {t('Ask me anything')}
+            </Typography>
+          </Box>
+        )}
         <Conversation
           scrollKey={messages.length}
           // 72px = floating capsule height (44px) + bottom offset (8px) +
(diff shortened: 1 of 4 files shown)


libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -418,15 +414,79 @@ export function AiChat({
...
+            <Typography
+              sx={{
+                fontSize: { xs: 22, sm: 26, md: 28 },
+                fontWeight: 500,
+                color: OVERLAY_HERO_FG,
+                textAlign: 'center',
+                lineHeight: 1.3
+              }}
+            >
+              {Array.from(t('Ask me anything')).map((char, i) => (
```
jaco-brink: **Concern:** `Array.from(t(...))` iterates Unicode code points and wraps each one in its own `display: inline-block` span. That breaks rendering for any script whose layout depends on neighbour relationships — and the translations for those scripts are already in the locale files:

- **Arabic** (`اسألني أي شيء`) — contextual letter shaping (initial/medial/final forms) and ligatures collapse to isolated forms → visually wrong Arabic.
- **Hindi/Devanagari** (`मुझसे कुछ भी पूछो`), **Bengali** (`আমাকে যেকোনো কিছু জিজ্ঞাসা করুন`), **Thai** (`ถามอะไรก็ได้`), **Burmese** (`ဘာမဆို မေးပါ`) — combining marks / matras get detached from their base consonants.

Mitigations, easiest first:
1. Split on whitespace and animate whole *words* — preserves all script behaviour, still gives a perceptible stagger for the typical 2–4 word translations.
2. Drop the per-char stagger entirely and animate the whole `Typography`.
3. Keep per-char only for known-safe scripts (Latin / CJK) and fall back to whole-string for the rest based on `journey.language.bcp47`.

Worth fixing before a journey in one of these languages picks this up.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -386,7 +377,12 @@ export function AiChat({
 
   const showDragHandle = isPanel && collapsible
   const showHeader = isPanel
-  const showOverlayClose = isOverlay && onClose != null
+  // Empty-state hero is overlay-only: gives the user a clear "this is
+  // the chat" signal when the overlay auto-opens with no messages yet
+  // (NES-1654). Hidden once a message exists, is being sent, or there's
+  // an error so it doesn't compete with conversation content.
+  const showOverlayHero =
+    isOverlay && messages.length === 0 && !isLoading && error == null
```
jaco-brink, line 385: **Concern:** PR description cites the frontend testing guideline (don't test static style/config edits) as cover for not adding a spec — but `showOverlayHero = isOverlay && messages.length === 0 && !isLoading && error == null` is **branching logic**, not a copy/style edit. That puts it squarely in the rule's own exception: _"add a test when the config drives branching logic, has a non-obvious transformation, or has caused real prior regressions."_

A short spec asserting:
- hero shows when overlay variant + empty messages + not loading + no error
- hero hides when `status === 'submitted'` / `'streaming'`
- hero hides once a message is present
- hero hides when `error != null`

…would guard the four-way condition without testing the wave styling itself. The existing `AiChat.spec.tsx` already has the `<AiChat variant="overlay" />` test scaffolding (`setChatState({ ... })`) — adding an `empty-state hero` describe block should be a few lines.

libs/journeys/ui/src/components/AiChat/chatStyles.ts
```diff
@@ -85,6 +85,9 @@ export const OVERLAY_CLOSE_BG = 'rgba(197, 45, 58, 0.72)'
 /** Hover state for the muted brand-red close control. */
 export const OVERLAY_CLOSE_BG_HOVER = 'rgba(197, 45, 58, 0.86)'
 
+/** Empty-state hero text colour on the dark overlay backdrop. */
+export const OVERLAY_HERO_FG = 'rgba(255, 255, 255, 0.85)'
```
jaco-brink: **Nit:** `OVERLAY_HERO_FG` is added inside the `--- PromptInput, overlay (floating) variant ---` section, but it's hero text — not an input style. The `--- Misc chat surfaces ---` block lower in the file (where `PLAIN_ASSISTANT_FG_ON_DARK` and `OVERLAY_FG_RETRY` live) is the more natural home for it. Pure organisation, no functional impact.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -418,15 +414,79 @@ export function AiChat({
...
+              justifyContent: 'center',
+              px: 4,
+              pointerEvents: 'none',
+              // One-shot wave animation on mount: each character lifts
+              // 8px then settles, staggered left-to-right. Draws the eye
+              // to the hero when the chat opens, then stays out of the
+              // way. The base (0%/100%) is translateY(0) so each char's
+              // resting state is its natural position — no fill-mode
+              // needed.
+              '@keyframes aiChatHeroWave': {
```
jaco-brink, line 493: **Nit:** The wave re-fires every time `showOverlayHero` flips false → true. The cap-hit `Start a new conversation` path runs `setMessages([])` + `clearError()`, so the hero remounts and the wave plays again. On repeated resets it can read as busy.

If you want it to feel deliberate ("plays once when the chat first opens, ever"), a small mount-scoped flag would do it — e.g. `const heroAnimatedRef = useRef(false)` set on first render where `showOverlayHero` is true, then gate the `animation` value to `'none'` on subsequent showings. Defensible to leave as-is too.

jaco-brink (commented):
## Review summary

**Concern (2):** per-character hero animation breaks complex scripts (Arabic/Hindi/Thai/Burmese/Bengali); new hero gating branch has no test coverage.
**Nit (2):** `OVERLAY_HERO_FG` placement in `chatStyles.ts`; wave re-fires on conversation reset.

**Overall:** Architecturally clean — moving the close button + `onClose` plumbing out of `AiChat` into `ChatOverlay` is the right call, all references are cleaned up, translation keys exist across all 22 locales, and `prefers-reduced-motion` is honoured. The top-inset math is well-commented and incidentally fixes an iPad safe-area issue. Main thing to address is the i18n regression from per-codepoint splitting before this lands behind a behaviour where users see Arabic/Hindi/etc.

Draft PR → submitted as COMMENT (no auto-request-changes without your call).

jaco-brink:
## Review feedback addressed (f4bce6b05)

All four self-review threads fixed in this push.

**Fixed:**
- `AiChat.tsx:470` (Unicode/script rendering) — Wave now staggers by **whole words** instead of code points. `Array.from(t(...))` was code-point-splitting the localised string and wrapping each char in `display: inline-block`, which would have visually broken Arabic shaping and Devanagari/Bengali/Thai/Burmese/Urdu/Nepali combining-mark layout. Translations for all of these scripts are confirmed present under `libs/locales/<bcp47>/libs-journeys-ui.json`, so this was a real (not theoretical) latent regression. Each word now stays in one text run; whitespace-less scripts (Thai/Burmese/CJK) cleanly fall out as a single span that wave-lifts as one unit.
- `AiChat.tsx:385` (missing spec) — Added an `overlay empty-state hero (NES-1654)` describe block to `AiChat.spec.tsx` covering the four inputs of `showOverlayHero` (overlay variant + empty messages + not loading + no error) plus the panel-variant negative case. Six new tests, all passing; existing 7 still pass. Added `data-testid="overlay-hero"` on the hero `<Box>` as a stable query target — the Box was already `aria-hidden` so no a11y impact.
- `AiChat.tsx:455` (wave re-fires on reset) — Added a `heroAnimatedRef` (`useRef(false)`) + `useEffect` gate so the wave is genuinely one-shot per AiChat mount. `handleStartNewConversation` (cap-hit reset) no longer re-fires the animation when `showOverlayHero` flips false → true; second and subsequent showings drop to `animation: 'none'`. Reduced-motion media query unchanged.
- `chatStyles.ts:89` (file organisation) — Moved `OVERLAY_HERO_FG` out of the `--- PromptInput, overlay (floating) variant ---` section and into `--- Misc chat surfaces ---`, sitting next to `PLAIN_ASSISTANT_FG_ON_DARK` and `OVERLAY_FG_RETRY` where the other dark-surface foreground colours live. No behaviour change.

**Verification:** `npx vitest run --config libs/journeys/ui/vitest.config.mts libs/journeys/ui/src/components/AiChat/AiChat.spec.tsx --coverage=false` → 13/13 pass.

jianwei1 (approved):

PR #9279, opened 2026-06-05
feat(NES-1588): about-this-chat notice page + chat-surface link

jaco-brink:
Closes [NES-1588](https://linear.app/jesus-film-project/issue/NES-1588/launch-privacy-notice-copy-on-chat-surface).

## What this does

Adds the user-facing data-disclosure surface that NES-1588 has been blocking the public chat launch on.

- **New page** at `/legal/about-chat` (`apps/journeys/pages/legal/about-chat.tsx`) that explains:
  - the chat is powered by an AI assistant
  - messages and replies are stored in a third-party analytics service (Langfuse) so the team can review conversations in aggregate
  - no name / email / account identifier is stored, conversations are not linked to identity
  - the only opt-out today is not using the chat
- **Inline link from `ChatHeader.tsx`** — appended to the existing subtitle:
  > Replies may not be perfect · **About this chat**
  
  Renders identically on mobile (PinnedChatBar) and desktop (ChatOverlay) because both share `ChatHeader`. Opens in a new tab so it never navigates the user away from the journey, including inside iframe embeds.

## Copy decisions worth knowing

- Surface microcopy intentionally avoids leading with "AI". The PRD already plans for the assistant to be honest when asked ("are you a bot?" → yes), so the surface doesn't need to confront the user with it before they've typed. *"Replies may not be perfect"* was already shipping; we only appended the link.
- Link label, page H1, and `<title>` are all *"About this chat"* — one canonical string, consistent across the click.
- No reference to the Cru privacy policy on the new page. The Cru policy doesn't disclose Langfuse chat-content storage, and folding it in would create the impression that it does. The page stands on its own.
- `whiteSpace: nowrap` on the link keeps *"About this chat"* tokenised so longer translations don't break mid-word. The surrounding `Typography` still wraps the whole subtitle to a second line when the combined string overflows — important for languages like German.

## Locale changes

- `libs/locales/en/apps-journeys.json` — 10 new English keys (page H1, `<title>`, three section headings, three body paragraphs, last-updated, back-link).
- `libs/locales/en/libs-journeys-ui.json` — one new key, `"About this chat"`, used by `ChatHeader`.
- Crowdin will pick up both on the next sync; no manual translation work in this PR.

## What I deliberately did NOT do

- No unit test for the page — `.claude/rules/frontend/apps.md` explicitly says don't unit-test copy strings.
- No edits to the Cru-linked footer in `pages/home/index.tsx` — out of scope.
- No new opt-out mechanism — page is honest that one doesn't exist yet.

## How to test on the Vercel preview

1. Open `<preview-host>/legal/about-chat` — verify the page renders with the Next Steps logo, sections, and back link.
2. Open any journey with `showAssistant = true` (or flip the LD flag) — verify the chat header subtitle reads *"Replies may not be perfect · About this chat"* and that clicking the link opens the page in a new tab.
3. Resize the chat to the mobile breakpoint and a narrow width to confirm the subtitle wraps gracefully (link wraps as a whole token to the next line if the combined string overflows).
4. Check in an in-app webview (Slack / Messenger / FB) — link should still open the page; iOS in-app webviews may render it in the same webview rather than a new tab, which is acceptable.

## Summary by CodeRabbit

* **New Features**
  * Added an "About this chat" legal page reachable from the chat UI and via hostname-aware routing.
  * Added an external "About this chat" link in the chat header and floating overlay input, plus a short "Replies may not be perfect" disclaimer.

* **UI**
  * Adjusted floating input layout and link styling for overlay vs panel surfaces.

* **Localization**
  * Added English copy for the new legal content and UI labels.

* **Tests**
  * Added/updated tests validating link presence and safe external-link attributes.


libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -662,17 +666,60 @@ export function AiChat({
...
-          <PromptInput
-            input={input}
-            onInputChange={setInput}
-            onSubmit={handleSubmit}
-            isLoading={isLoading}
-            onStop={stop}
-            disabled={isConversationCapped}
-            variant={isOverlay ? 'floating' : 'inline'}
-          />
+        <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
```
jaco-brink: **Concern (minor):** After the refactor to a vertical column, this `<Box sx={{ display: flex, alignItems: center, gap: 1 }}>` now wraps a single child (the `flex: 1, minWidth: 0` box → `PromptInput`). With one child, `alignItems` and `gap` are inert — leftover scaffolding from the row→column change. Can be flattened to a single Box to cut the extra nesting. Harmless, just noise.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -662,17 +666,60 @@ export function AiChat({
...
+              fontSize: 12,
+              lineHeight: '18px',
+              textAlign: 'center',
+              px: 1
+            }}
+          >
+            {t('Replies may not be perfect')}
+            {' · '}
+            <Link
+              href="/legal/about-chat"
```
jaco-brink, line 699: **Nit (minor):** The overlay-mode disclosure link has no test, while its panel-mode twin in `ChatHeader` now asserts href/target/rel. Optional — a cheap `variant="overlay"` render assertion here would lock in both the link attributes and the non-duplication guarantee (header link suppressed in overlay).

jaco-brink (commented):
## Review summary

Clean, well-documented PR — no Critical issues. I verified the two things most likely to break and both are correct:

- **Routing:** `proxy.ts` rewrites root-domain `/legal/about-chat` → `/home/legal/about-chat` and custom-host → `/[hostname]/legal/about-chat`, so both page files are required (not redundant), and the `[hostname]` variant DRY-re-exports the component. ✅
- **No link duplication:** `showHeader = isPanel = !isOverlay` cleanly gates the header link (panel) vs the caption link (overlay) — only one renders per mode. ✅

**Other positives:** `rel="noopener noreferrer"` on every `target="_blank"`; i18n keys in the correct namespaces with matching curly apostrophes; `ChatHeader.spec` strengthened to assert href/target/rel.

**Minor (non-blocking):** 2 inline notes below.

**Doc nit:** the PR description lists the page path as `apps/journeys/pages/legal/about-chat.tsx`, but the files are actually `pages/home/legal/about-chat.tsx` + `pages/[hostname]/legal/about-chat.tsx` — worth updating so reviewers find them.

(Draft PR → submitted as a comment, not an approval/request-changes.)

jaco-brink:
## Review feedback addressed (e4110006b)

**Fixed:**
- `AiChat.tsx:669` — Flattened the inert `display:flex; alignItems:center; gap:1` wrapper (plus its `flex:1; minWidth:0` child) that wrapped a single element after the row→column refactor. `PromptInput`'s root is already a `width:100%` form and the column container stretches its children, so both wrappers were dead scaffolding — `PromptInput` now renders directly. No layout change. (`eec2cd276`)
- `AiChat.tsx:703` — Added overlay-mode coverage for the about-this-chat disclosure link: a `variant="overlay"` render asserts `href`/`target`/`rel`, and both variants assert exactly one link renders — locking in the no-duplication invariant (header link suppressed in overlay, footer caption suppressed in panel). (`e4110006b`)

Both threads resolved. `AiChat.spec.tsx` passes (15/15); eslint clean on both files.

apps/journeys/pages/home/legal/about-chat.tsx
```diff
@@ -0,0 +1,106 @@
...
+            </Stack>
+            <Stack direction="row" spacing={2} justifyContent="center">
+              <Link
+                component={NextLink}
+                href="/"
+                underline="none"
+                variant="body2"
+              >
+                {t('Back to Next Steps')}
+              </Link>
```
jianwei1: **Concern:** The "Back to Next Steps" link uses `href="/"`. With `proxy.ts`:

- On the root domain, `/` → `/home` → the Next Steps landing. Label matches.
- On a custom hostname (e.g. `myjourney.com/legal/about-chat`), `/` → `/{hostname}` → `pages/[hostname]/index.tsx`, which resolves to the host's journey (or its journey list) — **not** the Next Steps landing.

Because the page renders with Next Steps branding (hardcoded logo + copy) regardless of how the user arrived, sending custom-host visitors to the custom-host root makes the label confusing: they land on a page that doesn't look like "Next Steps" at all.

Two reasonable options:
1. Drop the back link — the disclosure opens in a new tab via `target="_blank"`, so closing it is the natural return path and is host-agnostic.
2. Make the destination explicit (e.g. point at the configured root domain) so the label and behaviour agree on every host.

---

apps/journeys/pages/[hostname]/legal/about-chat.tsx
```diff
@@ -0,0 +1,26 @@
+import { GetStaticPaths, GetStaticProps } from 'next'
+import { serverSideTranslations } from 'next-i18next/pages/serverSideTranslations'
+
+import i18nConfig from '../../../next-i18next.config'
+
+// The page is hostname-agnostic — same static content under both the
+// default preview host (routed to `pages/home/legal/about-chat.tsx`)
+// and any custom hostname (routed here via Vercel's hostname rewrite).
+// We re-export the component and provide a getStaticProps that only
+// loads translations.
```
Comment (jianwei1) [inline apps/journeys/pages/[hostname]/legal/about-chat.tsx:11]: **Nit:** The comment attributes the rewrite to "Vercel's hostname rewrite", but the rewrite actually lives in `apps/journeys/proxy.ts` (a Next.js middleware), not Vercel routing. Worth correcting so the next reader who greps `vercel.json` isn't sent down the wrong path.

---

apps/journeys/pages/home/legal/about-chat.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { ThemeMode, ThemeName } from '../../../__generated__/globalTypes'
+import i18nConfig from '../../../next-i18next.config'
+import logo from '../../../public/logo.svg'
+
+function AboutChatPage(): ReactElement {
+  const { t } = useTranslation('apps-journeys')
+
+  return (
+    <>
+      <NextSeo title={t('About this chat | Next Steps')} nofollow noindex />
```
jianwei1: **Nit:** The translation key `'About this chat | Next Steps'` bakes the site-name separator into the key itself. Translators who want to swap the order (e.g. `{site} | {title}`, common in CJK styling) or use a different separator can't do so without forking the key. Not worth changing on its own — flagging only.

---

libs/journeys/ui/src/components/AiChat/chatStyles.ts
```diff
@@ -43,6 +43,26 @@ export const MUTED_FG = 'text.secondary'
...
+/**
+ * Link colour for caption-level "about/legal" links on the white
+ * panel/sheet surface (mobile chat sheet header).
+ *
+ * Concrete hex rather than a `'primary.main'` theme path: the chat
+ * inherits the journey card's theme, and the base dark palette
+ * resolves `primary.main` to near-white (#FEFEFE) — invisible against
+ * the panel's white background.
+ */
+export const PANEL_LINK_FG = brandRed
```
jianwei1, line 55: **Nit:** `PANEL_LINK_FG = brandRed` re-aliases an imported constant under a new export — the doc comment carries the substance, the indirection is only there to mirror `OVERLAY_LINK_FG`. Fine as-is; only worth tidying if more `*_LINK_FG` siblings appear later and the indirection stops paying for itself.

---

jianwei1 (commented):
## Review summary

**Concerns:** 1 — "Back to Next Steps" link behaviour on custom hostnames (see inline).
**Nits:** 3 — comment wording, translation-key shape, and a re-alias in `chatStyles.ts` (see inline).

Overall: clean change. The overlay refactor + tests lock in the no-duplication invariant across panel/overlay surfaces well, and the routing pattern matches the existing `proxy.ts` hostname convention. Locale parity for non-English will be handled in a follow-up per author confirmation.

---

libs/journeys/ui/src/components/AiChat/chatStyles.ts
jaco-brink: Agreed it's a thin indirection, but keeping it: `PANEL_LINK_FG` mirrors `OVERLAY_LINK_FG` so both surfaces read symmetrically at the call sites (`color: PANEL_LINK_FG` / `color: OVERLAY_LINK_FG`), and the doc comment is the real payload — it captures *why* we can't use `primary.main` here (the base dark palette resolves it to near-white against the panel's white background). Leaving as-is per your "fine as-is" read; will revisit if more `*_LINK_FG` siblings make the pattern earn a different shape.

jaco-brink:
## Review feedback addressed — round 2 (2685f5661)

Thanks @jianwei1 — four threads triaged:

**Fixed:**
- `home/legal/about-chat.tsx:87` — Dropped the "Back to Next Steps" link. `href="/"` is misleading on custom hosts (`proxy.ts` rewrites `/` → `/{hostname}` → the host's journey, not the Next-Steps landing the label promises). The page is always opened in a new tab (`target="_blank"` from both ChatHeader and the overlay link), so closing the tab is the natural, host-agnostic return — the link was redundant *and* wrong off the root domain. (`0096f6bf6`)
- `[hostname]/legal/about-chat.tsx:10` — Corrected the comment: the hostname rewrite is the `apps/journeys/proxy.ts` Next.js middleware, not Vercel routing. (`2685f5661`)

**Fixed (adjusted):**
- `home/legal/about-chat.tsx:24` — Your separator nit surfaced a real bug: `_app.tsx` already sets `titleTemplate="%s | Next Steps"`, so `title="About this chat | Next Steps"` rendered a doubled `… | Next Steps | Next Steps`. Changed to the bare `t('About this chat')` and let the central template append the suffix — which also fixes your underlying concern (separator/order now controlled in one place, not forked per key). (`0096f6bf6`)

**Skipped (kept as-is, per your read):**
- `chatStyles.ts:55` — `PANEL_LINK_FG = brandRed`: keeping the documented alias; it mirrors `OVERLAY_LINK_FG` and the doc comment captures why `primary.main` won't work here. See inline reply.

ESLint clean on the changed pages (0 errors). The three fixed threads are resolved.

jianwei1 (approved):
LGTM. All prior feedback addressed cleanly — the `titleTemplate` fix is a better answer than the original suggestion, and dropping the back link entirely sidesteps the host/branding mismatch. Tests, routing, and colour-token documentation are all in good shape.

---

PR #9283, opened 2026-06-08
fix(journeys): restore Preview button on public gallery cards (NES-1720)

csiyang:
## Problem

The Preview button was missing from every card on the public template-gallery page (`/template-gallery/[slug]`) — in both the **Explore** and **More** sections — while the sibling **Use** button still rendered. It was also missing from the admin collection-edit preview.

Fixes [NES-1720](https://linear.app/jesus-film-project/issue/NES-1720/preview-button-not-showing-up-on-public-gallery-page).

## Root cause

A regression from the NES-1694 restyle (#9250). That PR moved the card actions into the new shared `JourneyViewCardActions` component but **only carried over the "Use" button — the Preview button was dropped.** Both gallery sections (`FeaturedRow` = Explore, `JourneyViewCard` = More) and the admin preview render this one component, so the button disappeared everywhere at once.

Telltale signs it was meant to be there: the i18n keys `"Preview"` / `"Preview {{title}}"` and the `GALLERY_ACTION_SIZE` token comment ("Use / Preview action controls") were already present — just never wired to a button.

## Changes

- **`JourneyViewCardActions.tsx`** — restore the Preview `IconButton` (`Play3` icon) beside Use, opening the public viewer route `/<slug>` (`target=_blank`, `rel="noopener noreferrer"`, `aria-label="Preview {{title}}"`, `data-testid="GalleryTemplateCardPreviewButton"`). Restored in both the live and `decorative` (admin preview) branches. `flexShrink: 0` so the icon button can't be squeezed when Use stretches.
- **`JourneyViewCard.tsx`** (panel + overlay) and **`FeaturedRow.tsx`** — thread `itemSlug={item.slug}` through. No data-shape change: `slug` was already on `PublicGalleryPageItem`.
- **`PublicGalleryPage.spec.tsx`** — assert the Preview viewer link (`/template-0`), its `target`/`rel`, and Preview/Use count parity across Explore + More.

## Styling

Both card actions use the dark "almost black" token (`#26262E`, the base theme's `palette[900]`): **Use** is outlined dark, **Preview** is filled dark (hover to black) — matching the pre-restyle design. The brand-red accent remains on the eyebrows, section nav, and headers.

Exposed as a shared `GALLERY_ACTION_COLOR` token in `galleryTokens.ts` with a sync-note to the base-theme source. Hardcoded (rather than read from the theme) because the public gallery renders under the **website theme**, which doesn't expose that palette slot — the base `palette` object isn't exported, and journeysAdmin only surfaces it via its own MUI theme. This matches existing practice (yoga email components, `RadioOption/GridVariant`, and the pre-restyle gallery card all hardcode this hex).

## Responsive Use button (Explore section)

The Use button fills the row whenever the featured journeys are **stacked** — mobile (below `md`) on the live page, and always in the admin preview pane — and keeps its fixed 180px width only when the Explore rows sit side-by-side at `md`+. Implemented as a `'responsive'` option on `JourneyViewCardActions`' `fullWidth`, driven by `sx` breakpoints on the live page; the admin preview deliberately avoids viewport breakpoints (it's a ~287px pane inside a wide viewport) and is treated as stacked. More-grid cards are unchanged (always full width).

## Testing

`npx vitest run --config libs/journeys/ui/vitest.config.mts 'libs/journeys/ui/src/components/PublicGalleryPage/PublicGalleryPage.spec.tsx' --coverage=false` → **32/32 pass**. Responsive widths aren't assertable in jsdom (no layout) — covered by manual QA on the Vercel preview.


mikeallisonJS:
## Code Review (NES-1720)

A focused, well-documented regression fix that restores the dropped Preview button and adds a `'responsive'` width mode to `JourneyViewCardActions`. The change is small, well-tested (32/32 plus new assertions for the Preview link/target/rel and Use↔Preview parity), and the security basics on the new external link are correct. Two things are worth a look before merge: the failing `danger` CI check, and an unvalidated slug→href construction.

### Suggestions

| # | File | Issue | Category |
|---|------|-------|----------|
| 1 | CI | The `danger` check is **failing** (both runs) and `mergeable_state` is `blocked`. Everything else is green — likely a changelog/label/convention gate rather than code, but resolve before merge. | Process |
| 2 | `JourneyViewCardActions.tsx` | `previewHref` is built as `` `/${itemSlug}` `` with no guard on `itemSlug`. If it's ever empty the link silently points at `/` (homepage). More importantly, a slug starting with `/` (e.g. `//evil.com`) interpolates to a protocol-relative URL → off-site navigation. Slugs are server-controlled so risk is low, but a cheap `encodeURIComponent` / leading-slash strip makes it robust. | Security / Correctness |
| 3 | `galleryTokens.ts` | `GALLERY_ACTION_COLOR = '#26262E'` is a hardcoded hex (plus inline `#FFFFFF` / `#000000`) with a comment asking devs to manually keep it in sync with `palette[900]`. Pragmatic and explained, but a manual-sync token is fragile — consider a follow-up to expose the palette slot. | Maintainability |
| 4 | PR description | The "Styling note" says the Preview button is "brand-accent **(red)**", but the code fills it with `GALLERY_ACTION_COLOR` = `#26262E` (dark), hover `#000000`. Description contradicts the shipped color — worth updating so designers aren't misled. | Docs |

### What looks good
- **Link safety is correct**: the Preview `IconButton` uses `target="_blank"` + `rel="noopener noreferrer"`, and the new test asserts both.
- **Accessibility**: per-card `aria-label={t('Preview {{title}}', ...)}` disambiguates otherwise-identical controls; the decorative preview stays inside an `aria-hidden` wrapper.
- **Tests match intent**: asserting Preview count === Use count directly encodes the regression rather than a brittle fixed number.
- **`GALLERY_ACCENT` is not dead code** — still used in `AdminView`, `JourneyViewHeader`, `JourneyView`, `JourneyViewNav`, and the `FeaturedRow` header; removing it from the card-action call sites is correct.
- **`flexShrink: 0`** on the preview square is the right fix for the squeeze, and the `'responsive'` vs `true` prop docs (and why viewport breakpoints are avoided in the preview pane) are genuinely useful.

### Verdict
**Approve with minor changes** — sort out the failing `danger` check (#1) and add a small guard/encode on the slug→href (#2). The rest is non-blocking polish.

🤖 Review generated with Claude

csiyang:
Thanks for the review @mikeallisonJS! Disposition of the four points:

**1. Failing `danger` check** — resolved. The failures pre-dated the branch being updated with `main`; the latest Danger run on `5217030c7` passes and all checks are now green.

**2. Slug→href guard** — skipping this one deliberately. Slugs are validated and persisted server-side when the journey is edited, so by the time they reach this component they're already constrained — there's no realistic path for an empty or `//host`-shaped slug to arrive here. Happy to revisit if the slug source ever broadens.

**3. Hardcoded `GALLERY_ACTION_COLOR` hex** — double-checked, and the hardcode is unavoidable today: the base theme's `palette` object isn't exported (`brandRed` is the only exported raw colour token), and journeysAdmin only exposes `#26262E` as `palette.solid` through its own MUI theme — which the public gallery never sees, since it renders under the website theme. Hardcoding with a sync-note matches existing practice (yoga email components, `RadioOption/GridVariant`, and the pre-restyle gallery card all hardcode this same hex). Exporting the palette slot properly is a fair follow-up, but out of scope here.

**4. Stale "brand-accent (red)" styling note** — fixed. The PR description has been rewritten to describe the shipped dark (`#26262E`) treatment on both buttons, plus the responsive Use-button width behaviour added after the original description was written.

mikeallisonJS (approved):

PR #9286, opened 2026-06-08
feat(NES-1679): enforce per-card showAssistant on /api/chat (hard kill switch)

jaco-brink:
## NES-1679 — Backend: enforce per-card `showAssistant` on `/api/chat` (hard kill switch)

[Linear NES-1679](https://linear.app/jesus-film-project/issue/NES-1679) · World Cup Readiness

Makes `card.showAssistant` a **server-enforced** gate. Previously it was render-only: `/api/chat` never consulted it, so a viewer whose tab was already open kept chatting indefinitely after a creator flipped the card off. Now flipping it off stops chat for already-open tabs on their **next message** — the per-journey emergency kill switch Jaco asked for.

### What changed

**Server (`apps/journeys`)**
- `chatRequestSchema` now **requires `cardId`** (`z.string().min(1).max(128)`).
- New `src/libs/getCardChatEnabled` — queries the journey by `journeyId`, finds the `CardBlock` by `cardId`, returns `showAssistant === true`. Looking the card up *within* its journey also enforces card↔journey ownership (a stale tab can't keep chatting by spoofing a different `journeyId`).
- `/api/chat` calls it after the cheap local schema/size checks and **403s** (`{ error: 'chat disabled for this card', code: 'chat_disabled' }`) when the card is missing or `showAssistant !== true`. Logs a non-PII `chat_card_disabled` warn.

**Client (`libs/journeys/ui`)**
- `AiChat` sends the active `cardId` on every POST, read from the shared block history (`useBlocks()` + `getCardChild`) — the same source `Conductor`/`StepFooter` use. This covers **both** the mobile `PinnedChatBar` and the desktop `ChatOverlay` (both render `AiChat`) without threading a prop through each surface.
- `chat_disabled` is added to the non-retriable error codes, so the client hides Retry (re-firing would 403 again) — same treatment as the existing `not_found` flag-off case.

### Design decisions worth your eyes 👀

1. **Fail-open on infrastructure errors, fail-closed on definitive negatives.** The lookup runs on *every message*. A definitive "card off / not found / missing journeyId" → 403 (kill switch works). But a *gateway/network error* → allow + log (`chat_card_lookup_error`), so a transient gateway blip doesn't take chat down on every journey at once. The kill switch still works in every realistic scenario (flipping the toggle requires a working gateway). **This slightly weakens "hard" during a gateway outage — flag if you'd rather fail fully closed.**
2. **`fetchPolicy: 'no-cache'`** on the lookup — an in-memory Apollo cache would mask the creator's change and defeat the switch. This means **one gateway read per chat message**; the ticket's TTL-cache idea is left as a noted follow-up ("keep it simple unless load testing says otherwise").
3. **Heavy query.** Re-uses the existing `GET_JOURNEY` (full journey + all blocks) since there's no lighter single-card query exposed and the API route can only reach the gateway via Apollo (no Prisma from `apps/journeys`). A dedicated minimal `card(id){ showAssistant }` query would be a good perf follow-up.

### Deviations from the ticket (intentional)
- **Tests are Vitest, not Jest** — `apps/journeys` migrated to Vitest; the ticket said "jest" but I followed the actual runner.
- **Client `cardId` lives in `AiChat`**, not `AiChatButton`/`PinnedChatBar` as the ticket suggested — `AiChat` owns the transport and both surfaces render it, so this is one change instead of three and can't drift between surfaces.

### Tests
- `getCardChatEnabled.spec.ts` (8): enabled / `false` / `null` / card-not-in-journey / missing-journeyId / journey-not-found (closed) / infra-error (open) / no-cache+databaseId variables.
- `index.spec.ts` (+ kill-switch suite, 47 total): 400 when `cardId` missing, 403 + `chat_disabled` when disabled, lookup gets `{journeyId, cardId}` and streams when enabled, flag-off short-circuits before the lookup. Existing 43 tests updated to send `cardId`.
- `AiChat.spec.tsx` (18): request body includes the active `cardId`; undefined when no card active; `chat_disabled` hides Retry.
- No regressions: `AiChatButton` (5) and `Conductor` (62, incl. per-card-chat) green. `prettier` ✓, `eslint` ✓, `tsc -b` ✓.

### Manual verification (the "Done when")
1. Open a chat-enabled card in the viewer.
2. In the editor (another tab), flip `showAssistant` **off** on that card.
3. Send another message in the open viewer tab → it fails closed with a visible error and no Retry, while the global flag stays on for every other journey.

### Not in scope (per ticket)
Live "chat closed" push to stale tabs (needs SSE/WebSocket), removing `Journey.showAssistant` (NES-1624), per-region kill switch, mid-stream suppression of an in-flight response.

## Summary by CodeRabbit

* **New Features**
  * Per-card chat control: chat can be enabled/disabled per card; when disabled users see a clear “Chat has been turned off” message, cannot stream, and retry is hidden.
  * Chat requests now include the active card identifier so card-specific state is respected.

* **Bug Fixes**
  * Request validation tightened to require journey and card identifiers; returns specific errors (400/403/404) when missing or disabled.

* **Tests**
  * Added coverage for per-card kill switch, logging/trace adjustments, and updated request payloads.

### DIFF
--- a/apps/journeys/pages/api/chat/index.ts
+++ b/apps/journeys/pages/api/chat/index.ts
@@ -11,6 +11,7 @@ import { Langfuse, TextPromptClient } from 'langfuse'
 import type { NextApiRequest, NextApiResponse } from 'next'
 import { z } from 'zod'
 
+import { getCardChatEnabled } from '../../../src/libs/getCardChatEnabled'
 import { getFlags } from '../../../src/libs/getFlags'
 import {
   APOLOGIST_PROMPT_NAME,
@@ -74,7 +75,10 @@ const chatRequestSchema = z.object({
   messages: z.array(messageSchema).min(1).max(MAX_MESSAGES),
   language: z.string().max(64).optional(),
   sessionId: z.string().max(128).optional(),
-  journeyId: z.string().max(128).optional()
+  journeyId: z.string().max(128).optional(),
+  // Required (NES-1679): the server reads the card's `showAssistant` to enforce
+  // the per-card chat kill switch, so every request must say which card it is.
+  cardId: z.string().min(1).max(128)
 })
 
 type ParsedChatMessage = z.infer<typeof messageSchema>
@@ -194,7 +198,7 @@ export default async function handler(
     res.status(400).json({ error: 'invalid request', code: 'invalid_request' })
     return
   }
-  const { messages, language, sessionId, journeyId } = parsed.data
+  const { messages, language, sessionId, journeyId, cardId } = parsed.data
 
   const promptChars = totalMessageChars(messages)
   if (promptChars > MAX_TOTAL_CHARS) {
@@ -222,6 +226,24 @@ export default async function handler(
     res
       .status(400)
       .json({ error: 'request too large', code: 'conversation_capped' })
+    return
+  }
+
+  // Per-card kill switch (NES-1679): the card's `showAssistant` is enforced
+  // server-side so flipping it off in the editor stops chat for tabs that are
+  // already open, on their very next message. Runs after the cheap local
+  // schema/size checks so an obviously-bad request never reaches the gateway.
+  const chatEnabled = await getCardChatEnabled({ journeyId, cardId })
+  if (!chatEnabled) {
+    logger.warn(
+      { event: 'chat_card_disabled', journeyId, cardId, sessionId },
+      '[chat] blocked: chat disabled for card'
+    )
+    // `code` keeps the client's error-code contract: a disabled card is
+    // deterministic, so the client hides Retry (re-firing would 403 again).
+    res
+      .status(403)
+      .json({ error: 'chat disabled for this card', code: 'chat_disabled' })
     return
   }
 
--- a/apps/journeys/pages/api/chat/index.ts
+++ b/apps/journeys/pages/api/chat/index.ts
@@ -270,6 +270,12 @@ export default async function handler(
   // or the raw IP.
   const chatLogContext = {
     journeyId,
+    // `cardId` is recorded on every chat event (NES-1679) so the kill switch's
+    // allow path is queryable: on a "killed card still chatting" report the
+    // success/error events show which card the server actually consulted, which
+    // is the signal needed to tell a request-identity mismatch apart from a
+    // stale read — without a per-message log on the lookup itself.
+    cardId,
     language,
     ipCountry,
     sessionId,
--- a/apps/journeys/pages/api/chat/index.ts
+++ b/apps/journeys/pages/api/chat/index.ts
@@ -75,7 +75,12 @@ const chatRequestSchema = z.object({
   messages: z.array(messageSchema).min(1).max(MAX_MESSAGES),
   language: z.string().max(64).optional(),
   sessionId: z.string().max(128).optional(),
-  journeyId: z.string().max(128).optional(),
+  // Required (NES-1679): a missing journeyId is a malformed request, not a
+  // killed card. Requiring it returns 400 invalid_request instead of routing
+  // through the kill switch to a 403 chat_disabled (which would show users the
+  // "chat turned off" copy and pollute the chat_card_disabled signal). Matches
+  // the required cardId below.
+  journeyId: z.string().min(1).max(128),
   // Required (NES-1679): the server reads the card's `showAssistant` to enforce
   // the per-card chat kill switch, so every request must say which card it is.
   cardId: z.string().min(1).max(128)
--- a/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
+++ b/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
@@ -1,6 +1,9 @@
 import { IdType } from '../../../__generated__/globalTypes'
 
-import { CARD_LOOKUP_TIMEOUT_MS, getCardChatEnabled } from './getCardChatEnabled'
+import {
+  CARD_LOOKUP_TIMEOUT_MS,
+  getCardChatEnabled
+} from './getCardChatEnabled'
 
 const { mockQuery, mockLoggerWarn } = vi.hoisted(() => ({
   mockQuery: vi.fn(),
--- a/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
+++ b/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
@@ -1,6 +1,6 @@
 import { IdType } from '../../../__generated__/globalTypes'
 
-import { getCardChatEnabled } from './getCardChatEnabled'
+import { CARD_LOOKUP_TIMEOUT_MS, getCardChatEnabled } from './getCardChatEnabled'
 
 const { mockQuery, mockLoggerWarn } = vi.hoisted(() => ({
   mockQuery: vi.fn(),
@@ -97,6 +97,15 @@ describe('getCardChatEnabled', () => {
     expect(mockQuery).not.toHaveBeenCalled()
   })
 
+  it('returns false when the journey response is null (non-throwing not-found)', async () => {
+    mockQuery.mockResolvedValue({ data: { journey: null } })
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(false)
+    expect(mockLoggerWarn).not.toHaveBeenCalled()
+  })
+
   it('returns false when the journey is not found (definitive negative)', async () => {
     mockQuery.mockRejectedValue(new Error('journey not found'))
 
@@ -116,5 +125,33 @@ describe('getCardChatEnabled', () => {
       expect.objectContaining({ event: 'chat_card_lookup_error' }),
       expect.any(String)
     )
+  })
+
+  it('aborts and fails open (returns true) when the lookup hangs past the timeout', async () => {
+    vi.useFakeTimers()
+    // Simulate a hanging gateway: the query only settles when the abort signal
+    // passed through Apollo's fetchOptions fires.
+    mockQuery.mockImplementation(({ context }) => {
+      const signal = context?.fetchOptions?.signal as AbortSignal
+      return new Promise((_resolve, reject) => {
+        signal.addEventListener('abort', () => {
+          reject(new Error('The operation was aborted'))
+        })
+      })
+    })
+
+    const result = getCardChatEnabled({
+      journeyId: 'journey-1',
+      cardId: 'card-1'
+    })
+    await vi.advanceTimersByTimeAsync(CARD_LOOKUP_TIMEOUT_MS)
+
+    await expect(result).resolves.toBe(true)
+    expect(mockLoggerWarn).toHaveBeenCalledWith(
+      expect.objectContaining({ event: 'chat_card_lookup_error' }),
+      expect.any(String)
+    )
+
+    vi.useRealTimers()
   })
 })
--- a/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
+++ b/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
@@ -1,0 +1,120 @@
+import { IdType } from '../../../__generated__/globalTypes'
+
+import { getCardChatEnabled } from './getCardChatEnabled'
+
+const { mockQuery, mockLoggerWarn } = vi.hoisted(() => ({
+  mockQuery: vi.fn(),
+  mockLoggerWarn: vi.fn()
+}))
+
+vi.mock('../apolloClient', () => ({
+  createApolloClient: () => ({ query: mockQuery })
+}))
+
+vi.mock('../logger', () => ({
+  logger: { warn: mockLoggerWarn, error: vi.fn(), info: vi.fn() }
+}))
+
+function journeyWithBlocks(blocks: unknown[]): unknown {
+  return { data: { journey: { id: 'journey-1', blocks } } }
+}
+
+const enabledCard = {
+  __typename: 'CardBlock',
+  id: 'card-1',
+  showAssistant: true
+}
+
+describe('getCardChatEnabled', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('returns true when the card exists and showAssistant is true', async () => {
+    mockQuery.mockResolvedValue(journeyWithBlocks([enabledCard]))
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(true)
+  })
+
+  it('queries the journey by databaseId with no-cache so the toggle is read fresh', async () => {
+    mockQuery.mockResolvedValue(journeyWithBlocks([enabledCard]))
+
+    await getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+
+    expect(mockQuery).toHaveBeenCalledWith(
+      expect.objectContaining({
+        variables: expect.objectContaining({
+          id: 'journey-1',
+          idType: IdType.databaseId
+        }),
+        fetchPolicy: 'no-cache'
+      })
+    )
+  })
+
+  it('returns false when the card exists but showAssistant is false', async () => {
+    mockQuery.mockResolvedValue(
+      journeyWithBlocks([
+        { __typename: 'CardBlock', id: 'card-1', showAssistant: false }
+      ])
+    )
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(false)
+  })
+
+  it('returns false when showAssistant is null', async () => {
+    mockQuery.mockResolvedValue(
+      journeyWithBlocks([
+        { __typename: 'CardBlock', id: 'card-1', showAssistant: null }
+      ])
+    )
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(false)
+  })
+
+  it('returns false when the card is not part of the journey', async () => {
+    mockQuery.mockResolvedValue(
+      journeyWithBlocks([
+        { __typename: 'CardBlock', id: 'a-different-card', showAssistant: true }
+      ])
+    )
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(false)
+  })
+
+  it('returns false (and does not query) when journeyId is missing', async () => {
+    await expect(
+      getCardChatEnabled({ journeyId: undefined, cardId: 'card-1' })
+    ).resolves.toBe(false)
+    expect(mockQuery).not.toHaveBeenCalled()
+  })
+
+  it('returns false when the journey is not found (definitive negative)', async () => {
+    mockQuery.mockRejectedValue(new Error('journey not found'))
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(false)
+    expect(mockLoggerWarn).not.toHaveBeenCalled()
+  })
+
+  it('fails open (returns true) and logs on an infrastructure error', async () => {
+    mockQuery.mockRejectedValue(new Error('gateway exploded'))
+
+    await expect(
+      getCardChatEnabled({ journeyId: 'journey-1', cardId: 'card-1' })
+    ).resolves.toBe(true)
+    expect(mockLoggerWarn).toHaveBeenCalledWith(
+      expect.objectContaining({ event: 'chat_card_lookup_error' }),
+      expect.any(String)
+    )
+  })
+})
--- a/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
+++ b/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
@@ -150,5 +150,4 @@ describe('getCardChatEnabled', () => {
       expect.any(String)
     )
   })
-
 })
--- a/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
+++ b/apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
@@ -1,13 +1,11 @@
 import { IdType } from '../../../__generated__/globalTypes'
 
-import {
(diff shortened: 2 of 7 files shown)


apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
```diff
@@ -0,0 +1,82 @@
...
+  try {
+    const { data } = await apolloClient.query<GetJourney>({
+      query: GET_JOURNEY,
+      variables: {
+        id: journeyId,
+        idType: IdType.databaseId,
+        options: { status: JOURNEY_STATUS_EXCLUDE_DRAFT }
+      },
+      // Always hit the gateway: a cached value would defeat the kill switch.
+      fetchPolicy: 'no-cache'
```
jaco-brink, line 64: **Concern (reliability + perf):** two things on this per-message lookup:

1. **No timeout.** Apollo has no default query timeout, so a slow/hanging gateway blocks every chat send until the serverless function 504s — for all journeys, on every message. The fail-open path below only catches *thrown* errors, not hangs, so it does not help here. An `AbortController` / timeout would let a slow lookup fail fast and fall through to fail-open.
2. **Heavy query.** `GET_JOURNEY` pulls the whole journey graph (all blocks, 50+ fields) to read one boolean, on every message. Disclosed in the PR — just want it explicitly accepted with a tracked follow-up (TTL cache or a minimal `card(id){ showAssistant }` query).

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
```diff
@@ -0,0 +1,82 @@
...
+      {
+        event: 'chat_card_lookup_error',
+        journeyId,
+        cardId,
+        name: err?.name,
+        message: err?.message
+      },
+      '[chat] card showAssistant lookup failed — allowing chat (fail open)'
+    )
+    return true
```
jaco-brink, line 88: **Concern (design — needs product sign-off):** fail-open on infrastructure error means a gateway blip silently re-enables chat on a card that was killed. Deliberate and documented, but it softens the "hard" in hard kill switch for exactly the emergency scenario this feature targets. Flagging for an explicit decision rather than leaving it as an implementation default.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -251,6 +261,15 @@ export function AiChat({
   const journeyIdRef = useRef(journeyId)
   journeyIdRef.current = journeyId
 
+  // The active card id (NES-1679) lets the server enforce the per-card chat
+  // kill switch. Read from the same global block history the rest of the viewer
+  // uses (Conductor, StepFooter), so it tracks card navigation without
+  // threading a prop through PinnedChatBar / ChatOverlay.
+  const { blockHistory } = useBlocks()
+  const cardId = getCardChild(blockHistory[blockHistory.length - 1])?.id
```
jaco-brink, line 269: **Concern (edge, low-likelihood):** cardId is derived from `blockHistoryVar`. If a message is sent before Conductor populates it (e.g. effect ordering on an auto-sent `initialMessage`, since child effects run before parent effects), cardId is `undefined` → server `400 invalid_request` → generic "Something went wrong" with no retry and no recovery path. Likely unreachable since Conductor sets `blockHistoryVar` on mount, but worth confirming the ordering guarantee.

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  })
+
+  it('returns false (and does not query) when journeyId is missing', async () => {
+    await expect(
+      getCardChatEnabled({ journeyId: undefined, cardId: 'card-1' })
+    ).resolves.toBe(false)
+    expect(mockQuery).not.toHaveBeenCalled()
+  })
+
+  it('returns false when the journey is not found (definitive negative)', async () => {
```
jaco-brink, line 109: **Nit:** coverage gap — this tests a *thrown* journey-not-found, but not a non-throwing `{ journey: null }` response (the `data.journey?.` branch in the impl). One extra case would lock that path down.

jaco-brink (commented):
## Review summary (self-review)

CI is green and the change is well-tested. No blocking issues — a few items worth an explicit decision before this leaves draft:

- **Reliability:** add a timeout to the per-message card lookup so a slow gateway fails fast (inline).
- **Design sign-off:** confirm fail-open-on-infra-error is the desired behaviour for a kill switch (inline).
- **Perf:** accept the full-journey-per-message query + track a follow-up (TTL / minimal query) (inline).
- **Edge:** first-message cardId race (inline, low-likelihood).
- **Nit:** one missing test case (inline).

**Strengths:** card looked up within its journey (anti-spoof for free), `fetchPolicy: no-cache` for freshness, cardId sent from the single transport owner (AiChat) rather than threaded through surfaces.

Draft PR → posting as COMMENT (not request-changes).

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Addressed in `129ae332d`:

1. **Timeout** — added an `AbortController` with a 5s ceiling (`CARD_LOOKUP_TIMEOUT_MS`), passing its signal through Apollo's `context.fetchOptions.signal`. A hung gateway now aborts and falls through to the existing fail-open path instead of holding the request open until the function 504s. Added a fake-timer test that hangs the query on the abort signal and asserts it fails open + logs.
2. **Heavy query** — accepted for now; tracked as a follow-up in NES-1721 (minimal `card { showAssistant }` query or short TTL cache). Not a launch blocker.

Resolving.

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Decision: **keep fail-open** (accepted by design). The primary failure surface here is a global gateway blip; failing closed would couple chat availability on *every* journey to a per-request gateway read, so a transient blip would take down all chat at once — worse than briefly missing one targeted kill. For a deliberate per-card kill the card's `showAssistant` is already `false` in the data, so fail-open only re-enables chat in the narrow window where the lookup itself errors. Documented in the function's failure-policy comment. Resolving as accepted.

libs/journeys/ui/src/components/AiChat/AiChat.tsx
jaco-brink: Confirmed safe — the ordering concern can't be reached:

1. `initialMessage` is never supplied anywhere in the codebase, so the auto-send-on-mount path is never exercised.
2. `AiChat` is lazy-loaded via `dynamic(() => import(...))` behind an `open`/state gate in both `ChatOverlay` and `PinnedChatBar`, so it never mounts in Conductor's initial commit. Conductor populates `blockHistory` (its `setTreeBlocks` effect) well before the AiChat chunk loads and mounts, so `cardId` is always defined by the time a message can be sent.

No code change needed. Resolving.

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.spec.ts
jaco-brink: Fixed in `129ae332d` — added a case for the non-throwing `{ journey: null }` response, which exercises the `data.journey?.` branch (returns false, no warn log) that the thrown not-found test didn't cover. Resolving.

jaco-brink:
## Review feedback addressed (129ae332d)

All four self-review threads triaged and resolved.

**Fixed:**
- `getCardChatEnabled.ts:56` — **No timeout.** Added an `AbortController` (`CARD_LOOKUP_TIMEOUT_MS` = 5s) wired through Apollo's `context.fetchOptions.signal`. A hung gateway now aborts and falls through to the fail-open path instead of holding the request open until the function 504s. Covered by a fake-timer test that hangs the query on the abort signal and asserts fail-open + log.
- `getCardChatEnabled.spec.ts:100` — **Coverage gap.** Added a non-throwing `{ journey: null }` case exercising the `data.journey?.` branch (returns false, no warn) that the thrown not-found test missed.

**Accepted (by design):**
- `getCardChatEnabled.ts:80` — **Fail-open on infra error.** Kept fail-open. Failing closed would couple chat availability on every journey to a per-request gateway read, so a transient blip would take down all chat at once — worse than briefly missing one targeted kill. A deliberate per-card kill already has `showAssistant = false` in the data; fail-open only re-enables in the narrow window where the lookup itself errors.

**Confirmed safe (no change):**
- `AiChat.tsx:269` — **cardId-undefined edge.** Unreachable: `initialMessage` is never supplied anywhere in the codebase, and `AiChat` is lazy-loaded (`dynamic` import) behind an `open` gate, so it never mounts in Conductor's initial commit — `blockHistory` is always populated first.

**Tracked as follow-up:**
- `getCardChatEnabled.ts:56` — **Heavy `GET_JOURNEY` query for one boolean.** Accepted for now; optimization (minimal `card { showAssistant }` query or short TTL cache) tracked in [NES-1721](https://linear.app/jesus-film-project/issue/NES-1721/optimize-per-card-chat-kill-switch-lookup-minimal-card-query-ttl-cache). Not a launch blocker.

apps/journeys/pages/api/chat/index.ts
```diff
@@ -74,7 +75,10 @@ const chatRequestSchema = z.object({
   messages: z.array(messageSchema).min(1).max(MAX_MESSAGES),
   language: z.string().max(64).optional(),
   sessionId: z.string().max(128).optional(),
-  journeyId: z.string().max(128).optional()
+  journeyId: z.string().max(128).optional(),
```
jianwei1: **Concern (contract / observability):** `journeyId` is still `.optional()` here, but the new kill switch treats a missing `journeyId` as a chat-disabled 403.

Flow: schema accepts a body with no `journeyId` → `getCardChatEnabled({ journeyId: undefined, ... })` short-circuits to `false` → handler returns `403 { code: 'chat_disabled' }` → client shows _"Chat has been turned off for this part of the journey."_ and hides Retry.

Semantically the chat isn't off — the request just didn't say which journey to consult. Two effects:

1. **User-visible:** if a malformed client ever omits `journeyId`, users see the kill-switch copy when really the request was invalid.
2. **Observability:** the `chat_card_disabled` warn fires for what's actually a malformed request, so the log line stops being a clean signal of "creator flipped the toggle".

In practice `AiChat` always supplies `journeyId` from `useJourney()`, so this is a contract-vs-code mismatch more than a live bug. Two clean options:
- Tighten the schema: `journeyId: z.string().min(1).max(128)` (required) — matches the new requirement on `cardId` and 400s as `invalid_request` on malformed callers.
- Or have `getCardChatEnabled` distinguish "missing input" from "explicit off" and 400 here.

---

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
```diff
@@ -0,0 +1,99 @@
...
+ * belongs to the journey in the request, so a stale tab can't keep chatting by
+ * spoofing a different journeyId.
+ *
+ * Freshness matters: the lookup uses `fetchPolicy: 'no-cache'` so flipping the
+ * toggle off in the editor stops chat for already-open tabs on their next
+ * message — an in-memory Apollo cache would otherwise mask the change.
+ *
+ * Failure policy:
+ * - Definitive negatives (missing journeyId, journey not found, card not in the
+ *   journey, `showAssistant !== true`) → `false` (fail closed). This is the
```
jianwei1, line 42: **Concern (threat model — worth an explicit ack):** the doc here notes that looking the card up within its journey blocks the "spoof a *different journeyId*" attack. It's worth adding the symmetric note that the switch is still bounded by the client-provided `cardId` — a viewer who knows another card-id in the same journey could swap it in and bypass a targeted per-card kill (chat stays on for the un-killed cards in the same journey).

This is consistent with the stated scope (creator emergency stop, not malicious-actor defense) and the targeted card itself can't be re-enabled — but a one-line "we know" in the doc keeps the threat model honest and gives the next reader a clear pointer to the journey-level switch (NES-1624) for the broader case. No code change needed.

---

jianwei1 (commented):
## Review summary

**Critical:** none.
**Concerns:** 2 (inline) — contract mismatch on optional `journeyId` vs the new kill-switch path, and an explicit threat-model ack for client-asserted `cardId`.

**Strengths:** kill switch enforces card↔journey ownership for free; `fetchPolicy: 'no-cache'` matches the freshness contract; `AbortController` + 5s ceiling addresses the gateway-hang failure mode that try/catch alone misses (with a fake-timer test that actually exercises the abort path); `cardId` sourcing lives in `AiChat` (the single transport owner) so mobile/desktop surfaces can't drift; failure policy (fail-open on infra, fail-closed on definitive negatives) is named in code and disclosed in the PR description; test coverage is thorough across unit + integration + component.

CI is green on the latest run.

Direction looks good — concerns are worth an explicit decision but don't block.

---

apps/journeys/pages/api/chat/index.ts
jaco-brink: Fixed in `97710f65d` — tightened the schema to `journeyId: z.string().min(1).max(128)` (required), matching the already-required `cardId`. A missing `journeyId` now returns **400 `invalid_request`** instead of routing through the kill switch to a 403 `chat_disabled`, so the "chat turned off" copy no longer fires for malformed requests and the `chat_card_disabled` warn stays a clean "creator flipped the toggle" signal. `getCardChatEnabled` keeps its own `journeyId` guard as defense-in-depth. Added a test asserting missing `journeyId` → 400 (lookup not called), and updated the affected test helpers to supply `journeyId`. Resolving.

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Good call — added the symmetric note in the doc comment (`c19662a45`): the switch is bounded by the client-supplied `cardId`, so a viewer who knows another un-killed card id in the same journey could swap it in and keep chatting on those cards, while the killed card itself can't be re-enabled. Framed it as the intended scope (creator emergency stop, not malicious-actor defense).

One adjustment to your draft: I pointed at the global LaunchDarkly `apologistChat` flag as the journey-wide/everywhere hammer rather than NES-1624 — per NES-1679, NES-1624 is the *removal of `Journey.showAssistant`*, not a journey-level kill switch, so citing it could mislead the next reader. Resolving.

jaco-brink:
## Review feedback addressed (round 2 — `97710f65d`, `c19662a45`)

Both new threads from @jianwei1 triaged and resolved.

**Fixed:**
- `pages/api/chat/index.ts:78` — **journeyId contract.** `journeyId` was `.optional()`, so a missing id routed through the kill switch to a 403 `chat_disabled` (showing users the "chat turned off" copy and polluting the `chat_card_disabled` signal). Tightened the schema to `journeyId: z.string().min(1).max(128)` (matching the required `cardId`) so a missing id now returns **400 `invalid_request`**. `getCardChatEnabled` keeps its own guard as defense-in-depth. Added a missing-journeyId→400 test and updated the affected test helpers.

**Fixed (doc only):**
- `getCardChatEnabled.ts:42` — **threat-model note.** Documented that the switch is bounded by the client-supplied `cardId` (a viewer who knows another un-killed card id in the same journey could swap it in; the killed card itself can't be re-enabled) — a creator emergency stop, not a malicious-actor defense.
  - *Deviation from the suggestion:* pointed at the global LaunchDarkly `apologistChat` flag as the journey-wide hammer rather than NES-1624 — per NES-1679, NES-1624 is the *removal of `Journey.showAssistant`*, not a journey-level kill switch.

_Branch was rebased onto the latest PR head (picked up `61d67728b fix: lint issues`); all 58 tests in the two affected specs pass._

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Update: **reverted the AbortController timeout** in `766a6f829`. It caused a regression — the 5s ceiling aborted the heavy `GET_JOURNEY` lookup whenever it ran slow server-side, and the abort fell through to the fail-open `catch`, so a card with `showAssistant=false` still chatted (intermittent, latency-dependent — passed QA, broke under real gateway latency).

The original concern stands, but a timeout that **fails open** is wrong for a kill switch — it defeats the switch on exactly the slow-lookup case. Re-addressing it needs a fail-**closed** timeout (or moving off the heavy whole-journey query, NES-1721). Restored the QA-passing plain `no-cache` lookup for now.

jianwei1 (dismissed):
## Approving

Both prior concerns cleanly addressed:

- **`journeyId` schema** — tightened to required `z.string().min(1).max(128)` (`97710f65d`). Missing `journeyId` now 400 `invalid_request` before reaching the kill switch; new test asserts it.
- **Threat-model doc** — symmetric `cardId`-spoof note added (`c19662a45`), with the correct pointer to the LaunchDarkly `apologistChat` flag as the journey-wide hammer (NES-1624 is removal, not a switch — good catch on my draft).

Noted beyond my asks:

- AbortController timeout reverted in `766a6f829`. Reasoning checks out — a fail-open timeout defeats the switch on the slow-lookup case (kill the timeout, not the switch). Net is no timeout with serverless 504 as backstop until NES-1721 lands a lighter query.
- New `chat_card_lookup` info log closes a real diagnostic gap (stale-read vs. request-identity mismatch on the allow path).

CI green, including all five Playwright E2E suites. LGTM.

---

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Correction to my previous reply: **there was no actual regression.** The "kill switch not working" I chased was a testing-environment mixup — I was verifying on `stage` (which doesn't have this PR's code) instead of the Vercel preview. On the actual preview the per-card kill switch works correctly (toggling `showAssistant` off blocks chat). The timeout has been removed (net-zero in the diff); re-addressing the original hung-gateway concern with a fail-**closed** timeout is left as a follow-up.

jaco-brink:
### PR state for re-review (cb9bf2902)

Heads-up: the "regression" discussion above was a **false alarm** — I was accidentally testing on `stage` (which doesn't carry this PR's code) instead of the Vercel preview. Verified working on the preview: toggling a card's chat off blocks `/api/chat` as intended.

**Net change since the last review** is just the two requested items:
1. `journeyId` required on `chatRequestSchema` → missing id returns 400 `invalid_request` instead of 403 `chat_disabled`.
2. Threat-model doc note in `getCardChatEnabled` (cardId-bounded; creator emergency stop).

A per-lookup timeout I briefly added was reverted (net-zero in the diff) and a diagnostic log was removed, so the reviewable diff is back to **feature + those two changes**. Nothing else changed.

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
```diff
@@ -0,0 +1,90 @@
...
+        idType: IdType.databaseId,
+        options: { status: JOURNEY_STATUS_EXCLUDE_DRAFT }
+      },
+      // Always hit the gateway: a cached value would defeat the kill switch.
+      fetchPolicy: 'no-cache'
+    })
+
+    const card = data.journey?.blocks?.find(
+      (block) => block.__typename === 'CardBlock' && block.id === cardId
+    )
```
jianwei1, line 69: **Concern (minor — observability regression):** the `chat_card_lookup` info log on the allow path that landed in the previous round has been removed in this push. That log was the only signal recording what the gateway returned **on allow** (`journeyFound` / `cardFound` / `showAssistant` / `enabled`), which is exactly the data needed to triage a "killed card still chatting" report — without it, a stale allow can no longer be told apart from a request-identity mismatch from logs alone. The block path (`chat_card_disabled`) and error path (`chat_card_lookup_error`) still fire, so the gap is narrow but real.

Could you confirm the removal was intentional? If the trigger was per-message log volume (fair on a per-chat-message path), a sampled emit or block-path-only log would still close the diagnostic gap without the noise. Not a blocker either way — just want it on the record before I re-approve.

---

jianwei1 (commented):
## Re-review at 691af6b

Strict reduction vs `43d0a53` (which I approved): rebased on a newer main and dropped the `chat_card_lookup` info log + its test. Schema, threat-model doc, and core kill-switch logic are byte-identical.

**Concern:** 1 minor (inline) — the dropped info log was the only allow-path signal that distinguished a stale read from a request-identity mismatch. Worth confirming the removal was intentional.

**Critical:** none. Substance of the kill switch is unchanged; everything I approved last round still stands.

CI is in flight on this push — same green set as last time should re-confirm.

---

apps/journeys/src/libs/getCardChatEnabled/getCardChatEnabled.ts
jaco-brink: Good catch, and yes — removing the per-message `chat_card_lookup` info log was intentional. Full disclosure: it was added mid-investigation while I was chasing a "killed card still chatting" report that turned out to be a stage-vs-preview testing mixup (the lookup was never the problem), so it came out again to avoid a per-chat-message info log on the hot path.

But your underlying point is right — the allow path shouldn't be a blind spot. Closed it the lighter way in `a25fbaf62`: added `cardId` to the shared `chatLogContext`, so every chat event (`apologist_chat_completed` on allow + all the error events) now records **which card the server consulted**, alongside the `journeyId` already there.

Reasoning for not re-adding the lookup-level log: on the allow path `journeyFound` / `cardFound` / `showAssistant` / `enabled` are all *definitionally* true (anything else blocks via `chat_card_disabled`), so the only actionable triage signal is the **identity** — which journey+card was consulted. That's exactly what a request-identity mismatch vs. a stale read turns on, and it's now queryable from existing telemetry with zero added log volume. The block (`chat_card_disabled`) and error (`chat_card_lookup_error`) paths are unchanged. Resolving — shout if you'd still prefer the explicit lookup log.

jianwei1 (approved):
## Approving

Last round's observability concern is resolved in `a25fbaf62` — a stricter improvement than what I suggested.

Adding `cardId` to `chatLogContext` puts the card identity on every existing chat event (`apologist_chat_completed` + the convert / stream / pipe / sync error events) alongside `journeyId`, with **zero added per-message log volume**. The reasoning checks out: on the allow path the lookup branches are definitionally true, so the only actionable signal for "killed card still chatting" reports is which card+journey the server consulted — and that's now queryable from existing telemetry. Spec asserts `cardId` in `apologist_chat_completed`, so the contract is locked in.

Everything I approved at `43d0a53` still stands: required `journeyId`+`cardId` schema, threat-model doc, kill-switch core logic. No timeout remains intentional with the fail-closed follow-up tracked.

CI green across the board, including all five Playwright suites. LGTM.

---

PR #9288, opened 2026-06-10
chore: remove legacy api-journeys service code

mikeallisonJS:
All functionality has been migrated to api-journeys-modern. This removes the legacy NestJS application source, its CI deploy jobs, local dev wiring (Procfile, gateway hive dev composition, storybook deps), and updates rules/docs that referenced the legacy API.

The Terraform infrastructure (apis/api-journeys/infrastructure and the environment module blocks) is intentionally kept so the deployed service keeps running until it is turned off in the gateway as a follow-up step.

## Schema changes from recomposition

- `Query.journey` is now nullable — the modern subgraph returns `null` for the embedded+hostname routing path (legacy declared it non-null). apps/journeys call sites now handle the null case by returning `notFound`.
- `JourneyUpdateInput.showAssistant` appears in the recomposed gateway schema. This field already existed on the modern subgraph; main's committed gateway schema was simply stale. Benign drift, surfaced here for visibility.
- `VideoBlock.video` (deprecated, "use mediaVideo union instead") was served only by the legacy subgraph and is removed from the supergraph. Zero internal usages (all frontends use `mediaVideo`). **Pre-merge:** run a Hive schema-check against tracked operations to confirm no external consumers still query `VideoBlock.video`.

## Seed

The journeys seed has been relocated from the deleted `apis/api-journeys/db` to `libs/prisma/journeys/db` (with a `prisma-seed` target), since it bootstraps the journeys **database** (which lives on), not the removed service.


csiyang:
Ran a strict one-to-one audit of what's being deleted vs what replaces it. The migration coverage itself is genuinely complete — diffed all 3,060 schema members (types, fields, enum values, input fields, union members) between main's gateway schema and this PR's, plus mapped the non-GraphQL machinery (BullMQ consumers, mailChimp, health, seeds). Four findings:

### 🔴 1. CI blocker: `journey` query nullability fallout in `apps/journeys`

`lint-work` fails with ~30 type errors. Root cause: the modern subgraph's `journey` query is `nullable: true` (`journey.query.ts` — and the resolver genuinely returns `null` for the embedded+hostname path), while legacy declared `journey: Journey!`. With the legacy subgraph gone, codegen now emits `GetJourney_journey | null`, and the `apps/journeys` pages (`[journeySlug].tsx`, `[stepSlug].tsx`, `oembed.ts`, embed pages) were written against non-null. Since the modern resolver intentionally returns null, the fix is updating those call sites to handle it (plus committing regenerated codegen output) — not flipping the resolver to non-null.

### 🔴 2. Seeds deleted with no replacement

`db/seed.ts` + 11 seed files (`jfpTeam`, `onboarding`, `onboardingTemplates`, `playwrightUserAccess`, `quickStartTemplate`, the NUA journeys) are deleted, and no replacement exists in `api-journeys-modern` or `libs/prisma/journeys`. The seed is coupled to the **database** (prisma-journeys, which lives on), not the **service**. Live code still depends on seeded rows: the ACLs branch on `teamId !== 'jfp-team'`, and `TeamOnboarding.tsx` hard-codes the onboarding template ID that `TermsAndConditions` duplicates for every new account. Production/stage DBs already have these rows, so deploys are unaffected — but a fresh local DB has no bootstrap path. Suggest relocating the seed to `libs/prisma/journeys/db/` with a `prisma-seed` target there (mirroring how api-media keeps its seed next to its prisma domain).

### 🟡 3. `JourneyUpdateInput.showAssistant` drift hitchhiker

The recomposed gateway schema picks up `showAssistant` on `JourneyUpdateInput` — modern's subgraph already had it but main's committed gateway schema was stale. Benign, but worth a line in the PR body so the unrelated schema change isn't invisible.

### 🟡 4. `VideoBlock.video` removal — verify external consumers

`VideoBlock.video` (deprecated, "use mediaVideo union instead") was served only by the legacy subgraph and vanishes from the supergraph. Zero usages in our frontends (all migrated to `mediaVideo`), but no Hive schema-check ran on this PR — worth a manual Hive check against tracked operations before merge if any third parties query the gateway.

---

Everything else checks out one-to-one: all ~30 ported operations resolve to API_JOURNEYS_MODERN, the `[Journey]` → `[Journey]!` signature changes are output-narrowing (safe), the orphan `Translation` type has zero references anywhere, all BullMQ consumers already live in modern's `workers/`, and keeping the Terraform for the cutover window is the right call.

mikeallisonJS:
## Review feedback addressed (c89ded1b9)

Thanks @csiyang for the one-to-one audit — all four findings actioned.

**🔴 #1 — `journey` nullability CI blocker → Fixed** (`1b2e9b8`)
Regenerated `apps/journeys` codegen (the committed types were stale) and guarded the null case at each `getStaticProps` / `oembed` call site by returning `notFound` — mirroring the existing `isJourneyNotFoundError` handling rather than flipping the resolver to non-null, since the modern resolver intentionally returns `null` on the embedded+hostname path. `journeys:type-check` and `lint` now pass locally.

**🔴 #2 — Deleted seeds → Fixed (relocated)** (`c89ded1`)
Ported the seed to `libs/prisma/journeys/db/` with a `prisma-seed` target (`pnpm exec tsx libs/prisma/journeys/db/seed.ts`), mirroring api-media. Enum imports now resolve from the generated prisma client instead of the deleted GraphQL types (identical values, verified against `schema.prisma`). Seeds are type-checked against the prisma client via a relaxed `tsconfig.seed.json` so schema drift is still caught, while `src/` stays strict — the ported dev-only scripts keep their original loose typing instead of contorting Prisma create/update types across ~5k lines.

**🟡 #3 — `showAssistant` drift → Documented**
Added a line to the PR body noting the benign `JourneyUpdateInput.showAssistant` pickup (modern subgraph already had it; main's gateway schema was stale).

**🟡 #4 — `VideoBlock.video` external consumers → Noted as pre-merge step**
Confirmed zero internal usages (all frontends on `mediaVideo`). Flagged in the PR body as a manual Hive schema-check against tracked operations before merge — that check can't be run from here.

csiyang (dismissed):

csiyang (approved):

PR #9296, opened 2026-06-11
feat(NES-1727): mobile AiChat drawer opens via floating chat button + X close

jaco-brink:
Closes [NES-1727](https://linear.app/jesus-film-project/issue/NES-1727/aichat-replace-mobile-collapsed-barchevron-with-floating-chat-button-x)

## What this does

Replaces the always-docked collapsed chat bar + chevron drag handle on mobile with a **floating AI chat button** that lives in the same StepFooter row as the WhatsApp/Messenger buttons. The drawer now has two states (idle ~20% / active 80%), closes completely via an **X button** in the ChatHeader, and stays mounted off-screen so the conversation survives close → reopen.

### Mobile (xs)
- **Footer restored:** `Conductor` no longer swaps the `StepFooter` out for the `PinnedChatBar` — this was silently hiding the WhatsApp/Messenger buttons **and** host avatar/title on every chat-enabled card (and caused the flash-then-vanish on refresh when the LaunchDarkly flag resolved after first paint).
- The drawer renders alongside the footer and slides up **over** it while open — by design you can't swap to another chat mid-conversation; close first, then pick.
- Open state lives in `ChatOverlayProvider` — the same `open` the footer's `AiChatButton` toggles and the `expandChatByDefault` auto-open effect seeds (per-(journey, card) sessionStorage dedup unchanged).
- Reopening with messages present lands back on the 80% conversation view with history intact (drawer + `AiChat` stay mounted while closed).

### Desktop (sm+)
- `AiChatButton` is now a true sibling of the MessagePlatform buttons: 44×44, 8px gap, one evenly spaced group (was 40×40 in a separate box pushed out by the row's 16px gap).
- Overlay behaviour unchanged; the `ChatOverlay` surface only mounts on sm+ (xs is owned by the drawer — mounting both would run two parallel chat sessions).

### Removed
- `DragHandle` component (+ spec) and the `'collapsed'` sheet state.
- `OverlayContent`'s 176px clearance reserved for the docked bar.
- `PinnedChatBar`'s `initialExpanded` / `AiChat`'s `collapsible` + `initialCollapsed` props.

## Testing

- New `PinnedChatBar.spec.tsx` (provider-driven open/close, stays mounted on close, idle→active sheet state, admin/embed null).
- `AiChat`/`ChatHeader` close-button specs; `AiChatButton` xs-surface spec; `Conductor.perCardChat` footer-coexistence regression spec.
- Full suites green: `journeys-ui` 1372 passed (sole failure is the pre-existing `TemplateVideoPreview` flake — passes in isolation on this branch and on main), `journeys` 120 passed.
- `type-check` + `lint` clean for both projects. Next.js compile succeeds; full static generation needs Doppler secrets not available locally — Vercel preview is the authoritative build.

## QA on Vercel preview (mobile viewport, journey with WhatsApp + Messenger + AI chat)

1. Land on the card → no thin bar; three floating buttons (WhatsApp, Messenger, AI) + host info visible.
2. Refresh → no flash-then-vanish of the platform buttons.
3. Tap AI button → drawer opens small (~20%), X visible, no chevron; footer covered.
4. Send a message → drawer grows to ~80%, reply streams.
5. Tap X → drawer fully gone; button row + host info back.
6. Tap AI button again → reopens at 80% with conversation intact.
7. `expandChatByDefault` card → drawer already open small on arrival (once per tab).
8. Card without `showAssistant` → no AI button, no drawer.
9. WhatsApp/Messenger still open their apps.
10. Desktop → one evenly spaced 44×44 button group; AI button opens the overlay as before.

Note: the now-unused `Expand chat` / `Collapse chat` locale keys are left for the regular Crowdin sync to sweep.

## Summary by CodeRabbit

* **New Features**
  * Added a “Close chat” button to dismiss the chat interface (when supported), in both the header and panel flows.
* **Bug Fixes**
  * Improved mobile/desktop overlay behavior, including more consistent auto-opening rules.
  * Ensured chat stays mounted correctly when closing and reopening.
  * Refined pinned chat bar behavior so the footer and chat controls coexist reliably.
* **UI/UX Improvements**
  * Removed the chat sheet’s collapsed/drag-based behavior; the sheet now opens/closes cleanly.


edmonday (commented):
## Review summary

Clean, well-tested refactor — net −186 lines, dead `DragHandle`/`collapsed` machinery removed, and the inline comments explaining each decision are genuinely good. Restoring the WhatsApp/Messenger + host info on chat cards (and killing the LD-flag flash) is a nice catch. CI is green across the board.

A couple of things worth addressing before merge (inline):

**Concerns**
- Test-file naming drift — fold `Conductor.perCardChat.spec.tsx` into `Conductor.spec.tsx` here; raise a follow-up for the two `apologistChat` specs.
- AI button icon size — worth a quick visual confirm against Figma.

**Nit**
- Unnecessary `useCallback` on `PinnedChatBar` `handleClose`.

Nothing blocking — all addressable. 👍

apps/journeys/src/components/Conductor/Conductor.perCardChat.spec.tsx
```diff
@@ -361,41 +356,32 @@ describe('Conductor per-card chat', () => {
     })
   })
 
-  describe('mobile sheet initial state', () => {
-    it('passes initialExpanded=true to PinnedChatBar when card.expandChatByDefault is true (idle, ready to type)', async () => {
+  describe('mobile drawer + footer coexistence (NES-1727)', () => {
```
edmonday: **Concern:** This `Component.<feature>.spec.tsx` split is unique to the chat workstream on the frontend — only 3 files have it repo-wide (`Conductor.perCardChat`, `Conductor.apologistChat`, `StepFooter.apologistChat`), all from this effort. It may have been picked up from the API operation specs (`*.mutation.spec.ts` / `*.query.spec.ts`, ~157 files) — but that is an API-only convention; every other FE component co-locates everything in a single `Component.spec.tsx`.

**In this PR:** could you fold this file back into `Conductor.spec.tsx`? The PR already rewrites this describe block (`mobile sheet initial state` → `mobile drawer + footer coexistence`), so it is a low-cost spot to stop the pattern from spreading.

**Follow-up:** the other two (`Conductor.apologistChat.spec.tsx` + `StepFooter.apologistChat.spec.tsx`) are the same drift but out of scope here — would you mind raising a follow-up ticket to consolidate those into their base specs so it is reviewed on its own?

libs/journeys/ui/src/components/AiChatButton/AiChatButton.tsx
```diff
@@ -42,16 +50,18 @@ export function AiChatButton(): ReactElement | null {
...
+          // 44×44 matches the MessagePlatform buttons in ChatButtons so
+          // the footer chat group reads as one evenly-sized row.
+          width: 44,
+          height: 44,
           '&:hover': {
             backgroundColor: 'primary.dark'
           }
         }}
       >
         <MessageChatStarsIcon sx={{ fontSize: 40 }} />
```
edmonday, line 62: **Concern (visual confirm):** The button is now 44×44 to match the WhatsApp/Messenger buttons 👍. One thing worth eyeballing on the preview: `MessageChatStarsIcon` is set to `fontSize: 40`, and that icon is a 24-viewBox glyph authored edge-to-edge, so at 40px it nearly fills the 44px button — whereas the sibling platform icons (`MessageChatIcon` in `ChatButtons`) render at the default ~24px. The buttons are evenly sized, but the AI glyph may read noticeably larger/heavier than the platform glyphs next to it.

Could well be the intended look — just confirm that the three sit comfortably as one row. No change needed if it looks right.

libs/journeys/ui/src/components/PinnedChatBar/PinnedChatBar.tsx
```diff
@@ -139,69 +123,67 @@ const AiChat = dynamic(
...
-    initialExpanded ? 'idle' : 'collapsed'
-  )
+  const { open, setOpen } = useChatOverlay()
+  const [sheetState, setSheetState] = useState<AiChatSheetState>('idle')
 
   const handleSheetStateChange = useCallback((next: AiChatSheetState) => {
     setSheetState(next)
   }, [])
 
+  const handleClose = useCallback(() => {
```
edmonday: **Nit:** This `useCallback` is not load-bearing — `onClose` only ends up as an `onClick` in `ChatHeader` (`AiChat.tsx` passes it straight through), it is in no dependency array, and neither `AiChat` nor `ChatHeader` is memoized. `setOpen` is itself stable (`ChatOverlayProvider` wraps it in `useCallback([])`), so the wrapper never even changes identity. An unnecessary `useCallback` slightly miscommunicates intent — it implies the identity stability matters somewhere when it does not. Could inline:

```tsx
<AiChat
  onSheetStateChange={handleSheetStateChange}
  onClose={() => setOpen(false)}
/>
```

(Keep the sibling `handleSheetStateChange` wrapper — that one *is* in a `useEffect` dep array, so it is correct.)

libs/journeys/ui/src/components/AiChatButton/AiChatButton.tsx
jaco-brink: Good eye on the geometry — you're right that the numbers differ (the glyph fills ~63–67% of its viewBox, so at `fontSize: 40` the visible mark is ~25–27px vs the platform glyphs at ~22–24px).

I pulled the three up side-by-side on the preview to check the *rendered* result, and they sit comfortably as one row — the AI mark reads comparably to the Messenger glyph next to it (the WhatsApp phone is actually the most contained of the three, so the AI icon isn't the odd one out). The slightly larger footprint is load-bearing here: the sparkles need it to read as sparkles rather than specks, which is what distinguishes the AI affordance from a plain message bubble. `fontSize` scales the whole SVG, not just the padding, so dropping to ~24–30 shrinks the actual glyph and the stars lose their pop.

So I'm keeping the inherited NES-1726 `fontSize: 40` — confirmed it looks right on the preview, which was your "no change needed if it looks right" condition. Thanks for flagging it.

jaco-brink:
## Review feedback addressed (4662a2a0f)

Thanks @edmonday — three thoughtful threads, all triaged on merit.

**Fixed:**
- `PinnedChatBar.tsx:153` — Inlined the non-load-bearing `handleClose` `useCallback` to `onClose={() => setOpen(false)}`. Confirmed your analysis: it only ends up as `ChatHeader`'s `onClick`, is in no dep array, and `setOpen` is already stable. Kept `handleSheetStateChange` memoized — that one *is* in an `AiChat` `useEffect` dep array. (`de44da5`)

**Fixed (refactor) + follow-up raised:**
- `Conductor.perCardChat.spec.tsx` — Folded back into `Conductor.spec.tsx` as a nested `describe` and deleted the standalone file. Verified your claim: exactly 3 `Component.<feature>.spec.tsx` files exist repo-wide, all from this workstream. The `PinnedChatBar`/`ChatOverlay` stubs moved to file level (harmless to the non-chat tests — they render without the flag, so the bar never mounts); the folded tests use `variant: 'customize'`, so Conductor's journeyView effect early-returns and never touches the module-level geo `fetch`. `Conductor.spec.tsx` now runs 17 tests (5 base + 12 folded), all green. (`4662a2a`)
- Follow-up for the other two (`Conductor.apologistChat.spec.tsx` + `StepFooter.apologistChat.spec.tsx`): **NES-1733**, with the watch-outs I hit here written up.

**Challenged (see inline reply):**
- `AiChatButton.tsx:62` — Icon size. Your geometry is right (AI glyph ~25–27px vs platform ~22–24px), but I confirmed the *rendered* row on the preview: the three sit comfortably together, the AI mark reads comparably to Messenger, and the sparkles need that footprint to read as sparkles. Keeping the inherited NES-1726 `fontSize: 40` — which met your own "no change needed if it looks right" bar. Left the thread open for you.

**Verification:** `type-check` + `lint` clean for both `journeys` and `journeys-ui`; affected chat + Conductor suites all pass.

libs/journeys/ui/src/components/AiChatButton/AiChatButton.tsx
jaco-brink: the twinkle stars need the extra size to look good overal
image

libs/journeys/ui/src/components/AiChat/ChatHeader/ChatHeader.tsx
```diff
@@ -174,6 +183,22 @@ export function ChatHeader({
           </Link>
         </Typography>
       </Box>
+      {onClose != null && (
+        <IconButton
+          onClick={onClose}
+          aria-label={t('Close chat')}
+          tabIndex={0}
```
edmonday: **Nit:** `tabIndex={0}` is redundant on `IconButton` — it is already focusable. Harmless.

edmonday (commented):
## Review summary

Re-review after the changes addressing my last pass. All three prior items resolved:

- **Spec split** — `Conductor.perCardChat.spec` folded back into `Conductor.spec` as a nested describe; follow-up for the other two raised as NES-1733. ✅
- **Icon geometry** — confirmed on preview, `fontSize: 40` kept per the "no change needed if it looks right" condition. ✅
- **Non-load-bearing useCallback** — inlined to `onClose={() => setOpen(false)}`. ✅

New diff is clean — nothing blocking, two trivial nits inline. The `visibility 0s 280ms` delay (animate-then-remove-from-a11y-tree) is a nice touch, and test coverage on the new drawer behaviour is solid. CI green.

Approving.

edmonday (dismissed):
LGTM — prior review items all addressed, new diff clean, CI green. 🚀

libs/journeys/ui/src/components/AiChat/ChatHeader/ChatHeader.tsx
jaco-brink: Good catch — removed. IconButton is a native `<button>` so it's already focusable; the `aria-label` and native Enter/Space activation stay intact. (c637d6bac)

edmonday (approved):
Nice work!

PR #9297, opened 2026-06-12
feat(NES-308): single-page editor layout for desktop

csiyang:
## Summary

Replaces the desktop journey editor's three-slide carousel with a single-screen **LayeredView**: the journey flow map fills the screen, and card Content + Settings open in a right-anchored overlay drawer on top of it. Mobile (below `md`) keeps the existing Slider unchanged.

This is a rebuild of the approach prototyped in #8561, without the two problems that stalled it:

- **No component duplication.** The prototype copied the entire Slider tree (874 files). LayeredView imports the existing `Slider/JourneyFlow`, `Slider/Content`, and `Slider/Settings` components directly — the whole change is 25 files.
- **No global editor-state changes.** The prototype rewired `ActiveSlide` dispatches unconditionally, which broke mobile. Here, a new `EditorLayoutContext` (`'slider' | 'layered'`, defaulting to `'slider'`) lets each of the ~10 dispatch sites pick its slide value per layout: `isLayered ? <layered value> : <existing value unchanged>`. The `EditorProvider` reducer is completely untouched.

## Layered slide semantics (desktop)

| ActiveSlide | Meaning |
|---|---|
| `JourneyFlow` | drawer closed, flow full-screen |
| `Content` | drawer open, canvas only (staged reveal) |
| `Drawer` | drawer open, canvas + settings panel |

Selecting a step node opens the drawer at the canvas; selecting a block (or Add Block / Strategy / social preview) slides the settings panel in. Backdrop click / Esc closes the drawer.

## Mobile safety

- All existing specs pass with **zero changes to existing test expectations** — the layout context defaults to `'slider'`, so every existing code path is byte-for-byte unchanged.
- `EditorProvider` lib suite (36 tests) passes unmodified.
- Full `Editor/` suite: 1,660 tests passing (one pre-existing timezone-sensitive date spec fails locally in non-UTC timezones; file untouched by this PR).

## Testing

- New `LayeredView` spec: drawer open/close per slide, staged settings reveal, backdrop-close dispatch.
- Layered-mode cases added to Fab, StrategyItem, SelectableWrapper, Canvas, SocialPreview, JourneyFlow, useBlockCreateCommand, and Editor specs.
- ⚠️ Manual browser pass still to be done (drawer styling/sizing ported from the prototype).

## Rollout note

No feature flag — desktop users get the new layout on merge, per discussion.

## Release Notes

* **New Features**
  * Added a desktop “layered” editor layout with a persistent right-side settings drawer, while mobile remains in slider mode.
  * Updated block, FAB, and toolbar actions to open the correct panel for the current layout.
  * Enhanced canvas, goals, and social preview styling/behavior to match layered layout.

* **Refactor**
  * Centralized editor layout behavior and improved drawer/canvas coordination for smoother transitions.

* **Tests**
  * Added and updated unit, component, and end-to-end tests to cover layered behavior and drawer open/close flows.

### DIFF
--- a/apps/journeys-admin-e2e/src/pages/card-level-actions.ts
+++ b/apps/journeys-admin-e2e/src/pages/card-level-actions.ts
@@ -28,7 +28,24 @@ export class CardLevelActionPage {
     this.journeyName = testData.journey.firstJourneyName + this.randomNumber
   }
 
+  // The desktop layered editor keeps the card inside a closed modal drawer, so
+  // open it before touching the card iframe. No-op on the mobile slider, where
+  // every slide (including the card) stays mounted.
+  async ensureCardEditorOpen() {
+    if (
+      (await this.page.locator('div[data-testid="EditorCanvas"]').count()) > 0
+    ) {
+      return
+    }
+    await this.page.locator('[data-testid^="StepBlockNode-"]').first().click()
+    await this.page
+      .locator('div[data-testid="EditorCanvas"]')
+      .first()
+      .waitFor({ state: 'visible', timeout: sixtySecondsTimeout })
+  }
+
   async clickOnJourneyCard() {
+    await this.ensureCardEditorOpen()
     const iframes = this.page.locator(this.journeyCardFrame)
     const frame = await iframes.first().contentFrame()
     await frame
@@ -42,6 +59,7 @@ export class CardLevelActionPage {
   }
 
   async clickOnVideoJourneyCard() {
+    await this.ensureCardEditorOpen()
     const iframes = this.page.locator(this.journeyCardFrame)
     const frame = await iframes.first().contentFrame()
     await frame
@@ -59,6 +77,7 @@ export class CardLevelActionPage {
   }
 
   async clickAddBlockBtn() {
+    await this.ensureCardEditorOpen()
     await expect(
       this.page.locator('button[data-testid="Fab"]', { hasText: 'Add Block' })
     ).toHaveCount(1, { timeout: sixtySecondsTimeout })
@@ -103,6 +122,15 @@ export class CardLevelActionPage {
   }
 
   async clickDoneBtn() {
+    // The desktop layered editor has no 'Done' Fab; re-clicking 'Add Block'
+    // commits the inline edit while keeping the card drawer open (the verify
+    // steps that follow read the card frame). The mobile slider uses 'Done'.
+    if (
+      (await this.page.locator('div[data-testid="LayeredView"]').count()) > 0
+    ) {
+      await this.clickAddBlockBtn()
+      return
+    }
     await this.page
       .locator('button[data-testid="Fab"]', { hasText: 'Done' })
       .click()
@@ -478,6 +506,7 @@ export class CardLevelActionPage {
   }
 
   async deleteAllAddedCardProperties() {
+    await this.ensureCardEditorOpen()
     const iframes = this.page.locator(this.journeyCardFrame)
     const frame = await iframes.first().contentFrame()
     const properties = await frame
@@ -871,6 +900,7 @@ export class CardLevelActionPage {
   }
 
   async selectWholeFooterSectionInCard() {
+    await this.ensureCardEditorOpen()
     const iframes = this.page.locator(this.journeyCardFrame)
     const frame = await iframes.first().contentFrame()
     await frame
--- a/apps/journeys-admin-e2e/src/pages/journey-page.ts
+++ b/apps/journeys-admin-e2e/src/pages/journey-page.ts
@@ -261,7 +261,23 @@ export class JourneyPage {
         : testData.journey.secondJourneyName) + generateRandomNumber(3)
   }
 
+  // The desktop layered editor opens map-first with the card inside a closed
+  // modal drawer, so the card iframe is not mounted until a step node is
+  // clicked. The mobile slider keeps every slide mounted, so this is a no-op.
+  async openLayeredCardEditor(): Promise<void> {
+    if (
+      (await this.page.locator('div[data-testid="EditorCanvas"]').count()) > 0
+    ) {
+      return
+    }
+    await this.page.locator('[data-testid^="StepBlockNode-"]').first().click()
+    await expect(
+      this.page.locator('div[data-testid="EditorCanvas"]').first()
+    ).toBeVisible({ timeout: thirtySecondsTimeout })
+  }
+
   async enterJourneysTypography(): Promise<void> {
+    await this.openLayeredCardEditor()
     await this.page
       .frameLocator(this.journeyCardFrame)
       .first()
@@ -305,10 +321,46 @@ export class JourneyPage {
   }
 
   async clickDoneBtn(): Promise<void> {
+    // On the desktop layered editor the card lives in a modal drawer; closing
+    // it via its backdrop commits the inline edit and uncovers the toolbar.
+    // The mobile slider has no modal overlay, so tap the Fab instead.
+    if (
+      (await this.page.locator('div[data-testid="LayeredView"]').count()) > 0
+    ) {
+      await this.page
+        .locator('div[data-testid="LayeredViewDrawer"] .MuiBackdrop-root')
+        .first()
+        .click()
+      await expect(
+        this.page.locator('div[data-testid="EditorCanvas"]')
+      ).toBeHidden({ timeout: thirtySecondsTimeout })
+      return
+    }
     await this.page.locator('button[data-testid="Fab"]').click()
   }
 
+  // On the desktop layered editor an open card drawer is a modal whose backdrop
+  // covers the toolbar, so close it before reaching for a toolbar action. No-op
+  // on the mobile slider (no modal) or when the drawer is already closed.
+  async ensureLayeredDrawerClosed(): Promise<void> {
+    if (
+      (await this.page.locator('div[data-testid="LayeredView"]').count()) ===
+        0 ||
+      (await this.page.locator('div[data-testid="EditorCanvas"]').count()) === 0
+    ) {
+      return
+    }
+    await this.page
+      .locator('div[data-testid="LayeredViewDrawer"] .MuiBackdrop-root')
+      .first()
+      .click()
+    await expect(
+      this.page.locator('div[data-testid="EditorCanvas"]')
+    ).toBeHidden({ timeout: thirtySecondsTimeout })
+  }
+
   async clickThreeDotBtnOfCustomJourney(): Promise<void> {
+    await this.ensureLayeredDrawerClosed()
     await this.page.locator('button#edit-journey-actions').click()
   }
 
--- a/apps/journeys-admin/src/components/Editor/Editor.tsx
+++ b/apps/journeys-admin/src/components/Editor/Editor.tsx
@@ -39,7 +39,12 @@ export function Editor({
   initialState,
   user
 }: EditorProps): ReactElement {
-  const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'))
+  // noSsr defers the match to the client so the editor renders the correct
+  // surface on first paint instead of mounting <Slider /> then remounting
+  // <LayeredView /> (which would flash and remount React Flow) on desktop
+  const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'), {
+    noSsr: true
+  })
   const steps =
     journey != null
       ? (transformer(journey.blocks ?? []) as Array<TreeBlock<StepBlock>>)
--- a/apps/journeys-admin/src/components/Editor/Editor.tsx
+++ b/apps/journeys-admin/src/components/Editor/Editor.tsx
@@ -1,3 +1,5 @@
+import type { Theme } from '@mui/material/styles'
+import useMediaQuery from '@mui/material/useMediaQuery'
 import { ReactElement } from 'react'
 import { HotkeysProvider } from 'react-hotkeys-hook'
 
@@ -11,9 +13,11 @@ import { GetJourney_journey as Journey } from '../../../__generated__/GetJourney
 import { User } from '../../libs/auth'
 import { MuxVideoUploadProvider } from '../MuxVideoUploadProvider'
 
+import { EditorLayoutProvider } from './EditorLayoutContext'
 import { Fab } from './Fab'
 import { FontLoader } from './FontLoader/FontLoader'
 import { Hotkeys } from './Hotkeys'
+import { LayeredView } from './LayeredView'
 import { Slider } from './Slider'
 import { Toolbar } from './Toolbar'
 
@@ -35,6 +39,7 @@ export function Editor({
   initialState,
   user
 }: EditorProps): ReactElement {
+  const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'))
   const steps =
     journey != null
       ? (transformer(journey.blocks ?? []) as Array<TreeBlock<StepBlock>>)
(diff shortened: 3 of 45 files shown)


apps/journeys-admin/src/components/Editor/LayeredView/LayeredView.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+  it('keeps the off-screen settings panel inert on the content slide', () => {
+    renderLayeredView({ ...state, activeSlide: ActiveSlide.Content })
+
+    // on the content slide the settings panel is slid off-screen; it must stay
+    // pointer-events: none so it cannot intercept clicks meant for the backdrop
+    // while it is transitioning in/out
+    expect(screen.getByTestId('SettingsDrawer').parentElement).toHaveStyle(
+      'pointer-events: none'
+    )
+  })
```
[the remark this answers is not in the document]
csiyang: Addressed in 9db952416. The behavioral close is already covered by the adjacent test `dispatches journey flow slide on backdrop close`, which clicks `.MuiBackdrop-root` and asserts `activeSlide` transitions 2→0.

A click can't be made to genuinely "fall through" `pointer-events: none` in jsdom (no layout/hit-testing), so a fall-through assertion here would either click the backdrop directly (duplicating the test above) or pass for the wrong reason. I've renamed this test to describe what it actually pins — the paper being pointer-transparent while the settings panel stays interactive — and added a comment noting the jsdom limitation and pointing to the backdrop-close test. Real click-through routing was verified manually in-browser.

apps/journeys-admin/src/components/Editor/Editor.tsx
```diff
@@ -35,6 +39,7 @@ export function Editor({
   initialState,
   user
 }: EditorProps): ReactElement {
+  const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'))
```
jaco-brink: **Concern:** `useMediaQuery(breakpoints.up('md'))` returns `false` on the server and on the first client/hydration render, then flips to `true` after mount. Since the result now switches the *entire* editor tree (line 73, `<LayeredView />` vs `<Slider />`), desktop users render the mobile `<Slider />` first, then React tears it down and mounts `<LayeredView />` — a visible layout flash plus a full React Flow remount on every load/refresh.

This matches the repo's existing convention for layout-deciding breakpoints (`useBreakpoints.ts`, `SidePanelTitle.tsx`, `LanguageAutocomplete.tsx` all pass `noSsr`):

```ts
const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'), { noSsr: true })
```

The editor is fully client-interactive, so deferring this to post-mount is harmless.

apps/journeys-admin/src/components/Editor/Editor.tsx
```diff
@@ -55,17 +60,19 @@ export function Editor({
...
+              <FontLoader
+                fonts={[
+                  journey?.journeyTheme?.headerFont ?? '',
+                  journey?.journeyTheme?.bodyFont ?? '',
+                  journey?.journeyTheme?.labelFont ?? ''
+                ]}
+              />
+              <Hotkeys />
+              <Toolbar user={user} />
+              {mdUp ? <LayeredView /> : <Slider />}
```
jaco-brink, line 78: **Concern (rollout):** This swaps the whole desktop editing surface with no `useFlags()` gate, so reversal is a code revert + redeploy. The editor already gates surface area behind flags (`editorAnalytics`, `mediaLibrary`, `teamTemplateCollection`…), so a `layeredEditor` flag would give staged rollout + instant rollback nearly for free:

```ts
const { layeredEditor } = useFlags()
const layout = mdUp && layeredEditor ? 'layered' : 'slider'
```

The description's rollout note says this was decided "per discussion," so really this is a "confirm it's still the intent" given the blast radius across ~20 branched sites and the still-pending manual QA. No change needed if the team is settled on it.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.tsx
```diff
@@ -120,6 +121,7 @@ export function Properties({ block, step }: PropertiesProps): ReactElement {
   const selectedStep = step ?? state.selectedStep
 
   const mdUp = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'))
+  const { isLayered } = useEditorLayout()
```
jaco-brink, line 121: **Nit:** Anchoring here where `isLayered` is introduced, but this is about the settings-panel close handler further down (~line 205, not in the diff): it dispatches `activeSlide: mdUp ? ActiveSlide.JourneyFlow : ActiveSlide.Content`, keyed on `mdUp` rather than `isLayered`. It's the one dispatch site of ~10 that doesn't use `isLayered`. Behavior is correct *today* only because both derive from the same `md` breakpoint — but if layout is ever decoupled from the breakpoint (a feature flag, or a layered mobile mode), this silently diverges from every other site. Suggest keying it on `isLayered` for consistency: behavior-identical now, just self-consistent intent.

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/QuickControls/MoveBlock/MoveBlock.tsx
```diff
@@ -39,7 +41,10 @@ export function MoveBlock(): ReactElement {
         dispatch({
           type: 'SetEditorFocusAction',
           selectedStep,
-          selectedBlock
+          selectedBlock,
+          // in the layered desktop view, keep the settings panel open while
+          // reordering (block selection would otherwise force the canvas view)
+          ...(isLayered && { activeSlide })
```
jaco-brink, line 47: **Nit (test coverage):** This new layered branch (`...(isLayered && { activeSlide })`) isn't covered — `MoveBlock.spec.tsx` wasn't touched. Same gap in `useBlockDeleteCommand.tsx` (`isLayered ? activeSlide : ActiveSlide.Content`) and `SocialPreviewNode.tsx` (both branches → `Drawer` when layered). Peers like `useBlockCreateCommand`/`Fab`/`SelectableWrapper` did get layered tests, so this is just a gap. A one-line layered-mode assertion each (`<EditorLayoutProvider value="layered">` + assert `activeSlide`) would close it. Not a blocker.

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/utils/calculateDimensions/calculateDimensions.ts
```diff
@@ -5,6 +5,11 @@ import { EDIT_TOOLBAR_HEIGHT } from '../../../../../constants'
 export const CARD_WIDTH = 324
 export const CARD_HEIGHT = 674
 
+// height of the layered editor's floating paper: the card plus vertical chrome
+// (slug edit, footer, breathing room). The card/settings paper and the media
+// library overlay both use this so they stay vertically aligned.
+export const LAYERED_DRAWER_HEIGHT = CARD_HEIGHT + 200
```
jaco-brink, line 11: **Nit (organization):** `LAYERED_DRAWER_HEIGHT` is a layered-only constant living in the *slider's* canvas-dimension util, and it's imported back out by `LayeredView.tsx` and `Settings/Drawer/Drawer.tsx`. It would read more naturally next to the other layout constants in `Editor/constants.ts` (`DRAWER_WIDTH`/`DRAWER_GAP`/`EDIT_TOOLBAR_HEIGHT`). Minor; fine as a fast-follow.

jaco-brink (commented):
## Review summary

Solid execution of the NES-308 stretch goal — reusing the Slider components (874→41 files vs the old prototype) and leaving the shared `EditorProvider` reducer untouched are the right calls, the mobile path is provably unchanged (default `'slider'`), and CI is green. No correctness or security blockers found.

Two items worth resolving before merge (both low-effort), flagged inline:
1. Missing `noSsr` on the layout `useMediaQuery` → desktop flashes the mobile Slider + remounts React Flow on every load.
2. Confirm the no-feature-flag rollout is still intended given the blast radius (the description notes it was decided "per discussion").

The rest are maintainability / clarity / test-coverage nits that can land here or as fast-follow. The manual desktop QA noted as pending in the description should be completed before merge.

Two notes that don't map to a changed line, so couldn't be attached inline:
- **`ActiveSlide` enum doc comment** (`libs/journeys/ui/.../EditorProvider.tsx`, unchanged by this PR): the note _"Settings should only be set as active when on a mobile device"_ now contradicts layered desktop usage (`Drawer` is set on desktop). Worth refreshing when convenient. Bigger picture — overloading the carousel enum to mean drawer-open state is what drives the `isLayered ? Drawer : Content` ternary repeated across ~10 sites and the double-dispatches that "correct" the reducer; a derived `drawerState` (or reducer-owned `layout`) would centralize that mapping. Follow-up, not a blocker.
- **Stale `// TODO: Test dispatch via E2E`** (`SelectableWrapper.tsx`, pre-existing ~line 52): this PR adds the e2e page-object plumbing plus a layered `SelectableWrapper.spec.tsx` case that cover it, so the TODO can be deleted.

_(Generated from a 3-persona review — architecture, pragmatic engineering, security. Security came back low-risk with nothing exploitable.)_

apps/journeys-admin/src/components/Editor/Editor.tsx
csiyang: Done in 794442943 — added `{ noSsr: true }` to the layout breakpoint so the editor renders the correct surface on first paint (no <Slider />→<LayeredView /> remount/flash on desktop). Good catch.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/Properties.tsx
csiyang: Done in 794442943 — switched this dispatch to `isLayered` (and removed the now-unused `useMediaQuery`/`Theme`), so all ~10 layout-deciding dispatches now share one source.

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/QuickControls/MoveBlock/MoveBlock.tsx
csiyang: Done in 794442943 — added layered-mode tests (`<EditorLayoutProvider value="layered">` + an activeSlide assertion) for MoveBlock, useBlockDeleteCommand, and SocialPreviewNode.

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/utils/calculateDimensions/calculateDimensions.ts
csiyang: Keeping it in calculateDimensions for now: moving `LAYERED_DRAWER_HEIGHT` to `Editor/constants.ts` creates a circular import — `constants.ts` would import `CARD_HEIGHT` from calculateDimensions, which already imports `EDIT_TOOLBAR_HEIGHT` back from constants. Placing it next to `CARD_HEIGHT` (its only dependency) avoids the cycle. Happy to revisit if we reorganize the layout constants later.

apps/journeys-admin/src/components/Editor/Editor.tsx
csiyang: Confirming intent: shipping without a `layeredEditor` flag for now — decided per discussion. Noted the reversal cost given the blast radius; a flag can be a fast-follow if we want staged rollout. Leaving as-is for this PR.

jaco-brink: Confirming we're good to ship without a `layeredEditor` flag for now. The two considerations were how fast we could roll back and the user-facing impact if a regression slipped through — both lower/faster with a flag than with a code revert, which would simply take longer. We're accepting that residual time-to-recover risk given this is a low-risk, no-data UI change with a known one-file fallback in `Editor.tsx`.

jaco-brink:
## Review wrap-up

Where each item landed:

- **`noSsr` first-paint flash** — fixed (`794442943`); desktop now renders the right surface on first paint, no Slider→LayeredView remount. ✅
- **`mdUp` → `isLayered` (Properties close)** — fixed; all ~10 layout-deciding dispatches now share one source, and the unused `useMediaQuery`/`Theme` imports are gone. ✅
- **Layered-mode test coverage** (MoveBlock / useBlockDeleteCommand / SocialPreviewNode) — added, each asserting the active slide. ✅
- **`LAYERED_DRAWER_HEIGHT` location** — keeping it in `calculateDimensions` to avoid a circular import with `constants.ts`; sound rationale, agreed. ✅
- **Feature flag** — accepting no flag (see thread); considerations were rollback speed and user-facing impact, both lower with a flag than a revert, but acceptable for this change. ✅

Happy to approve once all checks are green.

apps/journeys-admin-e2e/src/pages/card-level-actions.ts
```diff
@@ -28,7 +28,24 @@ export class CardLevelActionPage {
...
+    if (
+      (await this.page.locator('div[data-testid="EditorCanvas"]').count()) > 0
+    ) {
+      return
+    }
+    await this.page.locator('[data-testid^="StepBlockNode-"]').first().click()
+    await this.page
+      .locator('div[data-testid="EditorCanvas"]')
+      .first()
+      .waitFor({ state: 'visible', timeout: sixtySecondsTimeout })
```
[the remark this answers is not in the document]
csiyang: The 60s wait is intentional: opening the layered drawer plays a slide/fade transition and then mounts the card iframe (FramePortal), which can exceed the 30s default on a cold preview deployment. It also matches the existing `sixtySecondsTimeout` usages throughout this file. Leaving as-is.

apps/journeys-admin-e2e/src/pages/card-level-actions.ts
```diff
@@ -28,7 +28,24 @@ export class CardLevelActionPage {
...
+  }
```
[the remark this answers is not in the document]
csiyang: These tests create a fresh journey with a single step card, so `[data-testid^="StepBlockNode-"].first()` is deterministic here — it just opens the card editor. The flow node has no user-facing label to scope to. Leaving as-is.

apps/journeys-admin-e2e/src/pages/journey-page.ts
```diff
@@ -261,7 +261,23 @@ export class JourneyPage {
...
+  // modal drawer, so the card iframe is not mounted until a step node is
+  // clicked. The mobile slider keeps every slide mounted, so this is a no-op.
+  async openLayeredCardEditor(): Promise<void> {
+    if (
+      (await this.page.locator('div[data-testid="EditorCanvas"]').count()) > 0
+    ) {
+      return
+    }
+    await this.page.locator('[data-testid^="StepBlockNode-"]').first().click()
+    await expect(
```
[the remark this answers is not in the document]
csiyang: Same as the card-level helper: the freshly-created journey has exactly one step card, so `.first()` is deterministic — it just opens the card editor before editing. Leaving as-is.

apps/journeys-admin-e2e/src/pages/journey-page.ts
```diff
@@ -305,10 +321,46 @@ export class JourneyPage {
...
+    // On the desktop layered editor the card lives in a modal drawer; closing
+    // it via its backdrop commits the inline edit and uncovers the toolbar.
+    // The mobile slider has no modal overlay, so tap the Fab instead.
+    if (
+      (await this.page.locator('div[data-testid="LayeredView"]').count()) > 0
+    ) {
+      await this.page
+        .locator('div[data-testid="LayeredViewDrawer"] .MuiBackdrop-root')
+        .first()
+        .click()
```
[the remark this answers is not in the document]
csiyang: Escape is not reliable here: after editing a block, focus is inside the card iframe (FramePortal), so an Escape keydown lands in the iframe document and never reaches the parent MUI modal — the drawer would not close. Clicking the backdrop closes it regardless of where focus is, and is what I validated end-to-end. The explicit hidden assertion stays right after it. Keeping the backdrop click.

jaco-brink (approved):

PR #9298, opened 2026-06-12
fix(NES-1728): pin stylis to emotion's bundled 4.2.0 to fix RTL journey crash

jaco-brink:
Fixes [NES-1728](https://linear.app/jesus-film-project/issue/NES-1728/rtl-rendering-error-in-arabic-chat-page)

## Problem

Every RTL-language journey (language code `ar`, `ar-*`, `fa`, `ur`, `he`, …) containing a text input crashes on load with an application-error screen, on stage and production today. Confirmed branch-independent in the ticket — repro is simply: Arabic-language journey + a Text Response block + open as visitor.

Console error:

```
TypeError: Cannot read properties of undefined (reading 'push')
  at … stylis Tokenizer.js / Middleware.js / Serializer.js
  at emotion-cache insert
```

## Root cause

Emotion compiles styles with its own bundled **stylis 4.2.0** (exact pin inside `@emotion/cache`). But `createEmotionCache` (`libs/shared/ui`) feeds that pipeline the `prefixer` + `stylis-plugin-rtl` resolved from the workspace's **stylis 4.3.6** — only when the cache is created in RTL mode, which `apps/journeys/pages/_app.tsx` does whenever the journey's language is RTL.

stylis 4.3.x added a `siblings` field to its CSS nodes. Nodes compiled by emotion's 4.2.0 don't have it, so when the 4.3.6 prefixer's `lift()` helper processes a bare `::placeholder` / `:read-only` rule it does `element.siblings.push(…)` → `undefined.push` → the page dies. LTR journeys never enter this code path, which is why only RTL journeys crash.

The mismatch has been latent since stylis was bumped to `^4.3` in #1864 (2023); it became reachable once bare `::placeholder` rules entered the journey viewer.

## Fix

Pin the workspace `stylis` to `4.2.0` — the exact version emotion ships and the pairing MUI documents for `stylis-plugin-rtl` — so one version serves the whole pipeline. This also fixes the identical pattern in `FramePortal` (`libs/journeys/ui`).

Deliberately a direct-dependency pin rather than a global pnpm override: the only other `stylis@4.3.6` consumer is `@mermaid-js/mermaid-cli` (dev-only diagram tooling, never in app runtime), and there's no reason to touch its internals.

The lockfile diff also contains a handful of `@babel/runtime` 7.28.4 → 7.29.2 snapshot dedupes — pnpm consolidating onto a version already present in the lockfile, no new packages.

## Test illustrating failure → fix

New regression spec `libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx` compiles bare `&::placeholder` and `&:read-only` rules through the RTL cache (the exact crashing recipe) and asserts the rtl plugin actually flipped `padding-left` → `padding-right`, so the pipeline can't silently regress to "no plugins".

**Before the pin** (stylis 4.3.6, original lockfile) — reproduces the production error exactly, LTR control passes:

```
× createEmotionCache > compiles ::placeholder and :read-only rules in RTL mode without crashing
  → expected [Function] to not throw an error but
    'TypeError: Cannot read properties of undefined (reading 'push')' was thrown
✓ createEmotionCache > leaves styles unflipped in LTR mode
Tests  1 failed | 1 passed (2)
```

**After the pin**:

```
✓ createEmotionCache > compiles ::placeholder and :read-only rules in RTL mode without crashing
✓ createEmotionCache > leaves styles unflipped in LTR mode
Tests  2 passed (2)
```

Also verified: `FramePortal.spec.tsx` (the other RTL cache consumer) and all 44 `libs/shared/ui/src/libs` spec files pass against stylis 4.2.0.

## QA

Per the ticket's confirmed repro recipe:

1. Journey whose **language** is any `ar`/`ar-*` entry (e.g. Arabic Modern Standard), with a **Text Response block** on a card → open as visitor. Expect: renders RTL, no application error.
2. Same journey with **chat enabled** (`showAssistant`) → chat input (`PromptInput`, the viewer's only bare `::placeholder`) renders without crashing.
3. LTR journey smoke test — styles unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated styling dependency to resolve potential right-to-left (RTL) rendering issues and ensure correct CSS property flipping during compilation.

* **Tests**
  * Added comprehensive test coverage validating RTL functionality, CSS property handling, and dependency version compatibility to prevent future regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -234,7 +234,7 @@
     "sharp": "^0.33.0",
     "short-unique-id": "^5.2.0",
     "slugify": "^1.6.6",
-    "stylis": "^4.3.0",
+    "stylis": "4.2.0",
     "stylis-plugin-rtl": "^2.1.1",
     "subscriptions-transport-ws": "^0.11.0",
     "swiper": "^12.0.0",
--- a/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
+++ b/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
@@ -40,10 +40,12 @@ describe('createEmotionCache', () => {
 
   it('keeps the workspace stylis pinned to the version @emotion/cache bundles', () => {
     const requireModule = createRequire(import.meta.url)
-    const workspaceStylisVersion: string =
-      requireModule('stylis/package.json').version
-    const emotionBundledStylisVersion: string =
-      requireModule('@emotion/cache/package.json').dependencies.stylis
+    const workspaceStylisVersion: string = requireModule(
+      'stylis/package.json'
+    ).version
+    const emotionBundledStylisVersion: string = requireModule(
+      '@emotion/cache/package.json'
+    ).dependencies.stylis
 
     expect(
       workspaceStylisVersion,
--- a/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
+++ b/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
@@ -26,7 +26,7 @@ const InputWithPseudoRules = styled('input')({
 })
 
 function getInsertedCss(): string {
-  return Array.from(document.head.querySelectorAll('style'))
+  return Array.from(document.head.querySelectorAll('style[data-emotion]'))
     .map((style) => style.textContent)
     .join('\n')
 }
--- a/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
+++ b/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
@@ -1,3 +1,5 @@
+import { createRequire } from 'node:module'
+
 import { CacheProvider } from '@emotion/react'
 import styled from '@emotion/styled'
 import { render } from '@testing-library/react'
@@ -34,6 +36,22 @@ describe('createEmotionCache', () => {
     document.head
       .querySelectorAll('style[data-emotion]')
       .forEach((style) => style.remove())
+  })
+
+  it('keeps the workspace stylis pinned to the version @emotion/cache bundles', () => {
+    const requireModule = createRequire(import.meta.url)
+    const workspaceStylisVersion: string =
+      requireModule('stylis/package.json').version
+    const emotionBundledStylisVersion: string =
+      requireModule('@emotion/cache/package.json').dependencies.stylis
+
+    expect(
+      workspaceStylisVersion,
+      'The workspace stylis (root package.json) must exactly match the ' +
+        'version @emotion/cache depends on. Mixing versions crashes every ' +
+        'RTL-language journey (NES-1728). If an @emotion upgrade changed ' +
+        'its bundled stylis, re-align the root package.json pin.'
+    ).toBe(emotionBundledStylisVersion)
   })
 
   it('compiles ::placeholder and :read-only rules in RTL mode without crashing', () => {
--- a/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
+++ b/libs/shared/ui/src/libs/createEmotionCache/createEmotionCache.spec.tsx
@@ -1,0 +1,66 @@
+import { CacheProvider } from '@emotion/react'
+import styled from '@emotion/styled'
+import { render } from '@testing-library/react'
+
+import { createEmotionCache } from '.'
+
+// NES-1728 regression: emotion compiles styles with its own bundled stylis
+// (pinned 4.2.0). The prefixer and rtl plugin passed in by createEmotionCache
+// come from the workspace's stylis. If those versions diverge (e.g. 4.3.x),
+// the 4.3.x prefixer reads `element.siblings` — a field 4.2.0 nodes don't
+// have — and crashes with "Cannot read properties of undefined (reading
+// 'push')" whenever a bare ::placeholder or :read-only rule compiles in RTL
+// mode. In production this took down every RTL-language journey containing a
+// text input. The workspace stylis must stay pinned to the version emotion
+// ships.
+const InputWithPseudoRules = styled('input')({
+  paddingLeft: '8px',
+  '&::placeholder': {
+    opacity: 0.4
+  },
+  '&:read-only': {
+    opacity: 0.6
+  }
+})
+
+function getInsertedCss(): string {
+  return Array.from(document.head.querySelectorAll('style'))
+    .map((style) => style.textContent)
+    .join('\n')
+}
+
+describe('createEmotionCache', () => {
+  afterEach(() => {
+    document.head
+      .querySelectorAll('style[data-emotion]')
+      .forEach((style) => style.remove())
+  })
+
+  it('compiles ::placeholder and :read-only rules in RTL mode without crashing', () => {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -234,7 +234,7 @@
     "sharp": "^0.33.0",
     "short-unique-id": "^5.2.0",
     "slugify": "^1.6.6",
-    "stylis": "^4.3.0",
+    "stylis": "4.2.0",
```
jianwei1, line 237: **Concern (minor):** Exact `"4.2.0"` is the right call today, but the real invariant is "match whatever stylis `@emotion/cache` bundles" — nothing enforces that automatically. If a future `@emotion/*` bump ships a newer bundled stylis, this workspace pin will silently re-diverge and the crash can resurface.

Note the regression spec only guards the *4.3-prefixer-compiling-4.2-nodes* direction (the `undefined.push` crash). The reverse mismatch after an emotion bump (4.2 prefixer on newer nodes) wouldn't necessarily throw, so the test may not catch it. Fine to ship as-is — worth a short `// keep in sync with @emotion/cache's bundled stylis` comment here so a future upgrader knows to re-check.

---

jianwei1 (commented):
## Review summary

Sound, minimal, well-targeted fix. Excellent root-cause write-up (the stylis 4.3.x `siblings`-field divergence from emotion's bundled 4.2.0), a strong regression spec that asserts the RTL pipeline actually flips physical properties rather than silently no-op-ing, and correct scoping (direct pin, dev-only `mermaid-cli` left untouched; `stylis-plugin-rtl` re-resolves to 4.2.0 cleanly).

**No code-level blockers.** One minor maintainability note inline on the pin.

### On the red `lint-work` (type-check) check — false positive, unrelated to this PR

I verified this rather than assume it, because it's the only blocking signal:

- The 31 failures are pure GraphQL nullability errors (`'data.journey' is possibly 'null'`, `GetJourney_journey | null not assignable`) in `apps/journeys/pages/api/oembed.ts`, `index.tsx`, `customize.tsx`, and `OnboardingTemplateCard.tsx` — **files this PR never touches**. A stylis pin cannot produce these.
- Installed deps with this PR's frozen lockfile (stylis 4.2.0 confirmed) and type-checked `journeys`, `journeys-admin`, `resources` — **clean**, both sequentially and in the parallel nx fan-out CI uses.
- Checked out the **exact CI merge ref** (`97affe6`, byte-identical to what CI tested), cleaned all `tsbuildinfo`, re-ran — **all three clean**.

Same source that fails in CI passes locally on an identical tree → CI-environment artifact, most consistent with a stale Nx Cloud cache: this PR changes root `package.json`, busting the cache for all 40 projects, so `type-check` actually executes (against outdated cached lib `.d.ts`) instead of being served a green cache hit as on branches that don't touch root deps.

**Recommendation:** re-run the `lint-work` job (or clear the Nx Cloud cache) rather than change any code. (The earlier `danger` failure was already superseded by three green `danger` re-runs.)

---

package.json
jaco-brink: Agreed on the gap — and since package.json can't carry comments, I went one step stronger than a comment: 75767a53 adds a spec that asserts the workspace stylis version exactly equals the `stylis` dependency `@emotion/cache` declares. That enforces the "match whatever emotion bundles" invariant automatically and catches **both** drift directions (workspace bump or a future `@emotion/*` bump shipping a newer bundled stylis), including the reverse-mismatch case the crash-repro test wouldn't catch. The failure message points the future upgrader at the re-alignment step. Also left a breadcrumb comment at the `stylisPlugins` site in `createEmotionCache.ts`.

jianwei1 (dismissed):
## Approved ✅

Correct, minimal, well-guarded fix. The follow-up commits address my earlier inline note: the same-stylis-version invariant is now both documented at the call site (`createEmotionCache.ts`) and enforced programmatically by the new `keeps the workspace stylis pinned to the version @emotion/cache bundles` test.

Verified, not just read:
- `@emotion/cache@11.14.0` declares an exact `stylis: "4.2.0"`, workspace resolves to `4.2.0`, and the full spec passes **3/3** via the nx vitest executor.
- The three apps that showed a red `lint-work` on the earlier commit type-check clean on the exact CI merge ref — that failure was a transient Nx-cache artifact, and `lint`/type-check is green on this head.
- Regression spec reproduces the exact crash recipe (bare `&::placeholder` / `&:read-only` through the RTL cache) and asserts the rtl plugin actually flips `padding-left` → `padding-right`, so it can't silently regress to a no-plugin pipeline.

One optional nit (not blocking): the pin test compares against emotion's *declared* dependency string with `.toBe()`, so if a future `@emotion` release ever declares a range (e.g. `^4.2.0`) instead of an exact pin, the test would fail spuriously. Low probability and the failure message is clear — fine to leave as a "re-verify on emotion bump" tripwire.

Recommend running the ticket's QA repro on the `journeys-9298` preview (Arabic journey + Text Response block, and with chat enabled) before merge as a final live confirmation, but nothing in the code blocks approval.

---

jaco-brink:
## Review feedback addressed (0ef92f2e8)

**Fixed:**
- `libs/shared/ui/.../createEmotionCache.spec.tsx:31` (CodeRabbit) — scoped `getInsertedCss()` to `style[data-emotion]` so unrelated injected CSS can't skew assertions; matches the `afterEach` cleanup selector. (0ef92f2e8)

**Fixed (adjusted):**
- `package.json:237` (@jianwei1) — the "keep in sync with @emotion/cache's bundled stylis" concern. JSON can't carry comments, so instead of the suggested comment the invariant is now **enforced by a test** (75767a53): the spec asserts the workspace stylis version exactly equals the `stylis` dependency `@emotion/cache` declares, catching both drift directions — including the reverse mismatch the crash-repro test wouldn't throw on. Breadcrumb comment added at the `stylisPlugins` site in `createEmotionCache.ts`. See inline reply.

**Notes:**
- Rebased onto the `autofix-ci` formatting commit (f046d35f) before pushing; no semantic changes.
- All 3 spec tests green locally; CI was fully green pre-push (the earlier `lint-work` failure resolved — it was a stale Nx Cloud cache artifact, as diagnosed in the review).
- QA's "Urdu/Hebrew still broken" reports were traced to the wrong deployment (`journeys-9292` preview, which lacks this fix) — details on [NES-1728](https://linear.app/jesus-film-project/issue/NES-1728/rtl-rendering-error-in-arabic-chat-page); re-test requested on the `journeys-9298` preview.

jianwei1 (approved):
## Approved ✅ (re-confirming on `0ef92f2`)

The only change since my prior approval is a one-line test refinement — `getInsertedCss()` now scopes to `style[data-emotion]` instead of all `<style>` tags, matching the `afterEach` cleanup so the assertion can't be perturbed by unrelated style elements. Production code, the `stylis` 4.2.0 pin, and the three test cases are unchanged.

Re-verified on this head:
- Spec passes **3/3** via the nx vitest executor.
- Full CI suite is green (`lint`/type-check, `danger`, all E2E, all deploy previews, GraphQL Hive, CodeQL).

No outstanding concerns. Across the review rounds the author has addressed every item — the invariant is documented at the call site, enforced by a version-pin test, and the test internals are now precise. Good to merge once the remaining required reviewer signs off; recommend the ticket's RTL QA repro on the `journeys-9298` preview as a final live check.

---

PR #9218, opened 2026-05-15
feat(journeys-admin): per-card showAssistant + expandChatByDefault editor toggles [NES-1557]

jaco-brink:
## Summary
- Adds an `AI chat` accordion to the card properties panel, gated by the `apologistChat` LaunchDarkly flag.
- Two MUI switches: `Show AI chat` (showAssistant) and, when enabled, `Open chat automatically` (expandChatByDefault).
- Toggling `Show AI chat` off also clears `expandChatByDefault` in the same mutation so the auto-open flag doesn't linger under a hidden chat button.
- Persists via the existing `cardBlockUpdate` mutation; wired through `useCommand` for undo/redo with optimistic response.
- Closes [NES-1557](https://linear.app/jesus-film-project/issue/NES-1557).

API + codegen for these fields are already on main from NES-1556 (#9145) and NES-1622 (#9178); this PR is editor-UI only — no schema, migration, or fragment changes.

## Test plan
- [x] `npx jest --config apps/journeys-admin/jest.config.ts --no-coverage 'apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/ChatAssistant/ChatAssistant.spec.tsx'` — 7/7 pass
- [x] `npx jest --config apps/journeys-admin/jest.config.ts --no-coverage 'apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/Card.spec.tsx'` — 38/38 pass
- [x] `npx nx run journeys-admin:lint` — 0 errors
- [x] `npx nx run journeys-admin:type-check` — clean
- [ ] Manual: with `apologistChat` flag off, accordion is hidden.
- [ ] Manual: with flag on, toggles render, persist, accordion summary reflects state (Off / On / On, auto-open), undo restores prior state.
- [ ] Manual: toggling `Show AI chat` off also clears `expandChatByDefault` server-side.

## Summary by CodeRabbit

* **New Features**
  * Added AI chat assistant configuration for card blocks behind a feature flag
  * New "Show AI chat" toggle and "Open chat automatically" option; chat accordion shown only when enabled

* **Tests**
  * Added comprehensive tests for conditional rendering, toggle behavior, update mutations, and undo flow

* **Localization**
  * Added English translations for new AI chat UI labels and states

[Review Change Stack](https://app.coderabbit.ai/change-stack/JesusFilm/core/pull/9218?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #9302, opened 2026-06-15
feat(NES-1734): rename AI chat toggles + make pop-open the default

jaco-brink:
## What

Renames the per-card **AI chat** editor toggles and flips the default so chat **pops open** for viewers.

| Editor setting | Viewer sees |
| --- | --- |
| **Enable AI chat** off | No chat on the card |
| Enable AI chat on, **Collapse chat** off *(default)* | Chat popped open |
| Enable AI chat on, **Collapse chat** on | Chat collapsed (icon only) |

- Toggle 1: `Show AI chat` → **`Enable AI chat`** (label only).
- Toggle 2: `Open chat automatically` → **`Collapse chat`**, meaning inverted, **off by default**.
- The card accordion summary now reads `Off` / `On` / `On, collapsed`.

Resolves **NES-1734**.

## How — backend stays put, a temporary FE mapper bridges

To keep the GraphQL schema / Prisma / kill-switch stable during the World Cup window, the backend fields are **unchanged** (`showAssistant`, `expandChatByDefault`). A new temporary mapper, `@core/journeys/ui/aiChatSettings`, is the single place that translates the editor's `enableAiChat` / `collapseChat` to those legacy fields and owns the inversion + "pop open by default" rule:

- `toAiChatSettings()` (BE→FE): `collapseChat = (expandChatByDefault === false)` — so `null` / `undefined` / `true` all mean "pop open".
- `toCardBlockAiChatInput()` (FE→BE): `expandChatByDefault = !collapseChat`.

The backend rename + mapper removal is tracked in **NES-1735** (Cleanup & Tech Debt); the mapper carries a code comment pointing there.

## Migration — zero data migration

No backfill runs. Each existing card's behaviour follows its stored `expandChatByDefault`, which the mapper reads as `collapseChat = (expandChatByDefault === false)`:

| Stored `expandChatByDefault` | How it got there | New viewer behaviour |
| --- | --- | --- |
| `null` | Chat-enabled but never touched by the #9218 toggles | **Pops open** (the new default) |
| `true` | Auto-open explicitly toggled on in #9218 | **Pops open** |
| `false` | Chat enabled via the #9218 editor — that editor saved `expandChatByDefault: false` on enable | **Stays collapsed** (`On, collapsed`) |

So this is **not** a blanket "every chat-enabled card now pops open" — only `null` and `true` cards do. Cards enabled through the #9218 editor (shipped 2026-05-19) without toggling auto-open hold `false` and keep starting collapsed. That **preserves their existing no-auto-open behaviour — not a regression**, but it does mean the pop-open default reaches only untouched (`null`) and explicitly-expanded (`true`) cards.

A one-time `false → null`/`true` backfill would be needed to make the `false` cohort pop open too; it's deliberately out of scope here (and `false` can't distinguish "just enabled" from a genuine collapse opt-out, so a blanket flip would override real opt-outs — better handled as a separate, deliberate decision rather than during the World Cup freeze).

## Scope vs NES-1727

This PR covers the editor labels, the saved setting + its default, and the **viewer reading** the new default (`Conductor` honours collapse via the mapper). The viewer's actual popped-vs-collapsed entry-point UI lives with **NES-1727**.

## Testing

- ✅ `aiChatSettings` mapper unit tests (9) — inversion + null handling + round-trip.
- ✅ `ChatAssistant` editor specs (7) — new labels, default-flip on enable, collapse save, undo.
- ✅ `Card` accordion specs (40) — `Off` / `On` / `On, collapsed`.
- ✅ `Conductor` per-card + apologist specs (18) — pop-open default auto-opens; collapse opt-in does not.
- ✅ `tsc -b` typecheck green for `journeys-ui`, `journeys-admin`, `journeys`.
- ✅ ESLint `--fix` clean on all changed files.
- ⏭️ Full Next.js production builds not run locally (CI-scale); `tsc -b` validated compilation.

## Note

The **"Collapse chat"** wording is Aaron's earlier (not-yet-locked) suggestion — built provisionally; it's a one-line label change if he picks something else.

## Summary by CodeRabbit

* **UI Improvements**
  * Renamed the AI chat controls to **“Enable AI chat”** and **“Collapse chat”** for clearer editor language.
  * Updated the accordion/button text to reflect the current AI chat state (including “On, collapsed” vs enabled defaults).

* **Behavior Updates**
  * Adjusted AI chat open/collapse behavior when preferences aren’t explicitly set, including a “no explicit choice” default that now opens the chat.
  * Ensured collapse preferences persist correctly when toggling AI chat on and off.


libs/journeys/ui/src/libs/aiChatSettings/aiChatSettings.ts
```diff
@@ -0,0 +1,65 @@
...
+ * `collapseChat` is true only when the card has explicitly opted into a
+ * collapsed start (`expandChatByDefault === false`). `null` / `undefined` /
+ * `true` all mean "pop open" — which is also what migrates pre-existing
+ * chat-enabled cards (stored as `expandChatByDefault: null`) to the new
+ * pop-open default with no data migration.
+ */
+export function toAiChatSettings(block: CardBlockAiChatFields): AiChatSettings {
+  return {
+    enableAiChat: block.showAssistant === true,
+    collapseChat: block.expandChatByDefault === false
```
jaco-brink, line 48: **Concern:** The PR description says existing chat-enabled cards "all have `expandChatByDefault: null`" and will auto-adopt pop-open. But the prior editor (NES-1557, #9218) saved `expandChatByDefault: false` on enable — its `handleShowAssistantChange` used `nextExpand = checked ? currentExpandChatByDefault : false`. Any card enabled through that editor holds `false`, which this mapper reads as `collapseChat: true` → it stays collapsed and the card summary shows "On, collapsed", not pop-open.

Worth noting this actually *preserves* those cards' old no-auto-open behaviour, so it isn't a regression — but it contradicts the "all pop open" migration note. Could you confirm production chat-enabled cards are genuinely `null` and not `false`? If broad pop-open is the goal, a one-time backfill (`false` → `null`/`true`) may be needed; if preserving explicit opt-outs is intended, the migration note in the description should be corrected.

jaco-brink (commented):
## Review summary

**Critical:** none
**Concern:** 1 (see inline)

**Overall:** Clean, well-scoped change. The inversion + "pop-open" semantics are isolated into a single documented, round-trip-tested mapper (`aiChatSettings`) that's the single source of truth across editor + viewer; the kill-switch (`hasAiChatButton` / `showAssistant`) stays intact; no orphaned reads of `expandChatByDefault` bypass the mapper; CI is green. Just one migration-data question to confirm before merge.

🤖 Reviewed with Claude Code

libs/journeys/ui/src/libs/aiChatSettings/aiChatSettings.ts
jaco-brink: Confirmed and corrected. I traced #9218: `currentExpandChatByDefault = cardBlock.expandChatByDefault === true` coerces `null`→`false`, and `handleShowAssistantChange` saved `nextExpand = checked ? currentExpandChatByDefault : false` — so enabling chat through that editor persisted `expandChatByDefault: false`. #9218 shipped to main 2026-05-19, so that cohort exists in production and the mapper reads it as `collapseChat: true` → `On, collapsed`, exactly as you flagged.

You're right that this *preserves* old no-auto-open behaviour rather than regressing it, so the code stays as-is. The inaccuracy was only in the migration note. I've rewritten that section to spell out the three cohorts (`null`/`true` → pop open, `false` → stays collapsed) and to state plainly that the pop-open default reaches only `null` and `true` cards — not a blanket flip.

A `false → null`/`true` backfill is left out of scope: `false` can't distinguish "just enabled" from a deliberate collapse opt-out, so a blanket flip would override genuine opt-outs — better as a separate deliberate decision than something run during the World Cup freeze.

jaco-brink:
## Review feedback addressed

**1 unresolved thread — resolved as a description fix (no code change):**

**Fixed (adjusted) — `libs/journeys/ui/src/libs/aiChatSettings/aiChatSettings.ts:48`** — *"all chat-enabled cards have `expandChatByDefault: null` → all pop open"* was inaccurate.

Investigation confirmed the concern:
- #9218's editor coerced `null`→`false` (`cardBlock.expandChatByDefault === true`) and saved `expandChatByDefault: false` on enable (`nextExpand = checked ? currentExpandChatByDefault : false`).
- #9218 shipped to main **2026-05-19**, so cards enabled through it hold `false` in production; the new mapper reads `false` as `collapseChat: true` → `On, collapsed`.
- Prisma confirms `expandChatByDefault Boolean?` has no DB default, so untouched cards are genuinely `null`.

The code is **correct as-is** — it faithfully preserves an explicit `false` as collapsed (old no-auto-open behaviour, not a regression). The only inaccuracy was the migration note, now rewritten to spell out the three cohorts (`null`/`true` → pop open, `false` → stays collapsed) and to state the pop-open default reaches only `null`/`true` cards — not a blanket flip.

A `false → null`/`true` backfill was considered and left out of scope: `false` can't distinguish "just enabled" from a deliberate collapse opt-out, so a blanket flip would override genuine opt-outs — better as a separate deliberate decision, not during the World Cup freeze. (Triage decision: *correct the description* over backfill.)

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/ChatAssistant/ChatAssistant.tsx
```diff
@@ -61,72 +61,62 @@ export function ChatAssistant(): ReactElement | null {
...
-    const execute: ChatAssistantValues = {
-      showAssistant: checked,
-      expandChatByDefault: nextExpand
-    }
-    const undo: ChatAssistantValues = {
-      showAssistant: currentShowAssistant,
-      expandChatByDefault: currentExpandChatByDefault
+    const execute: AiChatSettings = {
+      enableAiChat: checked,
+      collapseChat: current.collapseChat
```
jianwei1, line 99: **Concern (question — likely intended):** Re-enabling chat preserves `current.collapseChat`, so a card whose stored `expandChatByDefault === false` (the `collapseChat: true` cohort) comes back as **"On, collapsed"** rather than the advertised pop-open default.

This is internally consistent with the deliberate "disabling preserves the collapse setting" change, and persisting the opt-in across an off/on cycle is defensible. Two things worth a look:

1. The Collapse sub-toggle is hidden while chat is disabled (`Collapse in={current.enableAiChat}`), so the retained `true` is invisible to the editor until they re-enable — a small UX surprise.
2. The only "enable → pops open" test uses a fresh `expandChatByDefault: null` card; the `false`-cohort re-enable path isn't covered. A test asserting re-enable yields `expandChatByDefault: false` (or pop-open, whichever is intended) would lock the behaviour in.

Can you confirm this is the intended UX? If so, a test for the `false`-cohort re-enable would be a nice guard.

---

jianwei1 (commented):
## Review summary

Clean, well-tested rename + default-flip. CI is green and the abstraction (single temporary `aiChatSettings` mapper with a clear NES-1735 removal pointer) is the right call for bridging the rename without churning the schema during the freeze.

**Concern:** 1 (question, likely intended) — re-enabling the `expandChatByDefault: false` cohort yields "On, collapsed" rather than pop-open; path is untested.

No critical issues. Not blocking.

---

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Card/ChatAssistant/ChatAssistant.tsx
jaco-brink: Confirmed — this is intended.

**Re-enable preserves `collapseChat` (your main point):** This is the same deliberate "disabling preserves the collapse setting" design we settled in the now-resolved `aiChatSettings.ts` thread. A stored `expandChatByDefault: false` is treated as an explicit collapse opt-in, so we preserve it across an off/on cycle rather than blanket-reverting to pop-open. Pop-open remains the default only for the `null`/`true` cohorts — so a `false`-cohort card correctly comes back "On, collapsed".

**Point 1 (sub-toggle hidden while chat is off):** Also intended — we don't surface a sub-setting for a disabled feature. The retained `collapseChat: true` reappears the instant chat is re-enabled, which is the only place it's actionable.

**Point 2 (test gap):** Good call. Added a test in 4d82082a3 covering the `false`-cohort re-enable path: it renders a `{ showAssistant: false, expandChatByDefault: false }` card, toggles Enable AI chat on, and asserts the mutation saves `{ showAssistant: true, expandChatByDefault: false }` (stays collapsed). Locks the behaviour in. Thanks!

jaco-brink:
## Review feedback addressed (4d82082a3)

**Fixed:**
- `ChatAssistant.spec.tsx` — added a test for the `false`-cohort **re-enable** path: a disabled card storing `expandChatByDefault: false`, when Enable AI chat is toggled on, saves `{ showAssistant: true, expandChatByDefault: false }` (stays collapsed, preserving the explicit opt-in). The MockedProvider mock matches only those exact variables, so the test fails if the behaviour ever reverts to pop-open. Locks in the path the existing `null`-cohort "pops open" test didn't cover.

**Confirmed (intended, no code change — see inline reply):**
- Re-enabling preserving `collapseChat` ("On, collapsed" for the `false` cohort) is the deliberate "disabling preserves the collapse setting" design settled in the now-resolved `aiChatSettings.ts` thread: a stored `expandChatByDefault: false` is an explicit opt-in, preserved across an off/on cycle rather than blanket-reverting to pop-open. Pop-open is the default only for the `null`/`true` cohorts.
- The Collapse sub-toggle being hidden while chat is off is intended — we don't surface a sub-setting for a disabled feature; the retained value reappears the instant chat is re-enabled.

**Skipped:**
- `aiChatSettings.ts:48` thread — already resolved; no action.

All 8 ChatAssistant specs pass locally.

jianwei1 (approved):
## Review summary — Approved ✅

Re-reviewed after the latest changes. The prior Concern (re-enabling a `expandChatByDefault: false` card returns "On, collapsed" rather than pop-open) is fully resolved: confirmed as intended behaviour and locked in with a new regression test that asserts re-enabling the `false`-cohort saves `{ showAssistant: true, expandChatByDefault: false }`. The updated undo test stays consistent with the new pop-open default, and its mocked mutation variables line up with the mapper output.

No outstanding Critical or Concern items. CI is green (CLA, CodeRabbit, Danger) and all review threads are resolved.

Nicely handled — the feedback was addressed with a test that pins the exact edge case, not just a reply. The temporary `aiChatSettings` mapper keeps the FE↔BE field translation in one place with a clear NES-1735 removal pointer.

---

PR #9304, opened 2026-06-15
fix(NES-1736): reply in journey language by name + hardcode ESV translation

jaco-brink:
## Why

During Aaron's 11 June review, a non-English message got an English reply even though the Langfuse system prompt has a "respond in the user's language" instruction. Two code-side causes (this PR) plus one prompt-side cause (Langfuse, see below).

Closes NES-1736.

## What changed (code)

**1. Send the language *name*, not the BCP-47 code.**
The app was sending `journey.language.bcp47` (e.g. `"ur"`) into the prompt's `{{language}}` variable, so the model read *"respond in the ur language"* and fell back to English. We now send the human-readable name — resolved from `journey.language.name` (which the journey already carries) via the shared `extractLanguageNames` helper, with precedence `localName → nativeName → bcp47` as a last resort. The about-this-chat link keeps using the code, which it needs for locale copy.

**2. Hardcode the `{{translation}}` variable to ESV.**
The Langfuse prompt declares `Quote from the {{translation}} Bible.`, but the app never passed `translation`, so the literal `{{translation}}` placeholder leaked into the compiled prompt. We now always compile `{ translation: 'ESV' }` (plus `language` when known) and name ESV in the local fallback too. Centralised as `BIBLE_TRANSLATION` so it's a one-line swap later.

**3. Fallback wording.** The local fallback prompt (used when Langfuse is unavailable) now says *default to the journey language, but answer in the language the user types* — matching the intended Langfuse behaviour.

**4. Diagnostic logging for unresolved languages.**
The client now also sends the raw `languageBcp47` code alongside `language`. When the value reaching the prompt is *only* that raw code, the journey's language name didn't resolve, so the chat route logs `event: chat_language_unresolved` (pino → Datadog, `service:journeys`) with the code and the shared chat log context. This gives a queryable record of which codes lack a usable name so we can fix the data. Prompt behaviour is unchanged — the code stays the last-resort fill so `{{language}}` never leaks a literal placeholder.

## ⚠️ Required companion change — Langfuse prompt (not in this repo)

This PR cannot fully close the ticket on its own. In the Langfuse `apologist-world-cup-chat` prompt, the **OUTPUT STYLE** line must change:

```diff
- - Respond in the {{language}} language only.
+ - Default to responding in the {{language}} language. If the user writes their message in a different language, respond in that language instead.
```

The current `…only` hard-locks to the journey language and actively breaks the "user-typed language overrides" requirement. `{{translation}}` now receives a value, so that line works as-is.

## Testing

- `apps/journeys` chat route: **50/50** pass (translation + `chat_language_unresolved` warn assertions).
- `libs/journeys/ui` AiChat: **23/23** pass (language name vs code, prefers localized name, falls back to code, sends raw code for diagnostics).
- ESLint: 0 errors on changed files. Types verified against the generated `JourneyFields_language` type.

## Outstanding (QA)

- After the Langfuse prompt edit ships, QA in ≥2 non-English languages, noting any the model can't reliably produce (AC #4). Requires the live model — not coverable in unit tests.

## Notes

- Left the shared `JOURNEY_FIELDS` fragment unchanged on purpose — filtering `language.name` to English (`languageId: 529`) would ripple into journeys-admin. Current precedence already yields a usable language name.

## Release Notes

* **Enhancements**
  * Improved language resolution in chat requests with better fallback handling when language names are unavailable
  * Enhanced language detection to use readable language names when available, with graceful fallback to language codes
  * Refined system prompt behavior for more robust multi-language chat support


libs/journeys/ui/src/components/AiChat/AiChat.tsx
```diff
@@ -232,9 +234,27 @@ export function AiChat({
...
+
+  // Send the journey's language to the assistant as a human-readable name
+  // (e.g. "Urdu"), not the BCP-47 code: the model resolves a language name far
+  // more reliably than a code. The system prompt defaults to this language but
+  // still answers in whatever language the user actually types (NES-1736).
+  // Prefer the localized name, then the native name, then the code.
+  const { localName, nativeName } = extractLanguageNames(
+    journey?.language?.name ?? []
+  )
+  const language = localName ?? nativeName ?? languageBcp47 ?? undefined
```
jianwei1, line 266: **Concern (low):** `journey.language.name` is queried with no `languageId` filter (`journeyFields.tsx:26`; schema is `name(languageId: ID, primary: Boolean)`), so at runtime this array holds *every* translation of the language's name (native primary + English + Spanish + French + …). `extractLanguageNames` returns `localName = name.find(p => !primary)` — the **first non-primary entry in API order** — so the name we send is non-deterministic and not guaranteed to be the viewer-localized/English one the comment on line 245 implies.

This is functionally fine — any human-readable name beats a BCP-47 code for the model, which is the goal — but the comment overstates the guarantee. Suggest rewording "Prefer the localized name…" to something like "the first non-primary translation, else the native name, else the code." No code change needed.

---

apps/journeys/pages/api/chat/index.ts
```diff
@@ -503,11 +534,18 @@ function buildFallbackSystemMessage({
     'You are a helpful Christian apologist and spiritual guide.',
     'Be warm, empathetic, and conversational.',
     'Support your answers with relevant Bible passages when appropriate.',
+    `Quote from the ${BIBLE_TRANSLATION} Bible.`,
```
jianwei1, line 537: **Nit:** This line is appended unconditionally, so a non-English-defaulted journey gets "Default to responding in Spanish … Quote from the ESV Bible" — naming an English-only translation while responding in another language. The model will translate quotes and ESV-hardcoding is the ticket's intent, so this is acceptable; just flagging the English-translation/non-English-response pairing in the fallback path in case it matters for QA.

---

jianwei1 (commented):
## Review summary

Tight, well-scoped change — clear intent and an unusually thorough description. Verified the import path, the `extractLanguageNames` semantics, and the GraphQL fragment.

**Critical:** none.

**Concerns:**
- 1 inline (AiChat.tsx) — the language name sent is non-deterministic (first non-primary translation in API order), and the inline comment overstates it as "the localized name." Functionally fine; comment wording should match reality.
- **Process (not blocking the code):** this PR alone does *not* close the user-language-reply behavior — it fixes the `{{translation}}` placeholder leak and the code-vs-name issue, but "reply in the language the user types" stays broken until the companion Langfuse prompt edit (`…only` → `Default to…`) ships. The PR is transparent about this; flagging so whoever merges tracks the out-of-repo change and the QA item rather than treating green-on-merge as fixed.

**Nit:** 1 inline (index.ts) — ESV named unconditionally in the fallback for non-English journeys. Acceptable per ticket intent.

**Positive:** strong, behavior-focused test coverage (both warn branches included); `chat_language_unresolved` is a clean PII-free diagnostic reusing `chatLogContext`; `BIBLE_TRANSLATION` centralization prevents prompt/fallback drift.

Overall: good to merge once the Langfuse prompt companion change is confirmed.

---

libs/journeys/ui/src/components/AiChat/AiChat.tsx
jaco-brink: Good catch — the comment did overstate the guarantee. `journey.language.name` is queried without a `languageId` filter (`journeyFields.tsx`), so this resolves to the first non-primary translation in API order, not a viewer-localized name. Reworded the comment in ee91ae3d5 to describe the actual fallback chain (first non-primary → native/primary → BCP-47 code). No behavior change — any human-readable name still beats the code for the model.

apps/journeys/pages/api/chat/index.ts
jaco-brink: Confirmed, and intentional — ESV hardcoding is the ticket's intent (NES-1736), and the model translates quoted passages into the response language, so the ESV-named / non-English-response pairing is expected behaviour. Leaving the unconditional `Quote from the ESV Bible.` line as-is. Flagging the English-translation / non-English-response pairing for QA in the PR summary so it's on the radar during testing.

jaco-brink:
## Review feedback addressed (ee91ae3d5)

Two automated review threads (Claude Code) triaged. Both were low-severity; one comment-accuracy fix, one intentional-by-design.

**Fixed (adjusted):**
- `libs/journeys/ui/src/components/AiChat/AiChat.tsx:245` — the comment claimed "Prefer the localized name," but `journey.language.name` is queried without a `languageId` filter, so `extractLanguageNames` returns the *first non-primary translation in API order* — not a viewer-localized name. Reworded the comment to describe the real fallback chain (first non-primary → native/primary → BCP-47 code). No behavior change; the reviewer noted no code change was strictly needed, but the comment was genuinely inaccurate.

**Acknowledged — intentional, no change (thread resolved):**
- `apps/journeys/pages/api/chat/index.ts:537` — `Quote from the ESV Bible.` is appended unconditionally, so a non-English journey gets an English-named translation alongside a non-English response instruction. This is the ticket's intent (ESV hardcoding, NES-1736) and the model translates quoted passages into the response language.

**⚠️ QA flag:** For non-English journeys, the system prompt pairs `Quote from the ESV Bible` (English translation) with `Default to responding in <language>`. Expected behaviour is that scripture quotes are translated into the journey's language while remaining ESV-sourced — worth a spot-check during QA (e.g. a Spanish or Urdu journey) to confirm quotes render in-language.

jianwei1 (approved):
## Approving — code LGTM ✅

Both prior review items are resolved:
- **AiChat.tsx language-name comment** — reworded in `ee91ae3d5` to accurately describe the `first non-primary → native → BCP-47 code` fallback chain. No behavior change, which was the right call.
- **Unconditional ESV in fallback** — acknowledged as intentional (NES-1736); the model translates quoted passages into the response language, and it's flagged for QA.

Clean, well-tested change.

**Scope of this approval:** it covers the code in this repo only. Two things sit outside it and are *not* verified by this approval:
- The **Langfuse prompt companion edit** (`…only` → `Default to…`) is required to fully close NES-1736 — merging this code alone won't change the live reply-language behavior.
- **Live-model QA in ≥2 non-English languages** (AC #4) is still outstanding.

Worth confirming CI is green before merge.

---

PR #9321, opened 2026-06-20
fix(api-media): make team resolution best-effort on asset upload [QA-529]

csiyang:
## Problem

Custom **image** uploads in the NextSteps editor fail with `INTERNAL_SERVER_ERROR` from api-media ([QA-529](https://linear.app/jesus-film-project/issue/QA-529)). Video upload is unaffected.

## Root cause

The `createCloudflareUploadByFile` resolver resolves a `teamId` from the journey being edited via `maybeResolveTeamId`, which reads the **journeys database from inside api-media** (`@core/prisma/journeys/client`). That cross-database read is failing in production, and because the lookup *threw*, it took the whole upload down with it.

Two changes lined up to surface it:
- NES-1633 (#9272) added the `teamId` resolution + the api-media→journeys DB dependency. It stayed dormant because nothing sent a `journeyId`.
- NES-1634 (#9303) made the editor start sending `journeyId`, which activates the failing path on every editor image upload.

Video only escaped because its frontend doesn't send a `journeyId` yet — the same trap is dormant in the Mux resolvers.

## Fix (mitigation)

Make `maybeResolveTeamId` **best-effort**: if the team lookup fails for any reason (journeys DB unreachable, journey missing, or caller not a member), log a warning and return `null` so the upload still completes, tagged as a **personal** asset.

- The asset is still owned by the uploader via `userId`; a null team only means it won't surface in the team-shared library (`teamId` is already a nullable column and is not exposed in GraphQL).
- The **read** path (`getMyCloudflareImages` with an explicit `teamId`) keeps its strict `assertTeamMembership` check — that boundary guards against viewing another team's assets and is unchanged.
- The warning log keeps the underlying outage diagnosable in production.

No schema changes — this is pure resolver logic, so no codegen/migration is involved.

## Follow-up (not in this PR)

This stops the 500s but does not restore team tagging. The real fix is wiring api-media's connection to the journeys database in production (`PG_DATABASE_URL_JOURNEYS` secret / Terraform apply / security-group path). That's a call for @edmonday, who built NES-1633.

## Tests

Updated the specs that asserted "not a member → blocked" to "not a member → upload succeeds with `teamId: null`", and added a case for the lookup throwing (DB unreachable). All 64 tests across the three affected suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Uploads are now more resilient when team authorization information is unavailable, allowing operations to complete instead of failing.

### DIFF
--- a/apis/api-media/src/lib/journeysAccess/journeysAccess.spec.ts
+++ b/apis/api-media/src/lib/journeysAccess/journeysAccess.spec.ts
@@ -1,6 +1,7 @@
 import { GraphQLError } from 'graphql'
 
 import { journeysPrismaMock } from '../../../test/journeysPrismaMock'
+import { logger } from '../../logger'
 
 import {
   assertTeamMembership,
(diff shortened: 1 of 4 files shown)


edmonday (approved):
Verified the root cause and the fix — approving as a mitigation.

The mechanism checks out: `maybeResolveTeamId` → `resolveAuthorizedTeamId` does a cross-DB read into the journeys DB, and the unguarded throw propagated to INTERNAL_SERVER_ERROR on every editor image upload. Wrapping the lookup in try/catch → null genuinely stops the 500s, and the read path keeps its strict `assertTeamMembership` check. Same dormant trap in the Mux resolvers is correctly noted.

One correction to the follow-up section: `PG_DATABASE_URL_JOURNEYS` is already declared in `apis/api-media/infrastructure/locals.tf:36`, so the Terraform *code* isn't missing. The likely prod issue is the Doppler **prd** config missing that secret (which would have failed the `terraform apply` at the SSM-parameter step) or the security-group path to the journeys RDS being blocked — not a missing Terraform declaration.

Follow-up to restore team tagging tracked in NES-1743 (Media Library Picker → v1.1 team-shared visibility), to work through with Mike.

Behavioral note for whoever merges: the catch collapses 'journeys DB unreachable' and 'caller genuinely not a member' into the same 'tag as personal' path, so a real authz miss is now silent — fine for the editor flow, but worth tightening once the connection is restored (captured in NES-1743).

PR #9322, opened 2026-06-20
fix(api-media): use native fetch instead of node-fetch for image/unsplash [QA-530]

edmonday:
## Problem
api-media 500s on **custom image upload**, **AI image generation**, and **segmind generation** in stage/prod (not reproducible locally). Datadog:

```
Invalid response body while trying to fetch
https://api.cloudflare.com/client/v4/accounts/.../images/v2/direct_upload: Premature close
    at Gunzip.<anonymous> (.../node-fetch@2.7.0/.../node-fetch/lib/index.js:400:12)
  service: api-media
```

## Root cause
The fault is **`node-fetch@2.7.0`**, not app logic. Its gzip stream handling (the `Gunzip` frame above) throws "Premature close" when a gzipped response ends before it finishes reading. It's **environment-sensitive** — triggers on the `node:24-alpine` (musl) runtime + container networking, but not on the local glibc devcontainer, which is why it was never reproducible locally.

Two code paths were on node-fetch, in **two different ways** — and that's why a naive fix only half-works:

| | Unsplash | Cloudflare |
|---|---|---|
| How node-fetch got in | we passed it: `createApi({ fetch: nodeFetch })` | the **Cloudflare SDK** pulled it in itself |
| Fix that worked | remove the `fetch:` option → native fetch | remove option **+** `import 'cloudflare/shims/web'` |

For Cloudflare, removing our injection isn't enough: the SDK auto-detects the runtime at load and (seeing Node) loads `cloudflare/_shims/node-runtime.js` → `require("node-fetch")`, so it keeps using node-fetch internally. Importing `cloudflare/shims/web` **before** `cloudflare` makes its `setShims()` register Node 24's native `fetch`/`Response`/`FormData` and win over auto-detection.

**Why native fetch (undici) fixes it:** undici handles gzip decoding + keep-alive termination correctly, so the response stream completes instead of closing prematurely.

## Change
- `unsplash/service.ts` — drop the `fetch: nodeFetch` injection from `createApi`
- `cloudflare/image/service.ts` — drop the node-fetch injection **and** `import 'cloudflare/shims/web'` to force the SDK onto native fetch; `Response` → global type
- specs updated to use the global `Response`

## Why prior attempts missed it
- **#9321** ("best-effort team resolution") wrapped `maybeResolveTeamId` in try/catch — but the throw is in the Cloudflare HTTP call, *outside* that catch. Right instinct, wrong layer.
- **First commit here** fixed Unsplash (direct injection) but not Cloudflare (SDK-internal) — matching what we saw on stage: Unsplash gallery fixed, custom upload still 500ing. Resolved by the web-shim commit.

## Verification
- `nx type-check api-media` ✅
- api-media specs (cloudflare image, unsplash, segmind) ✅
- **Stage:** Unsplash gallery ✅ and custom upload ✅ confirmed working. AI generate + segmind use the same Cloudflare path.

## Out of scope
- Unsplash empty-gallery collection/key behavior (separate; not node-fetch related)

Linear: [QA-530](https://linear.app/jesus-film-project/issue/QA-530/nextsteps-image-upload-ai-generate-still-500-node-fetch-premature)

## Summary by CodeRabbit

* **Chores**
  * Updated internal API client runtime configurations and dependencies to leverage native runtime environments instead of external libraries, improving compatibility and reducing external dependencies.

### DIFF
--- a/apis/api-media/src/schema/cloudflare/image/image.spec.ts
+++ b/apis/api-media/src/schema/cloudflare/image/image.spec.ts
@@ -1,4 +1,3 @@
-import { Response } from 'node-fetch'
 import { type Mocked, type MockedFunction, vi } from 'vitest'
 
 import { CloudflareImage, ImageAspectRatio } from '@core/prisma/media/client'
--- a/apis/api-media/src/schema/cloudflare/image/service.spec.ts
+++ b/apis/api-media/src/schema/cloudflare/image/service.spec.ts
@@ -1,7 +1,6 @@
 import Cloudflare from 'cloudflare'
 import { APIPromise } from 'cloudflare/core'
 import clone from 'lodash/clone'
-import { Response } from 'node-fetch'
 import { vi } from 'vitest'
 import { mockDeep } from 'vitest-mock-extended'
 
--- a/apis/api-media/src/schema/cloudflare/image/service.ts
+++ b/apis/api-media/src/schema/cloudflare/image/service.ts
@@ -1,8 +1,13 @@
+// Force the Cloudflare SDK onto its "web" runtime so it uses Node's native
+// global fetch (undici) + native Response/FormData. Without this, the SDK
+// auto-detects the Node runtime and loads node-fetch@2, whose gzip stream
+// handling throws "Premature close" on the alpine/musl prod runtime. This
+// side-effect import MUST run before `cloudflare` so its setShims() wins over
+// the SDK's auto-detection. [QA-530]
+import 'cloudflare/shims/web'
+
 import Cloudflare from 'cloudflare'
 
-// Use Node's native global `fetch` (undici) rather than node-fetch@2, whose gzip
-// stream handling throws "Premature close" on the alpine/musl prod runtime. The
-// Cloudflare SDK defaults to the global fetch when none is supplied. [QA-530]
 export function getClient(): Cloudflare {
   if (process.env.CLOUDFLARE_IMAGES_TOKEN == null)
     throw new Error('Missing CLOUDFLARE_IMAGES_TOKEN')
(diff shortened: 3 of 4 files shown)


csiyang (approved):
Approving — correct root cause, minimal change, and the swap is behavior-safe.

**The fix is right.** Both the Cloudflare SDK and unsplash-js default to the global `fetch` when none is injected, so on the Node 24 runtime dropping `node-fetch` + `cloudflare/shims/node` is the clean way off node-fetch@2's musl gzip "Premature close".

**The `Response` swap is safe** (the part worth checking): the `Response` from `createImageFromText().asResponse()` is only ever passed straight through to the SDK upload (`images.v1.create({ file })`) — no node-fetch-specific methods (`.buffer()`, `.textConverted()`) are called on it. So moving to the global `Response` changes no behavior, and it actually removes a producer/consumer fetch mismatch (native end-to-end now). This lands exactly on the Datadog path (`createImageBySegmindPrompt → createImageFromText → createImageFromResponse`). Bonus: drops the `nodeFetch as unknown as typeof fetch` cast in unsplash.

**Non-blocking follow-up:** node-fetch@2 is still used elsewhere in api-media with the same musl risk — `lib/slack/chatPostMessage.ts`, `lib/videoCacheReset.ts`, `schema/shortLink/shortLinkDomain/shortLinkDomain.service.ts`. Worth a follow-up to migrate those (or drop the `node-fetch` dep entirely) so the next Premature-close doesn't surprise anyone.

**Caveat:** this only repros on the alpine/musl runtime, so no unit test can prove it — the stage-deploy confirmation you flagged is the real validation. Worth calling that out so reviewers don't expect a covering test.

PR #9323, opened 2026-06-22
fix(api-media): surface historical uploads in the media library picker

csiyang:
## Problem

The flag-gated media library (`getMyCloudflareImages`) only shows assets uploaded after the media-library backend landed (~2026-05-06). Older uploads are invisible in both the Custom and AI tabs, even though the rows still exist.

**Root cause:** `CloudflareImage.isAi` was added (NES-1627) as a nullable column with no value for existing rows, so every pre-feature upload is `isAi = NULL`. The Custom tab queries `isAi = false` and the AI tab `isAi = true`; in SQL `NULL = false` is not true, so historical rows match neither tab. New uploads set `isAi` explicitly, which is why only recent assets appear.

## Fix — handle `null` in the resolver (no data migration)

The Custom tab is "everything that isn't explicitly AI", so we treat `null` as not-AI **at query time**:

- **Custom tab** (`isAi: false`) → matches `false` **or** `null` → `OR [{ isAi: false }, { isAi: null }]`
- **AI tab** (`isAi: true`) → matches `true`
- arg omitted → no `isAi` filter

This surfaces historical uploads with **no schema change and no data migration** — `isAi` stays `Boolean?`.

### Why not backfill the data?

The first cut of this PR backfilled `NULL → false` and made the column `NOT NULL`. We dropped that approach: a full-table `UPDATE` + `SET NOT NULL` takes an `ACCESS EXCLUSIVE` lock on the production table, which is risky (and didn't reliably apply on stage). For this product, "a `null` upload is just a non-AI upload" is permanently true, so resolving it in the query is a complete fix, not a patch — and it carries zero data risk.

## Scope

- `apis/api-media/.../image.ts` — `getMyCloudflareImages` `where` reshaped to an `AND` of the userId/team predicate and a null-tolerant `isAi` predicate.
- `apis/api-media/.../image.spec.ts` — updated predicate assertions + a case proving a `null`-isAi historical row is returned under the Custom tab.

No GraphQL/schema/codegen changes (the `isAi` arg and field already exist; field stays `nullable: true`). Net diff vs `main` is just these two files.

## Verification

- All 84 api-media tests pass (`image.spec.ts`, `video.spec.ts`).
- Confirmed on stage: a user with both pre- and post-feature uploads now sees their historical images in the Custom tab.

## Summary by CodeRabbit

* **Bug Fixes**
  * Historical image uploads now correctly display in the media library's Custom Images tab after being previously hidden.

* **Documentation**
  * Added internal documentation explaining the fix for missing legacy image rows in media queries.


apis/api-media/src/schema/cloudflare/image/image.ts
```diff
@@ -101,10 +101,20 @@ builder.queryFields((t) => ({
...
+              ? { OR: [{ userId: user.id }, { teamId }] }
+              : { userId: user.id },
+            // Treat a null isAi as "not AI". Uploads created before the isAi
+            // column existed are stored as null, and `isAi = false` would
+            // exclude them (NULL != false in SQL) — keeping historical assets
+            // out of the Custom tab. Matching false-or-null surfaces them.
+            isAi === true
+              ? { isAi: true }
+              : isAi === false
+                ? { OR: [{ isAi: false }, { isAi: null }] }
```
edmonday, line 115: **Nit (non-blocking):** the no-filter branch contributes an empty `{}` into the `AND` array (`AND: [{ userId }, {}]`). Harmless in Prisma and the tests lock it in — noting only for the record.

edmonday (approved):
## Review summary

**Critical:** none
**Concerns:** none
**Overall:** Clean, well-reasoned fix. Approving.

Correct root-cause fix for SQL three-valued logic (`NULL = false → NULL`). Resolving it read-side rather than via a full-table `UPDATE` + `SET NOT NULL` is the lower-risk call given the table size and the `set -e` entrypoint silent-rollback failure mode. The top-level `AND` reshape correctly avoids the two-`OR`-keys collision with the existing userId/teamId predicate. Test adds a genuine `null`-isAi row case proving the legacy row is returned, not just a `where`-shape assertion. The abandoned backfill and the general pattern are well documented for future readers.

One cosmetic nit inline — non-blocking.

PR #9328, opened 2026-06-24
fix(journeys): load client-side translations under short locale keys [NES-1551]

csiyang:
## Problem

Client-side-rendered translations in `journeys-admin` (and `journeys`) silently fell back to **English for all non-English locales**. SSR masks it because translations get baked into the HTML, so the bug only surfaces in components that render purely on client interaction — e.g. `RadioOptionEdit`'s placeholder `t('Add your text here...')` showed English instead of the translated string.

### Root cause

Next.js URL locales are short codes (`fr`, `es`, …) but Crowdin writes translation folders in `libs/locales` with full region tags (`fr-FR`, `es-ES`, …). The configs used a **static `localePath`** string and leaned on `fallbackLng` (`fr: ['fr-FR']`) to bridge the gap. That works during SSR, but on the client i18next skips `fr-FR` because it isn't in `supportedLngs` (derived from `locales`). The store ends up:

```json
"fr":    { "apps-journeys-admin": {} },        // empty — what the client reads
"fr-FR": { "apps-journeys-admin": { ... } }    // populated — but ignored client-side
```

## Fix

Convert `localePath` to a **function** that resolves each short locale straight to its region folder at load time (matching the existing `apps/watch` / `apps/resources` pattern), so translations land directly under the short key — no fallback resolution needed. Adds the explicit `ns` array and `serializeConfig: false` required when `localePath` is a function.

Applied to both `apps/journeys-admin` and `apps/journeys` (the ticket flagged journeys as likely-affected; confirmed same pattern).

## Safety notes

- **journeys' `fallbackLng` is kept intact.** `pages/home/legal/about-chat.tsx` derives its `?lang=` language→folder map from it (NES-1724/NES-1731) and treats it as the single source of truth. The new `localePath` function *derives from the same object*, so the two can't drift.
- **Folder names pass through unchanged.** `about-chat` resolves `?lang=fr` to the folder `fr-FR` before calling `serverSideTranslations`; `fallbackLng[lng]?.[0] ?? lng` leaves already-resolved folders untouched. Verified: `fr`→`fr-FR`, `zh`→`zh-Hans-CN`, `zh-Hans-CN`/`fr-FR`→passthrough, `en`→`en`.

## Testing

- Verified both configs load and `localePath` maps every locale correctly (direct node check).
- Confirmed the ticket's key `"Add your text here..."` exists in `libs/locales/fr-FR/apps-journeys-admin.json`.
- Relying on CI for the Vitest suites (local Node version mismatch).

Closes NES-1551

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved translation loading with smarter language/region fallback and dynamic resolution of translation files.
  * Ensured journey-scoped fallback labels load correctly when switching locales, including safer handling of in-flight locale changes.

* **UI Enhancements**
  * Improved localization across editor settings/property values, card metadata, and common form labels (sign-in, registration, password reset).
  * Translated button/icon/alignment/typography values and video-processing error messages.

* **Tests**
  * Added/updated unit and integration tests covering locale fallback resolution and journey-scoped resource loading.


jianwei1 (commented):
## Review summary

**Concerns:** 3 items (2 inline, 1 below)
**Overall:** The root-cause fix is correct and well-designed — converting `localePath` to a function is the right approach and the BCP-47 canonicalisation in `resolveLocaleFolder` is solid. A few things to check before merge.

---

**Concern — icon label strings passed through `t()` without guaranteed translation entries** (`Button.tsx`, `SignUp.tsx`)

```tsx
value={t(
  icons.find(({ value }) => value === startIcon?.iconName)?.label ??
    'None'
)}
```

If `icons.find(...)` returns a label (e.g. `"Heart"`, `"Arrow"`), it's passed directly as a translation key. If that string isn't registered in `apps-journeys-admin.json`, `t()` returns the English key unchanged — same as before, but only accidentally. Worth confirming that icon name strings are (or will be) in the Crowdin source strings; otherwise wrapping them in `t()` creates a false expectation.

---

**Positive:** Clean separation of the `resolveLocaleFolder` utility, deriving `FOLDER_BY_LANGUAGE` from the existing `fallbackLng` to prevent drift is thoughtful, and the `about-chat` passthrough safety note in the PR description is a good call-out.

---

jianwei1 (commented):
**Concern — `journeys-admin` `fallbackLng` covers fewer locales than the `LOCALE_MAP` it replaces**

The old `LOCALE_MAP` (18 entries) mapped these locales that are now absent from `journeys-admin`'s new `fallbackLng` (13 entries):

| Short code | Old folder |
|---|---|
| `ar` | `ar-SA` |
| `hi` | `hi-IN` |
| `bn` | `bn-BD` |
| `am` | `am-ET` |
| `tl` | `tl-PH` |
| `ur` | `ur-PK` |
| `vi` | `vi-VN` |
| `my` | `my-MM` |

`loadJourneyLocaleResources.ts` derives its `FOLDER_BY_LANGUAGE` from the journeys-admin config. If a journey is authored in one of these locales and an editor opens it in the canvas, `resolveLocaleFolder` returns the short code unchanged (no mapping → passthrough), then tries to load e.g. `ar/libs-journeys-ui.json`. If that folder doesn't exist (only `ar-SA` does), the catch swallows the error and the canvas silently falls back to English — the same bug as before, but now for these 8 locales.

The `journeys` config has all of them. Is it intentional that `journeys-admin` doesn't? If all 18 folders exist in `libs/locales`, the admin config should list them all.

---

jianwei1 (commented):
**Concern — stale locale race condition (confirming CodeRabbit's finding)**

The existing unresolved CodeRabbit thread on this file is correct. If `locale` changes before the in-flight `loadJourneyLocaleResources()` promise resolves, the stale promise can still call `setResources`, keying resources under the old locale. Since the i18n instance is initialised with the new `locale`, those stale resources are effectively unreachable and the canvas falls back to English until a re-render triggers a fresh load.

The `memoizedLoadResources` useCallback doesn't help here — a new callback is created per `locale` change, but the old promise still holds a closure over the old `setResources`. A cleanup flag in `useEffect` is the right fix, as suggested in the CodeRabbit thread.

---

csiyang:
Thanks @jianwei1 — all three addressed:

**1. Icon labels passed to `t()` without guaranteed entries** — Verified non-issue. Every label in `Icon.tsx` is a static `t('…')` literal (`Arrow Left`, `Bible`, `Chat`, …), and each key is present in `apps-journeys-admin.json` across locales, so `t(label)` resolves rather than echoing the key. No change.

**2. `journeys-admin` `fallbackLng` covers fewer locales than the old `LOCALE_MAP`** — Confirmed regression, fixed in 6d8398d1e. Expanded `fallbackLng` to one entry per `libs/locales` folder (22, matching the `journeys` config), so `ar`/`hi`/`bn`/`am`/`tl`/`ur`/`vi`/`my`/`zh-Hant` journeys now resolve in the canvas. Added resolver-spec assertions for the newly-covered folders.

**3. Stale-locale race (confirming CodeRabbit)** — Fixed in d027843c3: a `cancelled` flag in the effect cleanup so a superseded load can't commit resources under the old locale.

Appreciate the review.

apps/journeys-admin/src/components/Editor/Slider/Content/Canvas/JourneyLocaleProvider/JourneyLocaleProvider.tsx
```diff
@@ -29,17 +29,13 @@ export function JourneyLocaleProvider({
...
+  const memoizedLoadResources = useCallback(async (): Promise<void> => {
+    return loadJourneyLocaleResources(locale, setResources)
+  }, [locale, setResources])
 
   useEffect(() => {
-    const directoryLocale = LOCALE_MAP[locale] || locale
-    void memoizedLoadResources(directoryLocale)
-  }, [locale, memoizedLoadResources])
+    void memoizedLoadResources()
+  }, [memoizedLoadResources])
```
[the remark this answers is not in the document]
csiyang: Fixed in d027843c3 — replaced the memoized loader with an effect carrying a `cancelled` flag in its cleanup, so a load that resolves after the locale changed no longer calls `setResources`. Thanks!

jianwei1 (dismissed):
All three concerns from the earlier review are resolved:

- **Stale locale race** — cancellation guard in `useEffect` cleanup is correct and clean.
- **Admin locale coverage** — `journeys-admin`'s `fallbackLng` now covers all 22 journey languages, matching `journeys`. The inline comment future-proofs it well.
- **Icon label `t()` wrapping** — harmless; `t()` returns the key as-is when no translation exists.

The core fix (converting `localePath` to a function, deriving `FOLDER_BY_LANGUAGE` from the same config as `about-chat`) is solid. CI green. 



---

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/StepBlockNode.tsx
```diff
@@ -131,14 +134,20 @@ export function StepBlockNode({
...
-              stepId={step.id}
-              block={block}
-              selected={isSelected}
-            />
-          ))}
+          {/* Block rows preview published content, so their default titles
+              (Submit/Button/Option/Subscribe) follow the journey's language to
+              match the canvas and viewer. The step row above ("Default Next
+              Step") is editor chrome and stays in the UI language. */}
+          <JourneyLocaleProvider locale={locale}>
```
jianwei1: **Concern (performance):** `StepBlockNode` renders once per step in the JourneyFlow map, so wrapping here mounts one `JourneyLocaleProvider` — and thus one `i18next` instance + one `loadJourneyLocaleResources` effect — per step node. For a journey with N steps that's N instances all loading the *same* journey locale, each running `createInstance()` + `init()` (twice: once before resources resolve, once after) plus its own `setState`. The dynamic `import()` is webpack-module-cached so the parse/fetch is paid once, but the instance creation and re-renders are per-node. The provider also mounts when `actionBlocks` is empty (the map renders nothing but the wrapper still initializes).

Cleaner would be to hoist a single `JourneyLocaleProvider` around the `JourneyFlow` nodes container so all nodes share one instance, or at minimum guard with `actionBlocks.length > 0 && (...)` so step-less nodes don't spin up an instance. Not a correctness issue — the integration test confirms it renders correctly — just avoidable per-node overhead.

---

jianwei1 (commented):
## Review summary

Solid fix. The root cause (short URL locale `fr` never populated in the client i18next store because translations landed under the region key `fr-FR`, which the client skips) is correctly addressed by converting `localePath` to a function that resolves each short locale straight to its region folder — matching the established `apps/watch`/`apps/resources` pattern. Deriving `FOLDER_BY_LANGUAGE` from the config's own `fallbackLng` is a good drift-prevention call, and it also fixes a latent inconsistency where `zh` resources were keyed under `zh-Hans` while the instance's `lng` was the passed locale. Test coverage is strong (folder resolution + casing, empty-folder no-op, and an integration test asserting real French strings).

**Critical:** none.

**Concern:** 1 — per-node `JourneyLocaleProvider` in `StepBlockNode` mounts one i18next instance + resource-load per step node (see inline). Purely avoidable overhead, not a correctness bug.

CI is fully green (build, test, lint, all E2E, all deploy previews).

---

apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/nodes/StepBlockNode/StepBlockNode.tsx
csiyang: Applied the minimum guard in a211bd700 — nodes without action blocks no longer mount a provider/instance.

Deliberately stopping short of the full hoist: wrapping the JourneyFlow container would put *all* node text ("Default Next Step →", duplicate menu, analytics labels, the "Image" card title) under the journey language, which contradicts the intended split — editor chrome stays in the interface language, only the block-preview rows follow the journey (that split is what QA v2 now documents). Doing the hoist correctly means a shared-instance context that only ActionButton consumes — real machinery for a mechanism that's scheduled for deletion in NES-1757 once default labels become stored content and the locale↔journey link is removed. The remaining per-node cost is one lightweight instance per node-with-buttons, with the dictionary import webpack-cached, so it felt like the wrong trade. Happy to do the shared-instance version if you think the interim window justifies it.

csiyang:
Thanks for the re-review @jianwei1 — closing the loop on both rounds:

**Your latest review:** the one performance concern is addressed with the minimum guard in a211bd700 (nodes without action rows no longer mount a provider). Full reasoning for stopping short of the container-level hoist is on the inline thread — short version: the literal hoist would put map chrome ("Default Next Step →", menus, analytics) under the journey language against the intended interface-vs-journey split, the correct version needs shared-instance machinery, and this whole mechanism is scheduled for removal in NES-1757 once default button labels become stored content and the locale↔journey link is dropped.

**Since your first review, for the record**, the PR's direction settled as: default/fallback text inside the journey (canvas, flow-map block rows, published viewer) consistently follows the **journey's** language — the earlier experiment moving the canvas to the UI language was reverted so existing journeys keep rendering exactly what they publish — and the flow-map block rows were brought in line with the canvas (the map/canvas mismatch QA flagged). Editor chrome stays in the interface language. The longer-term change (store defaults as real labels at creation + backfill existing empty labels + remove the journey-locale rendering) is specced in NES-1757.

CI on the guard commit is in progress (Danger already green); the previous head was fully green including all E2E.

jianwei1 (dismissed):
## Approved ✅

The core fix is correct and well-scoped: converting `localePath` to a function that resolves each short URL locale straight to its region folder populates the client i18next store under the short key (`fr`), fixing the silent English fallback for client-only-rendered strings. It follows the established `apps/watch`/`apps/resources` pattern and derives its folder map from the config's own `fallbackLng`, so it can't drift from `legal/about-chat`'s `?lang=` resolution.

All review feedback has been addressed:
- Stale-locale race → cancellation guard in the effect cleanup.
- Per-node `JourneyLocaleProvider` overhead → now guarded on `actionBlocks.length > 0`, so step nodes with no action rows don't mount an i18n instance.
- Full journey-language coverage → folder map expanded to one entry per `libs/locales` folder.

Test coverage is solid (folder resolution + casing, empty-folder no-op, and an integration test asserting real translated strings with the English-fallback path), and CI is fully green (build, test, lint, all E2E, deploy previews, CodeQL).

---

jianwei1 (approved):

PR #9329, opened 2026-06-24
fix(i18n): make extract-translations globs workspace-root-relative

csiyang:
## Problem

`nx extract-translations <app>` reported "✔ Extraction complete!" / exit 0 while updating **no catalog** — it silently extracted **zero keys**. New `t('…')` strings never reached `libs/locales/en/<namespace>.json`, so source catalogs quietly went stale (fallout from the i18next-parser → i18next-cli migration).

**Root cause:** the nx target runs from the **workspace root** (no `cwd`), but every app's `i18next.config.js` used **cwd-relative** `input` globs (`src/**`, `pages/**`). From the root those match nothing → 0 files scanned → 0 keys → "success" with no changes. It affected every app using this pattern.

Fixes [NES-1723](https://linear.app/jesus-film-project/issue/NES-1723).

## Changes

**1. Workspace-root-relative input globs** — all 6 configs (`journeys-admin`, `journeys`, `watch`, `resources`, `docs`, `libs/journeys/ui`) now prefix their globs with the project path (e.g. `apps/journeys-admin/src/**`), consistent with how the target runs and where `output` already writes.

**2. Recovered 18 silently-dropped keys** — running the corrected extraction surfaced keys that had been missing across 5 catalogs: `journeys-admin` +6 (incl. the [NES-1707](https://linear.app/jesus-film-project/issue/NES-1707) collection-dialog strings), `watch` +8, `journeys` +2, `resources` +1, `libs-journeys-ui` +1. All additive — no existing keys removed.

**3. Loud-fail guard** — new `i18next.assert-input.js` wraps each config's `input`. On `i18next-cli extract` it globs the patterns (Node 22 built-in `fs.globSync`, no new deps) and **throws** with a diagnostic (cwd + globs + issue ref) if 0 files match, so a future glob misconfiguration fails the target instead of silently passing. Scoped to `extract` only — `status`/`types`/`lint` are unaffected.

## Verification

- `nx run-many --target=extract-translations` across all 6 projects: green; the catalog diff is exactly the 18 recovered keys.
- Guard: passes on real extracts (no warning noise), exits 1 on a cwd-relative (0-match) config with a clear message, stays silent on `status`.
- `apps/arclight`'s `extract-translations` is an empty no-op (`commands: []`), so the 6 configs are the complete set.

## Summary by CodeRabbit

* **New Features**
  * Added new search-related text and loading/empty-state messaging in Watch.
  * Added new gallery, media collection, and chat-off messaging in Journeys and Journeys Admin.
  * Added updated wording for template gallery availability, embed-code help, and legal text formatting.

* **Bug Fixes**
  * Improved translation file handling so text extraction works reliably across apps and libraries, reducing missed strings during localization updates.


jianwei1 (commented):
## Review summary

**Concerns:** 3 items (see inline)
**Nit:** 1 item (see inline)

**Overall:** Solid, well-explained fix. The root cause diagnosis is precise and the workspace-root-relative glob approach is exactly right — consistent with how `output` already worked. Recovering 18 silently-dropped keys additively with no removals is safe. All 6 configs updated consistently. CI is fully green.

The inline comments are questions/suggestions rather than blockers — nothing here should prevent merging once the team has eyes on the `fs.globSync` experimental-API trade-off.

---

jianwei1 (commented):
**Concern — `fs.globSync` is experimental in Node 22**

The comment acknowledges the experimental warning and suppresses it, but `fs.globSync` can still change API between Node releases without a semver-major bump (it carries a stability caveat). If this runs on Node 23+ or a future LTS, behaviour could differ quietly.

Is the "no new deps" constraint firm enough to accept this risk? A safe drop-in would be the `glob` package (likely already a transitive dep in this monorepo), or just accepting the one-time experimental warning rather than suppressing it.

---

jianwei1 (commented):
**Concern — monkey-patching `process.emitWarning` is fragile**

The try/finally ensures it restores correctly in the single-config case. But if multiple configs are ever loaded concurrently (e.g. via a worker pool in future tooling), two concurrent patches would race and one wouldn't restore the original, masking unrelated warnings.

Today configs load sequentially so it works. A cleaner option: just let the experimental warning fire once (it's non-fatal and self-documenting), or use `--no-warnings` in the nx target command if noise is the concern.

---

jianwei1 (commented):
**Concern — `process.argv[2]` detection is implicit**

This correctly identifies the `extract` subcommand for `i18next-cli extract --config ...`, but the assumption is undocumented. If a future CLI version namespaces subcommands differently (e.g. `i18next-cli run extract`), this guard silently becomes a no-op on every invocation — the exact failure mode this PR was built to prevent.

A comment like `// i18next-cli places the subcommand at argv[2]: "i18next-cli <subcommand> [options]"` would make the assumption explicit and easy to catch when upgrading the CLI.

---

jianwei1 (commented):
**Nit — internal ticket reference in user-facing error message**

`\nSee NES-1723.` is useful for the team but opaque to external contributors or CI logs. Worth either appending the Linear URL or replacing with a self-contained description like `"Globs must be relative to the workspace root (the nx target runs there, not the project directory)."` so the error is actionable without needing Linear access.

---

jianwei1 (dismissed):
All previous concerns addressed in the latest commit:

- `fs.globSync` (experimental) → replaced with `glob` package's `globSync`, matching what i18next-cli itself uses
- `process.emitWarning` monkey-patching removed entirely
- `argv[2]` assumption is now documented with an upgrade note
- Error message is self-contained with a full Linear URL

Lockfile changes are routine deduplication from adding `glob@13`. Clean fix — good work.

---

jianwei1 (approved):
Lint fix looks good — the `TemplateGalleryPageMedia` → `TemplateGalleryPageMediaPublic` rename is consistent between the generated type and the test fixture. CI is fully green. Good to merge.

---

PR #9330, opened 2026-06-24
fix(NES-1636): defer team switch so copy-to-team AI translation completes

csiyang:
## Summary

Copying a journey to **another team** with the **AI translation** toggle on left the journey in the destination team **untranslated** ([NES-1636](https://linear.app/jesus-film-project/issue/NES-1636/ai-translation-does-not-work-when-copying-a-journey-to-another-team), Production).

Root cause is a lifecycle race in the shared `CopyToTeamDialog`. After `submitAction` resolved — which, on the translation path, only *starts* the AI-translate subscription — the dialog immediately called `updateTeamState`. That switches the active team and refetches `GetAdminJourneys`, which unmounts the source-team journey card and the `CopyToTeamMenuItem` that **owns the translation subscription**, before translation completes. The duplicate lands in the destination team in the source language, with no feedback.

## The fix — the team switch happens only on translation *success*

Ownership of the team switch moved to whichever component actually knows the outcome:

- **`CopyToTeamDialog`** no longer switches teams on the translation path at all. A **plain copy** (no translation) still switches to the destination team and closes immediately, exactly as before.
- **`CopyToTeamMenuItem`** and **`UseTemplateDeepLink`** switch the active team in their subscription **`onComplete`** (success only). On **`onError`** they clear the pending team and do **not** switch; closing/cancelling the dialog resets state without switching.
- **`CreateJourneyButton`** is unchanged — it navigates to the new journey by id, so it intentionally doesn't switch teams.

Because the switch now runs *after* the subscription completes (instead of right after it starts), the source card stays mounted through the translation, so it actually finishes — and a failed translation no longer drags the user to the destination team.

The shared switch logic (`setActiveTeam` + `updateLastActiveTeamId` + `GetAdminJourneys` refetch) is extracted into a small **`useUpdateActiveTeam`** hook, so the dialog and the two admin consumers share one implementation instead of duplicating it.

## What changed

| File | Change |
| --- | --- |
| `libs/journeys/ui/.../CopyToTeamDialog.tsx` | Stops switching on the translation path; plain-copy path unchanged |
| `libs/journeys/ui/src/libs/useUpdateActiveTeam/*` | New shared hook for the active-team switch + refetch |
| `apps/.../Team/CopyToTeamMenuItem.tsx` | Switches on `onComplete` (success only); no switch on error/cancel |
| `apps/.../UseTemplateDeepLink.tsx` | Same success-only switch before navigating |

## Test coverage

- `CopyToTeamDialog` spec asserts the dialog does **not** switch teams on the translation path (even after it closes).
- `CopyToTeamMenuItem` spec asserts **no team switch on a translation error**, alongside the existing success-path test.
- All consumer suites pass: `CopyToTeamDialog`, `CopyToTeamMenuItem`, `UseTemplateDeepLink`, `CreateJourneyButton`.

## Note

The unreachable `CopyToTeamDialog` inside `DuplicateJourneyMenuItem` (its dialog only opens when `activeTeam == null`, but the menu item only renders when `activeTeam != null`) is intentionally left untouched — the reachable "Copy to ..." flow is `CopyToTeamMenuItem`, which is what this PR fixes.

### DIFF
--- a/apps/journeys-admin/src/components/Team/CopyToTeamMenuItem/CopyToTeamMenuItem.spec.tsx
+++ b/apps/journeys-admin/src/components/Team/CopyToTeamMenuItem/CopyToTeamMenuItem.spec.tsx
@@ -47,6 +47,17 @@ describe('CopyToTeamMenuItem', () => {
   beforeEach(() => {
     handleCloseMenu.mockClear()
     refetchTemplateStats.mockClear()
+    // These result mocks are shared across tests; clear their call counts so
+    // assertions on whether the team switch fired aren't polluted by earlier
+    // tests in this file.
+    ;(
+      updateLastActiveTeamIdMock.result as MockedFunction<() => unknown>
+    ).mockClear()
+    ;(
+      updateLastActiveTeamIdMockForTranslation.result as MockedFunction<
+        () => unknown
+      >
+    ).mockClear()
     mockedUseTemplateFamilyStatsAggregateLazyQuery.mockReturnValue({
       query: [
         vi.fn(),
@@ -498,6 +509,9 @@ describe('CopyToTeamMenuItem', () => {
   })
 
   it('should handle translation errors', async () => {
+    // Clear leaked active-team state so TeamProvider doesn't fire its own
+    // mount-sync updateLastActiveTeamId and pollute the no-switch assertion.
+    window.sessionStorage.clear()
     const translateSubscriptionErrorMock = {
       request: {
         query: JOURNEY_AI_TRANSLATE_CREATE_SUBSCRIPTION,
@@ -593,6 +607,9 @@ describe('CopyToTeamMenuItem', () => {
       expect(screen.getByText('Translation failed')).toBeInTheDocument()
     })
     expect(handleCloseMenu).toHaveBeenCalled()
+    // The team must NOT switch when translation fails — the duplicate exists in
+    // the destination but the user should stay put (NES-1636 follow-up).
+    expect(updateLastActiveTeamIdMock.result).not.toHaveBeenCalled()
   })
 
   it('should call handleKeepMounted when provided', async () => {
--- a/apps/journeys-admin/src/components/Team/CopyToTeamMenuItem/CopyToTeamMenuItem.tsx
+++ b/apps/journeys-admin/src/components/Team/CopyToTeamMenuItem/CopyToTeamMenuItem.tsx
@@ -1,15 +1,11 @@
-import { useMutation } from '@apollo/client'
 import { useTranslation } from 'next-i18next/pages'
 import { useSnackbar } from 'notistack'
 import { ReactElement, useState } from 'react'
 
 import { CopyToTeamDialog } from '@core/journeys/ui/CopyToTeamDialog'
 import { useJourney } from '@core/journeys/ui/JourneyProvider'
-import { useTeam } from '@core/journeys/ui/TeamProvider'
 import { useJourneyAiTranslateSubscription } from '@core/journeys/ui/useJourneyAiTranslateSubscription'
 import { useJourneyDuplicateMutation } from '@core/journeys/ui/useJourneyDuplicateMutation'
-import { UPDATE_LAST_ACTIVE_TEAM_ID } from '@core/journeys/ui/useUpdateLastActiveTeamIdMutation'
-import { UpdateLastActiveTeamId } from '@core/journeys/ui/useUpdateLastActiveTeamIdMutation/__generated__/UpdateLastActiveTeamId'
 import CopyToIcon from '@core/shared/ui/icons/CopyTo'
 
 import { GetAdminJourneys_journeys as Journey } from '../../../../__generated__/GetAdminJourneys'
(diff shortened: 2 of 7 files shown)


jianwei1 (commented):
## Review summary

**Concerns:** 1 item (see inline)
**Overall:** Good fix for a real production race condition. Root cause analysis and PR description are excellent. CI fully green. The one open concern below is the same class of bug as the one being fixed — just on the error path — and is worth addressing before merge.

**Positive:**
- Excellent root cause analysis; the lifecycle race is clearly explained
- Fix is appropriately minimal — only timing changes, no structural rewrites
- New test correctly pins the deferral invariant (`UPDATE_LAST_ACTIVE_TEAM_ID` not called on submit, called once on close)
- `useCallback` on `updateTeamState` is correct since it's now a `useEffect` dep
- Removing `selectedTeamId` from `CopyToTeamMenuItem` is a net simplification

---

jianwei1 (commented):
**Concern:** Deferred team switch fires on error/cancel close, not just success.

`pendingTeamId` is set as soon as `handleSubmit` runs with translation enabled, but the effect commits the switch on *any* `open → false` transition. The error path in `CopyToTeamMenuItem.onError` calls `setDuplicateTeamDialogOpen(false)`, which closes the dialog while `pendingTeamId` is still set — so the team switches even though translation failed.

Concrete flow:
1. User submits with translation → `setPendingTeamId(values.teamSelect)` in dialog
2. Subscription errors → `onError` fires → `setDuplicateTeamDialogOpen(false)`
3. Dialog: `!open && pendingTeamId != null` → `updateTeamState` fires → **team is switched on error**

`pendingTeamId` is dialog-internal, so `CopyToTeamMenuItem` can't clear it before closing. A few options:

- Clear `pendingTeamId` inside the dialog's own `onError`/`onClose` path — e.g. expose a ref or callback so consumers can signal "close without committing"
- Guard the effect with an additional flag (e.g. `pendingTeamId` is only committed if a `translationComplete` boolean is also set)
- The simplest fix: only set `pendingTeamId` in the effect if translation is genuinely done — for example, have `CopyToTeamMenuItem.onComplete` call a prop like `onTranslationComplete(teamId)` instead of relying purely on the close timing

(This is the same concern CodeRabbit raised as unresolved — just adding context on the mechanism.)

---

csiyang:
Addressed @jianwei1's concern (deferred switch firing on error/cancel close) in `82efa4ab2`.

**Fix — the team switch now happens only on translation _success_.** Ownership moved so whichever component knows the outcome does the switch:

- `CopyToTeamDialog` no longer switches or defers on the translation path (a plain copy still switches + closes immediately).
- `CopyToTeamMenuItem` switches in its subscription `onComplete`; `onError` leaves the user put.
- `UseTemplateDeepLink` switches in `onComplete` before navigating; no switch on a translation error.
- `CreateJourneyButton` is unchanged (it navigates to the new journey by id, so it intentionally doesn't switch).

So the flagged flow — submit → translation errors → `onError` closes the dialog → team switches — no longer switches teams.

**Cleanup:** the switch logic (`setActiveTeam` + `updateLastActiveTeamId` + `GetAdminJourneys` refetch) was about to be duplicated across three files, so I extracted it into a shared `useUpdateActiveTeam` hook — the two admin consumers each drop ~15 lines of boilerplate.

**Tests:** the dialog now asserts it does **not** switch on the translation path, and `CopyToTeamMenuItem` asserts **no team switch on a translation error**. 76 tests pass across the dialog + all four `CopyToTeamDialog` consumers.

jianwei1 (commented):
## Follow-up review (new commits)

The concern from the previous review is fully addressed. The approach is cleaner than the original `pendingTeamId` + `useEffect` deferral:

**What changed:**
- `CopyToTeamDialog` no longer owns the team switch for the translation path at all — plain copy only. This removes the "fires on any close" ambiguity entirely.
- `useUpdateActiveTeam` extracts the shared mutation logic (setActiveTeam + updateLastActiveTeamId + refetchQueries) into one place, eliminating duplication between `CopyToTeamDialog`, `CopyToTeamMenuItem`, and `UseTemplateDeepLink`.
- Both translation consumers (`CopyToTeamMenuItem.onComplete`, `UseTemplateDeepLink.onComplete`) call `updateTeamState(selectedTeamId)` on success only; `onError` clears `selectedTeamId` without switching.

**Correctness check:**
- Success path: `onComplete` → `updateTeamState(selectedTeamId)` → correct, once, on success ✓
- Error path: `onError` → `setSelectedTeamId(null)`, no switch ✓  
- Cancel/close path: `onClose` resets `loading`, `translationVariables`, `selectedTeamId` — no switch ✓
- Plain copy: dialog switches immediately as before ✓
- `UseTemplateDeepLink` updated consistently with the same pattern ✓

**Tests:** New `CopyToTeamDialog` test verifies the dialog does not switch teams for the translation path even on close. New `CopyToTeamMenuItem` test pins no-switch on translation error. Both are the right invariants to assert.

**Note on CI:** `Deploy Preview (resources)` and `deploy-status` are failing on this run — the resources deploy failed on the previous run too before passing on a re-run, so this looks like an infra flake rather than a code regression. Worth re-running if it blocks merge.

**Note on PR description:** The description still refers to the `pendingTeamId` / `useEffect` approach — worth a quick update to reflect that the dialog now does nothing on the translation path and consumers own the switch.

No blocking issues. 


---

jianwei1 (approved):
Approved. All concerns addressed — clean fix, good test coverage, ready to merge.

---

PR #9335, opened 2026-06-25
fix(videos-admin): count algolia publish states from facets

tanflem:
## Summary

- add a Label column and single-select label filter to the Videos Admin library
- persist valid label filters in the URL and restore them into grid/GraphQL state
- reset pagination when filters change and ignore malformed or cleared label values without discarding valid filters

## Verification

- `pnpm exec vitest run --config vitest.config.mts src/libs/useVideoFilter/useVideoFilter.spec.tsx --coverage=false` — 19 tests passed
- `pnpm exec nx type-check videos-admin` — passed
- `pnpm exec nx lint videos-admin` — passed with 16 pre-existing warnings and no errors
- headed browser QA on the refreshed preview — pending (`agent-browser` is not installed in the current environment)

No additional operational monitoring required: this is a client-side admin filter backed by the existing `adminVideos` query. For the first day after deploy, the PR owner should validate that selecting and clearing each Label updates the URL and result set without console or GraphQL errors. Roll back this PR if label filtering returns mixed labels, drops unrelated filters, or produces repeated client/query errors.


tanflem:
Closing: the linked ticket VMT-326 was canceled on 2026-08-03, so this work is no longer needed. Reopen if the requirement comes back.

PR #9337, opened 2026-06-29
fix(journeys-admin-e2e): fix failing card level action e2e tests

sushmamalampati:
## Summary
- Fixed failing E2E tests broken by NES-1745 (editor drawer behaviour) and NES-308 (desktop layered editor) changes
- Reverted `clickDoneBtn()` to simple one-liner — Siyang's layered `if` block caused failures on desktop
- Fixed broken `ImageSource + div` selectors — thumbnail is now inside `ImageSource`, not a sibling
- Fixed `imageBlockHeaderDelete` selector to use `ImageBlockHeader` instead of `ImageSource`
- Added reopen step in `verifyImageGotChanged()` to handle NES-1745 auto-close of image library after upload
- Removed `closeToolDrawerForFooterImage()` and `clickleftSideArrowIcon()` calls — no longer needed in layered desktop view

## Files Changed
- `apps/journeys-admin-e2e/src/pages/card-level-actions.ts`
- `apps/journeys-admin-e2e/src/e2e/discover/card-level-actions.spec.ts`

## Tested in Local
Screenshot 2026-06-29 at 12 25 42 PM

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end flows to match the current editor behavior when working with images and footer content.
  * Improved coverage for image upload, gallery selection, and deletion by handling the image panel reopening automatically.
  * Adjusted footer card checks so validation continues smoothly without extra navigation steps.


csiyang (dismissed):

apps/journeys-admin-e2e/src/pages/card-level-actions.ts
```diff
@@ -294,11 +285,16 @@ export class CardLevelActionPage {
...
       this.page.locator(
-        'div[data-testid="ImageSource"] + div div[data-testid="ImageBlockThumbnail"] img'
+        'div[data-testid="ImageSource"] div[data-testid="ImageBlockThumbnail"] img'
       )
     ).not.toHaveAttribute('src', this.uploadedImgSrc, {
       timeout: sixtySecondsTimeout
     })
+    await this.page
+      .locator(
+        'div[data-testid="ImageSource"] button[data-testid="card click area"]'
```
jaco-brink, line 300: **Concern (non-blocking):** `verifyImageGotChanged()` now does two things — it checks the image changed **and** clicks to reopen the image panel. The method name only signals the check. It works here because every current caller needs the panel reopened next, but anyone who later calls this purely to assert will get a surprise click that moves the UI somewhere they didn't expect — and a confusing failure named like a harmless check.

Suggest pulling the reopen into its own step (e.g. `reopenImageSource()`) and calling it explicitly in the spec, so the verify stays a pure check. ~5 min, same clicks in the same order — CI will confirm it still passes.

apps/journeys-admin-e2e/src/e2e/discover/card-level-actions.spec.ts
```diff
@@ -225,7 +225,6 @@ test.describe('verify card level actions', () => {
     await cardLevelActionPage.clickSelectImageBtn() // Select Plus icon to add the Image
     await cardLevelActionPage.selectFirstImageFromGalleryForFooter() //Select first image from the gallery for the Footer section
     await cardLevelActionPage.valdiateSelectedImageWithDeleteIcon() //Verifying the Selected Image with Delete icon is display
-    await cardLevelActionPage.closeToolDrawerForFooterImage() // Close the tool drawer after selected the Image
```
jaco-brink, line 228: **Concern (non-blocking cleanup):** Removing this call leaves `closeToolDrawerForFooterImage()` (page object, ~line 1531) with no callers anywhere in the repo. Since this PR creates the orphan, simplest to delete the method definition here too — zero risk, nothing else references it.

apps/journeys-admin-e2e/src/e2e/discover/card-level-actions.spec.ts
```diff
@@ -240,7 +239,6 @@ test.describe('verify card level actions', () => {
     await cardLevelActionPage.validateWebsiteFooterSectionInCard(footerTitle) //Verifying the Selected Image, Display Title and menu icon are showing at the top of the card
     await cardLevelActionPage.expandJourneyAppearance('Menu') // clicking on the 'Menu' tab from the tab list of footer properties drawer
     await cardLevelActionPage.clickCreateMenuCardButtonInMenuFooter() //Click create menu card button to create menu card
-    await cardLevelActionPage.clickleftSideArrowIcon() //Click Left Chevron button to navigate to react flow panel
```
jaco-brink, line 243: **Concern (non-blocking cleanup):** Same pattern — removing this call orphans `clickleftSideArrowIcon()` (page object, ~line 446); no other callers remain after this PR. Safe to delete the method definition too.

jaco-brink (commented):
## Review summary — automation test fixes

Thanks for getting these green again. The scoping is right: the selector fixes (`ImageSource + div` → descendant) are applied consistently across all four image helpers, and dropping the `LayeredView` branch from `clickDoneBtn` is what the desktop editor needed. CI — including the Playwright journeys-admin run — is passing.

Two small, **non-blocking** follow-ups inline. Both are a direct result of this PR's own edits (not pre-existing), so they're fair to fold in here:
- Two page-object helpers are now orphaned because this PR removed their only callers.
- `verifyImageGotChanged()` now also performs a UI click that its name doesn't advertise.

**Out of scope (not this PR's job):** the page object has long-standing deviations from `.cursor/rules/e2e-testing.mdc` — CSS/`data-testid` selectors instead of user-facing labels, `waitForTimeout` hard waits, and try/catch fallback selector loops. These predate this PR and aren't part of getting automation passing again, so I'm explicitly **not** asking you to touch them here. Worth a separate cleanup ticket if the team wants one.

_FYI only, no action:_ the `clickDoneBtn` change reverts the `LayeredView` branch added under NES-308. CI is green so it looks correct — just flagging for whoever owns that desktop-editor work.

apps/journeys-admin-e2e/src/pages/card-level-actions.ts
jaco-brink: Concrete refactor — split the two responsibilities into two clearly-named methods:

**Page object (`card-level-actions.ts`):**
```ts
// Pure assertion — only checks the image changed, no side effects
async verifyImageGotChanged() {
  await expect(
    this.page.locator(
      'div[data-testid="ImageSource"] div[data-testid="ImageBlockThumbnail"] img'
    )
  ).not.toHaveAttribute('src', this.uploadedImgSrc, {
    timeout: sixtySecondsTimeout
  })
}

// NES-1745: the image library auto-closes after upload/selection,
// so reopen the ImageSource before the next step that needs it open
async reopenImageSource() {
  await this.page
    .locator(
      'div[data-testid="ImageSource"] button[data-testid="card click area"]'
    )
    .click()
}
```

**Spec (`card-level-actions.spec.ts`) — call the reopen explicitly where it's actually needed:**
```ts
await cardLevelActionPage.verifyImageGotChanged() // assert image changed
await cardLevelActionPage.reopenImageSource()     // reopen panel for the next step
await cardLevelActionPage.clickImageSelectionTab('Gallery')
// ...
await cardLevelActionPage.verifyImageGotChanged() // assert image changed
await cardLevelActionPage.reopenImageSource()     // reopen panel for the next step
await cardLevelActionPage.clickImgDeleteBtn()
```

Same clicks, same order — just renamed so each step does what its name says. The reader now sees the reopen happening in the spec instead of it being hidden inside a `verify` helper.

sushmamalampati: Done — split into a pure `verifyImageGotChanged()` assertion and a new `reopenImageSource()` method, and updated the spec to call both explicitly.

apps/journeys-admin-e2e/src/e2e/discover/card-level-actions.spec.ts
sushmamalampati: Done — removed `closeToolDrawerForFooterImage()` from the page object in the latest commit.

apps/journeys-admin-e2e/src/e2e/discover/card-level-actions.spec.ts
sushmamalampati: Done — removed `clickleftSideArrowIcon()` from the page object in the latest commit.

jaco-brink (approved):
Good work 🚀

PR #9339, opened 2026-06-30
fix(api-journeys): count card capture goals & paginate Plausible breakdown

csiyang:
Resolves [QA-537](https://linear.app/jesus-film-project/issue/QA-537/fix-team-templates-analytics-decline-capture-event-naming-mismatch).

Capture counts (Decisions for Christ, Gospel Started, …) on the team-templates breakdown report were declining over time even though they should only ever rise. Two distinct fixes:

## A. Fire registered capture goals from card events
Cards in `Step.tsx` fired the **raw** event labels (`decisionForChrist`, `custom1`, …) straight to Plausible, but only the `*Capture` names are registered goals — so card conversions were missing from the Plausible dashboard's conversions panel. Buttons/videos/radio questions were migrated to `fireCaptureEvent` in #9075; cards never were.

- `Step.tsx` now calls `fireCaptureEvent`, firing the registered capture goal (e.g. `christDecisionCapture`) consistently with the other block types.
- The team-templates report is unaffected — it maps raw→capture and is idempotent for capture names.

## B. Paginate the Plausible breakdown fetch
`getJourneyStatsBreakdown` issued a single un-paginated request, so Plausible returned only the **top 100 rows by visitors**. Low-volume capture rows sit at the bottom of that ranked list and silently dropped past rank 100 as a template family accumulated data — the prime suspect for the decline.

- Page through every row (`limit=1000`, looping `page=1,2,…` until a short/empty page) and combine results.
- An explicit `limit`/`page` from a caller is still honored with a single bounded request, so callers that intentionally request a slice are unaffected.

## Tests
- Updated `Step.spec.tsx` to assert the capture goal is fired and the raw label is not.
- New `service.spec.ts` covering single short page, full multi-page loop, exact-multiple boundary, explicit limit/page passthrough, key mapping, and error handling.
- All 62 `plausible` specs pass locally.

## Summary by CodeRabbit

* **New Features**
  * Enhanced journey analytics breakdown fetching with controlled pagination to handle large result sets and aggregate across all pages.
* **Bug Fixes**
  * Fixed card capture analytics so the correct capture goal is tracked (including proper key mapping) rather than the raw label.
  * Improved Plausible breakdown error handling by returning a clearer, user-facing error for invalid inputs.
* **Tests**
  * Added coverage for pagination edge cases and analytics capture behavior.

### DIFF
--- a/apis/api-journeys/src/schema/plausible/service.ts
+++ b/apis/api-journeys/src/schema/plausible/service.ts
@@ -13,6 +13,13 @@ type PlausibleBreakdownParams = {
   metrics: string
 } & typeof PlausibleStatsBreakdownFilter.$inferInput
 
+// Plausible's breakdown endpoint caps `limit` at 1000 rows per page. To fetch
+// more, callers must paginate with `page=1,2,…`. See
+// https://plausible.io/docs/stats-api#get-apiv1statsbreakdown
+const PLAUSIBLE_MAX_BREAKDOWN_LIMIT = 1000
+
+type PlausibleBreakdownRow = Record<string, number | string | null>
+
 export function buildJourneySiteId(journeyId: string): string {
   return `api-journeys-journey-${journeyId}`
 }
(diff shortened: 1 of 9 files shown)


apis/api-journeys/src/schema/plausible/service.ts
```diff
@@ -29,43 +45,107 @@ export function getPlausibleConfig(): {
...
+      if (rows.length < PLAUSIBLE_MAX_BREAKDOWN_LIMIT || newRows === 0) break
+      if (page === PLAUSIBLE_MAX_BREAKDOWN_PAGES) cappedWithMore = true
+    }
+    if (cappedWithMore) {
+      console.error(
+        '[getJourneyStatsBreakdown] hit max page cap; results may be truncated',
+        { siteId: resolvedSiteId, property: params.property }
       )
-    })
+    }
```
jianwei1, line 152: **Concern (low):** When every page is full and distinct, the loop stops at `PLAUSIBLE_MAX_BREAKDOWN_PAGES` (50k rows), logs to `console.error`, and returns a **truncated** set — a milder recurrence of the original "rows silently dropped" bug, just at a 500× higher threshold than the old top-100 default. Two things worth a conscious sign-off:

1. Is 50k distinct `(event, target, journeyId)` combinations reachable for your largest template families? If so, the aggregate would under-count, and `console.error` alone may not surface to alerting — consider routing the truncation signal to real observability.
2. **Nit:** `cappedWithMore` is a false positive at the exact boundary — when the total is exactly `50 × 1000`, page 50 returns a full page and sets the flag even though every row was fetched, logging "results may be truncated" spuriously. To be precise you'd only flag truncation after confirming a *further* page would have run.

Both are minor given this is a large improvement over the status quo and a deliberate tradeoff against Plausible's rate limits — just flagging for awareness, not blocking.

---

jianwei1 (commented):
## Review summary

Clean, well-tested fix — no blocking issues. CI is fully green and the prior CodeRabbit thread (unbounded loop) is resolved by the page cap.

I verified the two load-bearing claims in the description and both hold:

- **No double-count / no data loss across the transition.** `mapEventToCapture` in `transformBreakdownResults.ts` is idempotent for capture names (`custom1Capture` → `custom1Capture`) and folds raw names into the same bucket (`custom1` → `custom1Capture`), then **sums**. Historical raw rows and new capture rows merge cleanly, and the pagination de-dupe keys on the `templateKey` value (which differs between the two), so neither is dropped.
- **`Step.tsx` migration is behavior-preserving except for the intended event-name change.** `key`/`simpleKey`/`templateKey` are built identically to the old inline code (same empty `target`, same shape); only the event name shifts raw → capture goal. The fire stays inside `if (journey != null)`, so `journey.id` is safe.

**Concerns**
- 50-page cap can silently under-count very large families (see inline on `service.ts`).
- **Rate-limit exposure under load:** a single breakdown report now issues up to ~100 sequential Plausible calls (this report paginates twice via `Promise.all`; the aggregate query once). Plausible cloud is ~600 req/hr per team with a ~100 req/60s burst, shared across keys — concurrent dashboard loads on large families could trip 429s. The cap bounds the worst case; worth watching post-deploy.

**Nits** (see inline): false-positive truncation log at the exact page boundary.

**What's done well**
- Excellent, targeted coverage in `service.spec.ts` — short page, full multi-page loop, exact-multiple boundary, explicit limit/page passthrough, runaway-upstream de-dupe, cap-and-log, and all three error branches.
- Reusing `fireCaptureEvent` keeps cards consistent with the button/video/radio path from #9075 instead of duplicating keyify logic, and the `Record<BlockEventLabel, …>` map makes a missing mapping a compile error.
- Comments explain the *why* (Plausible's 1000-row cap, the cap rationale, the "ignore caller limit/page" decision).

Overall: good direction, the Concerns are operational edges rather than correctness problems in the diff.

---

apis/api-journeys/src/schema/plausible/service.ts
csiyang: Thanks — both fair calls. Signing off consciously on each:

**50k cap / under-count:** The report key is `(event, target, journeyId)`, where `target` collapses to `{'', 'link', 'chat'}` and `event` is a bounded set, so rows ≈ `journeys × ~15`. Hitting 50k needs ~3,000 journeys in a single family — reachable only for the most viral templates, and the cap drops only the lowest-visitor tail (rows are visitor-ranked, so the journeys that matter are always fetched first). It's a 500× improvement over the old top-100 limit.

We deliberately can't raise the cap much: this report already issues up to 50 pages × 2 breakdowns ≈ 100 sequential calls, which sits right at Plausible cloud's ~100 req/60s burst limit (600/hr per team). Raising it trades a rare tail truncation for 429s, which is worse. So 50 is a calibrated ceiling, not arbitrary — I've added a comment recording that rationale in `ff41c6ad5`.

**Observability:** agreed console.error is weak if this ever trips, but surfacing truncation as a first-class signal means a response field or metric — a schema change we intentionally avoided to keep the frontend/backend deploys decoupled. Leaving it as a log for now; worth revisiting if a family ever approaches the cap.

**`cappedWithMore` boundary nit:** correct, but it only fires at exactly 50×1000 distinct rows and the message already hedges with "may"; making it precise needs an extra probe fetch that defeats the cap's purpose. Leaving as-is.

jianwei1: Thanks for the thorough sign-off — all three points are settled for me. The row-count bound (`journeys × ~15`, visitor-ranked so only the low-visitor tail is ever dropped) and the burst-limit ceiling both check out, and the rationale comment in `ff41c6a` captures it well for future maintainers. Agreed on leaving observability and the boundary nit as-is. Resolving.

---

jianwei1 (approved):
## Approved ✅

Both root causes are fixed cleanly and the two load-bearing claims verify out independently:

- **No double-count or data loss across the transition** — `mapEventToCapture` is idempotent for capture names and folds raw names into the same bucket before summing, and the pagination de-dupe keys on `templateKey` (which differs between raw and capture rows), so neither is dropped.
- **`Step.tsx` migration is behavior-preserving** apart from the intended raw → capture goal event-name change; the fire stays guarded by `if (journey != null)`.

My two earlier concerns were low-severity operational edges and have been consciously resolved: the `PLAUSIBLE_MAX_BREAKDOWN_PAGES = 50` ceiling is calibrated against Plausible's ~100 req/60s burst limit (now documented in `ff41c6a`), and the tail-truncation only affects the lowest-visitor rows of families exceeding ~3,000 journeys — a 500× improvement over the old top-100. Observability-as-a-signal and the exact-boundary log nit are reasonable "won't fix" given the schema/deploy-coupling tradeoff.

Strong test coverage (multi-page loop, exact-multiple boundary, dedupe/runaway, cap-and-log, limit/page passthrough, all error branches) and CI is green. Good work.

---

PR #9340, opened 2026-07-01
fix(journeys-admin): use all-time range for template card analytics

csiyang:
Template cards showed a declining **Views** count: `TemplateAggregateAnalytics` sent `where: {}` to `templateFamilyStatsAggregate`, so the Plausible request carried no period and silently inherited Plausible's default 30-day rolling window. As high-traffic days aged out of that window, the number went down. The query now sends `period: 'custom'` with `2024-06-01,<today>` (the shared `earliestStatsCollected` constant), matching how `AnalyticsItem` and `TemplateBreakdownAnalyticsDialog` already build their all-time ranges.

Only Views is affected: Template uses and Responses on the card are computed from Postgres with no time window. The full audit of all four analytics surfaces (journey card, template card, template breakdown, analytics dialog) is documented on [QA-540](https://linear.app/jesus-film-project/issue/QA-540/verify-analytics-queries-use-all-time-date-range-instead-of-default-30), part of the QA-536 declining-numbers investigation.

Apollo mocks updated to the new variables in `TemplateAggregateAnalytics.spec.tsx` (deterministic `formatISO` mock, same pattern as `AnalyticsItem.spec`) and `JourneyCard.spec.tsx` (computed under the suite's faked system time).

---

Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved journey template analytics to always query a clearly defined custom date range, making displayed counts more consistent and accurate.
* **Tests**
  * Updated journey card analytics tests to match the new query variables and to control date formatting deterministically.
* **Documentation**
  * Added guidance on preventing Plausible integration issues caused by omitted `period`/`date`, including example before/after fixes.

### DIFF
--- a/apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCard.spec.tsx
+++ b/apps/journeys-admin/src/components/JourneyList/JourneyCard/JourneyCard.spec.tsx
@@ -1,5 +1,6 @@
 import { MockedProvider, MockedResponse } from '@apollo/client/testing'
 import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { formatISO } from 'date-fns'
 import { SnackbarProvider } from 'notistack'
 import { type MockedFunction } from 'vitest'
 
@@ -288,7 +289,12 @@ describe('JourneyCard', () => {
         variables: {
           id: publishedLocalTemplate.id,
           idType: IdType.databaseId,
-          where: {}
+          where: {
+            period: 'custom',
+            date: `2024-06-01,${formatISO(new Date(), {
+              representation: 'date'
+            })}`
+          }
         }
       },
       result: {
--- a/apps/journeys-admin/src/components/JourneyList/JourneyCard/TemplateAggregateAnalytics/TemplateAggregateAnalytics.spec.tsx
+++ b/apps/journeys-admin/src/components/JourneyList/JourneyCard/TemplateAggregateAnalytics/TemplateAggregateAnalytics.spec.tsx
@@ -1,5 +1,7 @@
 import { MockedProvider, MockedResponse } from '@apollo/client/testing'
 import { render, screen, waitFor } from '@testing-library/react'
+import { formatISO } from 'date-fns'
+import { type MockedFunction } from 'vitest'
 
 import {
   GetTemplateFamilyStatsAggregate,
@@ -17,6 +19,15 @@ vi.mock('next-i18next/pages', () => ({
   })
 }))
 
+vi.mock('date-fns', async () => {
+  return {
+    ...(await vi.importActual('date-fns')),
+    formatISO: vi.fn()
+  }
+})
+
+const mockFormatIso = formatISO as MockedFunction<typeof formatISO>
+
 const mockEnqueueSnackbar = vi.fn()
 
 vi.mock('notistack', () => ({
@@ -26,6 +37,10 @@ vi.mock('notistack', () => ({
 }))
 
 describe('TemplateAggregateAnalytics', () => {
+  beforeEach(() => {
+    mockFormatIso.mockReturnValue('2024-09-26')
+  })
+
   afterEach(() => {
     vi.clearAllMocks()
   })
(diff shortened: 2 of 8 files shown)


jianwei1 (commented):
## Review summary

Correct, well-tested, and well-documented fix for the surface it targets. The initial `TemplateAggregateAnalytics` query now mirrors the established `AnalyticsItem` all-time pattern, test determinism is handled correctly (mocked `formatISO` in one spec, faked system time in `JourneyCard.spec`), and the `docs/solutions/` write-up is genuinely useful.

One concern worth addressing before merge — it can't be left as an inline comment because the file isn't in this diff:

### Concern — the sibling call site still sends `where: {}`

`apps/journeys-admin/src/libs/useTemplateFamilyStatsAggregateLazyQuery/useTemplateFamilyStatsAggregateLazyQuery.ts:66`

The same hook that this fix touches runs `GET_TEMPLATE_FAMILY_STATS_AGGREGATE` a second time inside `refetchTemplateStats`, still with `where: {}`:

```ts
client.query({
  query: GET_TEMPLATE_FAMILY_STATS_AGGREGATE,
  variables: { id: templateId, idType: IdType.databaseId, where: {} }, // ← unchanged
  fetchPolicy: 'network-only'
})
```

`refetchTemplateStats` is called after duplicate / trash / restore / translate / copy-to-team operations across ~15 components, so this is a live path. Two consequences:

1. **It still carries the 30-day-window bug** this PR is closing out. The new doc explicitly says: _"Never send an empty `where` — it compiles, runs, and silently means 'last 30 days'."_ This call site violates that rule in the very file being fixed.
2. **It now writes to a different Apollo cache entry than the card reads.** Apollo keys `templateFamilyStatsAggregate` by its `where` args. The card now observes `where: { period: 'custom', date: ... }`; the refetch writes `where: {}`. Different `storeFieldName` → the refetch no longer updates the entry the card renders. Before this PR both used `where: {}` and shared a cache key, so a post-duplicate refetch refreshed the card's "Template uses" count; after this PR that refetch is effectively a UI no-op (stale count until remount) plus a wasted network request.

Given QA-540's goal is to make *every* template-stats query use an all-time range, this feels in scope. Suggestion: build the same `period: 'custom'` / `earliestStatsCollected` range in `refetchTemplateStats`, or extract a shared `buildAllTimeStatsWhere()` helper used by both call sites so they can't drift again. (The doc's "Prevention" note about making `period`/`date` required at the schema or defaulting in the resolver is the right longer-term follow-up.)

### Positive
- Faithful reuse of the `AnalyticsItem` pattern; import resolves correctly to `buildPresetDateRange/index.ts`.
- Both test-determinism approaches verified — no midnight-rollover flake in `JourneyCard.spec` thanks to `vi.setSystemTime`.
- Honest, high-quality solution doc (clear "What Didn't Work" and surface-by-surface audit).

### Note
- The `danger` check shows a red run alongside a later green one — looks like infra/label noise rather than the diff (tests, build, lint, E2E, CodeQL all green), but worth a glance.

Not blocking — posting as a comment for the author to weigh Concern #1.

---

jianwei1 (approved):
## Re-review — concern resolved ✅

Thanks for the quick turnaround. The earlier concern (the sibling `refetchTemplateStats` call site still sending `where: {}`) is fully addressed:

- Both call sites — the initial query in `TemplateAggregateAnalytics.tsx` and `refetchTemplateStats` in the hook — now share `buildAllTimeStatsFilter()`, so they can't drift and the split-cache-entry / silent-refetch-no-op problem is eliminated structurally.
- Nice call putting the helper in its own `libs/buildAllTimeStatsFilter` module rather than exporting it from the hook lib — that neatly avoids the `vi.mock` factory crash across the ~10 specs that mock the hook.
- The solution doc's new prevention bullet ("grep for every use of the query document; a `where` mismatch splits the Apollo cache entry") captures exactly the right lesson.

Verified the helper is exercised via `TemplateAggregateAnalytics.spec`'s exact-variable match, mock consistency holds for `formatISO` across both specs, and CI is now fully green (all test shards, build, lint, E2E, CodeQL, and `danger`).

Optional (non-blocking): the refetch path's `where` isn't directly asserted anywhere since callers mock `refetchTemplateStats` — drift risk is already gone via the shared builder, but a one-line hook-spec assertion would lock down the exact regression that first slipped through.

Approving.

---

PR #9342, opened 2026-07-02
fix(QA-538): let journey buttons work while AI chat is open

csiyang:
Once the AI conversational chat engaged on a journey card, every button outside the chat went dead on desktop: poll options and the nav arrows stopped responding, and only the chat frame stayed interactive ([QA-538](https://linear.app/jesus-film-project/issue/QA-538)).

Root cause: NES-1738 (JesusFilm/core#9308) shrank the ChatOverlay's visible dark surface to a bottom sheet so the journey card is revealed above the chat, but the overlay's invisible root wrapper stayed `position: fixed; inset: 0` at modal z-index, silently swallowing every click above the sheet. Since NES-1734 auto-opens the chat, journeys entered this state with no user action.

The fix is structural: the overlay container is now a bottom-anchored fixed band exactly as tall as the chat sheet (144px idle, 80% once the conversation is active — the heights it already animated between), with the dark click-to-close scrim filling the band behind the centred panel. Nothing chat-related exists above the sheet anymore, so the journey stays natively interactive by construction — no full-viewport layer, no pointer-events juggling. Same z-index, same look, same scrim-click-to-close; mobile is unaffected (the xs `PinnedChatBar` was already band-shaped). An adversarial review verified the geometry is equivalent across breakpoints, RTL, and the panel's internal absolute positioning.

Regression tests in `ChatOverlay.spec.tsx` pin the band geometry (bottom-anchored, 144px/80%, scrim and panel filling it) at the style level: jsdom does no hit-testing (and `Conductor.spec.tsx` mocks ChatOverlay away), which is why the existing nav-click tests never caught this.

---

Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the AI chat overlay to be bottom-anchored and only occupy the chat band, preventing it from blocking journey card clicks.
  * Improved the overlay’s compact/active sizing and ensured the dark backdrop and chat panel fill the band correctly.
* **Tests**
  * Strengthened UI tests to validate band height and overlay panel state in both compact and active modes, including backdrop and panel sizing.
  * Kept existing close interactions covered.
* **Documentation**
  * Added solution pages covering a cross-origin navigation reload race and the chat overlay click regression.

### DIFF
--- a/docs/solutions/logic-errors/handleaction-crossorigin-linkaction-reload-race-2026-07-03.md
+++ b/docs/solutions/logic-errors/handleaction-crossorigin-linkaction-reload-race-2026-07-03.md
@@ -1,0 +1,134 @@
+---
+title: 'handleAction LinkAction never opens on non-whitelisted origins (router.push + reload race)'
+date: 2026-07-03
+category: logic-errors
+module: journeys-ui/action
+problem_type: logic_error
+component: journeys-ui/handleAction
+symptoms:
+  - 'Link button ripples and fires analytics, but the link never opens; the page reloads in place instead'
+  - 'Same button works on your.nextstep.is and your-stage.nextstep.is but fails on Vercel preview deployments'
+  - 'No console error visible to the user (unhandled promise at most)'
+root_cause: logic_error
+resolution_type: code_fix
+severity: high
+related_components:
+  - journeys-ui/Button
+  - journeys-ui/RadioOption
+  - journeys-ui/SignUp
+  - journeys-ui/VideoTrigger
+tags:
+  - handleaction
+  - linkaction
+  - router-push
+  - cross-origin
+  - vercel-preview
+  - custom-domain
+  - navigation
+---
+
+# handleAction LinkAction never opens on non-whitelisted origins (router.push + reload race)
+
+> **Status:** diagnosed and source-verified (Next 16.2.5); fix not yet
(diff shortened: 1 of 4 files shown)


jaco-brink (approved):
Nicely done 🚀
Review found no concerns other than agents arguing about prose stuff in the docs.

I found the ui-bugs doc useful as it nicely shows the before/after better than the file diffs does.

e.g.
```tsx
// Before: full-viewport layer, children sized to the sheet
<Box sx={{ position: 'fixed', inset: 0, zIndex: (t) => t.zIndex.modal, ... }}>
  <Box onClick={onClose} sx={{ position: 'absolute', bottom: 0, height, ... }} />
  <Box sx={{ maxWidth: '48rem', height, ... }}><AiChat ... /></Box>
</Box>

// After: bottom-anchored band exactly as tall as the sheet; children fill it
<Box sx={{ position: 'fixed', bottom: 0, left: 0, right: 0, height, // 144px idle / '80%' active
           transition: HEIGHT_TRANSITION, zIndex: (t) => t.zIndex.modal, ... }}>
  <Box onClick={onClose} sx={{ position: 'absolute', top: 0, right: 0, bottom: 0, left: 0, ... }} />
  <Box sx={{ maxWidth: '48rem', height: '100%', ... }}><AiChat ... /></Box>
</Box>
```

PR #9345, opened 2026-07-05
feat(journeys-admin): add GTM analytics to template gallery collections [NES-1698]

csiyang:
## Summary

The new template gallery (Local Template Library) collected no usage analytics beyond default page views, so there was no way to tell whether teams actually create, curate, publish, or share collections. This PR instruments the full admin collection workflow with GTM events, following the `send*Event` helper pattern from NES-1662: one typed helper file ([sendCollectionEvent.ts](https://github.com/JesusFilm/core/blob/siyangcao/nes-1698-n2h-add-analytics-to-template-gallery/apps/journeys-admin/src/libs/sendCollectionEvent/sendCollectionEvent.ts)) with unit tests, called from 6 call sites. All events share the `team_collection_` prefix so the feature reads as one group in GA4.

## Events

| Event | Fires when | Parameters |
|---|---|---|
| `team_collection_create` | create mutation succeeds | `teamId`, `collectionId` |
| `team_collection_edit_dialog_open` | edit dialog opened from the card menu | `teamId`, `collectionId`, `collectionStatus` |
| `team_collection_publish` | publish mutation succeeds | `teamId`, `collectionId`, `collectionSlug` |
| `team_collection_template_added_via_drag` | server-accepted drop of a template into a collection | `collectionId`, `templateId` |
| `team_collection_template_added_in_dialog` | save that added a template via the dialog picker | `collectionId`, `templateId` |
| `team_collection_more_details_click` | "More details" expanded in the dialog | `collectionId` |
| `team_collection_menu_preview_click` | preview opened from the collection card menu | `collectionSlug`, `collectionId` |
| `team_collection_dialog_preview_click` | preview opened from the edit dialog | `collectionSlug` |
| `team_collection_publish_preview_click` | "View the page" from the publish success dialog | `collectionSlug` |
| `team_collection_copy_link_click` | public URL copied (either surface) | `location`, `collectionSlug` |
| `team_collection_description_update` | save that changed the description | `teamId`, `collectionId` |
| `team_collection_slug_update` | save that changed the slug | `teamId`, `collectionId`, `collectionSlug` |
| `team_collection_media_update` | save that changed media | `teamId`, `collectionId`, `mediaType`, `mediaProvider` |

## Design decisions

- Field-level events (description, slug, media) fire once per successful save that changed the field, not per keystroke, so counts read as "people who used the feature".
- `team_collection_template_added_via_drag` only fires after the server accepts a drop into a collection. Reorders within a collection and drag-outs to unsectioned don't count.
- Each preview surface is a distinct event (menu / dialog / publish success) so they show as separate rows in GA4. Copy-link stays one event with a `location` parameter.
- Event names stay within GA4's 40-character cap (`team_collection_template_added_in_dialog` sits at exactly 40).
- `team_collection_media_update` reports `mediaProvider` as the embed hostname (e.g. `www.youtube.com`) or `mux`. A full URL would exceed GA4's 100-character parameter value cap.

Events are inert until the matching Data Layer Variables, Custom Event triggers, and GA4 Event tags are added to the journeys-admin GTM container. Setup steps are in [06-analytics.md](https://github.com/JesusFilm/core/blob/main/apps/docs/docs/03-basics/06-analytics.md); that work is tracked as a manual follow-up on [NES-1698](https://linear.app/jesus-film-project/issue/NES-1698/n2h-add-analytics-to-template-gallery).

## Testing

New `sendCollectionEvent.spec.ts` covers every event including the three preview surfaces and the `mediaProvider` derivation (16 tests). All 263 tests across the `TemplateGalleryPageList` tree pass.

---

Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added analytics across the collection workflow, including create/edit-dialog open, “More details”, copy-link, and preview actions.
  * Captures template additions and drag-and-drop assignments when drops are accepted.
  * Extends submit analytics for media/description/slug updates, template add behavior, and publish success.
* **Documentation**
  * Added conventions for GTM/GA4 event design and Dev→Stage→Prod container replication.
  * Added guidance for registering GA4 custom dimensions.
* **Tests**
  * Added and updated Vitest coverage validating event names/payloads, including media provider parsing.


apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialog.tsx
```diff
@@ -366,6 +370,16 @@ function CollectionDialogContent({
...
+                          if (collection != null) {
+                            next
+                              .filter((id) => !values.journeyIds.includes(id))
+                              .forEach((templateId) =>
+                                sendCollectionTemplateAddEvent({
+                                  collectionId: collection.id,
+                                  templateId
+                                })
+                              )
+                          }
```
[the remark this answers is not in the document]
csiyang: Good catch — fixed in 1f97bfe9c. The event now fires in `useCollectionForm.handleSubmit`'s post-update success path, gated on `input.journeyIds !== undefined`, emitting one event per template id the save actually added (diffed against `collection.templates`). This matches the drag event's only-on-confirmed-persistence semantics, and a cancelled/discarded dialog no longer counts.

apps/journeys-admin/src/components/TemplateGalleryPageList/useDragEndHandler/useDragEndHandler.ts
```diff
@@ -293,18 +294,24 @@ export function useDragEndHandler(
...
-            // interpolating an empty name if the title can't be resolved.
-            const targetName = targetCollection?.title ?? returnedPage?.title
-            enqueueSnackbar(
-              targetName != null && targetName !== ''
-                ? t('Added to {{collection}}', { collection: targetName })
-                : t('Added to collection'),
-              { variant: 'success', preventDuplicate: true }
-            )
+          } else {
+            sendCollectionTemplateDragEvent({
```
jianwei1, line 298: **Concern (test coverage):** The `sendCollectionEvent` helper is exhaustively unit-tested, but the *wiring* here isn't. This event's correctness depends entirely on its placement — inside the `else` (accepted) branch under `if (targetCollectionId != null)`, so reorders and drag-outs to unsectioned don't count. That's exactly the kind of invariant a future refactor can silently break (e.g. moving the call up a scope would start counting rejected/reorder drops). `useDragEndHandler.spec.tsx` already exists with mock infra — worth a `vi.mock('../../../libs/sendCollectionEvent')` assertion that the drag event fires on an accepted cross-collection drop and does **not** fire on a rejected drop or an intra-collection reorder. Not blocking.

---

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/useCollectionForm/useCollectionForm.ts
```diff
@@ -326,6 +333,41 @@ export function useCollectionForm({
...
+            sendCollectionSlugUpdateEvent({
+              teamId,
+              collectionId: collection.id,
+              collectionSlug: values.slug
+            })
+          }
+          // Fire per template the save actually added (net of removals and
+          // re-adds within the session) — mirrors the drag event's
+          // only-on-confirmed-persistence semantics.
+          if (input.journeyIds !== undefined) {
```
jianwei1, line 359: **Concern (test coverage):** This template-add diff (fire once per id in `values.journeyIds` not already in `collection.templates`) is the other piece of non-trivial wiring with no regression guard — an off-by-one in the diff would over/under-count adds, and a pure reorder must emit nothing. `useCollectionForm.spec.tsx` already exists; a `vi.mock('../../../../libs/sendCollectionEvent')` assertion covering "adds one → one event", "reorder only → zero events", and "add + remove in one save → one event for the add" would lock in the intended semantics cheaply. Not blocking.

(Minor, no change needed: the `send*Event` calls sit inside the submit `try` block, so a throw from `sendGTMEvent` would surface a false "Couldn't save collection" snackbar after a successful save. Extremely unlikely in practice since it's just a `dataLayer` push — noting for awareness only.)

---

jianwei1 (commented):
## Review summary

Solid, low-risk analytics instrumentation — clean typed `send*Event` helper matching the NES-1662 precedent, correct fire-on-confirmed-persistence semantics throughout (I traced every event through its call site against the head source), and thoughtful GA4 cap handling (40-char event names, `mediaProvider` hostname to stay under the 100-char param cap). The `06-analytics.md` custom-dimensions step and the `docs/solutions/` learnings doc are genuinely valuable knowledge capture. The one prior review concern (template-add firing before Save) is already resolved correctly.

**Critical:** none.

**Concern (non-blocking):** The helper is well unit-tested, but the two pieces of non-trivial *wiring* — the drag event's placement in the accepted branch, and the dialog template-add diff — have no call-site regression guard. Both hooks already have specs + mock infra, so the assertions are cheap. See inline.

**Nits:** analytics calls inside the action `try` block (theoretical false-error snackbar); publish reports server `result.slug` while slug-update reports client `values.slug` (divergence unlikely). Both noted inline / for awareness only.

CI is fully green on this head. Nothing blocking here — recommend adding the two wiring tests before merge, but it's a judgment call.

---

apps/journeys-admin/src/components/TemplateGalleryPageList/useDragEndHandler/useDragEndHandler.ts
csiyang: Added in ebc06b4dc — new `analytics wiring (NES-1698)` describe in `useDragEndHandler.spec.tsx` with 4 tests: fires once (with `collectionId`/`templateId`) on a server-accepted drop into a collection, and does NOT fire on a silently-rejected drop, an intra-collection reorder, or a drag-out to unsectioned.

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/useCollectionForm/useCollectionForm.ts
csiyang: Added in ebc06b4dc — `analytics wiring (NES-1698)` describe in `useCollectionForm.spec.tsx` covering exactly the three cases you listed: adds one → one event (with the added `templateId`), pure reorder → zero events, add + remove in one save → one event for the add. On the try-block nit: agreed it's theoretical-only for a dataLayer push; leaving as-is.

csiyang:
@jianwei1 Both inline concerns are addressed (replies on the threads):

- `ebc06b4dc` — the two wiring test sets you asked for: drag event (accepted drop only; not rejected / reorder / drag-out to unsectioned) and the dialog add diff (add → 1 event, pure reorder → 0, add+remove in one save → 1 for the add).
- `696fedaad` — went one step further and covered the remaining non-trivial wiring: create (fires with the server-returned id only after the mutation succeeds), edit-dialog open, publish (asserts the **server-returned** slug, which also locks in the publish/slug-update divergence you noted), the three field-update events (fire only when that field was in the saved diff), plus a negative test that a title-only save fires nothing.

Left untested on purpose: the pure click-handler dispatches (preview ×3, copy-link, more-details) — payloads are covered by the helper spec and the wiring is a one-liner.

Ready for another look.

apps/journeys-admin/src/components/TemplateGalleryPageList/useDragEndHandler/useDragEndHandler.ts
jianwei1: Re-reviewed at `696fedaad3` — the 4 new tests in the `analytics wiring (NES-1698)` block genuinely guard the invariant, not just superficially: the assign mock returns real success data, the accepted-drop case asserts one call with the exact `{collectionId, templateId}`, and each negative (silent rejection, intra-collection reorder, drag-out to unsectioned) is paired with a `mock.result`/`reorderMock.result` assertion proving the path actually ran before checking the event wasn't fired. Non-vacuous. Nice coverage — resolving.

---

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/useCollectionForm/useCollectionForm.ts
jianwei1: Re-reviewed at `696fedaad3` — add-one / pure-reorder / add+remove are all covered non-vacuously (update mock returns success data so the post-mutation branch runs; the reorder case asserts `updateMock.result` fired once yet zero add events, which genuinely distinguishes reorder from add), and the field-update events assert both the changed field fires *and* the others don't. Publish correctly asserts the server slug. Coverage concern fully addressed — resolving.

One small FYI, not blocking and no need to act now: unlike the drag path, the form's add-event fires once `templateGalleryPageUpdate` resolves without throwing — it isn't re-checked against the server-returned `templates`. So the "mirrors the drag event's only-on-confirmed-persistence semantics" comment is a touch optimistic: a success-shaped response that silently dropped a `journeyId` would still emit the event. Only worth tightening if `templateGalleryPageUpdate` can silently no-op an id the way `assignJourney` can — if it can't, ignore this.

---

jianwei1 (approved):
## Approved ✅

Clean, well-reasoned analytics instrumentation. Every event fires on confirmed persistence, GA4 hard limits are handled deliberately (40-char event names, hostname-only `mediaProvider` under the 100-char cap), and the `06-analytics.md` custom-dimensions step + `docs/solutions/` learnings doc are genuinely valuable knowledge capture.

Both review concerns are fully resolved:
- **Drag-event wiring** — 4 non-vacuous tests lock in fire-on-accepted-drop and no-fire on silent rejection / intra-collection reorder / drag-out to unsectioned.
- **Template-add diff + field-update events** — add-one / pure-reorder / add+remove and per-field change detection all covered, with mutation mocks returning real success data so the assertions execute the success branch.

The author also added create/edit-open coverage in `TemplateGalleryPageList.spec.tsx` beyond what was flagged. CI is fully green.

One non-blocking FYI left on the form thread (add-event doesn't re-check the server-returned `templates` the way the drag path does) — optional, safe to merge as-is.

---

PR #9346, opened 2026-07-05
feat(journeys-admin): movetemplates in published collections + dnd imrpovements (NES-1703)

csiyang:
## What

[NES-1703](https://linear.app/jesus-film-project/issue/NES-1703/n2h-move-templates-into-published-journeys) — removes the guard rails on published collections in the template gallery and makes drag-and-drop more discoverable.

### Free moves in and out of published collections

- Removed the published guard from `useDragEndHandler` — drops into/out of published collections fire the assign/reorder mutations like any draft collection (no dialog, no snackbar)
- Removed the `publishedLock` prop plumbing and published-status droppable disabling; drag is gated only by in-flight state
- The backend never enforced this — frontend was the only gate

### Empty collections are publishable

- Removed the "Add at least one template before publishing" gate and the now-unused `journeyCount` prop; Publish is gated solely by the custom-domain `canPublish` check

### Drag and drop affordances

- **Move arrow on hover** — a multi-directional arrow fades in over the card image (opt-in `JourneyCard` prop, gallery only) and stays visible on the DragOverlay preview for the whole drag
- **Always-present drop placeholder** — every collection ends with a card-sized dashed "Drag templates here" tile, replacing the old empty-state caption; mirrors JourneyCard geometry via shared constants so its height matches real cards, darkens neutrally while a drag is active
- **Full-height All Templates drop zone** — the gallery column measures its offset and fills to the bottom of the viewport, with the unsectioned droppable growing into it, so dropping a template out of a collection doesn't need pixel accuracy

## Testing

- Updated specs: published-move rejection tests inverted to accept, publish-gate tests, CollectionCard empty-state test
- New `DraggableJourneysGrid.spec.tsx` covering the placeholder tile
- 263 tests pass across TemplateGalleryPageList + JourneyCard suites; typecheck clean
- QA requirements posted on the Linear ticket

## Summary by CodeRabbit

* **New Features**
  * Added clearer drag-and-drop visuals, including a move indicator on cards and a drop placeholder in empty areas.
  * Collections can now show a visible drop target even when they contain no templates.

* **Bug Fixes**
  * Improved drag-and-drop behavior so items can be moved more consistently across collections, including published ones.
  * Empty collections now keep their contents visible instead of replacing them with a separate caption.


apps/journeys-admin/src/components/TemplateGalleryPageList/TemplateGalleryPageList.tsx
```diff
@@ -208,6 +225,77 @@ export function TemplateGalleryPageList({
...
+  // The min-height is written straight to the node (not React state):
+  // a resize tick then costs one rAF-throttled style write instead of a
+  // full gallery re-render + a new emotion class per unique value. Safe
+  // because nothing else manages this node's inline style. innerHeight
+  // (not 100vh) so mobile browsers' URL-bar chrome is accounted for.
+  const [rootNode, setRootNode] = useState<HTMLDivElement | null>(null)
+  useEffect(() => {
+    if (rootNode == null || !visible) return
+    let frame: number | null = null
+    function measureGalleryMinHeight(): void {
```
jianwei1, line 250: **Concern (not blocking):** This viewport-fill measurement effect — `measureGalleryMinHeight`, `getScrollParent`, the content-relative offset math, and the ResizeObserver/rAF wiring — is the most substantive new logic in the PR and has no test coverage.

I traced the math and believe it's correct: the `+ scrollTop` in `topOffset` paired with `clientHeight` deliberately yields a scroll-invariant fill height, and the ResizeObserver→inline-style-write loop converges (writing `minHeight` doesn't change the node's top offset or the scroll parent's `clientHeight`, so it re-measures to the same value and settles). It's also genuinely awkward to test in jsdom (no layout, no `ResizeObserver` — correctly guarded here).

Not asking for a full test of the effect. Worth a quick discussion on whether the pure helper `getScrollParent` is worth extracting + unit-testing, or whether this is accepted as manual-QA/preview-verified layout code per the app's testing guidance. Either way is fine — flagging because it's ~60 lines of uncovered logic.

---

apps/journeys-admin/src/components/TemplateGalleryPageList/CollectionDialog/CollectionDialogFooter/CollectionDialogFooter.tsx
```diff
@@ -135,17 +130,12 @@ export function CollectionDialogFooter({
...
-  const publishBlocked = !canPublish || journeyCount === 0
-  const publishTooltip = !canPublish
-    ? (publishBlockedReason ?? '')
-    : journeyCount === 0
-      ? t('Add at least one template before publishing')
-      : ''
+  // anchored as the primary action. Publish is gated only by
+  // custom-domain teams that can't host gallery pages — an empty
+  // collection is publishable (it just renders no templates until some
+  // are added).
```
jianwei1, line 136: **Concern (product confirmation, not a code defect):** Combined with removing the published guard in `useDragEndHandler`, this changes two user-visible contracts that are worth an explicit product/QA sign-off:

1. A published collection can now be emptied entirely, and empty collections are publishable — so a **live public gallery page may render zero templates**.
2. Dragging a template into/out of a published collection now mutates the live page **immediately, with no confirmation dialog or toast**.

Both are stated as intended in the PR body and I verified they match the backend contract (`templateGalleryPageAssignJourney` / `templateGalleryPageReorderTemplate` resolvers both document "Allowed on both draft and published pages"). Just flagging that "live pages change instantly and silently" is the deliberate new behavior — assuming the Linear QA requirements cover it, no code change needed.

---

jianwei1 (commented):
## Review summary

Clean, well-reasoned, low-risk PR — no changes required to merge.

**Critical:** none. The riskiest change (deleting the published guard in `useDragEndHandler`) is backed by the backend contract — both the `templateGalleryPageAssignJourney` and `templateGalleryPageReorderTemplate` resolvers explicitly document "Allowed on both draft and published pages" (the reorder resolver even notes "the frontend gates the UX; the backend accepts unconditionally"). Existing silent-rejection handling means even a future backend policy change would degrade gracefully rather than break.

**Concerns:** 2 (both inline, both non-blocking) — the untested viewport-fill measurement effect, and a product/QA confirmation that live public gallery pages now change instantly and silently (and can be empty).

**Positives:**
- Backend claim verified rather than assumed — the frontend change is the correct half of a documented contract.
- Exceptional inline documentation on the non-obvious calls (scroll-invariant measurement, `useDndMonitor` over prop-drilling to preserve the grid's `memo`, `display: flex` over `display: contents` for `flexGrow` propagation, neutral darkening because `primary.main` is red).
- `JourneyCardSizer` + shared geometry constants keep the placeholder tile locked to real card height via a single source of truth.
- Behavior-change tests correctly inverted, plus a focused new `DraggableJourneysGrid.spec.tsx`; `handleDragCancel` closes the Escape/touchcancel overlay-lingering gap.

CI is green (the one failed early `danger` run was superseded by a passing re-run).

_Automated review — findings are advisory; a human should decide on approval._

---

jianwei1 (approved):
## Approved ✅

Code review looks good — no correctness, security, accessibility, or maintainability defects across the 13 changed files. The load-bearing change (removing the frontend published guard) is verified safe against the backend contract: both `templateGalleryPageAssignJourney` and `templateGalleryPageReorderTemplate` explicitly permit draft **and** published pages.

The two earlier concerns were product/QA-validation items rather than code defects, so I've resolved those threads — they'll be picked up in the manual QA pass on the preview (per the Linear QA requirements), not in code review.

Highlights: backend claim verified rather than assumed, excellent inline docs on the non-obvious layout/dnd decisions, shared card geometry via `JourneyCardSizer`, and correctly inverted behavior tests plus the new `DraggableJourneysGrid.spec.tsx`. CI is green.

---

PR #9349, opened 2026-07-07
fix(api-journeys): allow publishers to update template customization fields

Kneesal:
## Summary
- allow publisher-role users to manage template customization fields through the field-level ACL helper
- add regression coverage for publisher access on global and local templates, plus non-publisher denial
- capture the learning in docs/solutions and seed scoped template vocabulary in CONCEPTS.md

## Review
- Ran compound-engineering:ce-code-review in report-only style. No actionable findings. Residual risk noted: publisher access now explicitly follows the existing all-template publisher policy.
- Ran compound-engineering:ce-compound and validated the new solution doc frontmatter.

## Local verification
- Not run per request. CI/CD should be the verification source for this PR.
- Ran non-test checks only: git diff --check and docs frontmatter parser-safety validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Publishers can now update template customization fields as expected for global templates, without needing direct journey or team access.
  * Local template access rules remain unchanged, and non-publishers still require the appropriate access.

* **Documentation**
  * Added a glossary for key journey and template terms.
  * Updated the security issue write-up with the resolved behavior and guidance.

### DIFF
--- a/apis/api-journeys/src/schema/journey/journey.acl.spec.ts
+++ b/apis/api-journeys/src/schema/journey/journey.acl.spec.ts
@@ -367,7 +367,17 @@ describe('journeyAcl', () => {
       teamId: 'teamId'
     } as unknown as Journey
 
-    it('allows publishers to update customization fields on global templates', () => {
+    const localTemplateWithJourneyEditor = {
+      ...localTemplate,
+      userJourneys: [{ userId: user.id, role: UserJourneyRole.editor }]
+    } as unknown as Journey
+
+    const localTemplateWithTeamMember = {
+      ...localTemplate,
+      team: { userTeams: [{ userId: user.id, role: UserTeamRole.member }] }
+    } as unknown as Journey
+
+    it('should allow publishers to update customization fields on global templates', () => {
       expect(
         canManageTemplateField(globalTemplate, {
           ...aclUser,
@@ -376,7 +386,7 @@ describe('journeyAcl', () => {
       ).toBe(true)
     })
 
-    it('allows publishers to update customization fields on local templates', () => {
+    it('should allow publishers to update customization fields on local templates', () => {
       expect(
         canManageTemplateField(localTemplate, {
           ...aclUser,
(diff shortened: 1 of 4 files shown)


apis/api-journeys/src/schema/journey/journey.acl.ts
```diff
@@ -239,6 +239,9 @@ export function canManageTemplateField(
   const isLocalTemplate =
     journey.template === true && journey.teamId !== 'jfp-team'
 
+  if (journey.template === true && user.roles?.includes('publisher') === true)
```
edmonday: **Positive:** Good fix — this now mirrors `journeyAcl` line 83 ("Publishers can update any template"). Placing the publisher allow-rule before the local-template membership check is correct, and the `return false` default still denies non-publishers without access.

apis/api-journeys/src/schema/journey/journey.acl.spec.ts
```diff
@@ -346,4 +346,47 @@ describe('journeyAcl', () => {
       expect(can(Action.Export, journeyEmpty, user)).toBe(false)
     })
   })
+
+  describe('canManageTemplateField', () => {
```
edmonday, line 350: **Nit (optional):** Two small things on this new block:

1. Coverage — the reorder puts the publisher rule first. The publisher and non-publisher-deny cases are covered, but there is no test for a **non-publisher with a team/journey role on a local template** still returning `true`. That path is logically unaffected, but a test would lock in that the reorder did not regress local-template membership access.
2. Naming — these read `allows...`/`denies...`; project convention is to start `it` descriptions with `should`. Minor, non-blocking.

edmonday (commented):
## Review summary

**Critical:** none
**Concern:** none
**Nit:** 2 (see inline)

Verified the core change against `journey.acl.ts:83` (`Publishers can update any template`) — the fix genuinely aligns the narrower `canManageTemplateField` helper with the existing resource-level publisher policy, rather than introducing new access. Check ordering is correct (publisher allow-rule before the local-template membership rule; `return false` default still denies non-publishers).

Well-diagnosed root cause (two near-identical ACL helpers, only one on the mutation's execution path) and a genuinely useful `docs/solutions` write-up. CI green. Safe to merge.

apis/api-journeys/src/schema/journey/journey.acl.spec.ts
Kneesal: > there is no test for a **non-publisher with a team/journey role on a local template** still returning `true`

Addressed in 4a3c9e7cc9: added local-template journey-editor and team-member coverage for non-publishers, and renamed the new specs to the project’s `should...` style.

Kneesal:
Clarification: publisher-only customization-field access is scoped to **global templates** (`teamId === jfp-team`) in 228b9e4f2d. Local/team templates still require the user to have journey or team access, even when the user has the publisher role. The net ACL diff now adds that global-template allow rule and leaves the existing local-template membership path in place.

Kneesal:
Closing per request.

PR #9353, opened 2026-07-08
fix(journeys): customisable collection page templates

csiyang:
Open customize flow for customizable templates from Use button

Clicking "Use" on a customizable template in the public collection gallery now routes into the template customization flow instead of the admin "Copy to team" dialog, restoring the behavior from the previous system.

The admin deep-link handler also redirects customizable templates to /templates/[journeyId]/customize, and cross-page navigations no longer use shallow routing, which previously skipped getServerSideProps and surfaced 404s.

## Summary by CodeRabbit

* **New Features**
  * Templates marked as **customizable** now route to the guided customization flow (e.g., `/templates/<id>/customize`) across “Use” actions, public gallery cards, and admin previews.

* **Bug Fixes**
  * Improved customization navigation behavior: shows a loading overlay while determining the destination, then redirects appropriately for customizable vs non-customizable templates.
  * Updated navigation to avoid shallow in-app routing when handing off to the admin app and to correctly sanitize/construct admin URLs.

* **Tests**
  * Added/updated unit tests covering customizable routing and URL construction.

### DIFF
--- a/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
+++ b/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
@@ -1,3 +1,5 @@
+import Backdrop from '@mui/material/Backdrop'
+import CircularProgress from '@mui/material/CircularProgress'
 import { useRouter } from 'next/router'
 import { useTranslation } from 'next-i18next/pages'
 import { useSnackbar } from 'notistack'
@@ -95,7 +97,7 @@ function ActiveUseTemplateDeepLink({
     // Full navigation — not shallow. Shallow routing only applies to query
     // changes on the *same* page; using it across pathnames skips
     // getServerSideProps and surfaces a 404 on /templates/[journeyId]/customize.
-    void router.replace(`/templates/${journeyId}/customize`)
+    void router.replace(`/templates/${encodeURIComponent(journeyId)}/customize`)
   }, [journeyLoading, isCustomizable, journeyId, router])
 
   useEffect(() => {
@@ -255,7 +257,20 @@ function ActiveUseTemplateDeepLink({
     stripParamFromUrl()
   }, [loading, translationVariables, stripParamFromUrl])
 
-  if (journeyLoading || isCustomizable) return null
+  // Deep links land here directly, so show a blocking spinner instead of a
+  // blank page while we decide between the dialog and the customize redirect
+  // (and while the redirect itself is in flight).
+  if (journeyLoading || isCustomizable) {
+    return (
+      <Backdrop
+        open
+        data-testid="UseTemplateDeepLinkLoading"
+        sx={{ zIndex: (theme) => theme.zIndex.modal }}
+      >
+        <CircularProgress sx={{ color: 'common.white' }} />
+      </Backdrop>
+    )
+  }
 
   return (
     <CopyToTeamDialog
--- a/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
+++ b/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
@@ -53,7 +53,7 @@ interface ActiveUseTemplateDeepLinkProps {
 
 function ActiveUseTemplateDeepLink({
   journeyId
-}: ActiveUseTemplateDeepLinkProps): ReactElement {
+}: ActiveUseTemplateDeepLinkProps): ReactElement | null {
   const { t } = useTranslation('apps-journeys-admin')
   const { enqueueSnackbar } = useSnackbar()
   const router = useRouter()
--- a/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
+++ b/apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
@@ -1,7 +1,7 @@
 import { useRouter } from 'next/router'
 import { useTranslation } from 'next-i18next/pages'
 import { useSnackbar } from 'notistack'
-import { ReactElement, useCallback, useRef, useState } from 'react'
+import { ReactElement, useCallback, useEffect, useRef, useState } from 'react'
 
 import { CopyToTeamDialog } from '@core/journeys/ui/CopyToTeamDialog'
 import { useJourneyAiTranslateSubscription } from '@core/journeys/ui/useJourneyAiTranslateSubscription'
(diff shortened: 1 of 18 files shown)


libs/journeys/ui/src/components/PublicGalleryPage/JourneyView/JourneyViewCardActions.tsx
```diff
@@ -52,34 +58,20 @@ interface JourneyViewCardActionsProps {
...
-      .trim()
-      .replace(/^\/+/, '')
-      .replace(/\/+$/, '')
-    return `https://${sanitisedBase}/?useTemplate=${encodeURIComponent(itemId)}`
+function buildUseHref(
+  adminUrl: string,
+  itemId: string,
+  customizable?: boolean | null
+): string {
+  if (customizable === true) {
```
jianwei1, line 66: **Concern (verify — likely intended):** For a customizable template, the public gallery "Use" button now links straight to `admin/templates/<id>/customize`, skipping the `?useTemplate=` → `CopyToTeamDialog`/account-check path that non-customizable templates still use. I confirmed the target page has `getServerSideProps`, so this should gate auth there. Just want to confirm the intended flow for an **anonymous** visitor: clicking "Use" on a customizable template in the public gallery should bounce them to sign-in (with redirect back to `/templates/<id>/customize`) rather than dead-ending. The PR description says this restores prior behavior, so if that hand-off is handled by the customize page's GSSP, this is fine as-is.

---

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
```diff
@@ -53,7 +55,7 @@ interface ActiveUseTemplateDeepLinkProps {
 
 function ActiveUseTemplateDeepLink({
   journeyId
-}: ActiveUseTemplateDeepLinkProps): ReactElement {
+}: ActiveUseTemplateDeepLinkProps): ReactElement | null {
```
jianwei1: **Nit:** the return type is widened to `ReactElement | null`, but neither exit path returns `null` — the early return renders the `Backdrop` and the final return renders `CopyToTeamDialog`. `ReactElement` alone is accurate here. Harmless either way; feel free to leave it.

---

jianwei1 (commented):
## Review summary

Solid, well-scoped fix. Clicking "Use" on a customizable template now routes into the customize flow instead of the copy-to-team dialog, and cross-page navigations correctly drop `shallow: true` (I confirmed `/templates/[journeyId]/customize` has `getServerSideProps`, so shallow would have skipped it and 404'd — the fix is sound).

**Critical:** none.

**Concern:** 1 — verification question on the anonymous public-gallery path (inline on `JourneyViewCardActions.tsx`). Likely intended given this restores prior behavior.

**Nit:** 1 — inaccurate `ReactElement | null` return type (inline on `UseTemplateDeepLink.tsx`).

**What's done well:**
- The schemeless-URL guard was extracted into a tested `libs/adminTemplateLinks/` module and reused across the gallery cards and `UseThisTemplateButton`, replacing inline-only logic — with direct unit coverage of the valid/schemeless/whitespace/id-encoding cases.
- I checked the highest-risk failure mode (that `customizable` might not be selected by the feeding queries): it holds — the `JOURNEY_FIELDS` fragment, `GetTemplateGalleryPage`, and `GetAdminJourneys` all select it and the generated types carry it, so no build break and no silently-dead branch.
- Cross-app hand-off in `UseThisTemplateButton` is thoughtfully tested (origin-matches → in-app push, origin-differs → `location.assign`, schemeless env).

Nothing blocking here. CI was still running at review time — worth a glance once `test`/`build`/`lint-work` land.

---

libs/journeys/ui/src/components/PublicGalleryPage/JourneyView/JourneyViewCardActions.tsx
csiyang: Confirmed — the anonymous path is handled server-side by the customize page's `getServerSideProps` (`apps/journeys-admin/pages/templates/[journeyId]/customize.tsx`): when there is no user and `templateCustomizationGuestFlow` is off, it calls `redirectToLogin(ctx)`, which carries `/templates/<id>/customize` in the `redirect` param so the visitor lands back in the customize flow after sign-in. When the flag is on, `initAndAuthApp({ makeAccountOnAnonymous: true })` creates an anonymous account and the page renders the guest customize flow (`GuestCustomizeHeader`). So no dead end either way — the direct link is intentional and matches the restored behavior.

apps/journeys-admin/src/components/UseTemplateDeepLink/UseTemplateDeepLink.tsx
csiyang: Fixed in f8b3dfc3e — narrowed the return type back to `ReactElement`. The `| null` widening became stale once the loading state started rendering a `Backdrop` instead of returning `null`, so now every exit path returns an element. (This also supersedes CodeRabbit's earlier suggestion to add `| null`, which predated the spinner change.)

jianwei1 (commented):
## Re-review (head `f8b3dfc`)

Both prior findings are resolved — I've marked both threads resolved:

- **Nit (return type)** — fixed in `f8b3dfc`; `ActiveUseTemplateDeepLink` is back to `ReactElement` now that both exit paths render an element (`Backdrop` while loading/redirecting, `CopyToTeamDialog` otherwise). `UseTemplateDeepLink` correctly keeps `ReactElement | null`.
- **Concern (anonymous gallery path)** — confirmed by your writeup: the customize page's `getServerSideProps` handles it (`redirectToLogin` with the `/templates/<id>/customize` redirect param when signed-out and the guest flag is off; anonymous-account guest flow when it's on). No dead end — the direct link is intentional.

CI is green on this head (lint, lint-work, test 1–3/3, build, CodeQL, danger, and the completed Playwright E2E suites all passing; journeys E2E still running but no failures).

Nothing further from me — LGTM. 👍

---

jianwei1 (approved):
## Approved ✅

Clean, well-scoped fix. Clicking "Use" on a customizable template now routes into the customize flow (public gallery card → admin `/templates/<id>/customize`, admin deep-link redirect, and cross-app hand-off from `UseThisTemplateButton`), and cross-page navigations correctly drop `shallow: true` — verified that `/templates/[journeyId]/customize` has `getServerSideProps`, so shallow would have skipped it and 404'd.

Highlights:
- Schemeless-URL guard extracted into a tested `libs/adminTemplateLinks/` module and reused everywhere admin links are built.
- `customizable` is genuinely selected by the feeding queries (`JOURNEY_FIELDS`, `GetTemplateGalleryPage`, `GetAdminJourneys`) with generated types to match — no build break, no silently-dead branch.
- Thoughtful test coverage on the cross-app hand-off (origin-match → in-app push, origin-differ → `location.assign`, schemeless env) and the deep-link loading/redirect states.

Both earlier review threads resolved, and CI is green on `f8b3dfc`. 👍

---

PR #9363, opened 2026-07-09
feat(langfuse-export): English translation of non English chat, facets and previews

jaco-brink:
Closes NES-1762.

## Why

The insights explorer renders real Apologist chat conversations. About half the corpus is not English — Bengali, Spanish, Arabic, French, Afrikaans, Korean, Yiddish, Hindi — so an English-speaking stakeholder who opened `index.html` could not read it.

## What

An opt-in `--translate` pass. Translation is **additive**: the original text is preserved byte-for-byte and always rendered beneath the English, never replaced. English source text passes through untouched — no chip, no second block, no colour change.

Translation is **paid once**. A content-addressed cache (SHA-256 of the source string) means re-rendering the report costs nothing: rebuilding the 30-day production corpus is **288/288 cache hits, zero API calls**.

## The model lies about language, so nothing it says is taken as fact

Three deterministic guards, each written because the model actually got it wrong on this corpus:

| Guard | The lie it caught |
|---|---|
| `cannotBeEnglish()` | Real Bengali assistant replies came back tagged `lang:"en"` — and the wrong verdict was cached. Script is decidable; the item is re-asked alone, and if the model insists, nothing is recorded. A poisoned cache self-heals. |
| `scriptContradictsLanguage()` | A message tagged `es` rendered **"MACHINE-TRANSLATED FROM SPANISH" above visibly Arabic text**. The translation survives; the attribution does not, and the badge falls back to an unnamed `MACHINE-TRANSLATED`. 4 messages take this path. |
| keyword origin | Afrikaans `die` ('the') is also an English word, so the model called it English and it entered the facet rail as the English noun *death*. A term never once used in an English user message is marked with the language it was written in. |

Plus: a "translation" identical to its source is not a translation (`'sup'` came back tagged Spanish, echoed verbatim) — dropped, so no chip ever sits above identical text.

## Two facet bugs this work surfaced

**One language, two rows.** Production sends both `en` and `English` (and `es` / `Spanish, Latin American`) for the same language, so the rail split each across two rows: an analyst filtering "English" **silently missed 3 sessions**. `normalizeLanguageLabel()` folds them — 10 rows become 7, and the counts now sum to the session total.

**The facet answered the wrong question.** "Journey Language" is the journey's *configured* language, not what anyone typed. A new **`Language Typed`** facet counts **user turns only**:

| | Journey | Typed |
|---|---|---|
| Bengali | 20 | 11 |
| English | 18 | 31 |
| Spanish | 10 | 10 |
| Farsi | 1 | — (nobody typed it) |
| Yiddish / Korean | — | 2 / 1 (previously unfilterable) |

Hindi is excluded from `Language Typed` because its only appearance was the **assistant's** reply — a facet labelled "what people wrote" must not attribute to a human a language only the machine produced.

Glossing also exposed a latent gap: the keyword vocabulary was filtered by an *English* stopword list, so `que`, `puede`, `איר` sailed through where `that`, `can`, `you` never could. The gloss is now held to the same two bars an English token faces. 16 suppressed, and counted, not dropped silently.

## Design

Re-themed onto the Jesus Film Project design system. Body copy serif, chrome sans, warm-white editorial surfaces.

- **Navy `#424A66` means exactly one thing: machine translation.** It appears only directly above machine-written text and as the facet-rail dot. Selection uses the strong neutral so navy is never borrowed.
- **Brand red is reserved.** It appears only as the eyebrow and the focus ring. The eyebrow uses `--jfp-red-medium #CB333B` (5.15:1) because the system's `#EF3340` is **4.02:1 at 12px and fails AA** — documented in-code.
- The original recedes by size, never by legibility, and is never truncated in the conversation view.
- RTL (Arabic, Hebrew, Farsi, Yiddish) flips the rule and label to the side the text starts from.
- No dark mode (the system defines none), no gradients, no emoji, no web fonts. The bundle stays offline: no `http://`, no `src=`, two `</script>` tags, all corpus text via `textContent`.

## Verification

Built from a **one-time** 30-day production extract (131 traces, 56 sessions, 2026-06-10 → 2026-07-10). Langfuse was hit once; every subsequent build reads a snapshot.

A differential gate builds the bundle **with and without** `--translate` from that same snapshot and asserts:

```
zip manifest = {dataset.json, index.html, README.txt}   offline + structure OK
255 originals byte-identical                            0 facets invented, 0 dangling keys
139 translated (4 language unattributed), 116 as-is     no message names a language its script disproves
0 non-Latin messages left untranslated                  0 Latin-script messages look like missed translations
7 language facets, no duplicates, summing to 56         Language Typed == what users wrote
no session card out-claims its own conversation
```

`175 unit tests`, `tsc` clean, prettier clean.

The last bug the review found was mine and it was live on this branch: the guard that refuses to name a language the script disproves was applied to messages but **not** to the session card, so two sessions rendered "MACHINE-TRANSLATED FROM BENGALI" on the card above romanized Latin text while the detail view of that same message showed an unnamed badge. One message, two answers, one click apart. Fixed, and gate check `4c` now asserts a card can never out-claim its own conversation.

**LLM-as-judge**, three independent reviewers (monolingual-reader, design-system auditor, adversarial), scoring reader clarity and design adherence out of 10 each against rendered screenshots. Best round scored **18/20 and 18/20**; the adversarial reviewer's findings drove the last four commits — it is the reason `Language Typed` counts user turns only, and the reason a badge can no longer name a language the script disproves. A design auditor then caught that my own RTL fix did not work (`dir="auto"` was being resolved by the Latin word "Original"). Both are fixed and pinned by specs.

Reviewers disagreed on one point and I did not chase the last one: whether the selected-card border should be brand red. Two called red-on-selection a violation ("spending reserved red on chrome"); one wanted it back. It stays neutral.

## Known and accepted

- `god` and `dios`→`god` render as two adjacent keyword rows with the same English word. They are genuinely different filter keys over different sessions; merging would misstate the counts. The original term sits beside the gloss.
- Doubled text makes long Bengali/Arabic threads long. That is the cost of keeping the original present and authoritative.
- `output/.translation-cache.json` holds English renderings of real user messages. It is gitignored, chat-derived, and must be deleted alongside `.env`.

## Scope

`tools/langfuse-export/` only. No `.env`, no `output/`, no bundle artifacts committed.

## Summary by CodeRabbit

* **New Features**
  * Added optional machine translation previews for non-English messages and keyword facets, including RTL-aware rendering and a default-collapsed “Original” disclosure.
  * Introduced SHA-256 content-addressed translation caching with reuse and self-healing behavior, plus safe failure degradation.
  * Updated the offline insights explorer UI/UX to show machine-translated English (non-guaranteed accuracy) while preserving original text and improving language/facet labeling.

* **Documentation**
  * Expanded the export/explorer README with translation workflow, cache semantics, and detailed accuracy/attribution caveats.

* **Tests**
  * Expanded CLI, translation heuristics, dataset attribution, cache, and explorer rendering coverage.


tools/langfuse-export/src/pipeline/dataset.ts
```diff
@@ -84,13 +173,65 @@ export function buildDataset(
...
+    // in its own languages, and counting those made a 'Language Typed -> Hindi'
+    // facet whose single session had a Bengali-speaking user and a Hindi-speaking
+    // bot. A user message with no translation was English; one whose language we
+    // could not attribute contributes nothing rather than a guess.
+    const typedLanguages = new Set<string>()
+    if (translations != null) {
+      for (const message of messages) {
+        if (message.role !== 'user') continue
+        if (message.textEnglish == null) {
+          typedLanguages.add('English')
```
jaco-brink: **Concern:** Failed or absent translations are silently attributed to `Language Typed → English`. `withTranslation` leaves `textEnglish` unset both when a message is genuinely English *and* when its translation record is simply missing — and `translateTexts` catches batch/item failures and returns a non-null *partial* map, so a partial or total OpenRouter outage yields missing entries rather than a null map. Every un-translated Bengali/Arabic user turn then falls into this branch and is tagged English, so the new `Language Typed` facet (the headline "what people actually wrote") asserts English for a session whose original is visibly non-English right beside it. With a total failure, every session is tagged English. The comment at line 215 encodes the assumption ("A user message with no translation was English"). Deterministic fix: gate the English attribution on a positive signal — e.g. only add `'English'` when `!cannotBeEnglish(message.text)`, so a non-Latin message that failed to translate is never labelled English.

tools/langfuse-export/src/pipeline/facets.ts
```diff
@@ -197,6 +197,53 @@ function tokenize(text: string): string[] {
...
+  // 'es', 'es-419', 'pt_BR' -> the base subtag.
+  const code = trimmed.toLowerCase().split(/[-_]/)[0]
+  if (/^[a-z]{2,3}$/.test(code) && LANGUAGE_NAMES[code] != null) {
+    return LANGUAGE_NAMES[code]
+  }
+  // 'Spanish, Latin American' / 'Bengali (Indian)' -> the base language. The
+  // regional variant is dropped: the facet answers "which language", and the
+  // exact variant is still on the session (and in dataset.json).
+  const base = trimmed.split(/[,(]/)[0].trim()
+  return base.length > 0 ? base : trimmed
```
jaco-brink, line 244: **Concern:** The model's `lang` value is trusted verbatim, so one language can split into two facet rows. `normalizeLanguageLabel` maps a known 2–3 letter code to a display name, but falls through here to returning the raw string for anything else. If the model returns `"bengali"` on one item and `"bn"` on another (plausible, especially under the strict re-ask prompt "Identify the actual language"), you get two rows — `Bengali` and `bengali` — and the counts no longer sum cleanly. `parseTranslations` upstream also accepts any non-empty `lang` without validating it's a known code. A whitelist-or-drop on `lang` (unknown → treat as unattributed) closes it.

tools/langfuse-export/run.ts
```diff
@@ -189,13 +216,65 @@ async function main(): Promise<void> {
...
+      const strings = collectTranslatable(sanitised, facets)
+      const cache = loadCache(cachePath)
+
+      // Translate only the cache misses; each item is keyed by its content hash.
+      // A cached "this is English" verdict on non-Latin-script text is a lie an
+      // earlier run recorded — treat it as a miss so the cache self-heals.
+      const misses = strings.filter((text) => {
+        const cached = cache.get(cacheKey(text))
+        if (cached == null) return true
+        return cached.sourceLanguage === 'en' && cannotBeEnglish(text)
```
jaco-brink: **Concern:** This self-heal predicate — and the batching / re-ask / straggler orchestration it drives in `openrouter.ts` (`batchByCharBudget`, the `cannotBeEnglish` veto that refuses to "cache a lie", the individual straggler sweep) — is the riskiest logic in the PR and has zero unit tests; only the pure `parseTranslations` is covered. That's exactly where the two concerns above live. Several of these are pure and cheap to test: this predicate `cached.sourceLanguage === 'en' && cannotBeEnglish(text)` and `batchByCharBudget`'s char-budget / item-cap / oversize-single-item split are both trivially unit-testable. The README is candid that the OpenRouter client is manual-verified — closing this gap would make the self-heal and poisoned-cache promises regression-proof.

tools/langfuse-export/src/clients/openrouter.ts
```diff
@@ -115,6 +137,186 @@ export async function synthesizeThemes(
...
+    }
+  }
+}
+
+// Detect + translate a set of strings. Returns a Map keyed by each item's id.
+// A batch that fails after all retries is left out of the Map (those items stay
+// untranslated) and a warning is logged — the run never throws or fabricates.
+//
+// Input MUST be already-sanitised text — this is the module's trust boundary.
+export async function translateTexts(
```
jaco-brink, line 268: **Concern:** No fail-fast on a hard outage. On a total OpenRouter failure every batch exhausts `MAX_REQUEST_RETRIES` (4) with exponential backoff, is caught and logged, then *every* item becomes a straggler and is retried individually — again up to 4 attempts against the 60s timeout. A ~260-string corpus can grind for many minutes before it degrades to no-translation. Bounded, logged, and the original is preserved, so this is a UX/time issue, not a correctness one — but a simple circuit-breaker (if the first K batches all fail, stop retrying and degrade immediately) would fix it cheaply.

tools/langfuse-export/src/pipeline/explorer.ts
```diff
@@ -137,30 +332,128 @@ const VIEWER_JS = `
...
+    // label is pinned ltr and excluded; a known RTL language is stated outright.
+    var orig = el('div', 't-original');
+    orig.setAttribute('dir', RTL_CODES[code] ? 'rtl' : 'auto');
+    orig.style.display = 'none';
+    var olabel = el('div', 'o-label', name ? ('Original \\u00b7 ' + name) : 'Original');
+    olabel.setAttribute('dir', 'ltr');
+    orig.appendChild(olabel);
+    var otext = el('div', 'o-text');
+    otext.setAttribute('dir', 'auto');
+    renderMarkdown(otext, original);
```
jaco-brink, line 415: **Concern (borderline nit):** The original is rendered through the same `renderMarkdown` pass as the translation, which is at odds with the repeated "byte-for-byte / never altered / source of truth" framing. The *stored* bytes are intact (the JSON round-trip test proves it), but the *rendered* original reinterprets markdown: a user who typed a leading `> `, `- `, `* `, or `# ` has the marker stripped and the line reflowed as a quote/list/heading, and a lone `**` yields an empty `<strong>`. Real-world impact is low (chat users rarely type markdown, and assistant replies *are* markdown so rendering them is correct), but the pane labelled authoritative no longer shows exactly what was typed. Worth either softening the "byte-for-byte" copy for the rendered view, or not applying block markdown to the human original.

jaco-brink (commented):
## Review summary — two independent reviews, synthesized

Reviewed with two independent sub-agents (a correctness/security lens and a maintainability/test-quality lens) plus my own spot-checks of the load-bearing findings. **No Critical issues.** Both reviewers independently tried to break the XSS defense and could not, and I confirmed it too. This is careful, conscientious work; the notes below are correctness-robustness Concerns and cleanup Nits, none blocking. (Draft PR, so this review is a `COMMENT`.)

**Concerns (5, inline):** English misattribution on failed translation (the top one), the model's `lang` code splitting one language into two facets, the untested OpenRouter/self-heal orchestration, no fail-fast on an outage, and markdown applied to the "byte-for-byte" original.

**Nits (pattern-level):**
- **Dead / misleading code:** `isEnglishStopword` is exported and has its own tests but no production caller (`facets.ts:280`); `RTL_LANGUAGES` is a dead export whose comment claims the viewer uses it when the viewer has its own inline map (`types.ts:111`); a test asserts on the unused `--jfp-red-dark` token with a comment describing the selection state as red when the code deliberately uses the neutral (`explorer.spec.ts`); a redundant `.trim()` in the card-preview lookup (currently inert).
- **Three drifting language/script tables** (`LANGUAGE_NAMES`, viewer `LANG_NAMES`, `EXPECTED_SCRIPT`) — they have already drifted, and the Cyrillic/Greek/Thai/Japanese branches of `dominantScript` have no tests. The viewer's duplication is genuinely unavoidable (it's a string blob with no imports); worth a cross-linking comment plus a few more `scriptContradictsLanguage` cases.
- **Fixture fidelity:** several explorer specs use language *names* where the producer emits BCP-47 *codes*, so the preview-dedup path is never exercised against production-shaped data.

**What's genuinely strong:**
- The `</script>` XSS inertness is *proven*, not asserted — the JSON data-island escapes every `<` to `\u003c`, the hand-rolled markdown parser is `createTextNode`/`textContent`-only, and it is tested end-to-end with a real `` payload plus an exact-count assertion on raw `</script>` tags.
- Language guards are well-layered — `cannotBeEnglish` applied at three sites, `scriptContradictsLanguage` drops the *attribution* while keeping the translation, identical-source translations dropped, a SHA-256 content-addressed cache, and a self-heal that stays safe even when the model stubbornly repeats a wrong verdict (renders inertly, never a false "translated" chip).
- Tests are real (184 green) with meaningful assertions on hostile-input inertness, the user-turn-only typed-language filter, and the RTL `ltr`-pin; naming is intent-revealing; the README is honest about the manual-verification boundary.

**Process (Danger is red on non-code items):** the PR title is 102 chars (>100) and not lowercase per commitlint; the PR needs an assignee and a reviewer; big-PR warning; CLA is pending.

jaco-brink:
## Review feedback addressed (340aa195d)

Six threads fixed with tests; two consciously deferred. Full suite green (199 tests), `tsc` clean, Prettier clean.

**Fixed (correctness):**
- `dataset.ts:222` — A non-Latin user turn whose translation *failed* (item error or total OpenRouter outage) fell into the `textEnglish == null` branch and was tagged **English** in the Language Typed facet. English is now asserted only when `cannotBeEnglish(text)` does not rule it out; a Bengali/Arabic turn that failed to translate contributes nothing rather than a false claim.
- `run.ts:255` (CodeRabbit) — A poisoned `en` cache entry was treated as a miss but left in the cache, so a failed strict re-ask let the rebuild re-emit the false English attribution. It's now **evicted** when identified. Predicate extracted as the pure, tested `isPoisonedEnglish()`.
- `translation-cache.ts:76` (CodeRabbit) — `saveCache` now catches and logs `writeFileSync`/`mkdirSync` failures instead of propagating them; a persistence error no longer aborts the run and discards the freshly-fetched translations that already drive the in-memory dataset.

**Fixed (adjusted):**
- `facets.ts:244` — Addressed at the source: `parseTranslations` now **canonicalises** the model's `lang` (folds a full-name answer like `"bengali"` onto its code `bn`) so one language can't split into two facet rows. Deviated from the suggested hard whitelist-or-drop so a real-but-unlisted code (e.g. Tamil `ta`) is never silently dropped — unrecognised codes pass through unchanged, only nameable ones are unified.
- `run.ts:239` — Added the missing unit tests: `batchByCharBudget` (char-budget split / 30-item cap / oversize-single-item isolation), the extracted `isPoisonedEnglish` self-heal predicate, `canonicalLanguageCode`, the `saveCache` failure path, and a dataset guard for the non-Latin-not-English case.

**Fixed (maintainability):**
- `explorer.ts:368` (CodeRabbit) — `languageNames` never truncated (the 4-language test proves it); corrected the misleading comment and added the empty-list guard. Did not add truncation, which would contradict the existing test.

**Deferred (self-flagged low-priority, kept open):**
- `openrouter.ts:266` — circuit-breaker on a total outage. Non-correctness (bounded, logged, original always preserved); the added state isn't worth it for a manual batch tool over ~260 strings.
- `explorer.ts:414` — the collapsed original rendered through markdown vs the "byte-for-byte" framing. Stored bytes are intact (round-trip test proves it) and rendering the original is intentional and tested; only copy would change.

PR #9364, opened 2026-07-13
fix(NES-1760): sync upload variant languages

tanflem:
## Summary

Fixes upload-created video variants so they update the derived language discovery state after a successful published upload.

## Root Cause

The video upload worker had its own direct `VideoVariant` create/update path. That bypassed the normal side effects used by the GraphQL video variant mutations, so a variant could be playable by direct language lookup while the parent video's `availableLanguages` stayed stale.

## Changes

- Routes the upload worker through the shared `createOrUpdateVideoVariant` helper.
- Syncs published upload variants into `availableLanguages` and parent collection languages.
- Refreshes video and variant Algolia records and cache entries after upload finalization.
- Adds regression coverage for the worker path and ready-Mux resume path.
- Adds the HTML walkthrough plan under `plans/`.

## Validation

- `./node_modules/.bin/tsc -b apis/api-media/tsconfig.json`
- `../../node_modules/.bin/vitest run --coverage=false --config vitest.config.mts` from `apis/api-media` — 672 tests passed


tanflem:
Superseded by #9450, which re-lands this work as a single commit authored by a CLA-signed contributor (this PR was blocked on `license/cla` for commit c83f0fef). Phoebe's test-mock fix is preserved there via a Co-Authored-By trailer. The two duplicated `plans/*.html` files were dropped — they were 1266 of the 1416 added lines here.

PR #9367, opened 2026-07-14
chore: pilot TypeScript 7 type-checking on apis

mikeallisonJS:
## What

Pilots TypeScript 7 (the Go-native compiler) as the type-checker for all six `apis/*` workspaces, as phase 1 of a hybrid migration. `typescript` stays at 5.x for everything that consumes the TS compiler API (typescript-eslint, Next, Nx executors, editor default).

## How

- **`typescript7` pnpm alias** (`npm:typescript@^7.0.2`) + root `tsc7` script. Both packages ship a `tsc` bin, so `.bin/tsc` is ambiguous — the script (`node node_modules/typescript7/bin/tsc`) is the unambiguous entry point and the single line to change when the hybrid era ends.
- **Six api `type-check` targets** now run `pnpm tsc7 -b <tsconfig>`, replacing TS 5.9.
- **`tsconfig.base.json`**: removed `baseUrl`, made `paths` relative — TS 7 removed `baseUrl` (TS5102). Fully supported by TS 5 since 4.1; verified non-pilot projects still typecheck clean.
- **Api tsconfigs**: `module: commonjs` + inherited `node10` resolution (both removed in TS 7) → `esnext` + `bundler`, matching api-journeys' existing config. `nodenext` was rejected: it demands explicit `.js` extensions on dynamic `import()` in `src/workers/`, the wrong model for webpack-bundled code.
- **`.storybook/preview.d.ts`** (new): types the untyped `preview.js`. Fixes a pre-existing TS 5 error in api-users' stories program that was masked by stale `.tsbuildinfo` caches.
- **`docs/solutions/conventions/typescript-7-hybrid-migration.md`**: gotchas + remaining rollout phases (apps/libs/workers → editor flip → retire TS 5).

## Verification

- Clean (`.tsbuildinfo`-free) `tsc7 -b` passes on all six apis.
- `apps/journeys` (TS 5, non-pilot) typechecks clean against the modified base config; api-media's app config is clean under TS 5, so webpack builds are unaffected.
- Prisma clients must be generated for type-check to pass (unchanged requirement; documented).

## Summary by CodeRabbit

* **Bug Fixes**
  * Video player validation and missing-data errors now display as clear in-page alerts instead of unrendered status objects.

* **Documentation**
  * Added guidance for the TypeScript 7 hybrid migration, including rollout status, configuration requirements, and known cache considerations.

* **Chores**
  * Introduced TypeScript 7 type-checking support for API projects with dedicated configurations and workspace tooling.

### DIFF
--- a/.storybook/preview.d.ts
+++ b/.storybook/preview.d.ts
@@ -1,0 +1,39 @@
+import { ComponentType } from 'react'
+import { Loader, Parameters } from 'storybook/internal/csf'
+
+export declare const parameters: {
+  backgrounds: {
+    grid: { disable: boolean }
+    disabled: boolean
+  }
+  chromatic: { viewports: number[] }
+  controls: { disable: boolean }
+  viewport: { options: Record<string, unknown> }
+  apolloClient: { MockedProvider: ComponentType<Record<string, unknown>> }
+}
+
+export declare const globalTypes: {
+  theme: {
+    name: string
+    description: string
+    toolbar: {
+      title: string
+      icon: string
+      items: ['light', 'dark', 'all']
+    }
+  }
+}
+
+export declare const initialGlobals: {
+  theme: string
+}
+
+declare const preview: {
+  loaders: Loader[]
+  parameters: Parameters
+  globalTypes: typeof globalTypes
+  initialGlobals: typeof initialGlobals
+  tags: string[]
+}
+
+export default preview
--- a/apis/api-analytics/project.json
+++ b/apis/api-analytics/project.json
@@ -76,7 +76,7 @@
     "type-check": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "pnpm tsc7 -b apis/api-analytics/tsconfig.json"
+        "command": "pnpm tsc7 -p apis/api-analytics/tsconfig.ts7.json"
       }
     },
     "codecov": {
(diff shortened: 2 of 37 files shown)


csiyang (approved):
LGTM 🚀 — well-architected pilot. The shared TS5 `tsconfig.base.json` is left untouched (TS7 gets its own `tsconfig.base.ts7.json` + per-workspace `tsconfig.ts7.json`), so existing webpack/Next/vitest/editor builds carry zero risk. The `tools/typescript7` isolation for the `tsc7` bin collision is a clean solution.

Two minor notes, neither blocking:

1. **Duplicated `paths` map.** `tsconfig.base.ts7.json` now carries a copy of the alias map with a "keep in sync" comment. Drift is possible but self-catching (surfaces as a TS7 resolution error). Fine for the hybrid era — just flagging the maintenance cost.

2. **Body vs. diff + mild scope bleed.** The PR body says it modifies `tsconfig.base.json` (removing `baseUrl`), but the shipped diff creates a separate `tsconfig.base.ts7.json` instead — the shipped approach is safer, the body's just stale. Also the `arclight` page.tsx fix and `.storybook/preview.d.ts` addition ride along in an apis-scoped PR; both are real, correct fixes (the arclight one — an RSC returning plain objects instead of JSX — is a genuine pre-existing bug TS7 surfaced), just slightly beyond the stated scope.

PR #9380, opened 2026-07-20
fix(journeys-admin): group font dialog actions and align text response hint (QA-503)

sharonemathew:
**Issue**
[QA-503](https://linear.app/jesus-film-project/issue/QA-503/select-fonts-modal-cancel-button-misplaced-label-style-inconsistent) — On the Select Fonts modal, the Cancel button is pinned to the far-left while Confirm sits far-right (unusually wide gap), and the "Hint text" in the preview is indented so it doesn't line up with the "Label" or the input field.

The ticket's third item — "Label" text rendering in an inconsistent style in the preview — is **not addressed by this PR** and is left for a follow-up.

**Fix**
- **Footer buttons:** changed the action Stack from `justifyContent="space-between"` to `flex-end` with `gap={2}` (matching sibling dialogs), so Cancel and Confirm are grouped together at the bottom-right.
- **Hint text:** added `mx: 0` to the TextResponse `formHelperText` slot to remove MUI's default horizontal margin, so the hint aligns flush-left with the label and field. Note: `TextResponse` lives in the shared `libs/journeys/ui`, so this alignment change applies everywhere a text response hint renders (editor canvas and live journey player), not just the modal preview.
- **Close-while-saving guard:** Cancel and the title close button are disabled while the fonts mutation is in flight, and all close paths (backdrop click, Escape) route through a guard that no-ops mid-save — previously cancelling mid-save closed the dialog but the write still landed, firing a "Fonts updated" snackbar over a dismissed dialog.

**Test plan**
- Unit: `ThemeBuilderDialog.spec.tsx` asserts the action buttons share one flex row justified `flex-end` with Cancel preceding Confirm, and covers the close-while-saving guard (11 tests pass). The exact 16px gap is left to Chromatic since jsdom doesn't evaluate the sibling combinator.
- Unit: `TextResponse.spec.tsx` asserts the hint's computed horizontal margins equal the label's (verified red without the `mx: 0` fix, green with it).
- Manual: opened a Journey → text block → edit theme → Select Fonts; confirmed Cancel/Confirm grouped bottom-right and "Hint text" flush-left with "Label" and the input field (screenshot verified).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the theme builder dialog from closing while font changes are being saved.
  * Disabled dialog actions during saving to avoid interrupted updates.
  * Aligned text field hint messages with their labels for a cleaner layout.
  * Improved the footer layout so Cancel and Confirm actions are consistently positioned.


Jaxsbr (commented):
_Posted from the wrong GitHub account by mistake — inline comments have been removed. See the full review from @jaco-brink below._

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.spec.tsx
```diff
@@ -407,4 +440,55 @@ describe('ThemeBuilderDialog', () => {
...
+      </SnackbarProvider>
+    )
+
+    fireEvent.click(screen.getByText('Confirm'))
+
+    expect(screen.getByText('Cancel')).toBeDisabled()
+    expect(screen.getByTestId('dialog-close-button')).toBeDisabled()
+
+    fireEvent.click(screen.getByText('Cancel'))
+    expect(handleClose).not.toHaveBeenCalled()
```
jaco-brink: **⚠️ This test still passes even if the fix is deleted.**

**What I did:** I deleted the one-line safety check from the code (`if (saving) return` in `handleClose`) and re-ran the tests. All 11 still passed. That means if someone removes that safety check by accident in the future, no test will warn them.

**Why it doesn't catch it:** this test only clicks the **Cancel** button. While saving, Cancel is greyed out — and clicking a greyed-out button does nothing at all. So "the dialog didn't close" is guaranteed regardless of whether the safety check exists.

The safety check is really there for the **two ways a user can close a dialog without touching a button**:

- pressing the **Escape** key
- clicking the **dark area outside** the dialog

Neither of those is tested.

**✅ The good news — the code is correct.** I wrote a test for Escape + outside-click, deleted the safety check, and it failed (`expected "spy" to not be called at all, but actually been called 1 times` — the dialog closed mid-save, exactly the bug you set out to prevent). I put the safety check back and it passed, 12/12 green. **So this is a missing test, not a broken feature. No production code needs to change.**

**Suggested test to add** — I ran this exact code, red without the guard, green with it:

```tsx
it('should not close on escape or backdrop click while fonts are saving', async () => {
  const loadingMock = {
    delay: 60,
    request: {
      query: JOURNEY_FONTS_UPDATE,
      variables: {
        id: 'theme-id',
        input: {
          headerFont: FontFamily.Montserrat,
          bodyFont: FontFamily.Inter,
          labelFont: FontFamily.Nunito
        }
      }
    },
    result: {
      data: {
        journeyThemeUpdate: {
          __typename: 'JourneyTheme',
          id: 'theme-id',
          journeyId: 'journey-id',
          headerFont: FontFamily.Montserrat,
          bodyFont: FontFamily.Inter,
          labelFont: FontFamily.Nunito
        }
      }
    }
  }

  render(
    <SnackbarProvider>
      <MockedProvider mocks={[loadingMock]}>
        <JourneyProvider value={{ journey: mockJourney, renderMode: 'admin' }}>
          <ThemeBuilderDialog open onClose={handleClose} />
        </JourneyProvider>
      </MockedProvider>
    </SnackbarProvider>
  )

  fireEvent.click(screen.getByText('Confirm'))

  // escape and backdrop bypass the disabled buttons entirely, so these are
  // the paths the handleClose guard actually has to block
  fireEvent.keyDown(screen.getByRole('dialog'), { key: 'Escape', code: 'Escape' })
  expect(handleClose).not.toHaveBeenCalled()

  fireEvent.click(document.querySelector('.MuiBackdrop-root') as Element)
  expect(handleClose).not.toHaveBeenCalled()

  // once the mutation settles the dialog closes itself
  await waitFor(() => expect(handleClose).toHaveBeenCalled())
})
```

Run it from `apps/journeys-admin` with:

```bash
npx vitest run --config vitest.config.mts \
  'src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.spec.tsx' \
  --coverage=false
```

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.tsx
```diff
@@ -208,16 +215,22 @@ export function ThemeBuilderDialog({
         title: t('Select Fonts'),
         closeButton: true
       }}
+      slotProps={{ titleButton: { disabled: saving } }}
       dialogActionChildren={
-        <Stack direction="row" justifyContent="space-between" width="100%">
-          <Button variant="outlined" color="secondary" onClick={onClose}>
+        <Stack direction="row" justifyContent="flex-end" gap={2}>
```
jaco-brink: **Nit: `justifyContent="flex-end"` on this Stack has no effect — I measured it.**

**What I did:** rendered this dialog in a real browser (Chromium via Playwright) and measured exactly where the Cancel and Confirm buttons land, with and without this setting.

| Version | Cancel position | Confirm position | Gap |
| --- | --- | --- | --- |
| Your code (`justifyContent="flex-end"`) | 883px | 998px | 16px |
| Same, but `justifyContent` deleted | **883px** | **998px** | **16px** |
| The old buggy version (`space-between width="100%"`) | 176px | 998px | 723px |

The first two rows are pixel-for-pixel identical — so this setting isn't doing anything. The third row is the control: it shows the measurement really does catch layout changes, so the "identical" result is trustworthy rather than a blind test.

I repeated it across widths, in case the setting mattered on a squeezed screen:

```
viewport 1280px | A: Cancel@883 Confirm@998 | B: Cancel@883 Confirm@998 | IDENTICAL
viewport  768px | A: Cancel@507 Confirm@622 | B: Cancel@507 Confirm@622 | IDENTICAL
viewport  390px | A: Cancel@129 Confirm@244 | B: Cancel@129 Confirm@244 | IDENTICAL
viewport  320px | A: Cancel@ 59 Confirm@174 | B: Cancel@ 59 Confirm@174 | IDENTICAL
```

**Why:** the footer bar holding these buttons (MUI `DialogActions`) already computes to `justify-content: flex-end` on its own, and this Stack is only as wide as the buttons inside it.

**What actually fixed the bug** was removing `width="100%"`. In the same measurements the Stack's width drops from **928px → 221px** — that is what stops the buttons flying to opposite corners.

**Nothing is broken and the fix works correctly.** The only snag is that the new test asserts `justifyContent: 'flex-end'` as though that is what fixed it, which could mislead the next person to touch this. Worth knowing: a unit test can't catch this, because jsdom has no layout engine — it can only confirm the setting was *written*, never that it *did* anything.

Suggestion: either drop `justifyContent="flex-end"` (and its assertion), or keep it with a one-line comment noting it's defensive rather than load-bearing.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.tsx
```diff
@@ -208,16 +215,22 @@ export function ThemeBuilderDialog({
...
+            disabled={saving}
+          >
             {t('Cancel')}
           </Button>
           <Button
             variant="contained"
             color="primary"
             onClick={handleSubmit}
-            disabled={loading || createLoading}
+            disabled={saving}
```
jaco-brink, line 236: **Nit: while saving, the dialog looks frozen.**

Cancel greys out, the X greys out, Confirm greys out — and nothing spins or moves. On a slow connection the user gets a dead-looking modal they can't escape from, with no sign anything is happening.

MUI's `Button` has a built-in `loading` prop for exactly this, and the sibling dialog in this repo already uses it (`CollectionDialogFooter.tsx:90`):

```tsx
<Button
  variant="contained"
  color="primary"
  onClick={handleSubmit}
  loading={saving}
  disabled={saving}
>
```

Worth noting too: Apollo has no default timeout on mutations, so if a request genuinely hangs, the guard means there is no way out of the dialog at all. A spinner at least makes it clear the app is working rather than broken.

jaco-brink (commented):
## Review summary

Small, well-scoped fix — CI green, branch up to date with `main`, and the intent is clearly documented in the description (including the deferred third ticket item and the shared-library blast radius). **No blocking issues.** One test-coverage gap and two nits, all verified by running code rather than by eye.

**Concern (1):** the close-while-saving test passes even with the guard deleted — the guard itself is correct, but nothing protects it. Suggested replacement test included inline (verified red→green).

**Nits (2):** an inert `justifyContent` that the new test asserts as load-bearing (measured in a real browser, with a control), and a frozen-looking dialog during save.

**Verified while reviewing:**

- ✅ `TextResponse` hint-alignment test is genuinely meaningful — removed `mx: 0` and it failed with `expected '14px' to be '0px'`. Good guard on a shared-lib visual change.
- ✅ The close guard genuinely blocks Escape and backdrop — proven by removing it and watching a new test go red.
- 👍 `gap={2}` is the right mechanism here specifically because `StyledDialog` zeroes `marginLeft` on non-first `DialogActions` children, so MUI's default spacing wouldn't have applied.
- 👍 The guard matches the existing `CollectionDialog.guardedClose` precedent, and disabling the title X via `slotProps.titleButton` is actually more correct than `CollectionDialog`, which passes `loading` to `Dialog` (a no-op on the `dialogActionChildren` path).

🤖 Review assisted by Claude Code

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.spec.tsx
sharonemathew: Confirmed and fixed in 4e271a4e8 — thanks for running the deletion, that was the whole point.

Added your escape/backdrop test essentially verbatim. I also went one step further on the old test: rather than leave the tautological pair in place, I deleted the `fireEvent.click(Cancel)` → `not.toHaveBeenCalled()` assertions and retitled it `should disable the close affordances while fonts are saving`, which is what it actually proves. Its two `toBeDisabled()` assertions are real (they cover `disabled={saving}` and `slotProps.titleButton`), so those stayed.

Verified the same way you did — deleted `if (saving) return` and got your exact predicted failure:

```
× should not close on escape or backdrop click while fonts are saving
  → expected "spy" to not be called at all, but actually been called 1 times
```

Guard restored, 12/12 green. No production code changed for this one.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.tsx
sharonemathew: Agreed — dropped the prop and its assertion in 4e271a4e8.

Your measurement now reproduces on the deployed preview against the build with the prop actually removed, which is a stronger control than removing it locally:

```
stackJustifyContent: "normal"      ← prop gone
stackGap:            "8px"         ← unchanged
stackWidth:          "198.578px"   ← unchanged
```

Identical gap and width to the same dialog measured on the previous build. And `DialogActions` computes `justify-content: flex-end` on its own in production too, which is the mechanism behind your result.

**One wrinkle worth recording.** The gap is **8px** in the real app, not 16px — the base theme sets `spacing: 4` (`libs/shared/ui/src/libs/themes/base/tokens/spacing.ts:6`), so `gap={2}` resolves to 8px. Your 16px is what you get when the dialog renders *outside* the app's `ThemeProvider`, which a standalone Playwright mount would do. Doesn't affect your conclusion at all — the `IDENTICAL` comparison holds either way — but it did affect my test: I'd written `gap: '16px'`, which passed only because the spec has no `ThemeProvider` and falls back to MUI's default `spacing: 8`. That's the same species of misleading assertion you flagged here, so it's now theme-independent:

```ts
const { gap } = getComputedStyle(actions as Element)
expect(['', 'normal', '0px']).not.toContain(gap)
```

Still fails if `gap={2}` is deleted (verified), without hard-coding a number that contradicts production.

I also replaced the `justifyContent` assertion with one that guards what you identified as the real fix:

```ts
expect(actions).not.toHaveStyle({ width: '100%' })
```

Re-adding `width="100%"` turns that red — so the original bug is now covered, where previously nothing covered it. Plus a one-line comment on the Stack recording why `gap` and not margin (`StyledDialog` zeroes `marginLeft` on non-first `DialogActions` children), since that's the non-obvious bit you had to dig out.

apps/journeys-admin/src/components/Editor/Slider/Settings/CanvasDetails/Properties/blocks/Typography/ThemeBuilderDialog/ThemeBuilderDialog.tsx
sharonemathew: Added `loading={saving}` in 4e271a4e8, matching the `CollectionDialogFooter` precedent you pointed at.

Verified on the deployed preview (not just jsdom) — measured 400ms after clicking Confirm, sampled inside the page so there's no round-trip lag:

| | |
|---|---|
| `role="progressbar"` present | `true` |
| size | 16×16 |
| Confirm `.MuiButton-loading` | `true`, `disabled: true` |
| Cancel disabled | `true` |
| Title X disabled | `true` |
| Dialog open | `true` |

Since the save is normally sub-second, I held the `JourneyFonts` response via a temporary `window.fetch` wrapper to observe the state. Also added a spinner assertion to the unit test — removing `loading={saving}` fails it with `Unable to find an accessible element with the role "progressbar"`.

**Your Apollo-timeout point is not addressed, deliberately.** The spinner makes a slow save legible, but it does nothing for a genuinely hung request — the guard still leaves no exit. Holding the response for 12s was an unintentionally good demo: every close path was blocked for the full 12 seconds. That needs a real timeout or a bail-out affordance, which is a behaviour change beyond this nit, so I'd rather it be its own ticket than get smuggled into a QA-503 layout fix. Happy to raise one.

Two smaller things while I was in there:

- The spinner inherits the disabled text colour, `rgba(0, 0, 0, 0.26)`, so it's legible but faint against the greyed button. Could take a deliberate colour if you think it's too subtle.
- Separately and pre-existing: `storybook-addon-apollo-client` is a dependency and `.storybook/preview.js` sets `apolloClient: { MockedProvider }`, but the addon isn't in `main.ts`'s `addons` array — so the parameter is inert and both `ThemeBuilderDialog` stories crash with Apollo InvariantError 58 instead of rendering. Worth knowing because it means Chromatic is snapshotting an error box for this component, so the "leave the exact gap to Chromatic" plan in my PR description doesn't currently hold. Not touched here — it's a shared-config change well outside this ticket.

sharonemathew:
## Update — description corrections + review follow-up

Addressed all three of @jaco-brink's review comments in 4e271a4e8. A few claims in the
original description above are now superseded — correcting them here rather than silently
editing the body.

### Superseded in the original description

**1. The footer fix was mis-attributed.** The description says the Stack changed from
`justifyContent="space-between"` to `flex-end`. The `flex-end` was inert and has now been
removed. `DialogActions` already computes `justify-content: flex-end` itself, so a
content-width row lands at the trailing edge unaided — **dropping `width="100%"` is what
actually fixed it**. Verified on the deployed preview against the build with the prop
genuinely absent:


Same gap and width as before removal, so the prop was contributing nothing.
`gap` rather than margin because `StyledDialog` zeroes `marginLeft` on non-first
`DialogActions` children.

**2. The gap is 8px, not 16px.** The base theme sets `spacing: 4`
(`libs/shared/ui/src/libs/themes/base/tokens/spacing.ts:6`), so `gap={2}` resolves to 8px
in the app — measured 8px on the preview. 16px is what you get outside the app's
`ThemeProvider`, which is what the provider-less spec renders. The test no longer
hard-codes either value; it asserts a gap is set.

**3. "Left to Chromatic" doesn't hold.** `storybook-addon-apollo-client` is a dependency
and `.storybook/preview.js` sets `apolloClient: { MockedProvider }`, but the addon isn't
in `.storybook/main.ts`'s `addons` array — so the parameter is inert and both
`ThemeBuilderDialog` stories crash with Apollo InvariantError 58. Chromatic is
snapshotting an error box for this component, not the dialog. Pre-existing and out of
scope here; flagged for separate work.

### Added since

- **Save progress:** Confirm now uses MUI's `loading` prop, so a slow save shows a spinner
  instead of a dialog that greys out with no sign of activity (matching
  `CollectionDialogFooter`). Verified on the preview: `role="progressbar"` present, 16×16,
  with Cancel and the title X both disabled.
- **Escape/backdrop coverage:** the previous close-while-saving test passed even with the
  `if (saving) return` guard deleted — a disabled Cancel swallows its own clicks, making
  the assertion tautological. Replaced with a test that drives Escape and backdrop, the
  paths the guard actually blocks. Tautological assertions removed rather than left in.
- **Bug-guarding assertion:** the spec now asserts the action row is *not* `width: 100%`,
  which is the condition that caused the original bug. Re-adding `width="100%"` turns it
  red; previously nothing covered it.

### Test plan (updated)

- Unit: `ThemeBuilderDialog.spec.tsx` — **12 tests**. Every new assertion was verified by
  deleting the corresponding fix and confirming red: removing the guard →
  `expected "spy" to not be called at all, but actually been called 1 times`; removing
  `gap={2}` → gap assertion fails; re-adding `width="100%"` → width assertion fails;
  removing `loading={saving}` →
  `Unable to find an accessible element with the role "progressbar"`.
- Unit: `TextResponse.spec.tsx` — unchanged, still red without `mx: 0`.
- Preview (`journeys-admin-9380`): footer gap 8px / row 198.6px, hint flush with the input
  (both x=1002, helper margin `0px`), and the saving state showing the spinner with Cancel
  and the title X disabled.

### Known gap, not addressed

Apollo sets no mutation timeout, so a genuinely hung request still leaves no way out of
the dialog — the spinner makes it legible, not escapable. That needs a timeout or a
bail-out affordance, which is a behaviour change beyond this ticket. Happy to raise it
separately.

### CI

The two `Playwright E2E` failures (`journeys`, `resources`) seen earlier were flake — both
now pass on the unchanged commit `fe715bb8d`, and `journeys-e2e` had already passed on that
same SHA in a separate run. Worth noting `apps/journeys-e2e/src/e2e/journeys.spec.ts` has a
150s `toBeVisible` wait nested inside a 120s `test.setTimeout`, which makes failure
arithmetically certain on a slow preview cold-start — a latent flake source for unrelated
PRs.

jaco-brink:
Sanity-checked commit 4e271a4e8 against all three review threads, and re-checked the branch head (the merge commit after it dropped nothing). Every change matches its described fix:

1. **Escape/backdrop test** — new test `should not close on escape or backdrop click while fonts are saving` is in the spec, matching the suggested code; the old test was retitled to `should disable the close affordances while fonts are saving` with the tautological Cancel-click assertions removed. The `if (saving) return` guard is untouched, as described.
2. **`justifyContent="flex-end"` no-op** — prop and its assertion are gone; replaced by the theme-independent gap check and `expect(actions).not.toHaveStyle({ width: '100%' })`, which now guards the actual bug. The gap-vs-margin comment is on the Stack.
3. **Spinner** — `loading={saving}` added to Confirm alongside `disabled={saving}`, with the `progressbar` assertion in the unit test. Deferring the Apollo-timeout point to its own ticket is the right scope call.

CI is fully green (build, lint, all 3 test shards, all 5 Playwright E2E suites). ✅ Ready to merge.

jaco-brink (approved):
Good work Sharon 🚀

PR #9398, opened 2026-07-21
fix(journeys): use keyed pageview uniques for card visitor counts

csiyang:
## Problem

Card view counts on the journey map run 30–50% above the button clicks that lead to them. Since #9370, the trailing-slash merge **sums** unique visitors across the two Plausible pages per step (`.../stepId` and `.../stepId/`) — but on historical data those buckets overlap, so the sum double-counts.

## Root cause

Pre-#9370, the pageview URL appended the query string with a leading slash, so any visitor arriving with UTM/query params had their **pageviews** recorded on the trailing-slash pathname — while every **action event** (buttonClick, video events, the capture event fired on card view) always used the slash-free pathname. One visitor, one card visit, events on both "pages".

Verified against the Plausible CE **v2.1.1** source (the version we deploy): the `event:page` breakdown computes `visitors` as `uniq(user_id)` over **all** events on a pathname — custom events included, no `event:name` restriction (`lib/plausible/stats/breakdown.ex` → `base_event_query`, `lib/plausible/stats/base.ex`). So a query-string visitor who acted on a card is counted once per variant — twice per sum. Click counts (keyed by `event:props:simpleKey`) are pathname-independent and were always correct, which is exactly why views inflate relative to clicks.

## Fix

Derive each step's visitor count from the **pageview rows keyed by `event:props:simpleKey`** instead: the key is pathname-independent, so Plausible deduplicates visitors across the historical split — and it's the same dimension family as the click counts the cards are compared against. `event:page` rows remain the source for `timeOnPage` (Plausible only exposes it there) and the visitor fallback when keyed pageview data is absent (older journeys).

## Testing

- Updated `transformJourneyAnalytics.spec.ts`: expectation for the keyed-pageview override, plus new tests for the override and the no-key fallback.
- `npx vitest run --config libs/journeys/ui/vitest.config.mts` on `useJourneyAnalyticsQuery`: 18/18 pass.

https://claude.ai/code/session_01EPtWTi79cUJb67u6wcrTYh

---
_Generated by [Claude Code](https://claude.ai/code/session_01EPtWTi79cUJb67u6wcrTYh)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved journey analytics visitor counts by prioritizing deduplicated pageview data.
  * Prevented double-counting visitors across trailing-slash URL variants.
  * Preserved accurate, visitor-weighted time-on-page calculations.
  * Added fallback handling when keyed pageview data is unavailable.


edmonday (approved):
## Review summary

**Approved.** Clean, well-scoped fix for the journey card view-count over-count.

- Root cause verified against the deployed Plausible CE v2.1.1 source — sourcing per-step visitors from pathname-independent `event:props:simpleKey` pageview uniques correctly sidesteps the trailing-slash double-count.
- Presence-based override (zero-row authoritative) and the exit-count cap on both override and fallback paths are subtle and correctly tested.
- Thorough test matrix; `any` removed via typed Maps; solution doc updated.

Non-blocking notes: during reporting windows that straddle the key rollout, per-step visitors deliberately under-count (keyed era only) while summed exits span all history, so such steps can pin to a 100% exit rate — documented and accepted. Minor: new spec descriptions don't use the `should` prefix.

PR #9417, opened 2026-07-28
fix(api-media): resolve mediaUserRoles from auth context

### LINKED ISSUE #9416: api-media: UserMediaRole looked up by two different keys - mediaUserRoles and auth-scope publisher checks disagree
# UserMediaRole is looked up by two different keys — publisher checks can disagree between `me.mediaUserRoles` and runtime auth scopes

## Summary

api-media resolves media roles for the same user through **two different lookups keyed on different identifiers**:

1. **Auth-scope context** (gates fields like `VideoVariant.asset`): keyed by **Firebase uid** —
   [`apis/api-media/src/yoga.ts`](https://github.com/JesusFilm/core/blob/0c2153558c199f9a7c1ee929bbb7e847ca35cfe8/apis/api-media/src/yoga.ts#L43-L48)

   ```ts
   currentRoles:
     (await prisma.userMediaRole.findUnique({
       where: { userId: user.id } // user.id = Firebase uid from the JWT
     }))?.roles ?? []
   ```

2. **`AuthenticatedUser.mediaUserRoles` federation resolver** (what clients — including the videos-admin proxy gate — actually query): keyed by the **UserMediaRole primary key**, passing in the **api-users `User.id` uuid** —
   [`apis/api-media/src/schema/user/user.ts`](https://github.com/JesusFilm/core/blob/0c2153558c199f9a7c1ee929bbb7e847ca35cfe8/apis/api-media/src/schema/user/user.ts#L17-L29)

   ```ts
   mediaUserRoles: t.field({
     resolve: async (data) =>
       (await prisma.userMediaRole.findUnique({
         where: { id: data.id } // data.id = api-users User.id (federation key)
       }))?.roles ?? []
   })
   ```

`UserMediaRole` has `id String @id @default(uuid())` and `userId String @unique` ([`libs/prisma/media/db/schema.prisma`](https://github.com/JesusFilm/core/blob/0c2153558c199f9a7c1ee929bbb7e847ca35cfe8/libs/prisma/media/db/schema.prisma#L586-L592)). Its `id` is its **own** uuid, and its `userId` is the **Firebase uid** — neither column is declared to be the api-users `User.id`. Lookup (2) only succeeds when a row happens to have been seeded with `id` set to the api-users `User.id` (there is no mutation in the schema that creates these rows, so seeding convention is all that holds this together).

Because the two lookups use different keys, a row can satisfy one and not the other, and the two role checks then **disagree for the same user**.

## Observed impact

Operator account (granted `publisher` + `youtubeAdmin` per our setup checklist):

- **videos-admin sign-in works** — its proxy gates on `me { ... on AuthenticatedUser { mediaUserRoles } }` ([`apps/videos-admin/src/proxy.ts`](https://github.com/JesusFilm/core/blob/0c2153558c199f9a7c1ee929bbb7e847ca35cfe8/apps/videos-admin/src/proxy.ts)), i.e. lookup (2) finds `publisher`.
- **`VideoVariant.asset` (master video file) is denied on every variant** for the same account when queried through the gateway — that field is gated `withAuth({ isPublisher: true })` ([`apis/api-media/src/schema/videoVariant/videoVariant.ts`](https://github.com/JesusFilm/core/blob/0c2153558c199f9a7c1ee929bbb7e847ca35cfe8/apis/api-media/src/schema/videoVariant/videoVariant.ts#L330-L332)), i.e. lookup (1) does **not** find `publisher` for the Firebase uid on the JWT.

Net effect: the account looks like a publisher in videos-admin but cannot access any variant masters via the API. We hit this from Wheat (JesusFilm/youtube-studio), which downloads masters via `video → variants → asset.publicUrl`; every variant degrades to the downloads-tier fallback or "no master".

Clients that display *why* a master is hidden also can't trust `me.mediaUserRoles`, since it can report a different answer than the scope that actually gates the field.

## Suggested direction

Make both checks resolve roles from the same key. Options, in rough preference order:

1. Fix the `mediaUserRoles` resolver to key on the same identity the auth context uses. The resolver only receives the federation `id` (api-users `User.id`), so this likely means resolving via context for `me`, or adding/using a properly-declared api-users-id column on `UserMediaRole` and querying it explicitly.
2. At minimum: document the seeding invariant (`UserMediaRole.id` **must** equal api-users `User.id`, `userId` **must** equal the Firebase uid) and add a guarded grant path that enforces it, so a manually-granted row can't half-work.

Also worth an audit query: any `UserMediaRole` row whose `id` doesn't correspond to an api-users `User.id`, or whose `userId` doesn't correspond to that same user's Firebase uid, is currently a half-granted account.

## Environment

- Found by code inspection at `0c21535` plus live behavior from Wheat (JesusFilm/youtube-studio) against the production gateway, 2026-07-28.

mikeallisonJS:
Closes #9416

## Problem

api-media resolved the same user's media roles through **two lookups keyed on different identifiers**, so they could disagree:

- **Auth-scope context** (gates `VideoVariant.asset` via `withAuth({ isPublisher: true })`) reads `context.currentRoles`, computed in `yoga.ts` from `UserMediaRole.userId` = the **Firebase uid** (`context.user.id`).
- **`AuthenticatedUser.mediaUserRoles`** (what the videos-admin proxy gate queries) looked up `UserMediaRole` by `id: data.id` = the **api-users `User.id`** federation key — a different column that only matched by seeding coincidence.

Net effect reported in the issue: an operator could sign into videos-admin as a publisher (lookup via `mediaUserRoles` succeeded) yet be denied `VideoVariant.asset` on every variant through the gateway (the scope check disagreed), breaking master-file downloads from Wheat.

## Fix

Resolve `mediaUserRoles` from `context.currentRoles` — the exact same value the `isPublisher` scope check reads — so the two can never diverge (issue's option 1, "resolving via context for `me`").

```ts
resolve: (_data, _args, context) =>
  context.type === 'authenticated' ? context.currentRoles : []
```

The field is now caller-scoped by design: it returns the authenticated caller's own roles and no longer keys on the requested federation `id`. This cannot leak another user's roles, and `mediaUserRoles` is only ever queried for `me` in practice. Unauthenticated contexts return `[]`.

The GraphQL schema surface is unchanged.

## Tests

`user.spec.ts` now asserts roles come from the context lookup (`{ where: { userId: 'testUserId' } }`), explicitly asserts the old `{ where: { id: 'id' } }` lookup is **not** made, and covers the null/empty-roles path.

## Verification

- `nx run api-media:type-check` — pass
- `nx run api-media:lint` — pass
- `nx run api-media:generate-graphql` — schema diff empty (no surface change)
- `nx run api-media:test` — user spec green (one unrelated pre-existing `video.spec.ts` failure exists on `main`)

## Follow-ups (out of scope for this issue)

- `apis/api-languages/src/schema/user/user.ts` has the **identical** latent bug in `languageUserRoles` (keys on `id: data.id`). The videos-admin proxy relies on that field too — worth a companion fix.
- The issue's optional audit query for half-granted `UserMediaRole` rows (rows whose `id`/`userId` don't line up with the api-users user).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authenticated user media role resolution using the current request context.
  * Prevented incorrect role lookups based on user federation IDs.
  * Ensured users without media roles receive an empty roles list instead of an error.

### DIFF
--- a/apis/api-media/src/schema/user/user.spec.ts
+++ b/apis/api-media/src/schema/user/user.spec.ts
@@ -11,6 +11,8 @@ describe('user', () => {
     }
   })
 
+  const publicClient = getClient()
+
   const VIDEO_ROLES = graphql(`
     query VideoRoles {
       _entities(
@@ -54,6 +56,16 @@ describe('user', () => {
     const data = await authClient({
       document: VIDEO_ROLES
     })
+    expect(data).toHaveProperty('data._entities[0].mediaUserRoles', [])
+  })
+
+  it('should return no media roles for an unauthenticated request', async () => {
+    // Exercises the context.type !== 'authenticated' fallback: no auth context
+    // means no currentRoles, so the resolver returns [] without a DB lookup.
+    const data = await publicClient({
+      document: VIDEO_ROLES
+    })
+    expect(prismaMock.userMediaRole.findUnique).not.toHaveBeenCalled()
     expect(data).toHaveProperty('data._entities[0].mediaUserRoles', [])
   })
 })
--- a/apis/api-media/src/schema/user/user.spec.ts
+++ b/apis/api-media/src/schema/user/user.spec.ts
@@ -24,10 +24,10 @@ describe('user', () => {
     }
   `)
 
-  it('should return media roles', async () => {
+  it('should return media roles from the request context', async () => {
     prismaMock.userMediaRole.findUnique.mockResolvedValue({
-      id: 'id',
-      userId: 'userId',
+      id: 'mediaRoleId',
+      userId: 'testUserId',
       roles: [MediaRole.publisher],
       createdAt: new Date(),
       updatedAt: new Date()
@@ -35,11 +35,25 @@ describe('user', () => {
     const data = await authClient({
       document: VIDEO_ROLES
     })
+    // Roles are resolved from the auth context (keyed by Firebase uid via
+    // UserMediaRole.userId), the same source that gates publisher fields —
+    // never from a separate lookup on the federation id. See issue #9416.
     expect(prismaMock.userMediaRole.findUnique).toHaveBeenCalledWith({
+      where: { userId: 'testUserId' }
+    })
+    expect(prismaMock.userMediaRole.findUnique).not.toHaveBeenCalledWith({
       where: { id: 'id' }
     })
     expect(data).toHaveProperty('data._entities[0].mediaUserRoles', [
       MediaRole.publisher
     ])
+  })
+
+  it('should return no media roles when the user has none granted', async () => {
+    prismaMock.userMediaRole.findUnique.mockResolvedValue(null)
+    const data = await authClient({
+      document: VIDEO_ROLES
+    })
+    expect(data).toHaveProperty('data._entities[0].mediaUserRoles', [])
   })
 })
--- a/apis/api-media/src/schema/user/user.ts
+++ b/apis/api-media/src/schema/user/user.ts
@@ -1,5 +1,3 @@
-import { prisma } from '@core/prisma/media/client'
-
 import { builder } from '../builder'
 
 import { MediaRole } from './enums/mediaRole'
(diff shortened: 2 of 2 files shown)


mikeallisonJS:
## Review feedback addressed (0fe6edc93d)

**Fixed:**
- `apis/api-media/src/schema/user/user.spec.ts` — added a public (no-auth) request test asserting `mediaUserRoles` resolves to `[]` **without a DB lookup**, exercising the `context.type !== 'authenticated'` fallback branch of the resolver (CodeRabbit quick-win). Thread resolved.

All three `user.spec.ts` cases pass (`nx run api-media:lint` and the spec are green).

tanflem (approved):

PR #9425, opened 2026-07-31
feat(api-media): expose subtitle checksum manifest

lumberman:
## Summary

Core can now prove whether its authoritative subtitle metadata matches a downstream copy without sending every subtitle row on every audit. An interop-only manifest returns a versioned content snapshot, a root checksum, and per-video counts and checksums; a consumer requests row details only for buckets that differ.

The same database statement supplies the manifest, requested details, and all returned checksums. A guarded detail request rejects a stale discovery token with `SUBTITLE_SNAPSHOT_MISMATCH`, so a consumer cannot silently pair rows with a different source state.

## Contract and decisions

- Canonicalization version 1 hashes fixed-position JSON tuples as UTF-8 with SHA-256. It preserves nulls and stored Unicode, sorts IDs by UTF-8 bytes, and includes VTT/SRT version counters so a same-URL source revision changes the checksum.
- The root checksum commits to the total subtitle count and ordered per-video buckets. Requested zero-row videos receive explicit empty detail buckets, which makes delete-to-zero repair unambiguous.
- The query uses existing interop authorization, caps detail requests at 100 unique video IDs before the database read, and bypasses response caching. Existing public subtitle queries remain unchanged.
- The contract covers repair metadata and source version counters. It does not prove object-byte equality and does not retain historical database snapshots.

Session-settled decisions carried from planning: compute from live Core data (user-directed, over a Forge-only checksum); use hierarchical root and bucket checksums (user-directed, over one whole-table checksum); guard details with a content-addressed snapshot (user-directed, over independent reads); and version golden canonical bytes (user-directed, over implementation-defined hashing).

## Forge follow-up required

This endpoint is source-side infrastructure only. Forge must complete a separate PR before its health dashboard treats checksum reconciliation as adopted:

1. Send the existing Core interop token.
2. Implement canonicalization version 1 independently from the literal golden vectors.
3. Compare the root first, then request and repair only differing video buckets in batches of at most 100.
4. Stop on an unsupported version, retry boundedly on `SUBTITLE_SNAPSHOT_MISMATCH`, and fetch a final manifest after repair before reporting healthy.

## Validation

| Check | Result |
| --- | --- |
| Focused subtitle suites | 38 tests passed across 3 files |
| TypeScript | `api-media:type-check` passed with TypeScript 7 |
| Lint | `api-media:lint` passed with 0 errors; 15 existing warnings |
| Schemas | api-media generation and gateway composition passed |
| Client operation codegen | All 5 Apollo operation checks completed through equivalent Windows-compatible commands; no client operation references the new field |
| Full review | Correctness, standards, testing, maintainability, security, performance, API contract, reliability, and adversarial passes found no actionable issues |

Representative manifest-only build on the PR machine:

| Rows | Video buckets | Serialized GraphQL JSON | Pure build |
| ---: | ---: | ---: | ---: |
| 12,000 | 1,200 | 145,491 bytes | 17.41 ms |

The benchmark stays below the planned 512 KiB and 1 second limits. Wall-clock duration is recorded here instead of asserted in a flaky unit test.

- Search api-media traces and logs for `videoSubtitleChecksumManifest`, `SUBTITLE_SNAPSHOT_MISMATCH`, `BAD_USER_INPUT`, Prisma failures, and GraphQL authorization failures.
- Watch resolver p95/p99 latency, database query duration and pool utilization, GraphQL error rate, and response size after Core deployment and during the first three Forge nightly audits.
- Healthy signals: only interop callers succeed; manifest counts are stable absent editorial changes; normal responses remain within the measured size/latency envelope; snapshot mismatches are brief and resolve after a bounded retry.
- Failure signals: sustained resolver latency, pool saturation, repeated mismatch loops, unexpected authorization failures, or 5xx responses. Mitigate by keeping Forge checksum adoption disabled or pausing its scheduled audit, then revert this PR if Core load or correctness is affected.
- Owners: Core api-media for endpoint health; Forge synchronization owner for audit and repair behavior.

### Hierarchical content checksums with guarded reads

A root checksum answers whether anything differs. Per-video checksums identify where it differs. This avoids downloading the full source table when both systems already agree.

```mermaid
flowchart LR
  A["Fetch root and video buckets"] --> B{"Root matches?"}
  B -->|Yes| C["Audit complete"]
  B -->|No| D["Find differing video buckets"]
  D --> E["Request details with expected snapshot"]
  E --> F["Repair selected rows"]
  F --> G["Fetch final manifest"]
```

The expected snapshot is a content guard, not a historical handle: Core recomputes the current content and rejects the detail response if it changed since discovery. This pattern is useful for bounded reconciliation of authoritative data, but it is not appropriate when consumers need historical reads or proof of the bytes behind external URLs.

---

## Summary by CodeRabbit

* **New Features**
  * Added GraphQL support for retrieving video subtitle checksum manifests.
  * Manifests include versioning, snapshots, total counts, root checksums, per-video checksums, and optional subtitle details.
  * Added pagination and count queries for video subtitles.
  * Added snapshot validation to detect stale synchronization data.
  * Added limits and validation for requested video detail results.

* **Documentation**
  * Documented subtitle query behavior and checksum manifest capabilities.


tataihono:
Closing this one because the subtitle repair no longer needs a Core-side manifest/checksum endpoint.

We found Core already exposes the needed flat `videoSubtitles(where, offset, limit)` and `videoSubtitlesCount(where)` APIs in production. The Forge fix now keeps the repair one-sided: incremental flat subtitle sync, verified full Core ID inventory, full row-payload repair when Admin is missing Core IDs, and Core-only safe deletes.

So this PR is superseded by the Forge-only approach.

PR #9430, opened 2026-08-01
fix(journeys-admin): resolve unpublished variants in the video block picker

tanflem:
Related to #9464 (allow unpublished videos in journeys-admin) — makes the video block picker fetch the unpublished variant instead of failing to fetch it.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/VideoLibrary/VideoFromLocal/LocalDetails/LocalDetails.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/Settings/Drawer/VideoLibrary/VideoFromLocal/LocalDetails/LocalDetails.spec.tsx
@@ -1,5 +1,6 @@
 import { MockedProvider } from '@apollo/client/testing'
 import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { print } from 'graphql'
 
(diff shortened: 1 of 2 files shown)


tanflem:
**Why this is safe:** these values are already leaking into the builder, so this isn't exposing anything new — it's making resolution consistent with what the picker can already show.

The picker searches the **video-variants** Algolia index, whose records carry two distinct flags (`algoliaVideoVariantUpdate.ts:93-94`): `published` = the variant's own state, `videoPublished` = the parent video's.

The two entry points into the builder gate these differently:

- `pages/journeys/[journeyId].tsx:94` filters `published:true AND videoPublished:true`
- `pages/publisher/[journeyId].tsx:66` filters only `label:episode OR label:featureFilm OR label:segment OR label:shortFilm` — **no published gate at all**

So an unpublished variant already reaches the builder through the publisher route and can be selected there. `GET_VIDEO` in `LocalDetails.tsx` then applied its own published filter and returned a null variant, so a video the user could see and click failed to resolve.

`GET_VIDEO` is the only `variant(languageId:)` query in journeys-admin and there were no other `onlyPublished` usages in the app, so this changes one lookup and creates no inconsistency with a sibling call site. It does not widen what is reachable — it stops the detail lookup from dropping something the builder had already surfaced.

Kneesal (approved):

PR #9431, opened 2026-08-03
fix(journeys-admin): tighten video description spacing and align toggle (NES-1853)

sharonemathew:
## Issue

[NES-1853](https://linear.app/jesus-film-project/issue/NES-1853/video-details-description-block-loose-paragraph-spacing-and-misaligned) — the Video Details description block has loose paragraph spacing and a misaligned **Less** toggle.

## Fix

All in `VideoDescription.tsx`:

- **Split the description into real paragraph elements.** `component="p"` makes the block-level element the theme's `18px` caption leading needs — as an inline `span` it lost to the parent's strut and rendered at 24px. The paragraph gap becomes a `12px` margin instead of the full blank line `pre-line` produced.
- **Moved the toggle out of the text flow** to a flush-left sibling below it, so **More** and **Less** share the left edge with the description and the language chip. Lets the gradient background and absolute positioning be deleted.
- **Replaced `height: '70px'` with a 3-line clamp**, so the collapsed state cuts on a line boundary rather than mid-glyph.
- **Added `overflowWrap: 'anywhere'`** so long URLs wrap instead of pushing past the drawer.

Measured against the real JESUS description at the real 328px drawer width, identical in Chromium, Firefox and WebKit:

| | Before | After |
|---|---|---|
| Line spacing in a paragraph | 24px | **18px** |
| Gap between paragraphs | 48px | **30px** |
| Toggle indent | 74px expanded / 233px collapsed | **0px** |

## Test plan

- [x] 13 unit tests; full `VideoLibrary` suite 16 files, 150 passed, 1 skipped
- [x] Mutation-tested — the suite fails when the paragraph gap, the clamp, `-webkit-box-orient`, the line count, the toggle alignment, `component="p"` or `overflowWrap` is reverted
- [x] `tsc --noEmit` clean; `nx run journeys-admin:lint` 0 errors
- [x] Rendered against real gateway data in Chromium, Firefox and WebKit — console clean
- [ ] Visual check in the running editor — open **JESUS**, and **Magdalena** for the `&#13;` entity path

## Summary by CodeRabbit

- **Improvements**
  - Video descriptions now preserve paragraph spacing and line breaks more reliably.
  - Long descriptions collapse cleanly to three lines with improved URL wrapping.
  - Empty paragraphs and carriage returns are handled consistently.
  - The “Show more” control is better aligned with the description content.
  - Empty descriptions now display cleanly without unnecessary spacing or controls.


sharonemathew:
## Review notes

Measured, not estimated. Kept out of the description for brevity.

**Accepted, please don't file:**

- The clamp's ellipsis renders as `....` on ~20% of multi-paragraph descriptions (21/107 sampled) when the cut lands on a paragraph end. Both alternatives are worse — `max-height: 3lh` cuts mid-glyph, `text-overflow: clip` doesn't suppress it.
- Collapsed height varies 54–78px depending on where paragraph breaks fall. Always 3 lines.

**Coverage gaps** — no e2e here and Chromatic is off (`visual-test.yml` → `branches: [never]`), so the unit spec is the only net:

- `overflowWrap` is asserted on the first paragraph only; later paragraphs are a separate emotion class.
- `alignSelf`, `mt: '-4px'`, `whiteSpace: 'pre-line'` and `variant="caption"` are each removable with all 13 tests green.
- jsdom does no layout — every assertion pins a declaration, never a rendered box.

**Follow-ups:**

- **Dead "More"** — `showToggle` counts raw characters, the clamp counts rendered lines. ~140–179 chars, or any whitespace-padded description, shows a toggle that reveals nothing (measured: 81px → 81px, where `main` revealed 777px). One line to fix.
- **Sibling titles have the same URL exposure** — a long title renders 747px in a 278px column, overflowing the panel by 445px. The `overflowWrap` this PR adds is the only occurrence in the whole VideoLibrary tree.
- **No focus indicator on the toggle** — WCAG 2.4.7 AA. MUI styles `Mui-focusVisible` only for `contained` and no shared theme defines a focus rule, so the focused button is pixel-identical to the unfocused one. Pre-existing and app-wide; better fixed once in the theme.
- **No `dir="auto"`** — RTL descriptions render LTR with displaced punctuation. Reachable via YouTube descriptions.
- **CRLF bypasses the paragraph split** — `\r\n\r\n` keeps the old loose rendering. 0 occurrences in 400 gateway descriptions, but only English was reachable and Crowdin stores translations unnormalised.

jianwei1 (approved):

PR #9415, opened 2026-07-28
chore(deps): upgrade Nx toolchain 22→23 with minimal Vitest 3 bump (#9376)

mikeallisonJS:
## What

Implements **one step of the #9376 major-upgrade roadmap: the Nx toolchain family** (`nx` + all 16 `@nx/*`, `22.7.1 → 23.1.0`). Issue #9376's dead-dep cleanup already shipped in #9377; its roadmap left each big upgrade as an independent scoped effort — this is the Nx one.

`@nx/vitest@23` peers `vitest ≥3` (used by 21 projects), so Vitest is unavoidably pulled along. Per the agreed **"Nx + minimal Vitest"** scope, `vitest` + `@vitest/*` move to the lowest accepted 3.x (`3.2.7`) and **`vite` stays 7**.

## Scope control

`nx migrate` also opportunistically bumped TypeScript 6, typescript-eslint, storybook, vite 8, and webpack — **none required by Nx 23 peers, all reverted** to keep this scoped to the Nx family:

| Package | Kept at |
|---|---|
| `typescript` | `^5.5.4` (bump excluded — TS7 pilot is separate) |
| `typescript-eslint` / `@typescript-eslint/*` | `^8.26.0` |
| `vite` | `7.1.9` |
| `storybook` / `@storybook/*` | `^10.3.5` |
| `webpack` / `webpack-dev-server` | `5.104.1` / `5.2.2` |

`typescript@6.0.3` appears in the lockfile **only** as a private nested dep of `@nx/eslint@23` — every workspace project still resolves `typescript@5.9.3`.

Ran the Nx 23 codemods (all no-ops — no deep `@nx/*` internal imports, no `createNodesV2` usage); skipped the TS6/vite8/vitest4 migrations as out of scope.

## Three structural fixes the upgrade forced

[...]

Issue #9376, opened 2026-07-15
Prune dead dependencies (abandoned / deprecated / unimported) + major-upgrade roadmap


mikeallisonJS:
## Problem Statement

As a developer working in the `core` monorepo, I keep tripping over dependencies that no longer earn their place. Some are formally abandoned by their maintainers, some are deprecated releases, and some are declared in `package.json` but imported nowhere in our source. They bloat the install graph, produce deprecation warnings on every `pnpm install`, mislead anyone auditing the tree (a listed package looks "used"), and quietly widen our supply-chain surface. Separately, a large share of our dependencies are one or more **major** versions behind, and without a shared, sequenced view of which upgrades matter and in what order, these upgrades never get scheduled — they just accumulate risk until something forces a painful, all-at-once jump.

A full-tree audit (`pnpm outdated` across the workspace, cross-checked against actual source imports and the lockfile) found **271 outdated packages** — 109 major-behind, 113 minor, 48 patch — plus a small set of genuinely dead declarations. React 19 / Next 16 / TypeScript 5.9 are current, so the problem is concentrated in tooling, UI libraries, the AI SDK, backend SDKs, and abandoned leftovers.

## Solution

Two parts, one spec:

1. **Actionable now (this ticket's implementation):** Remove the dependencies that are genuinely dead — abandoned upstream, deprecated, or declared-but-unimported — so the install graph shrinks, deprecation noise disappears, and the manifest tells the truth about what we actually use. This is a zero-behavior-change cleanup: nothing in the running product should change, only the dependency manifest and lockfile.

2. **Documented roadmap (context for future tickets):** Capture a sequenced major-version upgrade roadmap so the largest, highest-risk upgrades (Nx, MUI, Vitest, the AI SDK, backend SDKs) are visible, ordered, and can each be picked up as its own scoped effort. The roadmap is reference material in this spec; it is explicitly **not** implemented here.

The developer's experience afterward: a clean `pnpm install` with no deprecation warnings for the removed packages, a `package.json` where every listed dependency is genuinely referenced, and a clear, agreed order in which the big upgrades should be tackled.

## User Stories

1. As a developer, I want abandoned packages removed from the manifest, so that I stop seeing "no longer maintained" deprecation warnings on every install.
2. As a developer, I want `subscriptions-transport-ws` removed, so that the abandoned Apollo v3 optional-peer package (which we import nowhere and have superseded with `graphql-ws`) stops appearing in our tree.
[...]

PR #9436, opened 2026-08-03
chore(deps): upgrade AI SDK 6→7 and providers 3→4 (#9376)

mikeallisonJS:
## What

The **AI SDK slice** of the #9376 major-upgrade roadmap.

> **AI SDK (lockstep):** `ai` (6 → 7), all `@ai-sdk/*` (3 → 4), `@openrouter/ai-sdk-provider`. Breaking API changes across the set.

Siblings already shipped: dd-trace/browser-rum (#9428), firebase client (#9426), firebase-admin (#9424), express (#9423), commitlint (#9418), dead-dep prune (#9377). Nx (#9415) is open.

| Package | From | To |
|---|---|---|
| `ai` | `^6.0.116` | `^7.0.49` |
| `@ai-sdk/google` | `^3.0.43` | `^4.0.31` |
| `@ai-sdk/openai` | `^3.0.53` | `^4.0.27` |
| `@ai-sdk/react` | `^3.0.164` | `^4.0.52` |
| `@ai-sdk/openai-compatible` | `^2.0.41` | `^3.0.20` |
| `@openrouter/ai-sdk-provider` | `^2.8.0` | `^3.0.0` |

Every package the roadmap names for this family, nothing from any other. `@ai-sdk/openai-compatible` is 2→3 and `@openrouter/ai-sdk-provider` 2→3, not the "3 → 4" the roadmap generalised to; `3.0.0` is the first openrouter release peering on `ai@^7`.

## The break the compiler could not see

This is the part that makes the PR more than a version bump.

v7 defaults `allowSystemInMessages` to `false` and throws when `messages` carries a `role: 'system'` entry:

```js
// ai@7.0.49/dist/index.js:2550
if (!allowSystemInMessages && messages.some((message) => message.role === "system")) {
  throw new InvalidPromptError2({
    prompt,
    message: "System messages are not allowed in the prompt or messages fields. Use the instructions option instead."
  });
}
```

`system` survives as a deprecated alias for `instructions`, so **the signature still type-checks**. I ran type-check across all four affected projects on the bumped packages with zero source changes: green. The full unit suite: green — because every existing spec `vi.mock`s the whole `ai` module, so a mocked `generateText` never validates anything.

Five call sites would have thrown in production on the first real request:

- `journeyAiTranslate.ts` (streamText)
- `translateJourneyMetadata.ts` ×2
- `translateCustomizationFields.ts` ×2

Each moved to `instructions`. The system prompt reaches the model identically — it is the same string in the same provider-level `system` slot, just routed through the option v7 requires.

### Tests that actually exercise the SDK

The existing mock-everything specs cannot catch this class of break, so the two new contract specs run the **real** `generateText` against a `MockLanguageModelV4` and assert on `doGenerateCalls[].prompt` — what the provider was really handed:

- `translateJourneyMetadata.contract.spec.ts`
- `translateCustomizationFields.contract.spec.ts`

They fail red on the pre-fix code with the exact `AI_InvalidPromptError` above, and they assert the outcome that matters (the system prompt reaches the model) rather than the option name, so they survive the next rename.

The `MockLanguageModelV4` stub lives in `@core/shared/ai/testModel`, deliberately **not** exported from `libs/shared/ai/src/index.ts` — it imports `ai/test` and has no business in a production bundle. It is one place because the provider-level result shape is itself versioned and changed in this very upgrade:

| Provider spec | v3 | v4 |
|---|---|---|
| `finishReason` | `'stop'` | `{ unified: 'stop', raw }` |
| `usage.inputTokens` | `number` | `{ total, noCache, cacheRead, cacheWrite }` |
| `usage.outputTokens` | `number` | `{ total, text, reasoning }` |

Core-level `FinishReason` is still the plain string union and core-level `usage.inputTokens` is still a number, so `apps/journeys`' `finishReason === 'error'` check and its Langfuse token reporting are untouched. Only code constructing a fake **provider** is affected, which is why this is confined to test helpers.

### `/api/chat` no longer accepts a client-supplied system message

`apps/journeys/pages/api/chat/index.ts` narrows the request schema's role enum from `['user','assistant','system']` to `['user','assistant']`.

Under v7 a client-sent system message flows through `convertToModelMessages` and then throws inside `streamText`, landing in the handler's outer `catch` as a **500 `upstream streamText failed`** — an upstream-fault code for what is a malformed request. Rejecting it up front returns the accurate `400 invalid_request`, which the client already treats as non-retriable via its existing error-code contract.

`allowSystemInMessages: true` was the strictly behaviour-preserving alternative. I chose rejection because the old path also let a client inject its own system instructions alongside ours. **This is an observable API contract change** — a request that previously reached the model now 400s. Our own `AiChat` client never sends system roles, so no first-party caller is affected.

### `getImageDescription` moves off the deprecated image part

`{ type: 'image', image: imageUrl }` → `{ type: 'file', mediaType: 'image', data: imageUrl }`.

v7 still accepts the image part but emits a runtime `DeprecationWarning` on every call. The two are not merely similar — a `file` part with string `data` runs the **identical** `convertToLanguageModelV4FilePart` path as the deprecated image part (`new URL(content)` first, inline-data fallback), so URL-vs-base64 resolution is unchanged. The new `getImageDescription.spec.ts` pins that: it asserts the URL reaches the model as `{ type: 'url', url }` rather than being inlined as base64, which is the one way this change could have gone silently wrong.

That spec also surfaced a detail worth knowing: with no `supportedUrls` declared, the SDK **downloads the image itself** rather than passing the link through. Long-standing behaviour, not a v7 change, but it is why the stub declares `supportedUrls` — otherwise the test hits the network.

## Lockfile scope

The `importers:` section changes on **exactly six specifiers** — the six above, nothing else. A plain `pnpm install` afterwards leaves the lockfile clean.

The bump **removes a duplicate**. Base had four parallel `ai` majors; this collapses the two v6 copies into one v7:

| | base | after |
|---|---|---|
| `ai` | 5.0.118, 5.0.154, 6.0.116, 6.0.168 | 5.0.118, 5.0.154, **7.0.49** |
| `@ai-sdk/gateway` | 2.0.24, 2.0.58, 3.0.66, 3.0.104 | 2.0.24, 2.0.58, **4.0.38** |
| `@ai-sdk/provider-utils` | 3.0.20, 3.0.22, 4.0.19, 4.0.23 | 3.0.20, 3.0.22, **5.0.18** |

The surviving `ai@5` / `@ai-sdk/*@2-3` copies are **not ours** — they come from Docusaurus's `@docsearch/react@4.2.0`, which bundles `@ai-sdk/react@2` and `ai@5`. Out of scope here.

New transitives pulled in by v7: `@ai-sdk/mcp`, `@workflow/serde`, `eventsource-parser`, `pkce-challenge`, `swr`, `undici@7`. No new peer warnings, no deprecation warnings on install.

v7 is **ESM-only** and requires **Node ≥ 22**. The devcontainer runs Node 24, and both consumers bundle it fine — `api-journeys` (webpack, `target: node`) compiles clean, and `journeys` reports `Compiled successfully`.

## Verification

Type-check and lint green for all four affected projects. Full unit suites:

| Project | Result |
|---|---|
| `api-journeys` | 1079 passed, 1 skipped |
| `journeys-ui` | 1471 passed, 7 skipped |
| `apps/journeys` | 214 passed |
| `shared-ai` | 118 passed |

Two failures are **pre-existing** — both confirmed identical with these changes stashed:

| Failure | Cause |
|---|---|
| `journeys:lint` (4 errors) | generated `dist/apps/journeys/.next/types/routes.d.ts` import + a floating promise |
| `journeys:build` | static export needs the GraphQL gateway on `localhost:4000`; compile step succeeds |

## Not in this PR

- **`@ai-sdk/react` 3→4 is the one part not exercised for real.** `AiChat.spec.tsx` mocks `useChat` wholesale, so only its type signature is verified. Runtime risk is low — `DefaultChatTransport` ships from `ai` and moves in lockstep with the hook — but a genuine streaming test of that hook does not exist, and writing one means hand-building a v7 UI-message SSE stream. Worth its own change.
- **`pipeUIMessageStreamToResponse` and `onFinish` stay on their deprecated aliases.** Both still work in v7 (removal is slated for the next major) and neither warns at runtime — only prompt content parts do, which is why the image part was worth fixing and these are not. Migrating to the stateless `toUIMessageStream` / `createUIMessageStreamResponse` helpers reworks the response path that the client's error-body contract depends on; that is a behavioural change, not a version bump.
- **`libs/shared/ai/vitest.config.mts` and `apis/api-journeys/vitest.config.mts` set neither `test.root` nor `test.include`**, so an unfiltered `npx vitest run --config <path>` from the repo root globs the *entire monorepo* — ~1470 spec files, mostly other projects' failures. Every run here was path-scoped to work around it. `docs/agents/testing.md` documents the root-level invocation as the standard one, so this is a real trap; `libs/journeys/ui` sets `test.root` and does not have it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of AI-powered journey translations and metadata analysis.
  - Improved image description processing, including clearer handling of failed requests.
  - Chat requests now reject unsupported system messages with clear validation errors.
  - Improved handling of prompts and image inputs for more consistent AI results.

- **Improvements**
  - Improved consistency across chat, translation, and image analysis experiences.
  - Added broader automated coverage for dependable AI responses and error handling.


mikeallisonJS:
## Review feedback addressed (7393f2a8c)

**Fixed (adjusted):**
- `apis/api-journeys/test/promptOf.ts` (new) — CodeRabbit was right that the `any` in `promptOf` drops validation of the migrated `messages[0]` shape. Rather than adding a third inline type, note that `translateJourneyMetadata.spec.ts` already carried a correctly-typed copy of this same helper — so the helper and its `MockTextCallOptions` type are now shared from `test/promptOf.ts` (already covered by `tsconfig.spec.json`'s `test/**/*.ts`).
  - `translateCustomizationFields.spec.ts` — dropped the `any` helper; also removed the eight `options: any` mock implementations it fed (now `unknown`).
  - `journeyAiTranslate.spec.ts` — all five `(mock.calls[n][0] as any).messages[0].content[0].text as string` walks replaced with `promptOf(...)`.
  - `translateJourneyMetadata.spec.ts` — local copy removed; the options-based signature also drops its two `as unknown` casts.

**Also fixed (not from a review thread, but blocking CI):**
- `apps/journeys/pages/api/chat/index.ts:439` — AI SDK 7 changed `pipeUIMessageStreamToResponse` from `void` to `Promise<void>`, so this upgrade left a floating promise. That was the single `@typescript-eslint/no-floating-promises` error failing `journeys:lint`. Now awaited, which is also the intent of the new signature — the handler stays alive until the response finishes streaming instead of resolving mid-pipe.

**Verified:** `tsc -p apis/api-journeys/tsconfig.spec.json` clean; api-journeys `journeyAiTranslate` suite 97/97 passing; `apps/journeys` chat route suite 51/51 passing; eslint + prettier clean on all touched files.

tanflem (approved):

PR #9449, opened 2026-08-05
fix(journeys-admin): use Montserrat in filter and sort dropdowns (NES-1217)

sharonemathew:
## Issue

[NES-1217](https://linear.app/jesus-film-project/issue/NES-1217/ui-bug-incorrect-font-in-dropdown-menus) — the **Sort By** and **Filter by status** menus showed their choices in Open Sans, while everything around them used Montserrat: the button that opens them, the tabs, the journey titles, and the 3-dot menu right beside them in the same toolbar.

The labels hardcoded `typography: 'body2'`, which the admin theme defines as Open Sans. The 3-dot menu was already correct because it goes through the shared `MenuItem` → `ListItemText`, which the theme overrides to Montserrat.

Not a font-loading or portal problem — the Google Fonts request returns both families fine.

## Fix

Menu labels now use Montserrat at Medium, and the 3-dot menu matches, so all three menus in the toolbar agree. Style-only, two files, no logic touched.

See the comment below for the detail — which weight and why, and why the 3-dot menu is changed on the menu rather than in the theme.

## Test plan

Open **Sort By**, **Filter by status** and the **3-dot menu** on each of:

- [ ] `/` — Team projects dashboard
- [ ] `/?type=templates` — Team templates tab
- [ ] `/` — **Shared with me**
- [ ] `/publisher` — Publisher (sort + 3-dot only)

Looking for:

- [ ] All three menus use Montserrat, matching the trigger and tabs — no Open Sans
- [ ] Menu labels are lighter than the trigger above them
- [ ] Selecting an option still applies the sort/filter and closes the menu
- [ ] Narrow viewport — same labels in the bottom drawer
- [ ] Nav drawer and editor toolbar unchanged (still SemiBold)

> [!NOTE]
> **Shared with me isn't in the ticket but is affected** — same components, same bug. The font change there is expected: a fix, not a regression.

**Automated:** 374 tests pass across the changed components and everything downstream (`JourneyList`, `StatusTabPanel`, `TemplateList`, `JourneyListView` incl. `SharedWithMeMode`); `tsc --noEmit` clean. Two pre-existing `LastModifiedDate` failures are unrelated — they fail identically on `origin/main` with these changes stashed.

## Summary by CodeRabbit

* **Style**
  * Updated journey list menu text to use a medium font weight.
  * Refined radio option labels with the intended typography and medium font weight.
  * Preserved existing menu border-radius styling.


sharonemathew:
## The fix in detail

Two style-only changes:

- **`RadioSelect.tsx`** — option labels go from `body2` (Open Sans) to `subtitle3` (Montserrat) at `fontWeight: 500`. Size unchanged at 14px.
- **`JourneyListMenu.tsx`** — the 3-dot menu drops to 500 too, so all three menus in the toolbar match.

### Why 500 and not `subtitle3`'s default 600

It keeps the options lighter than the trigger above them, so an open menu still reads as a list of choices rather than a row of buttons.

It's also the lightest weight available — `_document.tsx` loads `Montserrat:wght@500;600;700;800`, and a request for 400 silently resolves to 500 (CSS picks the nearest available weight rather than another family).

### Why the 3-dot menu is changed on the menu, not the theme

Its weight comes from `MuiListItemText.primary` in the admin theme, which every list in the app shares — nav drawer, editor toolbar, video lists, dialogs. Changing it there would have made this an app-wide typography change. Scoped to this one menu instead.

### Second commit

`1ad6315b8` is comment-only. `typography` expands to all of `subtitle3` including `fontWeight: 600`, and `sx` merges keys in insertion order — so `fontWeight: 500` wins only because it sits below. Reversing the two lines silently restores SemiBold, with no test failure. The comment marks it.

### For review

`subtitle3` plus a `fontWeight` override rather than a pure variant, because no existing variant is Montserrat at Medium/14px. Taking `subtitle3` as-is (600) is tidier if preferred, at the cost of bolder labels. A dedicated theme variant would be cleaner long-term — better as a follow-up than widening this PR.

jianwei1 (approved):

PR #9455, opened 2026-08-09
fix(journeys-admin): guard Make Template against rapid clicks (NES-1782)

sharonemathew:
Closes [NES-1782](https://linear.app/jesus-film-project/issue/NES-1782/make-template-button-allows-multiple-clicks-causing-duplicate)

## Issue

"Make Template" has no in-flight guard, so every click landing before the first one finishes starts its own duplicate → template chain. Rapid clicking creates duplicate templates.

The reported flash under Team Projects has the same cause: the duplicated journey is briefly a real project (`template: false`) until the second mutation flips it, and overlapping chains let a refetch catch it in that state.

## Fix

In `CreateTemplateItem.tsx`:

- `useRef` guard blocks re-entry while a chain is running
- `creating` state disables the menu item
- both reset in `finally`, so a failed mutation can't leave it stuck disabled

The ref is what stops the duplicate — `setState` doesn't apply until the next render, so two clicks in the same tick would both read the stale value. The mutation body is unchanged.

## Test plan

Two tests added. The 5 existing ones were all single-click, which is why this shipped.

- rapid clicks → one duplicate call, one template created
- menu item disables while in flight


sharonemathew:
## Follow-up: guard now survives dismissing the menu

Self-review turned up a remaining path to the reported bug, fixed in 27a1267ac.

### Problem

The in-flight guard lives in `CreateTemplateItem`, so it only holds while that component stays mounted.

In the JourneyCard menu, `keepMounted` starts `false` (`JourneyCardMenu.tsx:142`) and is only set by `ShareItem`, `CopyToTeamMenuItem` and `CopyToCollectionMenuItem` — the items that open dialogs. `CreateTemplateItem` never got it. Since `onClose` is wired to escape and backdrop clicks, dismissing the menu mid-flight unmounted the item and reset the guard. Reopening and clicking again created a second template.

This is the ticket's exact repro path — Team Projects → card context menu → Make Template.

The editor toolbar menu was never affected: `Menu.tsx:82` passes `keepMounted` unconditionally.

### Fix

`CreateTemplateItem` now claims `keepMounted` when creation starts, the same way those three items do. No new mechanism — it uses the one already in the menu.

```
isCreatingRef.current = true
setCreating(true)
handleKeepMounted?.()
```

Plus passing `handleKeepMounted` through from `DefaultMenu` to both instances.

### Verification

92 specs passed / 11 files · ESLint clean · `tsc` 0 errors. Added an assertion that `handleKeepMounted` fires on click.

Worth flagging: the test covers that the callback is called, not that the item genuinely survives a dismiss — that needs a real MUI Modal and a real escape keypress, so it stays a manual check.

### Also in this push

Merged the latest `main` (which now includes NES-1324 / #9451) and re-ran everything against it — still green.

### Still open from the self-review

Not addressed here, listed so they aren't lost:

- No `catch` on the mutation chain — a failure leaves an unhandled rejection and no user feedback
- If `journeyTemplate` fails after `journeyDuplicate` succeeds, an orphaned copy is stranded in Team Projects with no cleanup (pre-existing)
- The disabled item has no spinner or label change, so a multi-second operation reads as "dead" rather than "working"

sharonemathew:
## Follow-up: failures are now reported to the user

Closes the "no error handling" item from the self-review, in 7d6ea95a6.

### Problem

The mutation chain had a `try`/`finally` but no `catch`. `handleCreateTemplate` is `async`, so it returns a promise, and it's wired up as `onClick` — React discards whatever a click handler returns. A failed mutation therefore rejected into nothing: an unhandled rejection in the console, and nothing at all in the UI.

What the user saw on failure: click → item greys out → item un-greys → no template, no message. Indistinguishable from the click not registering.

Success was already loud (`enqueueSnackbar('Template Created')`). Failure was silent. That asymmetry was the defect.

Realistic triggers: network drop mid-chain, expired session, backend 500, team permission lost between page load and click.

### Fix

Added a `catch` before the existing `finally`, using the same pattern as `JourneyListContent.tsx:304`:

```
} catch (error) {
  if (error instanceof Error) {
    enqueueSnackbar(error.message, {
      variant: 'error',
      preventDuplicate: true
    })
  }
} finally {
  ...
}
```

The `finally` already handled recovery — the item re-enables rather than sticking disabled. Nothing exercised that branch until now.

### Verification

New test: `should surface an error and re-enable the menu item when creation fails`. Asserts the error message renders, the item is not left `aria-disabled`, and no navigation happens.

Confirmed it gates the fix — with the `catch` removed it fails on `Unable to find an element with the text: Failed to duplicate journey`.

93 specs passed / 11 files · ESLint clean · `tsc` 0 errors.

### Scope note

This was pre-existing, not introduced by this PR. Included because the ticket is about the button misrepresenting what it's doing, and a silent failure is the same problem in a different direction. Reasonable to disagree — happy to split it out if a reviewer would rather keep this PR to the click guard.

### Still open from the self-review

- If `journeyTemplate` fails after `journeyDuplicate` succeeds, an orphaned copy is stranded in Team Projects with no cleanup (pre-existing; needs a backend or cleanup story, not a client fix)
- The disabled item has no spinner or label change, so a multi-second operation reads as "dead" rather than "working"

csiyang (approved):
Verified the fix end to end:

- The ref-vs-state reasoning holds — `setState` only applies at the next render, so the ref is what actually blocks same-tick re-entry; `creating` drives the disabled affordance. Reset lives in `finally`, so a failed chain can't leave the item stuck disabled.
- The `keepMounted` hole was real: `JourneyCardMenu` starts `keepMounted` false, so dismissing the menu mid-flight would unmount the item and reset the guard. Claiming it on creation start matches how `ShareItem`/`CopyToTeamMenuItem` already do it, and the editor toolbar menu passes `keepMounted` unconditionally, so it was genuinely unaffected.
- The error snackbar matches the established `JourneyListContent`/`ActiveJourneyList` pattern exactly.
- Tests are well-built — the deferred-promise setup means the repeat clicks genuinely land while the first chain is in flight, and the error test asserts both the snackbar and recovery from the disabled state.

One non-blocking naming nit: consider `createInFlightRef` instead of `isCreatingRef`, matching the existing `writeInFlightRef` guard convention — it communicates that the ref is a synchronous re-entry guard rather than a mirror of the `creating` render state. (`creating` itself matches house style — it's the sibling of `loading`/`open`.)

PR #9457, opened 2026-08-09
fix(journeys-ui): wrap long card labels instead of overflowing the card

sushmamalampati:
## Summary
The "Tap to preview" label on the Media customize screen is centered under its card using absolute positioning with `white-space: nowrap` and no width constraint. English text ("Tap to preview", ~85px) fits inside the ~150px selected card, but longer translations overflow it on both sides — confirmed on the live site with Indonesian ("Ketuk untuk melihat pratinjau", ~195px), which overflowed the card's edges by ~22px on each side.

The label now sizes to its card's width (its actual positioned ancestor) and wraps instead of overflowing.

Fixes NES-1794.

## Test plan
- Added a regression test asserting the label has `width: 100%` / `overflowWrap: break-word` and not `white-space: nowrap`; confirmed it fails against the pre-fix code and passes after
- `tsc --noEmit` clean on `libs/journeys/ui`
- Full `TemplateCardPreview`/`TemplateCardPreviewItem` Vitest suite (23/23) passing
- Manually verified on the live site by applying the fix's exact CSS to the actual DOM node: label went from overflowing its 150px card by 22px on each side to fitting exactly within it, wrapping to 2 lines

---

Sonnet 5

## Summary by CodeRabbit

* **Bug Fixes**
  * Long selected card labels now wrap within compact card previews instead of overflowing.
  * Improved readability for labels containing lengthy words or text.


sushmamalampati:
Closing — reverting this fix per author's request; approach needs rethinking.

PR #9458, opened 2026-08-09
fix(journeys-admin): correct bulk action snackbar wording (NES-1780, NES-1874)

sharonemathew:
## Issue

Two related defects in bulk-action confirmation toasts. Both are wording-only; neither changes behaviour.

[NES-1780](https://linear.app/jesus-film-project/issue/NES-1780/templates-admin-bulk-actions-report-journeys-instead-of-templates) — On Templates Admin (`/publisher`), bulk actions confirm with a "... Templates" dialog then report success as "Journeys ...". The two messages contradict each other one click apart.

[NES-1874](https://linear.app/jesus-film-project/issue/NES-1874/bulk-unarchive-all-reports-restored-while-single-card-unarchive) — "Unarchive All" reports "... Restored". The button says Unarchive; the toast says Restored. Single-card Unarchive already said "Unarchived", so the bulk path was the odd one out.

Both have the same root cause: unarchive and restore-from-trash run the same `journeysRestore` mutation, and the bulk components took their wording from the mutation name rather than from the button the user pressed. The single-card components were written against the UI verb instead.

`JourneyListContent.tsx` showed it plainly — two branches, identical result:

```ts
} else if (status === 'archived') {
  messageKey = isTemplate ? 'Templates Restored' : 'Journeys Restored'
} else {
  messageKey = isTemplate ? 'Templates Restored' : 'Journeys Restored'
}
```

They are fixed together because fixing either alone creates a new mismatch — Templates drifting from Journeys, or bulk drifting from single-card.

## Fix

End state for every bulk action. Changed rows marked.

**Journeys (`/`)**

| Tab | Dialog | Snackbar | |
| --- | --- | --- | --- |
| Active | Archive Journeys | Journeys Archived | |
| Archived | Unarchive Journeys | Journeys **Unarchived** | ← |
| Archived | Trash Journeys | Journeys Trashed | |
| Trashed | Restore Journeys | Journeys Restored | |
| Trashed | Delete Journeys Forever | Journeys Deleted | |

**Templates (`/publisher`)**

| Tab | Dialog | Snackbar | |
| --- | --- | --- | --- |
| Active | Archive Templates | Templates Archived | |
| Archived | Unarchive Templates | Templates **Unarchived** | ← |
| Archived | Trash Templates | Templates **Trashed** | ← |
| Trashed | Restore Templates | Templates **Restored** | ← |
| Trashed | Delete Templates Forever | Templates **Deleted** | ← |

"Restored" is kept for restore-from-trash — correct there, and consistent with the single-card `RestoreJourneyDialog`.

Source changes:

| File | Before | After |
| --- | --- | --- |
| `ArchivedJourneyList.tsx:51` | Journeys Restored | Journeys Unarchived |
| `ArchivedTemplateList.tsx:51` | Journeys Restored | Templates Unarchived |
| `ArchivedTemplateList.tsx:67` | Journeys Trashed | Templates Trashed |
| `TrashedTemplateList.tsx:51` | Journeys Restored | Templates Restored |
| `TrashedTemplateList.tsx:66` | Journeys Deleted | Templates Deleted |
| `JourneyListContent.tsx:229` | archived branch → Restored | archived branch → Unarchived |

Four new keys in `libs/locales/en/apps-journeys-admin.json`, generated by `nx run journeys-admin:extract-translations`:

`Templates Restored` · `Templates Deleted` · `Templates Unarchived` · `Journeys Unarchived`

Without them `t()` echoes the key — looks right in English, never translated into the other 14 locales. `JourneyListContent` requests its keys dynamically (`t(messageKey)`), which the extractor cannot see; the literals in the `Archived*List` components are what register them.

E2E page objects updated to match (`publisher-and-templates-page.ts`). Those specs are `test.describe.fixme` so they do not gate CI, but they should not be left stale.

Frontend only. No mutation, schema, or API change.

## Test plan

Five new snackbar assertions across three spec files, plus two updated in `JourneyListContent.spec.tsx` — these components had no snackbar text coverage at all before. Each was confirmed failing before its fix and passing after.

| Check | Result |
| --- | --- |
| `src/components/JourneyList` + `src/components/TemplateList` | 316 pass, 1 skipped, 0 fail |
| `journeys-admin:type-check` | clean |
| eslint (changed dirs) | 0 errors |
| prettier (changed files) | clean |

Note: `LastModifiedDate.spec.tsx` fails locally in non-UTC timezones and passes in CI. It pins a fake UTC clock but writes dates as `'2021-12-10 23:59'` with no timezone, which JS parses as local time. Pre-existing on `main`, unrelated to this PR — verified passing under `TZ=UTC`.

Manual:

| Page | Step | Expected snackbar |
| --- | --- | --- |
| `/` | Archived → Unarchive All | Journeys Unarchived |
| `/` | Trashed → Restore All | Journeys Restored |
| `/publisher` | Archived → Unarchive All | Templates Unarchived |
| `/publisher` | Archived → Trash All | Templates Trashed |
| `/publisher` | Trashed → Restore All | Templates Restored |
| `/publisher` | Trashed → Delete All Forever | Templates Deleted |
| `/publisher` | Active → Archive All | Templates Archived (unchanged) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated confirmation messages to clearly distinguish between journey and template actions.
  * Renamed restore notifications to “Unarchived” for clearer status communication.
  * Added accurate success messages for bulk unarchive, trash, restore, and permanent deletion actions.
* **Localization**
  * Added translations for the updated journey and template status messages.


jaco-brink:
Optional cleanup (non-blocking): the five newly added snackbar tests largely duplicate the mutation tests immediately above them, including the provider setup, mocks, loading wait, click, and mutation flow. Could we fold each snackbar expectation into its existing unarchive/trash/restore/delete test instead? That would preserve the regression coverage while avoiding five additional render and mutation flows.

sharonemathew:
@jaco-brink done in 917fcb5a2.

Each snackbar expectation now lives in the unarchive/trash/restore/delete test above it, and the standalone tests are gone. Pure deletion: 103 lines removed, nothing added.

| File | Folded in |
| --- | --- |
| `ArchivedTemplateList.spec.tsx` | Templates Unarchived, Templates Trashed |
| `TrashedTemplateList.spec.tsx` | Templates Restored, Templates Deleted |
| `ArchivedJourneyList.spec.tsx` | Journeys Unarchived |

92/92 passing across the 6 affected specs, lint and type-check clean.

The only thing the separate tests bought was failure attribution by name — a broken toast now surfaces as `should unarchive all templates` failing on its last line rather than a test named for the snackbar. Cheap price for five fewer render and mutation cycles.

jaco-brink:
Non-blocking PR description correction: the Test plan currently says “Ten new snackbar assertions across four spec files.” Against main, this PR adds five new snackbar assertions across three files and updates two existing snackbar assertions in JourneyListContent.spec.tsx (making four touched spec files total). Could we update that sentence so the stated coverage matches the diff—for example: “Five new snackbar assertions across three spec files, plus two updated assertions in JourneyListContent.spec.tsx”?

sharonemathew:
@jaco-brink corrected, and it turned up a second stale number.

**Assertion count.** Your read is right. Against `main`: 5 new snackbar assertions across 3 spec files (`ArchivedJourneyList` ×1, `ArchivedTemplateList` ×2, `TrashedTemplateList` ×2), plus 2 updated in `JourneyListContent.spec.tsx` where the existing `Restored` assertions became `Unarchived`. Four files touched, three gaining new assertions. Updated to your wording.

The "ten" almost certainly came from counting the two end-state tables in the Fix section (5 journeys + 5 templates) rather than the diff — those tables describe every bulk action, not just the changed ones.

**Test count.** While checking, the pass count was also stale: 321 → **316**. The difference is exactly the 5 tests removed by 917fcb5a2 when the snackbar assertions were folded in. The description was written before that commit.

`0 fail` was correct as written — the two `LastModifiedDate` failures are timezone-dependent and pass under `TZ=UTC`, as the note below the table says. Re-verified just now:

```
TZ=UTC → Test Files  35 passed | 1 skipped (36)
              Tests  316 passed | 1 skipped (317)
```

apps/journeys-admin-e2e/src/pages/publisher-and-templates-page.ts
```diff
@@ -79,7 +79,7 @@ export class Publisher {
     await this.clickThreeDotBesideSortByOption()
     await this.selectThreeDotOptionsBesideSortByOption('Unarchive All')
     await this.clickDialogBoxBtn('Unarchive')
-    await this.verifyToastMessage('Journeys Restored')
+    await this.verifyToastMessage('Templates Unarchived')
```
jianwei1, line 82: **Concern:** one stale toast assertion left just above these — commenting here because line 71 is outside the diff.

`verifyAllTemplateMovedActiveToArchivedTab()` still asserts `'Journeys Archived'`:

```ts
// publisher-and-templates-page.ts:71
await this.verifyToastMessage('Journeys Archived')
```

But the Templates Active → Archive All path emits **`Templates Archived`** — `ActiveTemplateList.tsx:51` has done so since before this PR. The end-state table in the description lists that row as "Templates Archived (unchanged)", so the source is correct and only this assertion is behind. `verifyToastMessage` matches via `hasText` (substring), so `'Journeys Archived'` will never match `'Templates Archived'`.

Suggested change on line 71:

```ts
await this.verifyToastMessage('Templates Archived')
```

Not urgent — the enclosing block is `test.describe.fixme` (`publisher-and-templates.spec.ts:49`), so nothing here runs today, and this predates your change. Raising it because it's the same class of staleness this PR is fixing, in the same file, and whoever eventually re-enables the suite will hit a failure unrelated to their own change.

---

jianwei1 (commented):
One non-blocking concern on a stale e2e toast assertion — see inline. The source changes and the new unit test coverage look correct to me.

---

apps/journeys-admin-e2e/src/pages/publisher-and-templates-page.ts
```diff
@@ -68,7 +68,7 @@ export class Publisher {
     await this.clickThreeDotBesideSortByOption()
     await this.selectThreeDotOptionsBesideSortByOption('Archive All')
     await this.clickDialogBoxBtn('Archive')
-    await this.verifyToastMessage('Journeys Archived')
+    await this.verifyToastMessage('Templates Archived')
```
sharonemathew, line 71: Verified and fixed in 4ab73d6d7 — thanks, this was worth raising.

Checked each part rather than taking it on faith:

| Claim | |
| --- | --- |
| Line 71 asserted `Journeys Archived` | ✅ |
| Source emits `Templates Archived` (`ActiveTemplateList.tsx:51`) | ✅ |
| `hasText` is substring matching, so it could never match | ✅ |
| `test.describe.fixme`, so nothing runs today | ✅ |

One thing to add to your case for fixing it now: `verifyToastMessage` uses a **300000ms timeout**. So this would not have failed fast — it would have hung for five minutes before failing, on an assertion unrelated to whatever the next person was working on.

I also confirmed `verifyAllTemplateMovedActiveToArchivedTab` is called only from the templates spec (lines 74 and 83), so `Templates Archived` is unambiguous rather than a judgement call.

While in there I checked `journey-page.ts` for the same class of staleness after the NES-1874 change. It is clean — line 189 asserts `Journeys Restored` for Restore All **from Trash**, which NES-1874 deliberately left alone, and there is no bulk Unarchive All method in that file at all.

All five bulk-action assertions in the Publisher page object now match the source.

jianwei1 (approved):

PR #9459, opened 2026-08-10
fix: add missing space in Nepali/German "Delete {{ label }}" translation

sushmamalampati:
## Summary
- Fixes [NES-1789](https://linear.app/jesus-film-project/issue/NES-1789/card-remains-in-english-in-delete-card-context-menu-option-when-non): "Card" appeared in English in the "Delete Card" context menu option for non-English languages.
- Root cause investigation found this was actually already fixed for the code/most locales: the Nepali `"Card"` translation didn't exist when this ticket was filed (Jun 3), but a Crowdin sync on Jun 19 (`ff7c8f296`) added it. The code (`t('Delete {{ label }}', { label: t('Card') })`) was always correct.
- What remained: the `"Delete {{ label }}"` translation template itself is missing a space before the verb in **Nepali** and **German**, so the interpolated result is squished together (`कार्डमेटाउनुहोस्` / `Kartelöschen`) instead of properly spaced (`कार्ड मेटाउनुहोस्` / `Karte löschen`).
- This PR fixes just that spacing in `libs/locales/ne-NP/apps-journeys-admin.json` and `libs/locales/de-DE/apps-journeys-admin.json`.

**Out of scope (flagged, not fixed here):** 7 other locales (am-ET, bn-BD, hi-IN, my-MM, tl-PH, ur-PK, vi-VN) have no translation at all yet for `"Delete {{ label }}"` — that needs an actual translator via Crowdin, not a guessed fix from me.

## Test plan
- [x] Verified the exact interpolation (`t('Delete {{ label }}', { label: t('Card') })`) against the real locale JSON using a live i18next instance — confirmed output is now `कार्ड मेटाउनुहोस्` (ne) and `Karte löschen` (de).
- [x] Existing `DeleteBlock.spec.tsx` suite passes (8/8).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected German and Nepali translations for the delete action by adding the missing space between the item label and translated text.
  * Delete menu labels now display correctly in both supported languages, improving readability and consistency across the interface.


sushmamalampati:
Closing this PR — the remaining fix needs to happen in Crowdin itself (the source of truth for these translations), not as a direct edit to the synced JSON files here. A repo-only patch would get silently overwritten by the next automated Crowdin sync. This will be handled on the Crowdin side separately.

PR #7991, opened 2025-10-16
feat: add select available youtube subtitles, preview youtube video with subtitles 

jaco-brink:
## Summary by CodeRabbit

* **New Features**
  * Added subtitle selection for YouTube videos in the video editor
  * Introduced searchable subtitle language selector with "Off" option
  * Added notification when subtitles are available for a video
  * Implemented automatic YouTube closed captions fetching and configuration
  * Extended video block properties to support subtitle language selection

PR #9461, opened 2026-08-10
fix(journeys-admin): truncate long language names, pin Select button (NES-1860)

sharonemathew:
## Issue

[NES-1860](https://linear.app/jesus-film-project/issue/NES-1860/language-selector-ui-bugs-when-language-name-is-too-long) — two layout bugs in the video block's language UI when a language name is long.

1. **Source card:** the language name overflowed and was clipped by the card edge, with no ellipsis.
2. **Details footer:** the `Select` button's label wrapped onto two lines — 选择 rendered as 选 over 择.

Both causes were non-obvious:

- **Bug 1** — the truncation CSS was already there and inert. MUI 7 has no `caption` entry in `Typography`'s variant map, so `variant="caption"` renders a `<span>`, and `overflow` / `text-overflow` do nothing on an inline element.
- **Bug 2** — the chip and button are flex siblings that both shrink. A long chip label squeezed the button down to MUI's 64px minimum, too narrow for its label. CJK text can break between characters, so it wrapped; English `Select` is one unbreakable word, which is why only CJK locales showed it.

## Fix

Seven CSS properties. No behaviour change.

- `display: 'block'` on the caption in `SourceFromLocal`, `SourceFromYouTube` and `SourceFromMux`, so the existing ellipsis rules apply.
- `minWidth: 0` on the footer row, `minWidth: 0` + `flexShrink: 1` on the chip, `flexShrink: 0` on the button — the chip absorbs the squeeze, the button keeps its label on one line.

This is the pattern from `docs/solutions/ui-bugs/badge-alongside-webkit-line-clamp-title.md` (NES-1549), with the roles reversed.

Three notes for reviewers:

- **YouTube and Mux aren't in the ticket.** Same inert CSS, but short fixed labels — nothing visibly changes. This is hardening against a future label or translation change, not a fix for a live bug. Happy to drop them if you'd rather the diff stayed inside NES-1860.
- **No responsive breakpoint**, unlike the `JourneyDetails` precedent in `docs/solutions/ui-bugs/journey-details-chip-mobile-stacking-2026-05-24.md`. Once the button is pinned the row fits at every width this drawer takes — 272px, 280px, 327px — so a second layout would be maintenance for no gain.
- **NES-1568 is untouched.** Only `sx` lines changed. `onClick`, `disabled`, `startIcon` and the whole `VideoLanguage` block are unchanged, and both Apply regression tests still pass.

## Test plan

1. Set the app to 简体中文.
2. Journey builder → Video block → Video Properties → Video Source.
3. Pick a video with a long language name. It should truncate with an ellipsis inside the card.
4. Open the language selector, choose a long language, and confirm 选择 stays on one line.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video and subtitle language labels to truncate cleanly with ellipses when names are long.
  * Prevented long language names from wrapping, collapsing, or obscuring nearby Select buttons.
  * Enhanced layout behavior for local, custom, and YouTube video sources in narrow panels.
  * Preserved space for dropdown controls in subtitle selection.

* **Tests**
  * Added regression coverage and visual examples for long-language labels, multilingual names, and constrained layouts.


sharonemathew:
## Self-review — description corrected

Two fixes to the write-up, not the code. The commit is unchanged.

1. **Removed an invented number.** The case for including `SourceFromYouTube` / `SourceFromMux` claimed the longest `Custom Video` translation "leaves only ~18px of headroom in a 168px column". That was estimated from character counts, never measured — and it was carrying the argument for widening the diff beyond the ticket. Now says what's true: the longest translation fits, so those files are hardening rather than a fix. Weaker case, but the honest one. Happy to drop them.
2. **Disclosed a test gap.** Those same two files have no assertion covering their change, so a future refactor could undo it silently. Now called out in the test plan.

**Also found, out of scope:** `HostListItem.tsx:64-73` has the identical inert `caption` CSS around `{hostLocation}` — and that one is live, because the value is user-entered. A scan found four instances of this pattern repo-wide; it only catches CSS on the same element as the variant, so four is a floor. Tracked separately.

**Not yet verified:** the symptoms haven't been confirmed in a running app. The tests prove the CSS is applied, not that the text truncates or that 选择 stays on one line. Manual steps are in the test plan.

**Known trade-off:** truncated names have no tooltip, so sighted users can't read the full name. Screen readers still get the complete string from the DOM. Left as a product call rather than folded in here.

sharonemathew:
## Third fix — YouTube subtitle field (fa2d2c79f)

Fixes Esther's QA finding on NES-1860: a long subtitle language name ran under the dropdown arrow instead of truncating.

Two causes, both in one styling block in `YouTubeSubtitleSelector.tsx`:

1. MUI already truncates that field's text with an "…", but a `display: flex` override switched it off — `text-overflow` only works on a block element.
2. `padding: '16px 12px'` wiped out the 32px MUI reserves for the dropdown arrow, so the text ran underneath it.

Same shape as bug 1 in this PR — a `display` value quietly voiding truncation CSS that looks like it works.

**Fix:** drop the flex overrides so MUI's own truncation applies, and use `padding: '16px 32px 16px 12px'` so the arrow keeps its space. The duplicate `.MuiFilledInput-input` block was styling the same element, so it's folded into one rule. Both overrides came from the original feature (#7991), not from a later fix.

Also checked: journeys-admin has no RTL layout, so a plain `paddingRight` is correct. If RTL is ever added this needs `paddingInlineEnd`.

## Testing

250 tests pass; type-check and lint clean. One test covers the long composed name.

**No automated test for the layout.** I wrote one, then removed the fix to check it — it still passed. These styles sit behind a `& .MuiSelect-select` descendant selector, which jsdom's computed style doesn't apply, so `toHaveStyle` passes either way. Removed rather than leave coverage that only looks real; noted in the spec.

**Re-test:** try a subtitle language that's long *and* non-Latin (Farsi, Arabic) — mixed scripts are the harshest case.

csiyang (dismissed):

sharonemathew:
## Merged main — conflicts resolved (aec1a5cb4)

Main landed the MUI 7→9 upgrade (#9445). Its codemod rewrote the same `sx` blocks these fixes live in, causing three conflicts — `SourceFromLocal`, `SourceFromMux`, `SourceFromYouTube`.

All three were the same shape: the codemod moved `color="text.secondary"` into `sx`, while this branch added `display: 'block'` to that same `sx`. Both intents kept — `sx` now carries both. `LocalDetails.tsx` and `YouTubeSubtitleSelector.tsx` merged cleanly.

**The fixes were re-verified against MUI 9, not assumed.** All three were reasoned from MUI 7 internals, so I installed MUI 9.2.0 and re-checked each one. Still true in 9: `caption` is absent from `Typography`'s variant map (so it renders a `<span>`), `Button` keeps its 64px minimum with no `nowrap`, `Select` still truncates its value, and still reserves 32px for the arrow. So every fix remains necessary and correct.

**Checks**, run against MUI 9 after installing it — a run against the stale MUI 7 would have proved nothing: 250 tests pass, type-check clean, prettier clean.

Unchanged from earlier: the layout behaviour has no automated coverage. The MUI 9 upgrade is a further reason to re-test on the preview, since it touched hundreds of files well beyond this PR.

csiyang (approved):

PR #9472, opened 2026-08-12
fix(journeys-admin): capitalise domain nouns in empty states and dialogs (NES-1871)

sharonemathew:
## Issue

[NES-1871](https://linear.app/jesus-film-project/issue/NES-1871/text-style-inconsistency-title-case-vs-sentence-case) — Journey, Template, Project and Collection are proper nouns in this product, but the empty states and bulk-action dialogs on the team projects, team templates and publisher pages still rendered them in lower case.

## Fix

Capitalised those nouns across 8 components — 52 lines, all strings, no logic changed:

- Empty states: team projects, team templates, publisher active/archived/trashed, trashed journeys
- Bulk-action dialogs: Archive / Unarchive / Trash / Restore / Delete, both Journey and Template variants
- Remove-collection dialog

The copy exists twice — in `JourneyListContent` and again in the per-status list components — so both were updated, otherwise the fix would land on some screens and not others.

Two extras: dropped the duplicated word in *"archive a template `to to` delist"*, and fixed a spec regex that could no longer match the rendered text, which had made a "warning is hidden" assertion pass vacuously.

The English catalogue is regenerated with `extract-translations`. The 23 translated locales belong to Crowdin and are untouched, so these strings render in English for non-English users until Crowdin round-trips.

## Test plan

- [x] 31 spec assertions updated — written before the fix and confirmed failing first
- [x] `journeys-admin` suite passes (the 8 failing files are pre-existing timezone-fragile date tests; they pass under `TZ=UTC`)
- [x] type-check, lint and Prettier clean, re-verified after merging the MUI 9 upgrade
- [x] Preview deploy: walk each empty state and each dialog

## Summary by CodeRabbit

* **Style**
  * Standardized capitalization for Journeys, Templates, and Collections across management screens.
  * Improved wording in archive, trash, restore, deletion, and collection-removal messages.
  * Corrected duplicated wording in template guidance text.

* **Bug Fixes**
  * Updated empty states and confirmation dialogs to display consistent, polished messaging.
  * Aligned displayed messages and automated checks across journey and template management screens.


csiyang (changes requested):
Nice, clean capitalization pass — strings are consistent and the vacuous-regex catch in `CollectionUngroupDialog.spec.tsx` was a good one.

One change request before merge, on `libs/locales/en/apps-journeys-admin.json`: the capitalized strings were appended as new keys, but the old lowercase versions they replaced are still in the file as orphans (the extract tooling has `removeUnusedKeys: false` in the shared `i18next.config.base.js`, so re-running `extract-translations` won't prune them — it has to be a manual delete).

Could you remove the lowercase keys that this PR superseded? They're now dead — no source references them, so nothing renders them — but leaving them means the catalogue carries two copies of each string. That also includes the buggy dead one at line 1070, `"You can archive a template to to delist it from the Template Library."` (duplicated "to") — the fixed version is already in as a new key, so the orphan is safe to drop.

Scope this to just the keys this change orphaned, not a repo-wide cleanup — the broader `removeUnusedKeys` question is a separate decision. Everything else looks good to me.

sharonemathew:
@csiyang done in 78ecdf832.

Removed the 36 lowercase keys this change orphaned. Before deleting I checked each one for remaining references in any .ts/.tsx outside libs/locales — all 36 came back with zero, so nothing renders them. That includes the to to one you spotted at line 1070; the corrected string is already in as its own key.

Catalogue goes 1,395 → 1,359 keys, which puts it back exactly where it started before this PR — 36 added, 36 removed. New capitalised keys all verified still present, file still parses, tests green.

Kept it scoped to just these keys as you suggested — left removeUnusedKeys alone.

One heads-up: CodeRabbit left a comment asking for the opposite — to add all 37 new keys into the 22 non-English catalogues by hand. I haven't actioned that one. crowdin.yml sets libs/locales/en/*.json as the only source and the crowdin bot generates the rest, so hand-editing them would get overwritten on the next sync, and would write untranslated English into the translated files in the meantime. Happy to be told otherwise if there's history there I'm missing.

csiyang (approved):

PR #9473, opened 2026-08-12
fix(api-media): remove suppressed languages, emit numeric languageId (QA-375)

tanflem:
Related to #9474 — the Algolia-side half of the `hasVideos` suppression problem. The operator-owned flag itself lands in #9482, which closes that issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Language records in search results now expose language IDs in a consistent numeric format.
  * Preserved existing string-based record identifiers for compatibility.
  * Removed outdated language records when languages are unavailable or no longer contain videos.
  * Improved language index synchronization to keep searchable results aligned with available video content.
* **Maintenance**
  * Enhanced language reindexing to report both updated and removed records.

### DIFF
--- a/apis/api-media/project.json
+++ b/apis/api-media/project.json
@@ -163,7 +163,10 @@
     },
     "reindex-languages-algolia": {
       "executor": "nx:run-commands",
-      "dependsOn": ["prisma-languages:prisma-generate"],
+      "dependsOn": [
+        "prisma-languages:prisma-generate",
+        "prisma-media:prisma-generate"
+      ],
       "options": {
         "command": "pnpm exec tsx --tsconfig apis/api-media/tsconfig.app.json apis/api-media/src/scripts/algolia-helpers/reindex-languages-with-videos.ts"
       }
--- a/apis/api-media/src/lib/languages/updateLanguageInAlgolia.spec.ts
+++ b/apis/api-media/src/lib/languages/updateLanguageInAlgolia.spec.ts
@@ -116,7 +116,9 @@ describe('updateLanguageInAlgolia', () => {
     })
 
     it('emits languageId as a number while objectID stays a string', () => {
-      const record = buildAlgoliaLanguageRecord(createLanguage({ id: '143846' }))
+      const record = buildAlgoliaLanguageRecord(
+        createLanguage({ id: '143846' })
+      )
 
       // The rest of the index stores languageId numerically and arclight's
       // AlgoliaLanguageHit types it as a number, so a string here would make
--- a/apis/api-media/src/lib/languages/updateLanguageInAlgolia.spec.ts
+++ b/apis/api-media/src/lib/languages/updateLanguageInAlgolia.spec.ts
@@ -102,7 +102,7 @@ describe('updateLanguageInAlgolia', () => {
 
       expect(record).toEqual({
         objectID: '1234',
-        languageId: '1234',
+        languageId: 1234,
         bcp47: 'plu',
         iso3: 'plu',
         nameNative: 'Parikwaki',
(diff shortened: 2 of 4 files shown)


tanflem:
Correction to the commit message, and some prod context for the reviewer.

**Correction.** The commit message says the 4 string-typed records were written "since the sync fix deployed". That timing claim is not supported — the pre-QA-375 version of `buildAlgoliaLanguageRecord` also emitted `languageId: language.id` (a string), so a string value proves *api-media wrote the record*, not *when*. The rest of the message stands: the index is heterogeneous and this PR makes api-media's writes match everything else in it.

**Prod measurement** (read-only, `video-languages-prd`, 2319 records):

- Sampling the first 1000 records: **996 numeric, 4 string**. The 4 string ones are `20525` (Twi), `185257` (Saraiki, Megwhal), `4101` (Zhuang, Liujiang), `139188` (Zhuang, Central Hongshuihe).
- All 1000 sampled `objectID`s are numeric strings, so `Number(language.id)` cannot produce `NaN` for anything currently indexed.
- Algolia's `paginationLimitedTo` capped the scan at 1000, so "4" is a floor across the first page, not a total across all 2319.

**Why nothing is on fire today:** both arclight output paths coerce from `objectID` rather than the `languageId` field — `Number(hit.objectID)` at `_resources/index.ts:438` and `:489` — so the `z.number()` response schema is satisfied either way. The `languageId: number` declaration in `AlgoliaLanguageHit` (`_resources/index.ts:51`) is simply false for our records right now. This PR makes the declaration true again.

tanflem:
@Kneesal heads up — I expanded the scope of this PR after you were added, and the new part is the riskier half. Retitled to match. Happy to split it into its own PR if you'd rather review them separately.

**Added:** a delete path for the languages index (#9474).

Hiding a language is done by hand-editing `hasVideos` to `false`, but the Algolia sync had no delete anywhere in it — only `saveObjects`. So a language indexed while the flag was `true` stays searchable after it is turned back off.

That is live in prod today: **Twi (`20525`, 142 published variants) and Dari (`117014`) are both flagged `hasVideos: false` and both still searchable.** Five other suppressed languages are correctly absent.

Two changes:

1. `updateLanguageInAlgoliaFromMedia` now deletes the object when the language is missing or `hasVideos` is false, rather than skipping the write. This stops new records being stranded.
2. The reindex script gains a second pass that removes languages which have variants but `hasVideos: false`, to clear records stranded before this change.

**Why pass 2 is scoped to languages with variants:** the proper fix is to browse the index and delete anything not in the `hasVideos: true` set, but the api-media Algolia key returns `403 Method not allowed` for both `browse` and `query` (it can `deleteObject`/`deleteObjects`). So there is no way to ask which objectIDs are actually in the index. Sweeping all ~90k `hasVideos: false` languages would be ~90k pointless delete operations; scoping to those with variants is 7 in prod and covers the real failure mode. Deleting an absent objectID is a harmless no-op, so the narrower set costs nothing in correctness.

If someone can grant the api-media key `browse` ACL, pass 2 should become a real index-vs-DB reconciliation instead — worth doing, but it is an ops change, not a code one.

**Verified on stage:** 2241 reindexed, 5 suppressed languages swept, index total unchanged at 2241, control record `529` intact and now carrying a numeric `languageId`. Full api-media suite passes (785 tests).

**Not included:** clearing the two stranded prod records. That is a prod write and needs a deliberate run of the script against `prd`.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 24 checks green, no unresolved review threads, no conflicts. +205/-18 across 4 files. Removes suppressed languages and emits a numeric `languageId` (QA-375).

### What needs to happen next

Ready for review — no action needed from you. `@Kneesal` is already the requested reviewer.

PR #9151, opened 2026-05-05
refactor: remove child translation templates

mikeallisonJS:
Deleted the generated files for GetChildJourneysFromTemplateId and GetParentJourneysFromTemplateId, along with their associated hooks and mocks. Updated JourneyCardMenuItem, CopyToTeamMenuItem, LanguageScreen, and CopyToTeamDialog components to remove references to the deleted queries. This cleanup improves code maintainability and reduces unnecessary dependencies.

## Release Notes

* **Refactor**
  - Simplified template language selection and duplication workflows
  - Removed restrictions on copying and translating certain template scenarios

* **Chores**
  - Enhanced TypeScript module declarations for image file imports across applications

PR #9476, opened 2026-08-12
docs(context-map): absorb playbook question sequences and readiness gates (ENG-3708)

csiyang:
Finishes migrating the **AI Bug-Intake Playbook** into the per-area bug-diagnosis map files, so the playbook can retire as a source of truth for the covered areas (journeys-admin editor, journeys viewer). Docs-only: the two `CONTEXT-intake.md` files + the intake index.

Linear: [ENG-3708](https://linear.app/jesus-film-project/issue/ENG-3708/absorb-playbook-into-map-sections)

## What was absorbed, per type

The shipped sections already had signatures, one localizing question, look-first pointers, and `Handoff:` verdicts. The consistent gap was the playbook's **follow-up question sequences** and **per-type readiness gates**, plus some reporter vocabulary. Each touched section gains a `**Then ask:**` line (the ordered follow-ups) and a `**Ready when:**` line (the gate), matching the existing bold-label format.

| Type | Status | What this PR adds |
|---|---|---|
| T1 data semantics | absorbed remainder | which view; rows-created vs migration-shipped timing; working-vs-broken row pair; gate |
| T2 auth/sign-in | absorbed remainder | entry path, email/used-before, product/env, still-happening; gate. New trigger phrases ("can't log in", "works in incognito", "name shows Unknown") |
| T3 analytics | absorbed remainder | event→goal mapping, edited-after-collection, date range; gate reframed as ready-for-**human**-handoff (stays route-to-human per grill) |
| T4/T5 cache & drift | absorbed remainder | enumerate-all-feeding-mutations (NES-1199's nine paths), stale-view location, two-collection repro; separate T4/T5 gates |
| T6 rendering | absorbed remainder | annotated screenshot, device from reporter (never guess), cosmetic-vs-functional, image file format; gate |
| T7 canvas crash | not live — no section | the one-line not-live note in the index taxonomy is **folded into this PR** (from #9475, which closes in favour of this one); no intake section, per Siyang's 2026-08-13 call |
| T8 i18n/translation | absorbed remainder | exact language + JF ID, surface, public-vs-team; **quota/ops branch** (broad sudden failure ⇒ provider quota, not code); gate. Playbook's stale list counts (~14/~32/~46/~25) deliberately **not** imported — grill-corrected four-list model kept |
| T9 duplication collision | retired — kept as-is | none (resolved by removal, #9151; stale-report note already shipped) |
| T10 media/video | absorbed remainder | wrong-language signature + branch (video language vs journey language), BG-cover vs block / first card, what-changed-recently, old-WebKit rule-out; gate. New trigger "video plays the wrong language" |
| T11 permissions/ACL | absorbed remainder | role, exposed-vs-blocked, expected policy, invite resend; gate for the agent-able invite-email path; cross-team exposure explicitly never gated — escalate immediately. New triggers ("invite email never arrived", "I can see another team's data") |

Also: a cross-cutting "working-vs-broken pair is the highest-leverage artifact" heuristic added to the admin file preamble; new trigger phrases synced frontmatter ↔ `CONTEXT-MAP-intake.md`; `updated:` bumped.

Follow-up commits folded in after review discussion:

- the **T7 not-live taxonomy note** from #9475 (that PR closes in favour of this one)
- the admin **Integrations** section's `- Look first (fixer):` / `- Handoff:` labels bolded to match `CONTEXT-MAP-contract.md`'s reserved-label format (flagged by the ENG-3707 validator; the viewer Chat/AI section's missing Handoff line is left as-is deliberately — unreleased feature)

## Deliberately not absorbed

Harness/procedure content that belongs to the intake agent (ENG-3706), not the map: the 5-bucket triage flow, universal questions, repro-access handling (§6), escalate-not-nag, FAQ-deflection candidate, and §7 open questions (audit found them answered or absorbed elsewhere). Where the ENG-3685 grill corrections conflict with the playbook, the corrections win throughout.

## Verification

- Every T1–T11 type is now either embodied in a core-area file or carries an explicit not-live/retired note (T7 note included in this PR, T9 already shipped).
- Frontmatter `code_paths` on touched files verified to exist; no code paths added or changed.
- No app code, no skill changes.

## Summary by CodeRabbit

* **Documentation**
  * Expanded intake guidance for authentication, identity, invitations, access, translations, analytics, historical data, caching, and optimistic updates.
  * Added clearer triage questions and readiness criteria for stale views, quota issues, sign-in problems, invite-email failures, and data visibility.
  * Improved video troubleshooting for language mismatches, rendering issues, video types, first-card behavior, and working versus failing examples.
  * Added trigger phrases, preview-specific refresh guidance, refreshed documentation dates, and clarified editor/canvas block-state classification.

### DIFF
--- a/CONTEXT-MAP-intake.md
+++ b/CONTEXT-MAP-intake.md
@@ -10,8 +10,8 @@ trigger-phrases that route a report to it. The ENG-3707 accessor fetches an area
 
 ## Areas (NextSteps)
 
-- **journeys-admin** (creator/editor surface) | domain: `apps/journeys-admin/CONTEXT.md` | intake: `apps/journeys-admin/CONTEXT-intake.md` | triggers: "block won't save", "changes not showing in preview", "have to refresh", "can't translate into <language>", "template won't let me customize", "transfer ownership", "analytics numbers are wrong", "historical data disappeared"
-- **journeys (published viewer)** (audience surface) | domain: `apps/journeys/CONTEXT.md` | intake: `apps/journeys/CONTEXT-intake.md` | triggers: "button doesn't go to the next card", "goes to the wrong card", "video won't load", "video is slow", "image is cropped / wrong fit", "can't click / can't type on the card", "looks wrong when published", "custom domain / embed not working"
+- **journeys-admin** (creator/editor surface) | domain: `apps/journeys-admin/CONTEXT.md` | intake: `apps/journeys-admin/CONTEXT-intake.md` | triggers: "block won't save", "changes not showing in preview", "have to refresh", "can't translate into <language>", "template won't let me customize", "transfer ownership", "analytics numbers are wrong", "historical data disappeared", "can't log in", "name shows Unknown", "works in incognito", "invite email never arrived", "I can see another team's data"
+- **journeys (published viewer)** (audience surface) | domain: `apps/journeys/CONTEXT.md` | intake: `apps/journeys/CONTEXT-intake.md` | triggers: "button doesn't go to the next card", "goes to the wrong card", "video won't load", "video is slow", "video plays the wrong language", "image is cropped / wrong fit", "can't click / can't type on the card", "looks wrong when published", "custom domain / embed not working"
 
 _The two intake surfaces are where bugs are **reported** (creator editor, audience viewer). The
 backend (`api-journeys`) and shared kernel (`libs/journeys/ui`) are reached via each surface's
--- a/CONTEXT-MAP-intake.md
+++ b/CONTEXT-MAP-intake.md
@@ -10,8 +10,8 @@ trigger-phrases that route a report to it. The ENG-3707 accessor fetches an area
 
 ## Areas (NextSteps)
 
-- **journeys-admin** (creator/editor surface) | domain: `apps/journeys-admin/CONTEXT.md` | intake: `apps/journeys-admin/CONTEXT-intake.md` | triggers: "block won't save", "changes not showing in preview", "have to refresh", "can't translate into <language>", "template won't let me customize", "transfer ownership", "analytics numbers are wrong", "historical data disappeared", "can't log in", "name shows Unknown", "works in incognito", "invite email never arrived", "I can see another team's data"
-- **journeys (published viewer)** (audience surface) | domain: `apps/journeys/CONTEXT.md` | intake: `apps/journeys/CONTEXT-intake.md` | triggers: "button doesn't go to the next card", "goes to the wrong card", "video won't load", "video is slow", "video plays the wrong language", "image is cropped / wrong fit", "can't click / can't type on the card", "looks wrong when published", "custom domain / embed not working"
+- **journeys-admin** (creator/editor surface) | domain: `apps/journeys-admin/CONTEXT.md` | intake: `apps/journeys-admin/CONTEXT-intake.md` | triggers: "block won't save", "changes not showing in preview", "have to refresh to see it", "can't translate into <language>", "template won't let me customize", "transfer ownership", "analytics numbers are wrong", "historical data disappeared", "can't log in", "name shows Unknown", "works in incognito", "invite email never arrived", "I can see another team's data"
+- **journeys (published viewer)** (audience surface) | domain: `apps/journeys/CONTEXT.md` | intake: `apps/journeys/CONTEXT-intake.md` | triggers: "button doesn't go to the next card", "goes to the wrong card", "video won't load", "video is slow to load", "video plays the wrong language", "image is cropped / wrong fit", "can't click / can't type on the card", "looks wrong when published", "custom domain / embed not working"
 
 _The two intake surfaces are where bugs are **reported** (creator editor, audience viewer). The
 backend (`api-journeys`) and shared kernel (`libs/journeys/ui`) are reached via each surface's
--- a/CONTEXT-MAP-intake.md
+++ b/CONTEXT-MAP-intake.md
@@ -28,7 +28,7 @@ ENG-3686). Distilled from 2 years of #nextsteps-bugs history (AI Bug-Intake Play
 - **T4** cache / revalidation / stale page
 - **T5** client-side state sync / optimistic-update drift
 - **T6** rendering / visual / UI display
-- **T7** editor / canvas block-state
+- **T7** editor / canvas block-state — _no longer a live class (confirmed 2026-08-13, ENG-3704); no intake section_
 - **T8** i18n / translation / language lists
 - **T9** template duplication / language-id collision — _resolved by removal (#9151); reports stale_
 - **T10** media / video / external-content pipeline
--- a/apps/journeys/CONTEXT-intake.md
+++ b/apps/journeys/CONTEXT-intake.md
@@ -33,7 +33,7 @@ view (the "saved fine, wrong only in preview" branch handed over from journeys-a
 **Then ask:** an annotated screenshot (mark what's wrong on it); device/browser/OS **from the
 reporter** — never guess it from context (the guess is often wrong); cosmetic, or does it block
 interaction? For image bugs, also the image's file format.
-**Ready when:** annotated screenshot + the affected block/card + confirmed device/viewport.
+**Ready when:** annotated screenshot + the affected block/card + confirmed device/browser/OS/viewport.
 **Look first (fixer):** the shared Block Renderer + Card rendering in `libs/journeys/ui`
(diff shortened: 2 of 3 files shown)


apps/journeys/CONTEXT-intake.md
```diff
@@ -50,11 +55,17 @@ wires swipe/hotkey/button nav) → `libs/journeys/ui/src/libs/block/block.ts` (`
...
 **Status:** preloading is a **known weak spot** — attempted a few times, never much success. Treat
 "slow to load" as a known limitation, not a fresh regression, unless something clearly broke.
 **Localizing question (reporter):** which video source (library / YouTube / upload), which
 device/OS/browser, and a **working-vs-failing** example if possible.
+**Then ask:** background cover video or a video block — and is it the first card? What changed
+recently (an edit, a new upload, "started Monday" → bisect to a deploy)? For wrong-language: which
+language is set on the video vs on the journey? A very old browser/OS (e.g. a 7-year-old WebKit) ⇒
+likely not-a-code-bug — close with an explanation.
+**Ready when:** a failing-vs-working URL pair + video source + device/OS/browser + any recent
+change are recorded.
```
edmonday: This gate can't be met by a whole class of report.

Line 57 asks for a working-vs-failing example **if possible**. This line makes it mandatory. If a reporter says "none of my videos play", no pair exists — so the triager either keeps asking for something unobtainable, or hands off anyway and learns the gates are advisory. The second outcome is the costlier one; it weakens every other "Ready when" in the file.

Worth noting the gate also rejects the stronger evidence: "all videos fail" is more diagnostic than a pair.

```suggestion
**Ready when:** video source + device/OS/browser + any recent change are recorded,
plus a failing-vs-working URL pair where one exists — if every video fails, record
that instead; it is the stronger signal.
```

apps/journeys/CONTEXT-intake.md
```diff
@@ -50,11 +55,17 @@ wires swipe/hotkey/button nav) → `libs/journeys/ui/src/libs/block/block.ts` (`
...
 **Look first (fixer):** `libs/journeys/ui/src/components/Video/Video.tsx` (video.js setup, poster,
```
edmonday, line 70: `'video plays the wrong language'` now routes reports into this section, and the new **Then ask** localizes them well ("which language is set on the video vs on the journey"). But every pointer here is playback mechanics — `Video.tsx`, `InitAndPlay.tsx`, `BackgroundVideo.tsx`. None of them selects a language variant.

So the triager does the work, hands off, and the fixer opens three files that can't contain the bug. That's worse than the pre-PR state, where the report had no route and someone would have had to think about it.

The pattern you already used in the admin file in this same PR works here — naming the gap beats pointing at the wrong files:

> For non-block stale views … there is no single pointer yet — start from the enumerated feeding mutations …

Either a real pointer for the language path, or an explicit "no pointer yet".

apps/journeys/CONTEXT-intake.md
```diff
@@ -9,12 +9,13 @@ trigger_phrases:
   - 'goes to the wrong card'
   - "video won't load"
   - 'video is slow to load'
+  - 'video plays the wrong language'
   - 'image is cropped / wrong fit'
   - "can't click / can't type on the card"
   - 'looks wrong when published'
   - 'custom domain / embed not working'
 type_tags: [T6, T10]
```
edmonday: Language bugs are T8, but this stays `[T6, T10]`. T8 is claimed by the admin file (`[T1, T2, T3, T4, T5, T8, T11]`), whose T8 section covers translation lists and the AI-translate dialog — not video.

So phrase-routing sends a wrong-language video report here, and tag-routing sends it to admin. Neither has the answer. Suggest adding T8 here.

(Pairs with the **Look first** comment below — same finding, split so they can be resolved separately.)

CONTEXT-MAP-intake.md
```diff
@@ -10,8 +10,8 @@ trigger-phrases that route a report to it. The ENG-3707 accessor fetches an area
...
+- **journeys-admin** (creator/editor surface) | domain: `apps/journeys-admin/CONTEXT.md` | intake: `apps/journeys-admin/CONTEXT-intake.md` | triggers: "block won't save", "changes not showing in preview", "have to refresh", "can't translate into <language>", "template won't let me customize", "transfer ownership", "analytics numbers are wrong", "historical data disappeared", "can't log in", "name shows Unknown", "works in incognito", "invite email never arrived", "I can see another team's data"
+- **journeys (published viewer)** (audience surface) | domain: `apps/journeys/CONTEXT.md` | intake: `apps/journeys/CONTEXT-intake.md` | triggers: "button doesn't go to the next card", "goes to the wrong card", "video won't load", "video is slow", "video plays the wrong language", "image is cropped / wrong fit", "can't click / can't type on the card", "looks wrong when published", "custom domain / embed not working"
```
edmonday: Optional — small drift on lines you're already touching: the index says `"have to refresh"` and `"video is slow"`, while the frontmatter has `'have to refresh to see it'` and `'video is slow to load'`.

Pre-existing, and the five new phrases you added match exactly on both sides — so this is a two-word cleanup, not a request to audit the whole list.

edmonday (changes requested):
The "Then ask" / "Ready when" split is a good addition, and the admin preamble that defines the stop-asking rule makes the file much easier to follow.

Two inline comments to resolve before merge — both are about where the file *routes* people rather than how it reads: a readiness gate a whole class of report can't satisfy, and a new trigger phrase that lands on pointers which can't contain the bug. Thread 1 has a one-click suggestion. Happy to approve as soon as those two are settled.

Two optional nits below. One thing not worth changing, flagging only so it's a conscious call: the new blocks are indented under the preceding bullet in both files, so they render as part of that list item. It extends a pattern the old **Look first** lines already had, and the content self-labels (`**Then ask (T4, stale view)**`, `**Quota branch:**`), so nothing reads wrong.

Nit with no anchor (line 26 isn't in a diff hunk): the viewer preamble never defines "Then ask" / "Ready when" the way the admin one now does — one copied line would fix it.

apps/journeys/CONTEXT-intake.md
csiyang: Applied nearly verbatim in 72d085994 — pair where one exists, all-fail recorded as the stronger signal. Good catch on the gate rejecting the better evidence.

apps/journeys/CONTEXT-intake.md
csiyang: Fixed in 72d085994 with a real pointer rather than a "no pointer yet": wrong-language is resolved via the block's `videoVariantLanguageId` in `apis/api-journeys/src/schema/block/video/video.ts`, with variant-by-language selection in `apis/api-media/src/schema/video/video.ts` — verified in code, and the section now says explicitly that the playback files are the wrong place for this class.

apps/journeys/CONTEXT-intake.md
csiyang: Done in 72d085994 — via the heading (`· T8 (language variant)`) plus frontmatter, since the contract (`CONTEXT-MAP-contract.md`, #9478) requires `type_tags` to equal the union of heading tags; a frontmatter-only change would have failed the validator.

CONTEXT-MAP-intake.md
csiyang: Taken in 72d085994 — both index triggers now match their frontmatter phrases verbatim.

edmonday (approved):
All four threads resolved — thanks for the quick turnaround.

I verified the two new code pointers rather than taking them on faith, and both hold:

- `apis/api-journeys/src/schema/block/video/video.ts:113` — `primaryLanguageId: block.videoVariantLanguageId` really is where the block field feeds the video reference.
- `apis/api-media/src/schema/video/video.ts` — variant selection genuinely keys off `languageId`.

You went past what I asked here: I offered the "no pointer yet" escape hatch and you traced the actual mechanism instead. Tagging the heading `T10 · T8 (language variant)` is a nice touch too — the dual ownership is now visible without opening the frontmatter.

One non-blocking note for a future pass. While verifying, this pattern showed up across every variant field in `api-media/src/schema/video/video.ts`:

```ts
query: ({ languageId }) => ({ where: { languageId: languageId ?? '529' } })
```

A silent fallback to 529 (English) when no `languageId` arrives looks like a strong candidate for the exact symptom this section now routes — "plays the wrong language", specifically playing English when something else was set. A clause pointing at it would turn the pointer from a location into a first hypothesis. Not a condition of this PR.

The optional preamble nit is still open (the viewer file doesn't define "Then ask" / "Ready when" the way the admin one now does) — fine to leave or fold into a later pass.

PR #9477, opened 2026-08-12
chore: scoped lint-changed script + agent-gated pre-push lint hook (ENG-3691)

csiyang:
Closes ENG-3691.

## What

Two pieces:

1. `pnpm lint:changed` (`tools/scripts/lint-changed.sh`): scoped ESLint over changed files only, grouped by nearest `eslint.config.mjs` and run once per workspace from the repo root with `--config`, so flat-config path patterns resolve exactly as under `nx lint`. Workspaces lint in parallel, capped at `LINT_CHANGED_JOBS` (default 4). Modes: default (all changes vs merge-base with `origin/main`, incl. uncommitted/untracked), `--committed` (branch commits vs a freshly fetched `origin/main` — what the pre-push gate lints), `--staged`, `--fix` (fixes the working tree; re-stage/re-commit yourself). Conflicting mode flags exit 2.
2. An **agent-gated pre-push hook** (`.husky/pre-push`): when an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), the push is blocked until `lint:changed --committed` passes. Manual human pushes are not gated — for humans the script is advisory, documented in root `AGENTS.md`.

## Why this shape (measured)

The ticket deferred hook-vs-convention to measured efficiency. Timings on real files (M-series Mac, ESLint 9 flat config, type-aware rules via `projectService`):

| Run | Time |
| --- | --- |
| 2 files in `apps/journeys-admin`, cold | 21.5s |
| 1 file in `apps/journeys-admin`, warm | 7.4s |
| 1 file in `apis/api-journeys`, warm | 5.2s |
| Two workspaces, parallel, incl. `origin/main` fetch | 7.0s wall |
| Two workspaces forced serial (`LINT_CHANGED_JOBS=1`) | 14.4s |
| Non-agent push (hook skips) | <1s |
| No lintable changes (incl. fetch) | <1s |

The cost is the per-workspace TypeScript program build for type-aware rules — paid on every invocation regardless of file count. ~8–20s (tens of seconds cold) is too slow for a per-commit hook (the ticket's "few seconds" bar), but fine once per push, and only agents — whose loops this is for — are forced through it. No `lint-staged`, zero new dependencies.

## Contract and edge cases

- `--committed` refreshes `origin/main` when the network allows (best-effort `git fetch origin main --quiet`; offline, the base may be stale — but an offline push fails moments later anyway) before computing the merge-base, so a stale clone can't widen the diff onto unrelated code. If the merge-base still can't resolve (fresh clone, missing remote), it exits with an actionable message instead of a bare git error.
- The gate deliberately lints the branch diff vs `origin/main`, not the literal ref-range of the push; pushes to forks/other remotes — and branches cut from `stage`, which may over-lint — are judged against `origin/main` all the same (documented in `AGENTS.md`). `LINT_CHANGED_JOBS` caps parallel workspaces (default 4; validated).
- Filenames with spaces/glob chars are handled (array-based throughout, `set -euo pipefail`); a lint worker dying without a result fails the run loudly rather than silently passing the gate.
- Warnings don't fail (parity with CI's `nx lint`). Files outside any workspace config are printed as "skipped" — parity with `nx lint`, which only lints projects.

## Verified

- Deliberate error-severity violation (incl. in a `space test.tsx` filename), agent env set → push blocked, output names file/line/rule; `--fix` → passes.
- Same push without agent env → sails through in <1s.
- Missing `origin/main` (scratch repo) → actionable failure message.
- `--staged --committed` → conflict error, exit 2. `LINT_CHANGED_JOBS=1` serializes (14.4s vs 7.0s parallel, identical findings).
- Pre-existing branch-name/commitlint hooks untouched; every push on this PR went through the new hook.

## Escape hatch

`git push --no-verify` skips the hook (documented in the hook and `AGENTS.md`). Skipping just defers to autofix.ci/CI, which stay untouched as the backstop. Prettier is deliberately excluded: autofix.ci owns formatting.

## Summary by CodeRabbit

* **Chores**
  * Added automated lint checks for changed JavaScript and TypeScript files before pushes initiated by supported coding agents.
  * Added optional autofix support and clearer reporting for skipped or failed checks.
  * Human-initiated pushes remain unaffected, with an emergency bypass available when needed.

* **Documentation**
  * Documented changed-file linting commands, autofix usage, CI formatting responsibilities, comparison behavior, and bypass instructions.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,9 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# Agent-gated lint (ENG-3691): agents must lint what they push; manual human
+# pushes are not gated (run `pnpm lint:changed` yourself anytime). Escape
+# hatch for emergencies: `git push --no-verify`.
+if [ -n "${CLAUDECODE:-}" ] || [ -n "${CURSOR_AGENT:-}" ]; then
+  tools/scripts/lint-changed.sh --committed
+fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,7 +35,7 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 
 Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), `.husky/pre-push` runs the `lint:changed` script on committed changes. It lints only changed files, scoped per workspace; full `nx lint` is far too slow. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime. [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
 
-Contract of the gate: it lints the branch diff against a freshly fetched `origin/main` — not the literal ref-range of the push. Pushes to other remotes or forks are judged against `origin/main` all the same.
+Contract of the gate: it lints the branch diff against `origin/main`, refreshed when the network allows. It does not lint the literal ref-range of the push. Forks, other remotes, and branches cut from `stage` are all judged against `origin/main`, so a `stage`-cut branch may over-lint (use `git push --no-verify` if that blocks you). `LINT_CHANGED_JOBS` caps parallel workspaces (default 4).
 
 ### Documented Solutions
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 
 ### Lint before push
 
-Agents are lint-gated at push: `.husky/pre-push` runs `pnpm lint:changed --committed` (scoped ESLint over changed files only; full `nx lint` is far too slow) when an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`) — manual human pushes skip it, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime; [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
+Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), `.husky/pre-push` runs the `lint:changed` script on committed changes. It lints only changed files, scoped per workspace; full `nx lint` is far too slow. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime. [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
 
 ### Documented Solutions
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,10 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 - Use descriptive variable and function/const names.
 - Define TypeScript types; avoid `any`.
 
+### Lint before commit
+
+Run `pnpm lint:changed` before committing — scoped ESLint over changed files only (`--fix` applies autofixes; full `nx lint` is far too slow). Deliberately not a git hook; [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
+
 ### Documented Solutions
 
 The context map (`CONTEXT.md`, and `CONTEXT-intake.md` when diagnosing) is the primary knowledge source — rely on it by default. `docs/solutions/` is a **secondary, opt-in** archive of past problem write-ups (bugs, best practices, workflow patterns), organized by category with descriptive filenames and YAML frontmatter (`module`, `tags`, `problem_type`).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,9 +31,9 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 - Use descriptive variable and function/const names.
 - Define TypeScript types; avoid `any`.
 
-### Lint before commit
+### Lint before push
 
-Run `pnpm lint:changed` before committing — scoped ESLint over changed files only (`--fix` applies autofixes; full `nx lint` is far too slow). Deliberately not a git hook; [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
+Agents are lint-gated at push: `.husky/pre-push` runs `pnpm lint:changed --committed` (scoped ESLint over changed files only; full `nx lint` is far too slow) when an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`) — manual human pushes skip it, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime; [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
 
 ### Documented Solutions
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,8 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 
 Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), `.husky/pre-push` runs the `lint:changed` script on committed changes. It lints only changed files, scoped per workspace; full `nx lint` is far too slow. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime. [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
 
+Contract of the gate: it lints the branch diff against a freshly fetched `origin/main` — not the literal ref-range of the push. Pushes to other remotes or forks are judged against `origin/main` all the same.
+
 ### Documented Solutions
 
 The context map (`CONTEXT.md`, and `CONTEXT-intake.md` when diagnosing) is the primary knowledge source — rely on it by default. `docs/solutions/` is a **secondary, opt-in** archive of past problem write-ups (bugs, best practices, workflow patterns), organized by category with descriptive filenames and YAML frontmatter (`module`, `tags`, `problem_type`).
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "demo": "nx demo",
     "lint": "nx run-many --target=lint",
     "lint:fix": "nx run-many --target=lint --fix",
+    "lint:changed": "tools/scripts/lint-changed.sh",
     "prettier": "npx prettier .",
     "prettier:fix": "npx prettier . --write",
     "e2e": "nx e2e",
--- a/tools/scripts/lint-changed.sh
+++ b/tools/scripts/lint-changed.sh
@@ -27,11 +27,12 @@ FIX_ARGS=()
 for arg in "$@"; do
   case "$arg" in
     --staged | --committed)
-      if [ "$MODE" != "all" ]; then
+      new_mode="${arg#--}"
+      if [ "$MODE" != "all" ] && [ "$MODE" != "$new_mode" ]; then
         echo "conflicting options: --$MODE and $arg" >&2
         exit 2
       fi
-      MODE="${arg#--}"
+      MODE="$new_mode"
       ;;
     --fix) FIX_ARGS=(--fix) ;;
     *)
@@ -41,6 +42,12 @@ for arg in "$@"; do
       ;;
   esac
 done
+
+MAX_JOBS=${LINT_CHANGED_JOBS:-4}
+if ! [ "$MAX_JOBS" -gt 0 ] 2>/dev/null; then
+  echo "LINT_CHANGED_JOBS must be a positive integer (got '$MAX_JOBS')" >&2
+  exit 2
+fi
 
 diff_names() {
   git diff --name-only --diff-filter=ACMR "$@"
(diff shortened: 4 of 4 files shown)


edmonday:
## Code review — two-axis (Standards / Spec)

**Fixed point:** merge-base `62d9368` (`origin/main`) · 2 commits · 4 files · +135 −0
**Spec source:** ENG-3691 lives in Linear, which was not reachable from the review session (no GitHub issue, connector unauthorized). The PR description was used as the spec of record.

## Standards

### Hard violations

**1. `tools/scripts/lint-changed.sh` — word splitting on filenames**
`for f in $FILES`, `for f in $SKIPPED`, `for ws in $WORKSPACES`, and `$ws_files` passed unquoted to `eslint`. Any path with a space (or glob char — no `set -f`) is split or expanded. The inline `# shellcheck disable=SC2086 # ws_files is a list of paths without spaces` records the assumption but does not enforce it. The repo has no shellcheck config, so tooling does not catch this.

**2. Missing `set -euo pipefail`** — deviates from the established convention in `tools/scripts/reset-stage.sh` and `tools/scripts/check-prisma-migrations.sh` (both `set -euo pipefail`). This script uses `set -u` only. Consequence: `cd "$(git rev-parse --show-toplevel)" || exit 1` is guarded, but `mktemp -d` failing, or `cut`/`grep` failing mid-pipeline, proceeds silently.

**3. `origin/main` is not guaranteed fetched.**

```sh
BASE=$(git merge-base HEAD origin/main) || exit 1
```

On a fresh/shallow clone, or with a remote named other than `origin`, this exits 1 with only git's stderr and — via `.husky/pre-push` — blocks the push with no explanation. A stale `origin/main` silently widens the diff. No `git fetch`, no diagnostic.

**4. ASD-STE100 prose rules ("descriptive sentences under 25 words", "one idea per sentence").** The `AGENTS.md` addition is a single ~75-word sentence with four em-dash/parenthetical clauses. Split it.

### Not violations (verified)

- `. "$(dirname "$0")/_/husky.sh"` is **correct** here: `package.json` pins `husky: ^8.0.3` and both existing hooks (`pre-commit`, `commit-msg`) use the same preamble.
- `pnpm` is right (`packageManager: pnpm@10.33.4`).
- `#!/bin/bash` matches every other script in `tools/scripts/`.
- `mktemp -d` + `trap … EXIT` cleanup present and correct.
- `"$OUTDIR"/*.out` cannot fail to match: the loop is guarded by the non-empty `$WORKSPACES` check, and each subshell creates its `.out` via redirection.

### Judgement calls

- **Doc drift**: `AGENTS.md` says the hook runs `pnpm lint:changed --committed`; `.husky/pre-push` invokes `tools/scripts/lint-changed.sh --committed`. Equivalent today, drifts tomorrow.
- **Duplicated code**: the `merge-base` + `git diff --name-only --diff-filter=ACMR` shape appears in both the `committed` and `all` branches. Compute `BASE` once.
- **`grep "^$ws|"`**: `$ws` is interpolated as a BRE, so `.` in a workspace path matches any character. Use `grep -F` with an anchor, or restructure.
- **Opaque names**: `i` / `$i.out` / `$i.code` index files by loop counter, not workspace. Failure output cannot be traced back to a workspace without counting.

## Spec

**Caveat:** the authoritative ticket was not reachable, so "missing requirement" findings are limited to what the PR body claims.

### (a) Missing or partial

1. **"pre-existing branch-name/commitlint hooks untouched"** — true and verified: `.husky/pre-commit` (validate-branch-name) and `.husky/commit-msg` (commitlint) are unmodified, and no `pre-push` existed before. Preamble matches husky `^8.0.3` (`prepare: husky install`). No finding.
2. **"scoped ESLint over changed files only"** — partial. The script walks up only until `d = "."`, and there is no root `eslint.config.mjs`. Changed JS/TS outside a project dir (e.g. `tools/langfuse-export`, `tools/load-test`, repo root) is printed under "skipped" and silently passes. All 42 lint-target projects do have `eslint.config.mjs`, so the gap is confined to non-project files.
3. **"flat-config path patterns resolve exactly as under `nx lint`"** — holds. Workspace configs use root-relative patterns (`files: ['apps/journeys-admin/**/*.{ts,tsx,js,jsx}']`), consistent with running from the repo root with `--config`. No `maxWarnings` anywhere, so "warnings don't fail" is parity-correct.
4. **Exit-code propagation** — correct. The `if` block's status is the script's, so a non-zero lint blocks the push; non-agent pushes exit 0.

### (b) Scope creep

None material. `--staged` is unused by the hook and undocumented in `AGENTS.md`, but the PR body claims it.

### (c) Implemented but looks wrong

1. **`--committed` = "what a push would send"** — it is not. It uses `git merge-base HEAD origin/main` against the **local, never-fetched** `origin/main`. On a stale clone the base is old, so after a merge from main the diff includes hundreds of files the agent never touched — the hook then lints unrelated workspaces (minutes, not 10.6s) and can block the push on someone else's violation. A push actually sends `<remote-ref>..HEAD`; husky passes those refs on stdin, which the hook ignores.
2. **No `origin/main` ref** (fresh clone, renamed remote) → `merge-base` fails → `exit 1` → push blocked with a bare git error and no explanation.
3. **`for f in $FILES` / unquoted `$ws_files`** — filenames with spaces or glob characters break silently.
4. **`--staged --committed` together** — last flag silently wins; no conflict error.
5. **`--staged --fix`** fixes the working tree without re-staging, so the committed content can still fail.

---

**Summary:** Standards — 4 hard violations, 4 judgement calls; worst is unquoted word splitting over file lists in `lint-changed.sh`. Spec — 1 partial requirement, 5 implementation defects; worst is `--committed` resolving against a never-fetched `origin/main`, which breaks the "what a push would send" claim and can block a push on unrelated code.

edmonday:
## Re-review after `eeef9f9`

## Fix before merge

**1. `cd` guard is gone — `tools/scripts/lint-changed.sh:21`**
`cd "$(git rev-parse --show-toplevel)"`. If `rev-parse` fails, `cd ""` returns 0, so `set -e` does not fire. The script then lints from the wrong directory. Restore `|| exit 1`.

**2. PR description is stale**
It was written before `eeef9f9`. It now misses the `--committed` network fetch, the exit-2 on conflicting flags, and the `--staged --fix` re-stage caveat. The 10.6s hook timing was measured without a fetch in the path. This matters because the description is the only spec on record (ENG-3691 is not reachable from the review session).

## Worth fixing

**3. Output filename collision — `tools/scripts/lint-changed.sh:130,145`**
`tag=${ws//\//-}` maps `apps/a-b` and `apps/a/b` to the same name. Two parallel runs overwrite each other, and a lint failure can vanish so the push gate passes. No collision exists today. Use the loop index instead.

**4. `cat "$OUTDIR/$tag.out"` is unguarded — `tools/scripts/lint-changed.sh:146`**
Under `set -e`, a dead subshell means `cat` fails and the push is blocked with zero output — the same silent-block you just fixed elsewhere.

**5. `--committed` still is not "what a push would send"**
The hook ignores the remote name and refs husky passes on stdin, so a push to a fork or a second remote is judged against `origin/main`. Also `|| true` on the fetch hides an offline failure. Either narrow the claim in the docs or read stdin.

## Nits

- `AGENTS.md` prose is still one dense paragraph with passive voice — ASD-STE100 asks for one idea per sentence.
- `--committed --fix` writes fixes the push does not contain; only `--staged` carries a caveat.
- Parallelism is unbounded — a wide refactor spawns one type-aware TypeScript program per workspace at once.
- `git diff --name-only --diff-filter=ACMR` still repeats three times; `tag=` is computed in both loops.

## Closed since last round

Word splitting, `set -euo pipefail`, the BRE in `grep`, the doc drift in `AGENTS.md`, the bare `exit 1` message, and conflicting `--staged --committed` flags are all fixed. Skipping files with no workspace config is fine — nothing skipped is lintable under `nx lint` either.

csiyang:
Applied in `d9af0e41f`:

- **cd guard:** restored via assignment-first (`REPO_ROOT=$(git rev-parse --show-toplevel); cd "$REPO_ROOT"`) — `cd "$(...)" || exit 1` wouldn't fire since `cd ""` returns 0 in bash; a failed assignment under `set -e` does.
- **Collision + unguarded cat:** output files keyed by loop index + workspace tag; a worker with no result file now fails the run with an explicit message instead of silently passing (or blocking) the gate.
- **Parallelism:** capped at `LINT_CHANGED_JOBS` (default 4); verified serialization with `=1` (14.4s vs 7.0s parallel, identical findings).
- **PR description:** rewritten to match current behavior (fetch, flag conflicts, `--fix` caveat, new timings).
- **`--committed` semantics:** took the "narrow the claim" option — docs now state the gate lints the branch diff vs freshly fetched `origin/main`, not the push's ref-range; forks/second remotes are judged the same. stdin-refs felt like over-engineering while agents only push to `origin`. Offline-tolerant fetch kept: an offline push fails moments later regardless.
- **Nits:** `diff_names()` dedup and single tag computation done; `--fix` caveat now covers all modes. Kept the AGENTS.md paragraph's terse compound style — it's short sentences now, and full ASD-STE100 rewriting fights the file's existing voice.

edmonday:
## Re-review after `d9af0e4`

Almost there. One new blocker, and it is a small one.

## Fix before merge

**1. `LINT_CHANGED_JOBS=0` hangs the push forever — `tools/scripts/lint-changed.sh:133`**

```sh
while [ "$(jobs -rp | wc -l)" -ge "$MAX_JOBS" ]; do sleep 0.2; done
```

Tested: with `MAX_JOBS=0`, `0 -ge 0` is true even with no jobs running. The gate busy-waits forever with no output and no timeout. `0` is the natural way to ask for "no parallelism". Clamp to a minimum of 1.

**2. `AGENTS.md` now over-claims**

> it lints the branch diff against a freshly fetched `origin/main`

The fetch is `|| true` with stderr discarded, so offline the base is silently stale. Say "refreshed when the network allows".

## Worth fixing

**3. Branches cut from `stage` fail their own push**
The gate always uses `merge-base HEAD origin/main`. A branch off `stage` gets a base at `main`, so it lints everything `stage` carries ahead of `main` — code the agent never touched. Not handled, and the new contract paragraph covers forks and other remotes but not other base branches.

**4. Non-numeric `LINT_CHANGED_JOBS` spams a raw shell error**
`LINT_CHANGED_JOBS=abc` prints `[: abc: integer expression expected` once per workspace, then silently runs unbounded. Validate the value once, next to the clamp in item 1.

**5. `LINT_CHANGED_JOBS` is missing from `AGENTS.md`**
It is in the script header and the PR body, but not the doc agents actually read.

## Nits

- New `AGENTS.md` sentence is passive with two ideas around an em dash: "Pushes … are judged against `origin/main` all the same."
- `$i-${ws//\//-}` is now written out at three sites instead of two.
- PR body describes the same run as "~7-9s wall" in the table and "7.0s parallel" under Verified.
- `--staged --staged` reports "conflicting options" and exits 2. Harmless, slightly wrong.

## Closed this round

The `cd` guard (verified: a standalone assignment does trip `set -e` on bash 3.2), the output-filename collision, the unguarded `cat`, the silent `RAW=` failure, the `--fix` re-stage wording, and the duplicated `git diff` call via `diff_names()`. The job throttle itself works — 6 jobs at `MAX_JOBS=2` ran in 3.48s, and `jobs -rp` behaves inside a command substitution in a non-interactive script.

On the spec side the rewritten PR body now matches the code line by line, and the "parity with `nx lint`" claim checks out: no root ESLint config, no project globs root files, so CI does not lint `dangerfile.ts` or `checkly.config.ts` either. A dying worker does fail loudly rather than passing the gate.

csiyang:
Applied in `802aae81e`:

- **`LINT_CHANGED_JOBS` (your 1 + 4, CodeRabbit's inline):** validated as a positive integer immediately after arg parsing — `0`, negatives, and non-numeric values exit 2 with one clear message before any worker starts. Went with reject-over-clamp: someone writing `=0` means "no parallelism", but silently running serial when they may have meant "unbounded" felt worse than telling them.
- **AGENTS.md (your 2 + 5):** now says "refreshed when the network allows", documents `LINT_CHANGED_JOBS`, and covers `stage`-cut branches in the contract paragraph.
- **Stage-cut branches (your 3):** documented, not implemented — multi-base detection is real complexity for a flow agents don't use (branches come off `main`; `stage` is the bot-managed merge target), and the failure mode is over-linting with `--no-verify` as the documented out, never under-linting.
- **Nits:** repeated same-flag is now a no-op (only true conflicts exit 2); `tag` computed once; PR body timing figures reconciled (7.0s).

edmonday (approved):
## Re-review after `802aae8` — LGTM

Clean. Nothing left that blocks a merge.

## All round-3 items closed

`LINT_CHANGED_JOBS` validation was tested against `0`, `abc`, `-1`, `4.5`, `1e3`, `0x10`, `1;echo pwned`, `4 4`, whitespace, and a 20-digit value — every one exits 2 with one clear message before any work starts. The passing edge values (`" 4 "`, `08`, `+4`, max int64) all reach the throttle and exit it correctly, so the infinite busy-wait is gone in every case. `--bogus` still exits 2 first, so the argument contract is unchanged.

`AGENTS.md` now says "refreshed when the network allows", which matches the best-effort fetch. `LINT_CHANGED_JOBS` is documented there with the right default. Repeated identical flags pass; genuinely conflicting ones still exit 2. The PR body timings are now consistent.

Documenting the `stage`-cut branch case rather than fixing it is the right call: `stage` is rebuilt from `main` by `tools/scripts/reset-stage.sh`, so cutting work from it is already an anti-pattern, and recent branches are all `user/ticket-*`.

## Optional polish

- The `stage` remedy in `AGENTS.md` leads with `git push --no-verify`, which teaches gate bypass. "Rebase onto `main`" is the better first answer, with `--no-verify` as fallback.
- The PR body's What section still says `--committed` uses a "freshly fetched" `origin/main`, while its own Contract section and `AGENTS.md` both say best-effort. Soften for consistency.
- That third `AGENTS.md` sentence is 28 words, passive, and carries three ideas plus a parenthetical — the only ASD-STE100 miss left.
- `tag="$i-${ws//\//-}"` is computed in both loops. Legible as-is.
- `LINT_CHANGED_JOBS=99999999999999999999` reports "must be a positive integer" when the real reason is bash 3.2 range. Rejecting is right; the wording is slightly off.
- Verified section does not yet list the new rejection path or the repeated-flag case. One line each.

None of these need code changes.

PR #9482, opened 2026-08-13
feat(api-languages): add operator-owned searchable flag to Language (QA-375)

### LINKED ISSUE #9474: Language.hasVideos is overloaded: operator suppression cannot be made to stick
## Summary

`Language.hasVideos` is overloaded. It is both a derived fact ("this language has video content") and the operator's visibility control ("hide this language"). There is no other flag on the model:

```prisma
model Language {
  id        String  @id
  hasVideos Boolean @default(true)
  // ...no visibility/publish flag
}
```

Per @tanflem, disabling a language is done by hand-editing `hasVideos` to `false`. **That decision cannot currently be made to stick** — two automated writers overturn it.

## What `hasVideos` gates

Suppression is not search-only. It hides a language across the public languages API too:

| Surface | Location |
|---|---|
| Algolia languages index | `apis/api-media/src/lib/languages/updateLanguageInAlgolia.ts` |
| `languages` query (default filter) | `apis/api-languages/src/schema/language/language.ts:182` |
| `languagesCount` (default filter) | `apis/api-languages/src/schema/language/language.ts:216` |
| `Country.countryLanguages` + `languageHavingMediaCount` | `apis/api-languages/src/schema/country/country.ts:50,61` |

## What overturns it

**1. `ensureLanguageHasVideosTrue` — automatic, high frequency.**
Flips `false -> true` and pushes to Algolia on every variant write. Called from `apis/api-media/src/schema/videoVariant/videoVariant.ts:541` (`videoVariantCreate`) and `:723` (`videoVariantUpdate`). Any edit to any variant of a suppressed language un-hides it everywhere in the table above.

**2. `wess-languages-import` — manual, occasional.**
`apis/api-languages/src/scripts/wess-languages-import.ts:186` upserts `update: { hasVideos: row.hasVideos }` whenever the feed supplies a value, so an import re-enables whatever the feed considers to have videos. Run via `nx run api-languages:wess-languages-import`; not scheduled.

**3. Humans**, by hand. The intended mechanism.

## Why the flip cannot simply be deleted

The same import **creates** languages with `hasVideos: row.hasVideos ?? false` (`:172`). So there is a large, legitimate population of `hasVideos: false` languages that genuinely have no content yet and *should* be promoted automatically when they gain a variant — roughly 89,965 of the 92,285 languages in prod are `hasVideos: false` with zero variants. Removing the auto-promote would strand all of them.

The two meanings genuinely need separating; there is no way to infer intent from the single boolean.

## Current prod state

7 languages have published variants but `hasVideos: false`. Two are still in `video-languages-prd` and searchable:

| objectID | name | published variants | in index? |
|---|---|---|---|
| 20525 | Twi | 142 | **yes — searchable** |
| 117014 | Dari (فارسی) | 2 | **yes — searchable** |
| 12104 | Guhu-Samane | 75 | no |
| 169877 | Savara Sora | 62 | no |
| 2993 | Pennsylvania Dutch | 62 | no |
| 19861 | Psikye | 62 | no |
| 139769 | Blang | 62 | no |

## Proposed solution

Add an operator-owned field that no automated process writes:

```prisma
model Language {
  hasVideos  Boolean @default(true)   // derived; variant hook + WESS may write
  searchable Boolean @default(true)   // operator-owned; nothing automated writes it
}
```

- Visibility becomes `hasVideos && searchable` at all four surfaces above.
- `hasVideos` goes back to being purely derived, so the auto-promote and WESS stay as they are.
- Data migration: the 7 languages above get `searchable = false`. Their `hasVideos` can then be corrected to `true` by the normal path, since they do in fact have videos.
- The ~89,965 `hasVideos: false` languages with no variants keep `searchable = true` and stay hidden by `hasVideos` alone, correctly.

### Decisions needed before implementing

1. **Name and polarity** — `searchable` (default `true`) vs `hidden` / `suppressed` (default `false`).
2. **Scope** — should all four surfaces respect it, or Algolia only to start? Operators currently expect "disabled" to mean disabled everywhere, which argues for all four.
3. **Affordance** — expose it on the admin languages query and add a videos-admin toggle, or keep it a DB edit for now (but a safe one that nothing overwrites)?

## Related

PR #9473 adds the delete path so that a suppressed language is removed from Algolia rather than left stranded. That is necessary but not sufficient: until the field above exists, clearing the two prod records only lasts until the next variant write on those languages. **The prod cleanup should wait for this.**

tanflem:
Closes #9474

> **Note:** this PR targets `tannerfleming/qa-375-algolia-language-id-type` (the head of #9473), not `main`. GitHub only honours closing keywords on PRs that target the default branch, so the link above will **not** auto-close #9474 as written — retarget this PR to `main` once #9473 merges, or close #9474 by hand.


tanflem:
**Stacked on #9473** — base is `tannerfleming/qa-375-algolia-language-id-type`, so this diff shows only the new work. Merge #9473 first and this will retarget to `main`.

Closes the durability half of #9474.

### Why a new column rather than fixing the flip

`hasVideos` had two jobs and could not keep the second one. Two automated writers overturn an operator's decision to hide a language:

- `ensureLanguageHasVideosTrue` flips `false -> true` on every variant write (`videoVariant.ts:541`, `:723`)
- `wess-languages-import` writes `hasVideos` whenever the feed supplies it (`wess-languages-import.ts:186`)

Deleting the flip is not an option: the same import **creates** languages with `hasVideos: row.hasVideos ?? false` (`:172`), so ~89,965 of prod's 92,285 languages are legitimately `false` with no content and must still auto-promote when they gain a variant. There is no way to infer intent from one boolean.

`searchable` (default `true`) is operator-owned; nothing automated writes it. `hasVideos` goes back to being purely derived, so the variant hook and the import are unchanged — only comments were added there marking `searchable` as off-limits.

### Visibility is now `hasVideos && searchable`

Applied at all four surfaces that gate on the flag:

| Surface | File |
|---|---|
| Algolia index (sync + reindex) | `updateLanguageInAlgolia.ts` |
| `languages` default filter | `language.ts:182` |
| `languagesCount` default filter | `language.ts:216` |
| `Country.countryLanguages`, `languageHavingMediaCount` | `country.ts:50,61` |

Admin queries deliberately keep showing hidden languages — that assertion is unchanged on purpose in `language.spec.ts`.

### Verification

- Migration applied to a scratch Postgres by replaying every languages migration in order: `searchable` lands `NOT NULL DEFAULT true`, composite index `Language_hasVideos_searchable_idx` created. Prisma 7's `migrate diff` silently no-ops in this environment, hence the direct check.
- Generated GraphQL schema is **unchanged** — the field is intentionally not exposed yet.
- api-media 787 tests, api-languages 113 tests, lint clean on all three projects, type-check clean.

### Deliberately not included

- **No GraphQL exposure / admin toggle.** Setting the flag stays a DB edit for now, but a durable one. Worth a follow-up ticket so operators can see and set it without SQL — right now they cannot even tell which languages are hidden.
- **No prod data change.** `api-media:backfill-language-searchable` is added but defaults to a dry run; it marks languages that have content but `hasVideos: false` as `searchable: false`, preserving today's visibility exactly. It needs a deliberate `--apply` run after this merges.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 6 checks green, no unresolved review threads. +247/-50 across 17 files, adding an operator-owned `searchable` flag to `Language` (QA-375). Reviewers already requested: `@Kneesal`, `@tataihono`, `@mikeallisonJS`.

### What needs to happen next

Ready for review. Pairs with #9473 on QA-375 — worth telling the reviewers they should land together.

PR #9484, opened 2026-08-13
chore(deps): upgrade Vitest 3→4 and Cloudflare workers pool (#9376)

mikeallisonJS:
## What

The **Vitest slice** of the #9376 major-upgrade roadmap.

> **Vitest (lockstep):** `vitest`, `@vitest/coverage-v8|ui|snapshot`, `@vitejs/plugin-react`, `vitest-mock-extended` (2 → 4). Worth doing before adding more specs given the active jest→vitest migration.

#9415 (Nx) took this as far as a "minimal Vitest 3 bump" and left the major for its own PR. Siblings already shipped: dead-dep prune (#9377), commitlint (#9418), express (#9423), firebase-admin (#9424), firebase client (#9426), Datadog (#9428), AI SDK (#9436), react-email (#9441), MUI (#9445). Cut from `main`, independent of the open one-off sweep (#9446).

| Package | From | To |
|---|---|---|
| `vitest`, `@vitest/coverage-v8`, `@vitest/snapshot`, `@vitest/ui` | `3.2.7` | `4.1.10` |
| `vitest-mock-extended` | `^2.0.2` | `^5.1.1` |
| `@cloudflare/vitest-pool-workers` | `^0.7.1` | `^0.21.3` |

`vitest-mock-extended` peers on `vitest >=4.0.0` and the Cloudflare pool on `vitest ^4.1.0`, so both move in lockstep rather than as separate bumps. `@cloudflare/vitest-pool-workers@0.7.1` already peered on `vitest 2.0.x - 3.0.x` while the repo sat on 3.2.7, so it was out of range before this PR too.

**Zero product behaviour change.** I captured a full pass/fail baseline on Vitest 3 before touching anything; test counts match it exactly in all 22 Vitest workspaces (618/3769 in journeys-admin, 184/1478 in journeys-ui, 163/1101 in api-journeys, …).

### Constructor mocks must be `function`, not an arrow

Vitest 4 spies support `new`. An arrow has no `[[Construct]]`, so any mock standing in for a constructor now throws `... is not a constructor` — usually at *import* time, since these live in `vi.mock` factories, which fails the whole suite rather than one test.

```ts
// ✗ Vitest 4: throws where the subject calls `new Queue(...)`
Queue: vi.fn().mockImplementation(() => ({ add: vi.fn() }))

// ✓
Queue: vi.fn(function () {
  return { add: vi.fn() }
})
```

~20 sites converted: bullmq `Queue`/`QueueEvents`, `Mux`, `Cloudflare`, `S3Client` and the AWS `*Command` objects, `Redis`, `ApolloClient`, `ShortUniqueId`, firebase `GoogleAuthProvider`/`FacebookAuthProvider`/`OAuthProvider`, and the browser globals `IntersectionObserver`, `TextTrackList` and `Blob`.

One shared file — `apis/api-media/test/bullmqMock.ts` — accounted for **37 of the 40 failing api-media suites** on its own.

### Coverage reporting would have silently degraded

The subtlest part, and the one worth reviewing closely. Vitest 4 removed `coverage.all` and `coverage.extensions`, and its default `coverage.exclude` is now `[]` (verified against `coverageConfigDefaults`; Vitest 3 shipped 17 entries).

Left alone, the cobertura reports Codecov consumes would have changed in two directions at once:

- **Untested source disappears.** Without `coverage.all`, a report covers only files some test happened to import. `libs/yoga` dropped from 47 files to 6 — the percentage jumps for no real reason and untested code stops being visible.
- **Non-source starts counting.** With the exclude default emptied, `.d.ts` files, build output and `test/` helpers get reported as source. `apps/short-links` picked up `next-env.d.ts`/`index.d.ts`; `apis/api-media` picked up its `test/*Mock.ts` helpers.

Every coverage-enabled config now carries an explicit `include`/`exclude` reproducing the Vitest 3 report shape, verified file-by-file (`libs/yoga` 47, `apps/short-links` 8, `apis/api-media` 313, with zero `.d.ts`/`test/`/spec/`dist` entries across all 20 reports).

One trap for anyone porting more of Vitest 3's defaults: its `**/\0*` glob for rollup virtual-module ids matches **every** path under Vitest 4's matcher and empties the report entirely. `**/virtual:*` and `**/__x00__*` cover those ids instead.

### `vitest.workspace.ts` removed

Vitest 4 removed the `workspace` option outright. The file was already unreferenced — added with the jf-proxy worker in #5350 and never wired up.

### `workers/jf-proxy` — Cloudflare pool 0.13 rewrite

Pool 0.13 (the release that added Vitest 4 support) made three changes that hit this worker:

- `defineWorkersConfig` from `/config` → a `cloudflareTest()` Vite plugin, with `poolOptions.workers` passed directly to it.
- `fetchMock` removed from `cloudflare:test`; upstream now points at stubbing `globalThis.fetch` or using MSW.
- `isolatedStorage` removed — storage isolation is per test *file* now, not per test. The spec's cache assertions already use distinct keys per test, so nothing changed there.

Rather than rewrite 45 assertions against a raw `fetch` stub, `workers/jf-proxy/test/fetchMock.ts` wraps `vi.spyOn(globalThis, 'fetch')` in the one-shot-interceptor API the spec already used (origin + path + method matching, undici's order-insensitive query comparison, `assertNoPendingInterceptors`). **The spec diff is a one-line import change**, so the 45 tests still assert exactly what they asserted before.

The config also had to be renamed `.ts` → `.mts`: the plugin is ESM-only, and the old config was CJS-loaded, so it could not `require` it.

## Not included: `@vitejs/plugin-react` 4 → 5

The one member of the roadmap's Vitest family I left behind, deliberately.

v5 ships exports-only — no `main`, no `types`, just `exports["."] → ./dist/index.js`. The repo's `tsconfig.base.json` uses `moduleResolution: "node"`, which ignores `exports`, so `react()` resolves as an error-typed call and fails `@typescript-eslint/no-unsafe-call` in every `vitest.config.mts` that uses it.

I confirmed this reproduces against **unmodified, pristine** config files, so it is the bump and not anything in this PR. It also surfaced intermittently at first, which cost me a couple of wrong turns — `nx reset` is needed for trustworthy lint results after `pnpm install` swaps `node_modules`.

Fixing it properly means moving `moduleResolution` off `node`, which belongs with a Vite upgrade rather than here. Vitest 4 does not require it — nothing in this PR depends on it.

## Incidental

- **`apps/watch/pages/api/blurhash.spec.ts` was already failing on `main`** (10 tests — jsdom's `AbortSignal` is not undici's, so `fetch` rejected). It passes on Vitest 4; the module-runner rewrite resolves the identity mismatch. Not fixed by me, just no longer broken.
- `docs/agents/testing.md` refreshed: the workspace table listed 10 of the 22 configs and omitted `workers/` and `tools/` entirely. Added the `--root` pitfall (Vitest's `root` defaults to the shell CWD, so running a whole workspace from the repo root sweeps the entire monorepo — this bit me while collecting the baseline) and a Vitest 4 notes section carrying the rationale above.
- The CI shard-count comment in `main.yml` said "Vitest 3"; I verified the underlying limitation still applies on 4 and reworded rather than removing the workaround.

## Verification

| Gate | Result |
|---|---|
| `nx run-many --target=type-check --all` | 39/39 |
| `nx run-many --target=lint --all` | 40/40 |
| All 22 Vitest workspaces, coverage enabled | green, counts match the Vitest 3 baseline |
| `nx test` executor path (`--ci --max-workers N --shard=X/3`) | green, incl. the Cloudflare workers pool |

The executor path is checked separately because my per-workspace runs used the Vitest CLI, and CI goes through `@nx/vitest:test` — which calls `startVitest`/`parseCLI` directly and could have broken independently.

Two pre-existing local-environment notes for reviewers: `prisma-media`/`api-media` type-check needs `nx run prisma-media:prisma-generate` first (stale generated client, unrelated), and the gitignored `next-env.d.ts` files fail lint if a Next build has run locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored comprehensive coverage reporting, including untested source files while excluding generated, dependency, test, and configuration files.
  - Improved Vitest 4 test execution, sharding reliability, Cloudflare Workers testing, and constructor-based service mocks.

- **Documentation**
  - Updated testing guidance for coverage, workspace configuration, constructor mocks, and Workers test setup.

- **Chores**
  - Upgraded testing tools and removed obsolete workspace configuration.
  - Standardized test configuration across applications and shared libraries.


csiyang (dismissed):
Exceptionally well-documented upgrade — every claim I spot-checked held up:

- **Constructor mocks**: all conversions are mechanically correct (`vi.fn(function () { return obj })` under `new` returns `obj` by constructor-return-object semantics). The one conversion that dropped a parameter — `useIsInViewport.spec.tsx` losing `(callback)` — is safe: the tests recover the callback via `mock.calls[0]`, never from the implementation.
- **jf-proxy `test/fetchMock.ts`** faithfully reproduces the undici subset the spec uses: origin normalization, order-insensitive query matching, one-shot consumption, and unmatched requests throwing (preserving the old `disableNetConnect` guarantee). The one divergence — `assertNoPendingInterceptors` clearing the queue before throwing — is an improvement in `afterEach`, since it prevents cascade failures.
- **Coverage**: the restored exclude list is a verified subset of Vitest 3's defaults, and the `**/\0*` virtual-module trap is documented rather than ported.

One question, non-blocking:

**Was the 20× duplication of the coverage `include`/`exclude` block a deliberate Nx-cache-correctness choice?** The obvious dedup — a shared file imported by every config — is subtly unsafe as-is: `@nx/vitest:test` inputs in `nx.json` are `{projectRoot}/**/*` + `sharedGlobals` (only `babel.config.json`), so a repo-root shared file would not be hashed and editing it would leave stale test caches. If the inlining was deliberate for that reason, a line saying so in `docs/agents/testing.md` (next to "keep the two lists identical") would stop a future cleanup PR from walking into it. If not deliberate, a shared file + a `{workspaceRoot}/...` entry in the vitest target's inputs would remove the hand-sync rule. Either resolution is fine by me.

mikeallisonJS:
## Review feedback addressed (01bd4e2a6)

**Fixed:**

- `apis/api-media/src/workers/dataExport/service/service.spec.ts:39` — S3 command mocks now assign onto `this` instead of returning `args`, so `command instanceof HeadObjectCommand` matches.

CodeRabbit was right, and the consequence was slightly worse than stated: a constructor that returns an object discards the instance, so **both** `mockS3Send` branches fell through to `Promise.resolve({})`. `should handle case when file does not exist in R2` was exercising the file-*exists* path and never reached `backupExistingFile`'s catch branch.

Two notes beyond the finding:

- **Pre-existing, not introduced here.** On `main` the mock was `vi.fn().mockImplementation((args) => args)`, which loses identity the same way; the `function`-form conversion in this PR preserved the existing behaviour exactly. The Vitest 4 constructor-mock work surfaced it rather than caused it.
- **Identity alone doesn't finish the job.** The test's existing assertions (`HeadObjectCommand` called, `PutObjectCommand` called) hold true in *both* branches, so it would still pass if the fix regressed. I added the two that actually distinguish them — the `No existing file found at` log, and `CopyObjectCommand` not being called. Both fail when run against the old mock, which is the check that makes this meaningful rather than cosmetic.

`S3Client`'s mock still returns a replacement object deliberately: nothing does `instanceof S3Client`, the service only calls `.send()`, so identity is irrelevant there and returning the object is what makes it work.

Verified: dataExport spec 4/4, api-media 78 files / 779 tests, lint and type-check green.

**Challenged / skipped:** none — this was the only unresolved thread.

csiyang (approved):

PR #9485, opened 2026-08-13
test(journeys-admin-e2e): pin Publisher empty-state copy casing (NES-1876)

sharonemathew:
[NES-1876](https://linear.app/jesus-film-project/issue/NES-1876/add-e2e-coverage-for-journeys-admin-empty-state-and-bulk-action-copy)

The three Publisher empty-state assertions used string `hasText`, which Playwright matches case-insensitively. `No Templates to display.` and `no templates TO DISPLAY` both passed, so a casing regression couldn't be caught.

Two helper strings had no assertion at all.

## Fix

One file: `apps/journeys-admin-e2e/src/pages/publisher-and-templates-page.ts`

Three assertions moved from `hasText` to `getByRole('heading', { name, exact: true })`:

- `No Templates to display.`
- `No archived Templates.`
- `Your trashed Templates will appear here.`

`exact: true` is what pins the casing — `getByRole` alone is also case-insensitive by default.

Added two previously unasserted helper strings:

- `Archived Templates are delisted from the Template Library.`
- `Trashed Templates are moved here for up to 40 days.`

## Test plan

- [x] `tsc --noEmit` — clean
- [x] `nx run journeys-admin-e2e:lint` — 0 errors
- [x] `prettier --check` — passes

## Heads up

**These assertions don't run yet.** `publisher-and-templates.spec.ts` is `test.describe.fixme` throughout, so the file is skipped and CI goes green without executing any of this. It's prep — the casing is pinned for when the suite is re-enabled.

So NES-1876's "lowercase it and watch it go red" and "green 3× in a row" aren't met here, and can't be by an e2e-only change.

Not in scope:

- The 18 bulk-action dialog bodies — only reachable via skipped specs.
- The three discover-page locators in `journey-page.ts` — that page no longer has status tab panels, so the selectors match nothing. Likely why `active-archived-trash.spec.ts` is skipped.

Follow-up ticket to re-enable the skipped suites to come.

## Summary by CodeRabbit

* **Tests**
  * Improved verification of empty states across active, archived, and trashed template views.
  * Added checks for explanatory text in archived and trashed states.
  * Confirmed that trashed views contain no template cards.

## Summary by CodeRabbit

* **Tests**
  * Improved verification of empty states across active, archived, and trashed template views.
  * Added checks for explanatory text in archived and trashed views.
  * Continued validating that the trashed view contains no template cards.


csiyang (dismissed):
Traced every assertion through to the rendered DOM and they all check out:

- Strings match `ActiveTemplateList`/`ArchivedTemplateList`/`TrashedTemplateList` on `main` verbatim, punctuation included.
- Roles are right: the empty-state messages are `Typography variant="subtitle1"` → `<h6>` → `heading` role, and the helper strings are `variant="caption" component="div"`, so `getByText` (not `getByRole`) is the correct choice there.
- The container selectors still resolve: `publisher/index.tsx` → `TemplateList` → `StatusTabPanel` → shared `TabPanel`, which emits `id="{name}-tabpanel-{index}"` / `aria-labelledby="{name}-tab-{index}"`, matching the `*-status-panel-tabpanel` / `*-status-panel-tab` substrings.
- Dropping the `h6` CSS selector for role-based locators is what `docs/agents/e2e-testing.md` asks for.

Understood that the spec is `test.describe.fixme` so this is prep that doesn't execute yet — appreciate the upfront heads-up in the description. Optional nit for the follow-up that re-enables the suite: the outer `div[id*="...-status-panel-tabpanel"]` containers could become `getByRole('tabpanel', { name: 'Archived' })` etc., completing the role-based migration.

sharonemathew:
refactor(journeys-admin-e2e): scope Publisher assertions by tabpanel role (NES-1876)

Addresses @csiyang's review nit. The three empty-state checks still located
their container by raw markup (`div[id*="archived-status-panel-tabpanel"]`),
so only half of each assertion was role-based. Switch to
`getByRole('tabpanel', { name })`, completing the migration
docs/agents/e2e-testing.md asks for.

Two side benefits: inactive panels carry the `hidden` attribute, so
`getByRole` skips them and can only ever resolve the panel on screen; and the
selector now reads as the tab the user clicked.

Note the trashed panel's accessible name is 'Trash', not 'Trashed' — it comes
from the tab label in StatusTabPanel, not the panel's own id.

Also drop a duplicated selector in verifyAllTemplateAreDeletedFromTrashTab,
which repeated the container path instead of reusing the extracted locator.

csiyang (approved):

PR #9490, opened 2026-08-14
chore: remove deprecated watch and watch-modern apps

mikeallisonJS:
## Summary

`apps/resources` is now the sole owner of the `/watch` surface. This removes the deprecated `apps/watch` and `apps/watch-modern` apps, their Playwright suites (`watch-e2e`, `watch-modern-e2e`), and the watch PRDs, along with every reference pointing at them. 615 files, ~56k lines, and 4 projects out of the Nx graph.

## Two cross-app references that would have broken

Both were pre-existing copy-paste bugs reaching from a live app into `apps/watch`, so the deletion surfaced them rather than caused them:

| File | Was | Now |
|---|---|---|
| `apps/resources/src/components/Videos/testData.generator.ts` | Read and wrote `apps/watch`'s generated `testData.ts` | Reads and writes the resources copy that already existed |
| `apps/player/project.json` | `serve-all` ran `nx serve watch-modern` | `nx serve player`, matching every other app's `serve-all` |

The generator one meant `nx generate-test-data resources` was regenerating watch's fixtures instead of its own. Its `.prettierignore` negation moved with it.

## Everything else

- `tools/scripts/codegen.sh`: ran `nx codegen watch`, which errors with the project deleted
- `.github/workflows/`: the e2e choices and matrix apps, plus all stale `WATCH_*` env plumbing. `WATCH_VERCEL_PROJECT_ID` / `WATCH_MODERN_VERCEL_PROJECT_ID` in `e2e-tests.yml` and `app-deploy.yml`, and `DOPPLER_WATCH_TOKEN` / `DOPPLER_WATCH_MODERN_TOKEN` / `DOPPLER_WATCH_ADMIN_TOKEN` in `main.yml` and the 4 `ecs-frontend-deploy-*` workflows (caller and reusable-workflow sides together, so the `secrets:` contracts stay matched). No `WATCH` references remain under `.github/`
- `.storybook/main.ts`: the watch story globs and spread. Watch was never in `libs/shared/storybook`'s `implicitDependencies`, so there was nothing to update on that side
- `.gitignore`: the watch-modern i18n locales rule
- `CONTEXT-MAP.md`: both app entries, 5 relationship bullets, and the deploy / GQL-consumer / storybook-drift lists
- `apps/CONTEXT-e2e.md` (11 suites down to 9), `docs/agents/testing.md`, `docs/ai-foundations.md`, and the review-pr command and skill

## Action required outside this repo

Merging this does not take `/watch` down by itself, but three things need repointing or retiring:

1. **JF Proxy `WATCH_PROXY_DEST`** still routes `/watch*` to the Watch deployment. Production `/watch` will 404 once those Vercel deployments go away, unless it is repointed at Resources first. This needs to land *before* the Vercel projects are deleted. The worker code needs no change, since the destination is an opaque per-environment hostname.
2. The **`watch` and `watch-modern` Vercel projects**.
3. The GitHub secrets themselves: **`WATCH_VERCEL_PROJECT_ID`**, **`WATCH_MODERN_VERCEL_PROJECT_ID`**, **`DOPPLER_WATCH_TOKEN`**, **`DOPPLER_WATCH_MODERN_TOKEN`**, **`DOPPLER_WATCH_ADMIN_TOKEN`**. Nothing references them after this PR, but the secrets still exist.

## Deliberately left alone

`libs/shared/ui-modern` now has zero importers, since watch was its only consumer. It is not dead code though: `apps/resources` vendors its own copies of 4 shadcn components (Select, Dialog, Accordion, Button) plus a byte-identical `cn` helper, so the real situation is duplication, not disuse. The lib stays and the context map records that. Pointing resources at it is separate work.

Historical entries under `docs/plans/` and `docs/solutions/` keep their stale `apps/watch` references. They are archives of past work, not live docs.

## Verification

Rebased onto the Vitest 4 upgrade (#9484), which had just added a `vitest.config.mts` to both deleted apps and rewrote the `docs/agents/testing.md` workspace table. Those conflicts resolved to the deletion, and the doc table keeps main's new area-grouped form minus the 2 watch entries.

`nx show projects` resolves cleanly, the 3 remaining `codegen.sh` targets all resolve, `pnpm lint:changed` passes, Prettier passes on every edited file, and the full `apps/resources` suite is green (607 passed, 28 skipped).

---

Claude Code


csiyang (commented):
Swept the whole tree on this branch for references to the deleted projects (excluding the deliberately-kept `docs/plans`/`docs/solutions`/`prds` archives). The deletion is thorough — the two cross-app fixes check out (the generator now reads/writes resources' own `testData.ts`, and player's `serve-all` matches every other app), `app-deploy.yml`'s matrix is dynamic so deploys drop out automatically, and the context-map/testing-doc rewrites are accurate. Two leftovers, one real:

**Concern — `tools/scripts/codegen.sh` line 3 still runs `nx codegen watch`.** With the project deleted this now errors ("Cannot find project 'watch'"). No CI or package.json caller — it's a manual helper — but it's exactly the reference class this PR exists to remove. One-line deletion. (The file isn't in the diff, so flagging here rather than inline.)

**Nit — stale `WATCH_*` env plumbing in untouched workflows.** `app-deploy.yml` still passes `WATCH_VERCEL_PROJECT_ID`/`WATCH_MODERN_VERCEL_PROJECT_ID` in two places (this PR cleaned the identical lines in `e2e-tests.yml`), and the four `ecs-frontend-deploy-*` workflows declare/pass `DOPPLER_WATCH_TOKEN`/`DOPPLER_WATCH_MODERN_TOKEN`/`DOPPLER_WATCH_ADMIN_TOKEN` even though only arclight deploys through them. All dead-but-harmless today, but since this PR's action list includes deleting those GitHub secrets, the references should go either here or in that follow-up.

Also noting for the record: the failed resources Deploy Preview is a prerender error on `/watch/[part1]` (`/es/watch/spanish-castilian.html` — a data fetch returned HTML instead of JSON during SSG). This PR's only resources build inputs are a comment line in generated test data and generator file paths, neither imported by pages, so it looks like an upstream API flake during static generation rather than this diff — but it needs a green re-run before merge.

mikeallisonJS:
## Review feedback addressed (c43c7f5f9)

Thanks @csiyang, both leftovers were real. Fixed the Concern and took the nit further than flagged.

**Fixed:**
- `tools/scripts/codegen.sh:3` — dropped `nx codegen watch`. Good catch, and squarely my miss: this line showed up in my initial sweep and I didn't act on it. Confirmed the remaining three targets (`journeys`, `journeys-admin`, `journeys-ui`) all still resolve.

**Fixed (adjusted — wider than flagged):**
- The `WATCH_*` env plumbing sweep found one file beyond the five you named. `main.yml:58-60` carries the same three `DOPPLER_WATCH_*` tokens as the ecs workflows, so stopping at the named files would have left a follow-up sweep with hits. Cleaned all 6:

| File | Removed |
|---|---|
| `app-deploy.yml` | `WATCH_VERCEL_PROJECT_ID`, `WATCH_MODERN_VERCEL_PROJECT_ID` (both preview and prod jobs) |
| `main.yml` | the 3 `DOPPLER_WATCH_*` tokens |
| `ecs-frontend-deploy-{prod,stage}.yml` | the 3 `DOPPLER_WATCH_*` pass-throughs |
| `ecs-frontend-deploy-{prod,stage}-worker.yml` | the 3 `DOPPLER_WATCH_*` `required: true` declarations and their pass-throughs |

  Caller and reusable-workflow sides were removed together so the `secrets:` contracts stay matched. `grep -rn WATCH .github/workflows/` now returns nothing.

  Note `DOPPLER_WATCH_ADMIN_TOKEN` was already dead before this PR — no `watch-admin` project has ever existed in this repo. Removed it anyway since it's the same class and same lines.

**Noted, no action:**
- The resources Deploy Preview prerender error on `/watch/[part1]` — your read was right, and it has since gone green. All checks currently pass, zero failing, zero pending. Agreed it was an upstream flake during SSG rather than this diff.

The action list in the PR body still stands: deleting the GitHub secrets themselves, the two Vercel projects, and repointing JF Proxy's `WATCH_PROXY_DEST` at Resources are all outside the repo. The `WATCH_PROXY_DEST` repoint should land *before* the Vercel projects are deleted, or production `/watch` 404s in the gap.

csiyang (approved):
Both items from my review are addressed in c43c7f5f, and the sweep went further than flagged — thanks for catching `main.yml`, which my own grep missed.

Independently verified the delicate part of the workflow cleanup: since the reusable `ecs-frontend-deploy-{prod,stage}-worker.yml` workflows had their `secrets:` contract shrunk, I diffed each worker's `required: true` list against what its caller passes — zero mismatches on both sides. Also confirmed nothing outside `.github/` (no `project.json` `fetch-secrets` target etc.) reads `DOPPLER_WATCH_*`, so no runtime consumer is left holding a now-absent env var. `git grep WATCH .github/` returns nothing on this branch.

Deploy preview is green on re-run and every check passes, confirming the earlier prerender failure was an SSG flake rather than this diff.

Endorsing the sequencing note for whoever executes the outside-repo action list: repoint JF Proxy's `WATCH_PROXY_DEST` at Resources *before* deleting the watch Vercel projects, or production `/watch` 404s in the gap.

Issue #9468, opened 2026-08-11
Phase 1: Catalog-wide parent-language audit and repair tooling


tanflem:
> *This was generated by AI during triage.*

Part of #9381 — **phase 1 of 3**.

Blocked by #9382, which delivers the narrow, single-container version of this same repair. Build on its `parentVariantsOnly` behaviour; do not reimplement parent-Variant creation a third time.

## Context

A qualifying published Child Video with at least one published Variant in a language requires a corresponding generated parent Variant for every parent relationship. One qualifying Child is sufficient — not every Child must carry that language. Today this invariant is maintained only as a best-effort side effect of publishing, individual parent failures are swallowed, and there is no way to ask the catalog where it is violated.

QA-554 is the reported instance: on the Do You Ever Wonder series, Kurmanji Standard (`20770`), Tajik, and Pashto Eastern Afghan should each expose eight published Episodes, but parent-language state and downstream indexes do not reflect the Child Variants that exist.

## Scope

An idempotent audit and repair tool that scans every Video relationship across the whole catalog.

- **Dry run is the default.** Apply mode requires an explicit option and repairs only deterministic gaps.
- **Discovery is relationship-based, not Label-based** — series, collections, feature-film containers, and any future container Label are covered whenever they own Children.
- The invariant applies whether the parent is Draft or published; parent state must be internally consistent before any future publication.
- Coverage spans every reconstructable processing stage: Variant validity, Mux readiness where applicable, required parent Variants, required Downloads, and both Video and Variant Algolia records.
- Unverifiable state is reported as `unknown`. The audit never infers success.
- Parent repair atomically creates or normalizes the generated Variant and updates the parent Video's available languages; Algolia updates follow the database transaction.
- **A repair is not complete until it is indexed.** Index failures are retained as incomplete work, reported, and produce a non-zero result, while safe database changes remain in place for idempotent retry.
- Generated parent Variant identity stays deterministic per parent and language, consistent with the existing uniqueness invariant.
- Apply mode may normalize an existing Variant **only** when it is clearly a generated parent Variant. Any Variant with real media indicators — Mux association, stream URLs, duration, or Downloads — is ambiguous and must be reported without mutation.
- Repair output identifies each Video, parent, language, Variant, action, and result.

Run it in production against Do You Ever Wonder first, then catalog-wide.

## Acceptance criteria

- [ ] A global dry run reports proposed and ambiguous changes without writing.
[...]

PR #9491, opened 2026-08-14
feat(api-media): add parentVariantsOnly publish mode and download-size backfill

### LINKED ISSUE #9382: Repair zero-size Downloads and recover missing parent language Variants
> **Scope update — 2026-08-11 (triage).** The *prevention* half of the reported incident has already shipped. `f6a209445` ("fix(api-media): stop persisting zero size/bitrate Mux download renditions", VMT-239) stops `createDownloadsFromMuxAsset` from writing zero-value `size`/`bitrate` when Mux marks a static rendition `ready` before its metadata has propagated, applies the same guard to the sd/high fallback-resolution lookup, and extends the `mux-videos` backfill script to re-fetch already-corrupted zero-value rows.
>
> **What remains open in this issue is unchanged in substance but narrower in framing:**
>
> 1. The `parentVariantsOnly` mutation mode and its operator command (`6_Acts` recovery) — not started.
> 2. The standalone, global Download-size backfill command covering **Mux, R2, and legacy URLs** — not started. The `mux-videos` script fix is Mux-only, opportunistic, and not a bounded/resumable/audited global repair, so it does not satisfy this deliverable.
>
> New zero-size rows should no longer appear from the Mux path; this issue is now about repairing the **existing** backlog and recovering parent language Variants. Verified against `origin/main` at triage time.

## Problem Statement

Publishers and downstream Arclight consumers cannot rely on every existing Download having usable byte-length metadata. Some Mux-backed Downloads exist and are reachable but have `size` set to zero or null, causing Arclight to return `sizeInBytes: 0`. The reported case was the English Variant (`529`) of LUMO Acts episode 2 (`6_Acts0402`): its generated low and high Mux Downloads returned HTTP 200 with real byte lengths, while Arclight reported zero-byte sizes. This prevents legacy download automation from treating the files as valid even though the media itself does not need to be reuploaded.

The LUMO Acts base series also depends on empty, non-downloadable parent Variants for language discovery. Its language set should equal the union of languages represented by Variants on its direct published children. Existing publishing behavior can create these parent Variants, but the current mutation cannot recover missing parent Variants when the series is already published.

Operators need two safe, separate tools: one to audit and recover missing parent language Variants through the existing publishing mutation, and another to repair null or nonpositive sizes on existing Downloads from authoritative provider data. Both operations need dry-run behavior, bounded execution, audit evidence, and no unrelated media mutations.

## Solution

Extend the existing child-publishing mutation with an explicit `parentVariantsOnly` mode. This mode will compare a parent Video with its direct published children, report missing empty parent Variants during dry-run, and recover only those missing Variants during apply by reusing the existing tested parent-Variant creation logic. A reusable operator command will invoke this mode for a supplied parent Video ID; the initial production audit target is `6_Acts`.

Add a standalone api-media Download-size backfill command. It will find every existing Download whose size is null, zero, or negative; classify its provider deterministically; resolve an authoritative positive byte length; and conditionally update only the size. Mux is the primary provider, R2 is the backup for legacy media, and unlinked non-Mux URLs are legacy sources. The command will default to dry-run, require an explicit apply flag for writes, process globally by default in resumable batches, and produce a redacted JSONL audit report plus a provider summary.

## User Stories

1. As a publisher, I want the LUMO Acts base series to expose every language present on its direct published children, so that consumers can discover the series in every available language.
2. As a publisher, I want episode language sets to remain independent, so that recovery does not incorrectly require every episode to contain the same languages.
3. As an operator, I want to audit a parent Video by ID, so that the same language-integrity check can be reused beyond LUMO Acts.
4. As an operator, I want the parent-language audit to consider only direct published children, so that its result matches the existing parent synchronization rule.
5. As an operator, I want a dry-run to list missing parent language Variants, so that I can review the intended recovery before changing production.
6. As an operator, I want missing parent Variants recovered through the existing publishing mutation and parent-Variant creation behavior, so that slug, publication, and language semantics are not duplicated in a maintenance script.
7. As an operator, I want existing parent Variants left unchanged, so that recovery is idempotent and does not disturb valid catalog data.
8. As an Arclight consumer, I want every existing Download to report a positive verified byte length, so that download automation can validate and retrieve the file.
9. As a 5fish download-robot operator, I want Mux fallback Downloads to carry their real sizes, so that missing distro Downloads do not make otherwise valid media unusable.
10. As an operator, I want the size backfill to include published, unpublished, downloadable, and non-downloadable Variants, so that invalid metadata does not remain hidden until later publication.
11. As an operator, I want Mux Downloads resolved from Mux rendition metadata first, so that the primary provider remains the primary source of truth.
12. As an operator, I want a verified HTTP-length fallback for Mux when rendition metadata lacks a filesize, so that reachable zero-size Downloads such as the reported LUMO Acts files can still be repaired.
13. As an operator, I want R2-backed Downloads resolved from their linked R2 asset metadata, so that legacy backup files use their connected provider record.
14. As an operator, I want unlinked legacy URLs checked with lightweight HTTP metadata requests, so that recoverable legacy rows can be corrected without downloading entire media files.
15. As an operator, I want unverifiable or conflicting sizes skipped and reported, so that the backfill never writes guesses or copied sibling sizes.
16. As an operator, I want dry-run to be the default, so that an accidental invocation cannot modify production.
17. As an operator, I want an explicit apply flag, so that production writes express deliberate intent.
18. As an operator, I want global processing by default, so that all null and nonpositive Download sizes can be repaired in one controlled operation.
19. As an operator, I want optional filters for focused verification, so that a known Download, Variant, or provider can be tested before a global run.
20. As an operator, I want stable pagination and resumable batches, so that a large run can recover from interruption without restarting from the beginning.
21. As an operator, I want provider-specific concurrency limits, so that Mux and legacy hosts are not overwhelmed while inexpensive R2 metadata can be processed efficiently.
22. As an operator, I want one failed row to be recorded and skipped, so that isolated provider failures do not abort the entire repair.
23. As an operator, I want conditional writes that only update a row while its size remains null or nonpositive, so that concurrent corrections are never overwritten.
24. As an operator, I want the backfill to change only the Download size, so that URLs, qualities, providers, assets, and Download existence remain unchanged.
25. As an operator, I want a redacted row-level audit report, so that every proposed or applied correction can be traced without exposing credentials or signed URLs.
26. As an operator, I want a final summary grouped by provider and outcome, so that I can assess coverage and unresolved debt quickly.
27. As a maintainer, I want the production dry-run reviewed before apply, so that the blast radius and provider failures are understood.
28. As a maintainer, I want post-apply verification against both provider data and public Arclight output, so that database correction demonstrably fixes consumer behavior.
29. As Dan's support contact, I want `6_Acts0402` English 720p retried by the download robot after repair, so that the original incident has an end-to-end acceptance check.

## Implementation Decisions

- Build two separate operator workflows: parent-language reconciliation and Download-size backfill. Do not combine them into one command.
- Extend the existing child-publishing mutation with a `parentVariantsOnly` mode rather than creating parent Variants directly from an operator script.
- In `parentVariantsOnly` dry-run, compute and return language IDs found on direct published child Variants but missing from the parent Video's empty Variants.
- In `parentVariantsOnly` apply, invoke the existing parent-Variant creation behavior for only the missing languages. Do not republish Videos or child Variants, and do not modify existing parent Variants.
- Provide a reusable parent audit/operator command parameterized by parent Video ID. Run it initially for `6_Acts`.
- Add a standalone TypeScript Download-size backfill command registered as an api-media Nx target.
- Select every existing Download where `size` is null or less than or equal to zero. Include rows regardless of Variant publication or downloadable state.
- Classify the authoritative provider deterministically: a Mux stream URL is Mux-backed; a non-Mux Download with a linked asset is R2-backed; remaining URLs are legacy.
- For Mux, prefer the matching static rendition `filesize`. If unavailable, verify the final Mux URL with HTTP HEAD and then a one-byte Range request when necessary.
- For R2, use the positive content length on the linked R2 asset.
- For legacy URLs, use final HTTP `Content-Length`, falling back to a one-byte Range request that exposes the total length. Never stream the full file merely to calculate its size.
- Treat only a positive, provider-connected byte length as repairable. Conflicting, missing, invalid, or unreachable sources must be skipped and reported.
- Default to dry-run. Require an explicit `--apply` flag for writes.
- Process all eligible rows globally by default. Optional focused filters may support validation but must not be required for the global operation.
- Use stable ID-based pagination, bounded database batches, resumable progress, and conservative provider-specific concurrency.
- Apply a conditional size-only update. If the row is no longer null or nonpositive at write time, record it as already corrected and do not overwrite it.
- Emit a redacted JSONL record for each candidate containing Download ID, Variant ID, provider, prior size, verified size, outcome, and normalized error code. Do not include credentials, signed query strings, or sensitive full URLs.
- Emit a final summary by provider and outcome, including total candidates, repairable rows, applied rows, already-corrected rows, skipped rows, and failures.
- Preserve the existing Download's URL, quality, provider association, asset association, and identity.
- The production workflow is: parent audit dry-run, optional parent recovery apply, size-backfill dry-run and review, size-backfill apply, database verification, Arclight verification, then download-robot retry.

## Testing Decisions

- Prefer the existing GraphQL mutation seam for parent reconciliation. Tests should execute the mutation behavior and assert externally visible results rather than testing private query-shape helpers.
- Add mutation coverage proving that `parentVariantsOnly` dry-run reports missing language IDs without writing.
- Add mutation coverage proving that apply creates only missing empty parent Variants through existing behavior and leaves existing Variants unchanged.
- Add mutation coverage proving that only languages from direct published children participate; unpublished children and deeper descendants do not.
- Add mutation coverage proving that `parentVariantsOnly` does not publish Videos or child Variants.
- Test the Download-size backfill at its command/service boundary with mocked Prisma and provider clients, asserting outcomes and writes rather than internal helper implementation.
- Cover Mux rendition filesize resolution, Mux HTTP fallback, R2 content-length resolution, legacy HEAD resolution, Range fallback, provider disagreement, invalid lengths, and unreachable sources.
- Cover null, zero, and negative candidate sizes as well as exclusion of positive sizes.
- Cover dry-run, explicit apply, conditional concurrent-update protection, pagination, resumability, provider-specific failure isolation, and redaction.
- Verify idempotency: a second apply after a successful run must perform no additional size updates.
- Before production apply, run a global production dry-run and review its summary and failure report.
- After apply, query production to confirm no repairable Download remains null or nonpositive; any remaining row must have a corresponding failure record.
- Verify applied rows against their authoritative provider values.
- Verify the public Arclight response for `6_Acts0402` / `529` reports positive correct low and high sizes, then have the 720p download retried by the original robot.

## Out of Scope

- Correcting the known nonzero-but-incorrect distro Download sizes; these require a separate audited migration because of their larger blast radius.
- Creating missing Downloads or distro Downloads.
- Deleting or replacing existing Downloads.
- Changing Download URLs, qualities, providers, asset links, dimensions, or bitrates.
- Reuploading source media.
- Adding persistent Mux download-request lifecycle tracking to VideoVariantUpload.
- Preventing premature Download creation in the ingestion workflow.
- Preventing *new* zero-size Mux renditions from being persisted — already shipped in `f6a209445` (VMT-239). This issue covers repair of the existing backlog only.
- New proactive monitoring, dashboards, or alerts for Download completeness and correctness.
- Requiring every LUMO Acts episode to have the same language set.
- Recursively deriving parent languages from descendants beyond direct published children.

## Further Notes

- The original LUMO Acts incident was `6_Acts0402`, language `529`, requested quality 720p. After the missing Mux Downloads were generated, the low and high URLs existed but Arclight still reported `sizeInBytes: 0`.
- At investigation time, the low file's verified HTTP length was 243,808,898 bytes and the high file's verified HTTP length was 1,296,505,318 bytes.
- Separate production analysis found 142,771 linked distro Downloads with nonzero sizes that differ from linked R2 content lengths. This is deliberately excluded from the immediate backfill and should remain visible as follow-up data debt.
- Mux is the primary Download provider. R2 is the backup for legacy media; other sources are legacy debt.
- Download rows must represent available files, not provider-request progress. Persistent request/readiness tracking is a later design effort and should use existing VideoVariantUpload lifecycle records rather than blank Download rows.

tanflem:
Closes #9382

## Summary

Implements both remaining deliverables from the LUMO Acts zero-size Download / missing parent language Variant PRD. The prevention half already shipped in `f6a209445` (VMT-239, stops zero-value Mux rendition metadata from being persisted); this closes the repair half.

### 1. `parentVariantsOnly` publish mode

Extends the existing `videoPublishChildren` mutation with a `parentVariantsOnly` value on `VideoPublishMode`, rather than a standalone script — slug/publication/language semantics stay in one tested place.

- **Dry-run** returns `missingParentLanguageIds`: language IDs present on a **direct published** child Variant but absent from any existing parent Variant. Unpublished children and deeper descendants never contribute a language.
- **Apply** reuses the existing tested `handleParentVariantCreation` helper for only the missing `(childVideoId, languageId)` pairs. It never republishes a Video or child Variant and never touches an existing parent Variant (the helper is itself a no-op once a Variant exists for that language) — so reruns are idempotent.
- New `recover-parent-variants` Nx script/target wraps this mode for a supplied parent Video ID (default `6_Acts`, the initial production target). Dry-run by default; `PARENT_VARIANT_RECOVERY_APPLY=true` to write.

### 2. Download-size backfill

New standalone `backfill-download-sizes` Nx target + `lib/downloadSizeBackfill/` module:

- Selects every Download with `size` null or `<= 0`.
- Classifies the authoritative provider deterministically: Mux stream URL → Mux (even over a stale linked R2 asset); non-Mux with a linked asset → R2; else → legacy.
- Mux: prefers the matching static rendition's `filesize`, falling back to HTTP HEAD then a one-byte Range request. R2: uses the linked asset's positive content length. Legacy: HEAD then Range. Never streams a full file to measure it — Range requests abort the connection immediately after headers.
- Dry-run by default; `DOWNLOAD_SIZE_BACKFILL_APPLY=true` to write. Global processing by default, with optional `downloadId`/`videoVariantId`/`provider` filters for focused validation.
- Stable ID-cursor pagination, bounded batches, per-provider concurrency limits, resumable via a saved cursor file.
- Conditional, size-only write (`updateMany` scoped to still-null/nonpositive rows) — a row corrected concurrently is recorded as `alreadyCorrected` instead of overwritten. One row's failure never aborts the batch.
- Redacted JSONL audit report (Download ID, Variant ID, provider, prior/verified size, outcome, error code — no URLs or credentials) plus a provider/outcome summary.

## Key decisions

- Kept the two tools fully separate per the PRD, no combined command.
- "Missing" parent language = present on a direct published child Variant but absent from **any** existing parent Variant (empty or not) — matches `handleParentVariantCreation`'s own idempotency check, so an ambiguous non-empty parent Variant is left alone rather than silently touched.
- HTTP Range responses only trust `Content-Range`'s total on a 206; a 200 response (Range ignored) falls back to `Content-Length`, and the connection is always aborted right after headers arrive.

## Testing

- `videoPublishChildren.mutation.spec.ts`: 4 new tests for `parentVariantsOnly` (dry-run reporting, direct-published-children scoping, apply via the existing helper, idempotency) — 17/17 passing.
- `lib/downloadSizeBackfill/*.spec.ts`: 38 new tests covering provider classification (including provider-disagreement determinism), HTTP HEAD/Range resolution (incl. unreachable/invalid-length), Mux rendition + HTTP fallback, R2 resolution, legacy resolution, the concurrency limiter, and the service boundary (selection, dry-run vs apply, conditional writes, pagination/resumability, per-row failure isolation, redaction, filters).
- `pnpm exec nx affected -t lint,type-check,test --base=origin/main` — clean (23 affected projects, api-media: 821/821 tests, 0 lint errors).
- Regenerated `apis/api-media/schema.graphql` and `apis/api-gateway/schema.graphql` for the new field/enum value.

## Out of scope / follow-up

This PR ships the tooling only. The production runbook (parent audit dry-run → optional apply for `6_Acts` → size-backfill dry-run/review → apply → DB verification → Arclight verification for `6_Acts0402`/`529` → download-robot retry) still needs to be executed by an operator after merge. #9468 (catalog-wide parent-language audit/repair) should build on the `parentVariantsOnly` mode landed here rather than a second path.

## Summary by CodeRabbit

* **New Features**
  * Added parent-variant recovery to identify missing language variants, preview changes, and create them when applied.
  * Added download-size repair tooling for Mux, R2, and legacy media, with filtering, batching, resumable processing, dry-run support, and audit reporting.
  * Added GraphQL fields and modes to report missing parent language IDs.
* **Documentation**
  * Added usage guidance for parent-variant recovery and download-size repair commands.
* **Tests**
  * Added comprehensive coverage for size resolution, concurrency, recovery, filtering, pagination, and failure handling.

### DIFF
--- a/apis/api-gateway/schema.graphql
+++ b/apis/api-gateway/schema.graphql
@@ -3032,6 +3032,10 @@ type VideoPublishChildrenResult @join__type(graph: API_MEDIA)  {
   publishedVariantsCount: Int
   dryRun: Boolean
   videosFailedValidation: [VideoPublishChildrenUnpublishedVideo!]!
+  """
+  Language IDs present on a direct published child Variant but missing an empty parent Variant. Populated by parentVariantsOnly mode.
+  """
+  missingParentLanguageIds: [ID!]
 }
 
 type VideoPublishChildrenUnpublishedVideo @join__type(graph: API_MEDIA)  {
@@ -3821,6 +3825,7 @@ enum VideoPublishMode @join__type(graph: API_MEDIA)  {
   childrenVideosOnly @join__enumValue(graph: API_MEDIA) 
   childrenVideosAndVariants @join__enumValue(graph: API_MEDIA) 
   variantsOnly @join__enumValue(graph: API_MEDIA) 
+  parentVariantsOnly @join__enumValue(graph: API_MEDIA) 
 }
 
 enum Service @join__type(graph: API_MEDIA)  {
--- a/apis/api-gateway/schema.graphql
+++ b/apis/api-gateway/schema.graphql
@@ -3036,6 +3036,10 @@ type VideoPublishChildrenResult @join__type(graph: API_MEDIA)  {
   Language IDs present on a direct published child Variant but missing an empty parent Variant. Populated by parentVariantsOnly mode.
   """
   missingParentLanguageIds: [ID!]
+  """
+  Subset of missingParentLanguageIds whose parent Variant was successfully created. Populated by parentVariantsOnly mode when dryRun is false; a language missing from this list relative to missingParentLanguageIds failed and was logged.
+  """
+  recoveredParentLanguageIds: [ID!]
 }
 
 type VideoPublishChildrenUnpublishedVideo @join__type(graph: API_MEDIA)  {
--- a/apis/api-media/project.json
+++ b/apis/api-media/project.json
@@ -182,6 +182,20 @@
         "command": "pnpm exec tsx --tsconfig apis/api-media/tsconfig.app.json apis/api-media/src/scripts/run-parent-variant-audit.ts"
       }
     },
+    "recover-parent-variants": {
+      "executor": "nx:run-commands",
+      "dependsOn": ["prisma-media:prisma-generate"],
+      "options": {
+        "command": "pnpm exec tsx --tsconfig apis/api-media/tsconfig.app.json apis/api-media/src/scripts/recover-parent-variants.ts"
+      }
+    },
+    "backfill-download-sizes": {
+      "executor": "nx:run-commands",
+      "dependsOn": ["prisma-media:prisma-generate"],
+      "options": {
+        "command": "pnpm exec tsx --tsconfig apis/api-media/tsconfig.app.json apis/api-media/src/scripts/backfill-download-sizes.ts"
+      }
+    },
     "populate-crowdin-ids": {
       "executor": "nx:run-commands",
       "options": {
--- a/apis/api-media/schema.graphql
+++ b/apis/api-media/schema.graphql
@@ -1346,6 +1346,11 @@ type VideoPublishChildrenResult {
   publishedVariantsCount: Int
   dryRun: Boolean
   videosFailedValidation: [VideoPublishChildrenUnpublishedVideo!]!
+
+  """
+  Language IDs present on a direct published child Variant but missing an empty parent Variant. Populated by parentVariantsOnly mode.
+  """
+  missingParentLanguageIds: [ID!]
 }
 
 type VideoPublishChildrenUnpublishedVideo {
@@ -1358,6 +1363,7 @@ enum VideoPublishMode {
   childrenVideosOnly
   childrenVideosAndVariants
   variantsOnly
+  parentVariantsOnly
 }
 
 enum VideoRedirectType {
--- a/apis/api-media/schema.graphql
+++ b/apis/api-media/schema.graphql
@@ -1351,6 +1351,11 @@ type VideoPublishChildrenResult {
   Language IDs present on a direct published child Variant but missing an empty parent Variant. Populated by parentVariantsOnly mode.
   """
   missingParentLanguageIds: [ID!]
+
+  """
+  Subset of missingParentLanguageIds whose parent Variant was successfully created. Populated by parentVariantsOnly mode when dryRun is false; a language missing from this list relative to missingParentLanguageIds failed and was logged.
+  """
+  recoveredParentLanguageIds: [ID!]
 }
 
 type VideoPublishChildrenUnpublishedVideo {
--- a/apis/api-media/src/lib/downloadSizeBackfill/classifyProvider.spec.ts
+++ b/apis/api-media/src/lib/downloadSizeBackfill/classifyProvider.spec.ts
@@ -1,0 +1,42 @@
+import { classifyProvider } from './classifyProvider'
+
+describe('classifyProvider', () => {
+  it('classifies a Mux stream URL as mux', () => {
+    expect(
+      classifyProvider({
+        url: 'https://stream.mux.com/abc123/720p.mp4',
+        assetId: null
+      })
+    ).toBe('mux')
+  })
+
+  it('classifies a non-Mux URL with a linked asset as r2', () => {
+    expect(
+      classifyProvider({
+        url: 'https://legacy.example.com/file.mp4',
+        assetId: 'asset-1'
+      })
+    ).toBe('r2')
+  })
+
+  it('classifies a non-Mux URL without a linked asset as legacy', () => {
+    expect(
+      classifyProvider({
+        url: 'https://legacy.example.com/file.mp4',
+        assetId: null
+      })
+    ).toBe('legacy')
+  })
+
+  it('deterministically prefers mux over a conflicting linked asset', () => {
+    // Provider disagreement: a Mux stream URL alongside a leftover R2
+    // asset link from an earlier migration. The URL a client actually
+    // fetches from must win.
+    expect(
+      classifyProvider({
+        url: 'https://stream.mux.com/abc123/720p.mp4',
+        assetId: 'stale-r2-asset'
+      })
+    ).toBe('mux')
+  })
+})
--- a/apis/api-media/src/lib/downloadSizeBackfill/classifyProvider.ts
+++ b/apis/api-media/src/lib/downloadSizeBackfill/classifyProvider.ts
@@ -1,0 +1,18 @@
+import type { BackfillProvider, DownloadCandidate } from './types'
+
+export const MUX_STREAM_BASE_URL = 'https://stream.mux.com/'
+
+/**
+ * Deterministically classifies which provider is authoritative for a
+ * Download's byte length. A Mux stream URL always wins, even if the row
+ * also carries a linked R2 asset (data debt from an earlier migration) —
+ * the URL a client actually fetches is what must be verified. A non-Mux
+ * Download with a linked asset is R2-backed; everything else is legacy.
+ */
+export function classifyProvider(
+  download: Pick<DownloadCandidate, 'url' | 'assetId'>
+): BackfillProvider {
+  if (download.url.startsWith(MUX_STREAM_BASE_URL)) return 'mux'
+  if (download.assetId != null) return 'r2'
+  return 'legacy'
+}
--- a/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.spec.ts
+++ b/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.spec.ts
@@ -30,4 +30,13 @@ describe('createConcurrencyLimiter', () => {
       })
     ).rejects.toThrow('boom')
   })
+
+  it.each([NaN, Infinity, -Infinity, 1.5, 0, -1])(
+    'rejects a non-finite, non-integer, or non-positive limit (%p)',
+    (limit) => {
+      expect(() => createConcurrencyLimiter(limit)).toThrow(
+        'createConcurrencyLimiter requires a finite positive integer'
+      )
+    }
+  )
 })
--- a/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.spec.ts
+++ b/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.spec.ts
@@ -1,0 +1,33 @@
+import { createConcurrencyLimiter } from './concurrencyLimiter'
+
+describe('createConcurrencyLimiter', () => {
+  it('never runs more than `limit` tasks concurrently', async () => {
+    const limit = createConcurrencyLimiter(2)
+    let active = 0
+    let maxActive = 0
+
+    const task = async (): Promise<void> => {
+      active++
+      maxActive = Math.max(maxActive, active)
+      await new Promise((resolve) => setTimeout(resolve, 5))
+      active--
+    }
+
+    await Promise.all(
+      Array.from({ length: 6 }, () => limit(task))
+    )
+
+    expect(maxActive).toBeLessThanOrEqual(2)
+  })
+
+  it('propagates a task result and rejection', async () => {
+    const limit = createConcurrencyLimiter(1)
+
+    await expect(limit(async () => 'ok')).resolves.toBe('ok')
+    await expect(
+      limit(async () => {
+        throw new Error('boom')
+      })
+    ).rejects.toThrow('boom')
+  })
+})
--- a/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.ts
+++ b/apis/api-media/src/lib/downloadSizeBackfill/concurrencyLimiter.ts
@@ -5,7 +5,12 @@
  * R2 metadata lookups can run with more headroom).
  */
 export function createConcurrencyLimiter(limit: number) {
-  const boundedLimit = Math.max(1, limit)
+  if (!Number.isInteger(limit) || limit < 1) {
+    throw new Error(
+      `createConcurrencyLimiter requires a finite positive integer, got ${limit}`
+    )
+  }
+  const boundedLimit = limit
   let active = 0
   const queue: Array<() => void> = []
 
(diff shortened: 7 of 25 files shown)


tanflem:
Phoebe update for #9382: 1 new commit(s) pushed to this branch.

tanflem:
Fixed the **Danger** failures (PR title now matches commitlint convention, assignee and reviewer set) — that check is now green.

The **autofix.ci** failure is entirely the \`lint branch name\` step (\`validate-branch-name\`) — ESLint/Prettier/type-check within that same job all pass with 0 errors. The branch name \`phoebe/issue-9382\` doesn't match the required pattern (the \`issue\` prefix segment is 5 chars; the convention caps ticket prefixes at 2–4, per `AGENTS.md`). This is structural to how this branch/PR was created, not something introduced by this PR's diff, and fixing it would require renaming the branch — which conflicts with keeping this PR's identity (`phoebe/issue-9382`) intact for the automation driving it. The immediately-preceding Phoebe PR (#9489, branch `phoebe/issue-9483`) hit the identical failure and was left with it unresolved for the same reason, so I'm leaving this branch name as-is rather than renaming out from under the PR.

All substantive gates (lint, type-check, test, build) are green — confirmed both on GitHub Actions and locally via `nx affected -t lint,type-check,test --base=origin/main` (23 projects / 46 lint+type-check tasks, 14 test projects incl. `api-media`, 0 errors).

tanflem:
Phoebe attempted an idle CI fix for PR #9491 but could not resolve the failing checks. The branch was left unchanged. A human should investigate the CI failures.

tanflem:
## Review feedback addressed

**Fixed:** 10 thread(s)
**Challenged:** 0 thread(s)
**Skipped:** 0 thread(s)

All 10 unresolved CodeRabbit threads identified real issues; all were fixed (one — the exit-code/logger consolidation on `recover-parent-variants.ts` and `backfill-download-sizes.ts` — required a slightly different shape than the literal suggestion to fit the existing return types, noted below):

- **`concurrencyLimiter.ts`**: reject `NaN`/`Infinity`/fractional/non-positive limits instead of silently bypassing the concurrency bound.
- **`httpSize.ts`**: added a 10s deadline (`AbortController` + timer) to both HEAD and Range requests so a host that accepts the connection but never sends headers can't hold a concurrency slot indefinitely.
- **`resolveMuxSize.ts`**: only trust a rendition whose `status` is exactly `ready` (not `preparing`); require a strictly numeric `filesize` instead of `parseInt`'s permissive parsing (which accepted `"500bytes"`, `"1e3"`, etc).
- **`service.ts` / `backfill-download-sizes.ts`**: validate `batchSize` as a finite positive safe integer at both the service and command boundaries, so an invalid value fails fast instead of looping forever with no progress.
- **`videoPublishChildren.mutation.ts` (`parentVariantsOnly`)** — *fix-adjusted*: instead of scoping `handleParentVariantCreation` (not straightforward — it discovers parents internally), we now call the lower-level `createEmptyParentVariant` directly, scoped to the requested parent `id`, so a child shared by multiple parents can never write to a sibling parent. Per-language outcomes are tracked and returned via a new `recoveredParentLanguageIds` GraphQL field instead of always reporting the pre-write missing list as recovered.
- **`recover-parent-variants.ts`**: reports attempted vs. recovered counts and sets a nonzero exit code when any recovery fails; switched to the structured Pino logger; sets `process.exitCode` instead of calling `process.exit(1)` so `prisma.$disconnect()` in `finally` always completes.
- **`backfill-download-sizes.ts`**: the saved cursor is now bound to the run's apply mode and filters (a dry-run cursor is no longer accepted by an apply run and vice versa); batch records are written with drain-aware backpressure and the report stream's completion is awaited before advancing the cursor or exiting; switched to the structured Pino logger.

Added/extended regression tests for every fix (invalid concurrency limits, HTTP timeout aborts, `preparing`/malformed Mux filesize, invalid batch sizes, multi-parent write scoping, partial-recovery-failure reporting). Regenerated `apis/api-media/schema.graphql` and `apis/api-gateway/schema.graphql` for the new `recoveredParentLanguageIds` field.

Verification: `tsc7` (type-check), `lint:changed`, and the full `api-media` Vitest suite (844/844) all pass on the pushed commit.

tanflem:
## CI investigation: \`autofix.ci\` (lint / lint-work) failure

Both failing checks (\`lint\`, \`lint-work\`) come from the same root cause: the **"lint branch name" step** inside \`lint-work\`, which runs \`pnpm exec validate-branch-name -t $GITHUB_HEAD_REF\`. Every other step in that job — Prettier, Type Check, ESLint, Extract Translations, Subgraph Check — passes cleanly.

The branch \`phoebe/issue-9382\` doesn't match the repo's \`validate-branch-name.config.js\` pattern: the \`username/ticket-id-short-description\` alternative caps the ticket-id segment at 2–4 chars (\`[a-z0-9]{2,4}\`), and Phoebe's \`issue\` segment is 5 chars. This is a pre-existing mismatch between how the Phoebe host names branches and the repo's branch-naming policy — **it's not caused by this PR's code**, and I confirmed the other open Phoebe PR (#9489, branch \`phoebe/issue-9483\`) fails the identical check for the identical reason.

I verified locally that the actual code gates are clean:
- \`pnpm exec nx affected -t lint,type-check --base=origin/main\` — pass
- \`pnpm exec nx affected -t test --base=origin/main\` — pass

I can't fix this from within the PR: renaming the branch isn't possible without closing/reopening the PR, and I'm not modifying the shared \`validate-branch-name.config.js\` policy as part of this feature PR since that's a repo-wide naming-convention decision affecting all future Phoebe-authored PRs, not something scoped to this change.

Filed **#9492** to track fixing the pattern (or the Phoebe host's branch-naming scheme) so future Phoebe PRs don't hit this.

PR #9493, opened 2026-08-14
chore: remove orphaned libs/shared/ui-modern

mikeallisonJS:
Was stacked on #9490 (`chore: remove deprecated watch and watch-modern apps`), which has since merged. Rebased onto `main`; this PR is now standalone and contains only the `ui-modern` removal.

## Why

`libs/shared/ui-modern` was the shadcn/Tailwind component kit built for `watch-modern`. With #9490 merged, **nothing in the repo imports `@core/shared/ui-modern`** — it was 21 files in `watch-modern` and nowhere else. `CONTEXT-MAP.md` already carried the note "has no consumers today — its only consumer was the removed legacy Watch app." This deletes it rather than leaving a lib that lints, type-checks, and shows up in the Nx graph for no one.

## What changed

- Deleted `libs/shared/ui-modern/` (project config, 18 vendored shadcn components, the `cn` helper, design tokens, the `add-shadcn-component.sh` script, docs — plus a stray `textarea.tsx.back`).
- Dropped the `@core/shared/ui-modern/*` path alias from `tsconfig.base.json` and `tsconfig.base.ts7.json`.
- Removed the `Shared UI Modern` entry from `CONTEXT-MAP.md`.
- Removed the seven npm dependencies nothing else used: `@radix-ui/react-checkbox`, `@radix-ui/react-popover`, `@radix-ui/react-slider`, `@radix-ui/react-switch`, `@radix-ui/react-tabs`, `@radix-ui/react-tooltip`, `cmdk`.

## Dependencies deliberately kept

`clsx`, `tailwind-merge`, `lucide-react`, `class-variance-authority`, `@radix-ui/react-{accordion,dialog,select,slot}` all stay — `apps/resources` has its own local shadcn-style components (`src/components/{Button,Accordion,Dialog,Select}`, `src/libs/cn`) that use them.

## Verification

- `grep -rn "ui-modern"` across the tree returns nothing outside the deleted directory.
- `pnpm nx show projects` resolves (41 projects) and no longer lists `ui-modern`.
- `pnpm install --lockfile-only` produces a deletion-only lockfile diff (−260 lines, no version churn elsewhere).
- Pre-push `lint:changed` passed.

Re-verified after the rebase against current `main`: each of the seven removed dependencies has zero remaining references anywhere in the repo (including per-workspace `package.json` files), and `main` has not touched `package.json` or `pnpm-lock.yaml` since the original merge-base, so the lockfile diff still applies cleanly.

## Note

If the Tailwind migration restarts, `apps/resources` is now the de facto reference for the shadcn pattern in this repo, and the components are recoverable from git history.

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed the shared modern UI library and its reusable components, including buttons, forms, dialogs, menus, alerts, tabs, and tooltips.
  - Existing applications relying on these shared UI exports may require updates.

- **Documentation**
  - Removed setup, usage, component inventory, and testing documentation for the library.

- **Chores**
  - Removed associated styling, project configuration, generation tooling, and workspace path mappings.


csiyang (approved):
Verified independently against the branch rather than the description:

- **No dangling references.** `ui-modern` appears nowhere on the tree outside the doc archives, including the non-import surfaces I swept specifically — tailwind content globs, storybook config, `.prettierignore`, Nx `implicitDependencies`, workflows.
- **All seven removed packages have zero remaining references** (excluding the lockfile) and zero remaining lockfile importer entries, so the lockfile was genuinely regenerated.
- **All eight kept packages have live importers** — `clsx` (3 files), `tailwind-merge` (1), `lucide-react` (5), `class-variance-authority` (1), and one each for `@radix-ui/react-{accordion,dialog,select,slot}`, all in `apps/resources`. The retention list is exact.
- **Lockfile diff is deletion-only as claimed** — the only two `+` lines are transitive `@babel/runtime` 7.29.2→7.29.7 re-resolutions in the snapshots section, not a version bump in `package.json`.

Nothing to flag. Nice catch on the `textarea.tsx.back` stray, and the "resources is now the shadcn reference / recoverable from git history" note is the right framing for anyone who restarts a Tailwind track later.

PR #9498, opened 2026-08-17
fix(tooling): add prettier pass to lint:changed so autofix.ci stops fixing formatting

csiyang:
## Why

We added the pre-push lint gate (ENG-3691, #9477) to stop autofix.ci committing `fix: lint issues` on PRs. It didn't work, and #9495 shows why: the gate passed, then autofix.ci pushed two formatting commits anyway.

The gate ran **ESLint only** — but ESLint is structurally incapable of reporting a formatting problem in this repo. We use `eslint-config-prettier`, whose entire purpose is to switch every formatting rule off, and there is no `eslint-plugin-prettier` in any workspace config. Formatting belongs to Prettier alone.

Every autofix.ci commit I could find across recent PRs was pure Prettier output. Not one was an ESLint fix:

| PR | File | What autofix.ci changed |
|---|---|---|
| #9495 | `docs/agents/github-projects.md` ×2 | Markdown table re-alignment |
| #9478 | `CONTEXT-MAP-contract.md` | Markdown table re-alignment |
| #9473 | `updateLanguageInAlgolia.spec.ts` | Line-wrap at printWidth 80 |
| #9459 | `DeleteBlock.spec.tsx` | Line-wrap at printWidth 80 |
| #9457 | `TemplateCardPreview.spec.tsx` | Un-wrapped a line that fit |

So the gate was checking the thing that never fires and missing the thing that always does.

A second problem compounded it. Changed files were narrowed to `.ts/.tsx/.js/.jsx/.cjs/.mjs` *before* linting, and dropped silently — not even reported as skipped. #9495 was docs-only, so every file was discarded and the gate printed `✅ - no changed lintable files`: a green tick for zero work.

## What changed

`lint:changed` now mirrors every autofix.ci step that can write a file, in the same order, so a clean run locally means autofix.ci has nothing left to commit.

1. **Prettier** over every changed file of any type — `.md`, `.json`, `.yaml`, `.css`, not just JS/TS. Under a second.
2. **ESLint** over changed JS/TS, scoped per workspace, exactly as before. 5–20s per touched workspace.
3. **i18next extraction** over changed projects, grouped by nearest `i18next.config.js` the same way ESLint groups by `eslint.config.mjs`. 2–3s per touched project.

`i18next-cli` has exactly the modes needed: `--dry-run --ci` reports drift by exit code without touching the working tree, and plain extraction under `--fix` writes it, as the bot would.

### What is deliberately left to CI

Only the steps that can actually produce a commit are worth gating. Checked which of them write tracked files:

| autofix.ci step | Writes tracked files? | Gated locally |
|---|---|---|
| `prettier` | yes | ✅ |
| `lint --fix` | yes | ✅ |
| `extract-translations` | yes — 300 `libs/locales/*.json` | ✅ |
| `codegen` | yes — 606 `__generated__` files | ❌ see below |
| `prisma-generate` | no | n/a |
| `type-check`, `subgraph-check` | no — report only | n/a |

**`codegen` is deliberately left to CI.** It only works where `apollo` is installed globally — the devcontainer does it in `.devcontainer/post-create-command.sh` (`npm i -g nx foreman apollo graphql`) and autofix.ci does the same. Gating it would mean requiring that global install on every machine that pushes, which we decided against. Run `nx codegen <project>` in the devcontainer after changing a GraphQL schema or query.

Details worth calling out:

- **`.prettierignore` needs no special handling.** Prettier honours it for explicitly-passed paths, so the existing carve-outs (`.claude/skills/`, `.agents/skills/`, `libs/locales/`, generated files) keep working for free. Verified with a deliberately-malformed probe file: ignored inside `.claude/skills/`, caught at the repo root.
- **`--ignore-unknown` is required.** Without it a changed `Dockerfile` or `LICENSE` exits 2 with "No parser could be inferred" and hard-fails the whole push.
- **Untracked directories are skipped.** `git ls-files --others` collapses them to a directory path; the `-f` guard drops those, so nested worktrees don't get swept in.

### Toolchain resolution (fixes a pre-existing bug)

The script called `pnpm exec eslint`. A checkout made with `git worktree add` has no `node_modules` of its own, and pnpm treats each worktree as its own root because it has its own `pnpm-workspace.yaml` — so `pnpm exec` there fails with `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL` instead of falling back.

That means **the ESLint step was already broken for any worktree push touching JS/TS** — it just never surfaced, because the JS/TS filter meant most worktree pushes exited before reaching it. Adding a Prettier pass that runs on *every* push would have turned that latent breakage into a wall.

Both binaries now resolve from `node_modules/.bin`, falling back to the main checkout via `git rev-parse --git-common-dir`. Same dependency set, and both tools resolve their config from the target file, so running them against worktree paths is correct.

## Verification

Reproduced #9495 exactly. Checked out the pre-autofix commit (`e65c2a68`) into a fresh worktree with no `node_modules`:

- **Before:** `✅ - no changed lintable files`, exit 0 — push allowed, autofix.ci then committed the fix.
- **After:** `🛑 - formatting issues`, exit 1 — push blocked at the right moment.
- **`--fix` output is byte-identical** to what autofix.ci committed in `ee85fab1`. `diff` is empty.

Also covered: unknown-parser files don't exit 2; a workspace TS file runs both passes and ESLint now actually executes from a worktree; a clean tree exits 0.

## Scope

The gate is still **agent-only**, and manual human pushes still skip it; `--no-verify` still bypasses everything. Codex sessions are now recognised alongside Claude Code and Cursor, matched on the `CODEX_` prefix rather than specific variable names — the documented ones are conditional (`CODEX_SANDBOX` is set only while the sandbox is active) and nobody here runs Codex to confirm the full set. A missed name fails silently by leaving those pushes ungated, which is the exact bug being closed; a false positive only runs a lint we wanted anyway.

Verified all six i18n projects are in sync on `main`, so the new pass cannot fire spuriously on unrelated work.

## Summary by CodeRabbit

* **Documentation**
  * Updated contributor guidance for the pre-push validation workflow.
  * Clarified formatting, linting, translation checks, scopes, limitations, and fix-mode behavior.
  * Documented which checks run automatically and which require local action.

* **Developer Experience**
  * Expanded changed-file validation to cover formatting, JavaScript/TypeScript linting, and translation extraction.
  * Added check-only and auto-fix modes with aggregated results.
  * Replaced the automatic pre-push hook with an explicit validation workflow.


tools/scripts/lint-changed.sh
```diff
@@ -1,7 +1,20 @@
 #!/bin/bash
 
 # Lint only changed files (full `nx lint` is far too slow for commit-time feedback).
-# Groups files by their nearest eslint.config.mjs and runs ESLint once per
+# Mirrors the two passes autofix.ci runs, in the same order, so a clean run here
+# means autofix.ci has nothing left to commit:
```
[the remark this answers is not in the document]
csiyang: Good catch — applied in `873de128`.

You are right that it overclaimed. `codegen`, `prisma-generate` and `extract-translations` all run in `autofix.ci` and none of them are covered by this gate, so "nothing left to commit" was wrong. `AGENTS.md` already spelled that out in its "Not covered locally" paragraph, so the headline was contradicting its own caveat a few lines down.

Took your wording verbatim in both places, plus the same sentence in the PR description which had the identical problem:

- `tools/scripts/lint-changed.sh:4-5`
- `AGENTS.md:38`

Docs-only, no behaviour change — the gate still passes clean.

tools/scripts/lint-changed.sh
```diff
@@ -13,15 +34,53 @@
...
+  MAIN_ROOT=$(dirname "$common")
+  if [ -d "$MAIN_ROOT/node_modules/.bin" ]; then
+    BIN_DIR="$MAIN_ROOT/node_modules/.bin"
+    echo "note: using the main checkout's toolchain ($MAIN_ROOT/node_modules)"
+    return 0
+  fi
+  return 1
+}
+
+if ! resolve_bin_dir; then
```
jianwei1, line 107: **Concern:** this resolves the tool *binaries* from the main checkout, but not the modules their *configs* pull in — so worktree support ends up depending on where the worktree lives.

Each tool loads a config from the worktree, and those configs resolve bare specifiers from the worktree's own directory chain:

- `.prettierrc` → `"plugins": ["prettier-plugin-tailwindcss"]`
- `apps/*/eslint.config.mjs` → `libs/shared/eslint/next.mjs` → bare plugin imports
- `apps/*/i18next.config.js` → `i18next.assert-input.js` → `require('glob')`

Nested worktree (`core/.worktrees/x`): Node's upward walk finds `core/node_modules` and everything works. Sibling worktree (`git worktree add ../core-feature`): nothing resolves. Reproduced against a checkout with no `node_modules`, prettier invoked from elsewhere — the same shape as a sibling worktree:

```
$ /elsewhere/node_modules/.bin/prettier --check --ignore-unknown 'apps/journeys/pages/home/[journeySlug].tsx'
[error] Cannot find package 'prettier-plugin-tailwindcss' imported from /path/to/checkout/noop.js
exit=1
```

Exit 1 is indistinguishable from a real formatting failure, so the gate prints `🛑 - formatting issues (prettier, files named above)` while naming no files. ESLint hits the same wall and exits 2 → `🛑 - lint failed (file and rule named above)` with no file and no rule. Both are dead ends short of `--no-verify`.

Flagging it mainly because the verification in the description can't separate the two cases: the reported "After" result for the worktree run is `🛑 - formatting issues`, exit 1 — byte-identical to what this failure produces. The `--fix` diff against `ee85fab1` is what actually rules it out for that run, which suggests your worktrees are nested.

Either preflight it (check `$MAIN_ROOT` is an ancestor of `$REPO_ROOT`, or that the plugin resolves) and say so explicitly, or document that worktrees have to be created inside the main checkout.

---

tools/scripts/lint-changed.sh
```diff
@@ -13,15 +34,53 @@
...
+  echo "🛑 - could not find node_modules/.bin" >&2
```
jianwei1, line 108: **Concern:** `resolve_bin_dir` checks that `node_modules/.bin` is a directory, but never that the three binaries are actually in it.

On a partial install (`--prod`, a pruned image) the invocation exits 127, and the script folds that into `🛑 - formatting issues` or `🛑 - extracted translations are out of date`. Same shape as the worktree case above: an infrastructure failure wearing the mask of a content failure, sending you hunting for a formatting bug that doesn't exist.

Cheap guard while you're already here:

```bash
for bin in prettier eslint i18next-cli; do
  [ -x "$BIN_DIR/$bin" ] || { echo "🛑 - $bin missing from $BIN_DIR; run \`pnpm install\`" >&2; exit 1; }
done
```

---

AGENTS.md
```diff
@@ -33,10 +33,24 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
 
 ### Lint before push
 
-Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), `.husky/pre-push` runs the `lint:changed` script on committed changes. It lints only changed files, scoped per workspace; full `nx lint` is far too slow. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime. [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
+Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`, `CURSOR_AGENT`, `CODEX_SANDBOX`, or `CODEX_SESSION_ID`), `.husky/pre-push` runs the `lint:changed` script on committed changes. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime.
```
jianwei1: **Concern:** this documents a narrower trigger than the hook implements.

The line lists four specific variables including `CODEX_SANDBOX` and `CODEX_SESSION_ID`, but `.husky/pre-push` matches `env | grep -q '^CODEX_'` — the whole namespace. The PR description argues at length for prefix matching *because* the specific names are unreliable, and then this line reinstates the guess. Suggest something like "…or any `CODEX_*` variable" so the doc matches the code and carries the reasoning.

Related, same sentence: "Manual human pushes skip the gate" stops holding for a human with any `CODEX_*` var exported. Verified:

```
$ env -i CODEX_HOME=/x sh pre-push-condition
GATED
```

`CODEX_HOME` is a plausible thing to have in a shell profile. By your own reasoning a false positive only runs a lint we wanted anyway, so this is a docs fix rather than a behaviour one — but the promise as written isn't quite true.

---

tools/scripts/lint-changed.sh
```diff
@@ -173,8 +264,72 @@ for ws in "${WS_LIST[@]}"; do
...
+    fi
+  done
+  if [ "$seen" -eq 0 ]; then I18N_LIST+=("$cfg"); fi
+done <<<"$RAW"
+
+I18N_FAILED=0
+if [ ${#I18N_LIST[@]} -eq 0 ]; then
+  echo "no changed files in a project with translations"
+else
+  I18N_ARGS=(--quiet)
```
jianwei1, line 350: **Nit:** under `--fix` this runs *project-wide* extraction, so `pnpm lint:changed --fix` can write keys into `libs/locales/en/*.json` for code the developer never touched, if the project carries any drift.

It mirrors what autofix.ci does and you've verified main is in sync, so it's a note rather than a defect — but the usage comment at the top only warns that `--fix` touches "the working tree", which reads as "the files you changed". One line there would set expectations.

Confirmed the flag choice itself is right, incidentally: `anyFileUpdated` is computed from `result.updated` *before* the `isDryRun` write guard (`extractor/core/extractor.js:65-84`), so `--dry-run --ci` genuinely reports drift rather than silently passing. That was the one place this change could have no-op'd the way the ESLint-only gate did.

---

jianwei1 (commented):
## Review summary

**Critical:** none. **Concerns:** 3 (inline). **Nit:** 1 (inline).

The diagnosis is right and the fix lands where the bug actually is. I verified the mechanism rather than reading it — downloaded `i18next-cli@1.56.7` and read its extractor, installed `prettier@3.6.2` and ran it against this repo's real `.prettierrc`/`.prettierignore`, and ran the script end-to-end in a stub-binary lab repo including a sibling git worktree.

### What holds up

- **`--dry-run --ci` genuinely reports drift.** `anyFileUpdated` is computed from `result.updated` before the `isDryRun` write guard, so the flag pair does what you claim. This was the one place the change could have silently no-op'd the same way the ESLint-only gate did.
- **`--ignore-unknown` is genuinely required, and safe at both edges.** Against the real config: `.devcontainer/Dockerfile` without it → exit 2; with it → exit 0; an all-`.prettierignore`d file set → exit 0. That last one matters — `/sync-ce` pushes touch only `.claude/skills/`, and they don't spuriously fail.
- **249 tracked paths contain `[...]`** (Next.js dynamic routes). Prettier stats literal paths before globbing, so they pass through intact.
- **i18n grouping is complete.** All five `i18next.config.js` input globs are self-contained under their own project dir, so nearest-config grouping can't miss a project. The "sixth" project (`apps/arclight`) has an `extract-translations` target with an empty `commands` array — a no-op, so five is full parity with CI.
- **`extract --config <path>` after the subcommand does reach `program.opts()`** — commander routes subcommand-unknown options to the parent (verified against commander 14), matching how the nx targets already invoke it.
- **Control flow verified with stub binaries**, including the exact #9495 regression: a docs-only change now format-checks and exits 1 instead of printing `✅ - no changed lintable files`. `--fix` correctly drops `--dry-run --ci`; the bash 3.2 empty-array guards behave; the `${arr[@]+"${arr[@]}"}` idiom preserves elements containing spaces.

### The thread connecting the three concerns

This gate's job is to be trusted when it blocks a push. Right now three different "the tooling is broken" situations — sibling worktree, missing binary, unresolvable plugin — all print the same message as "your code is wrong", and two of them name no file. That's the property most likely to teach people to reach for `--no-verify` by reflex, which is the failure mode this whole line of work is trying to close.

### Not raised

ARG_MAX exposure on very large diffs, key-order flake from `sort: false` (bounded by `removeUnusedKeys: false`), the `env` multiline false positive, and `--fix` exiting 0 after rewriting (pre-existing for `eslint --fix`). CI is green across all 24 checks; the existing CodeRabbit thread is resolved and not re-litigated.

---

tools/scripts/lint-changed.sh
csiyang: Fixed in `3924543e` — good catch, and you were right that my worktrees are nested, which is exactly why the verification could not have caught this.

`resolve_bin_dir` now refuses when the worktree is not inside the main checkout, rather than handing back binaries whose configs cannot resolve:

```
🛑 - this worktree is not inside the main checkout, so tool plugins cannot resolve
    worktree:      /private/tmp/.../sibling
    main checkout: /Users/siyangcao/CodingProjects/core
    create worktrees under the main checkout (.claude/worktrees/<branch>),
    or run `pnpm install` here to give this checkout its own node_modules
```

Reproduced your failure first (`Cannot find package 'prettier-plugin-tailwindcss'`), then confirmed the guard fires on a sibling worktree and stays out of the way for a nested one. It exits directly rather than returning, so the generic "could not find node_modules/.bin" no longer prints on top of it — that message would have been wrong here, since it *was* found, just not reachable.

Went with the preflight rather than documenting the constraint because the repo convention is `.claude/worktrees/` (nested) anyway, so this only fires for a case that should not arise.

tools/scripts/lint-changed.sh
csiyang: Fixed in `d82ff434` — took your snippet essentially as written.

```bash
for required_bin in prettier eslint i18next-cli; do
  if [ ! -x "$BIN_DIR/$required_bin" ]; then
    echo "🛑 - $required_bin missing from $BIN_DIR" >&2
    echo "    run \`pnpm install\`" >&2
    exit 1
  fi
done
```

Verified against a simulated pruned install (a `.bin` containing only `prettier`): it now reports `🛑 - eslint missing from ...` instead of folding a 127 into a formatting complaint.

Your framing of it as "an infrastructure failure wearing the mask of a content failure" is the thing that made both this and the sibling-worktree case click as the same bug — worth naming, since the script has several places where a non-zero exit gets attributed to the file rather than the environment.

AGENTS.md
csiyang: Resolved by deleting the hook entirely in `d82ff434`, so there is no longer a trigger to document or a promise to soften.

Your `CODEX_HOME` point was the more interesting half — "manual human pushes skip the gate" was not true, and the prefix match I had argued for is what made it untrue. That pushed the design question rather than the wording: what we actually want is not to *block* a push but to apply the fixes and carry them with it, and a hook fundamentally cannot do that. Git resolves which commits to push before hooks run, so a commit created inside the hook is left behind. Confirmed on a throwaway repo — the hook committed, the remote received only the original, and the branch silently ended up ahead.

So the gate is now an agent instruction at the top of `AGENTS.md`: run `pnpm lint:changed --fix`, commit what it changes, push it together. Humans are ungated by construction, which is what we wanted, and all the environment detection is gone.

tools/scripts/lint-changed.sh
csiyang: Kept the behaviour, documented it — and the discussion moved it from "acceptable side effect" to the intended design.

The principle we settled on: anything autofix.ci writes is generated output, not opinion. It is already being committed, just by a bot after the fact. So pulling in a `t()` string someone else left unextracted is correct — it is precisely what the bot would have committed to the PR anyway, and it belongs in a human-authored commit instead.

That also removed `--dry-run`. The check-only mode existed to avoid touching the tree, but under this framing the fixes should be applied and committed, so `--fix` is the path agents take. `AGENTS.md` now says extraction is project-wide rather than per-file, and why that is deliberate.

Thanks for reading `extractor/core/extractor.js` to confirm `anyFileUpdated` is computed before the `isDryRun` guard. That was the one place this could have silently no-opped exactly like the ESLint-only gate did, and I had not verified it that far down.

AGENTS.md
```diff
@@ -31,11 +31,35 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
...
-### Lint before push
+### Lint before push (agents)
 
-Agent pushes are lint-gated. When an agent environment is detected (`CLAUDECODE`/`CURSOR_AGENT`), `.husky/pre-push` runs the `lint:changed` script on committed changes. It lints only changed files, scoped per workspace; full `nx lint` is far too slow. Manual human pushes skip the gate, and `git push --no-verify` skips everything. Run `pnpm lint:changed [--fix]` yourself anytime. [autofix.ci](https://autofix.ci) stays the CI backstop and owns Prettier formatting.
+**Agents: before every `git push`, run this and push the result as one unit.**
 
-Contract of the gate: it lints the branch diff against `origin/main`, refreshed when the network allows. It does not lint the literal ref-range of the push. Forks, other remotes, and branches cut from `stage` are all judged against `origin/main`, so a `stage`-cut branch may over-lint (use `git push --no-verify` if that blocks you). `LINT_CHANGED_JOBS` caps parallel workspaces (default 4).
+```bash
+pnpm lint:changed --fix        # applies formatting, lint fixes and translation extraction
+git add -A && git commit -m "chore: lint fixes"   # only if it changed anything
```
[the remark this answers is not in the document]
csiyang: Fixed in `ec23a5170`. Correct and worth more than its Major rating here — the whole design now rests on agents following that snippet literally, so a snippet that does not match its own comment undermines the change itself.

```bash
# Commit your own work first, so anything dirty afterwards is a generated fix.
pnpm lint:changed --fix

if [ -n "$(git status --porcelain)" ]; then
  git add -A && git commit -m "chore: lint fixes"
fi

git push
```

Guarded the commit on `git status --porcelain` as suggested, and added a line telling the reader to commit their own work first, since `git add -A` will otherwise sweep it in. Went with that rather than a hard clean-worktree precondition: by the time you are pushing, your own work is committed, so the tree should already be clean and the only dirty files are generated ones. Verified both branches — dirty commits, clean skips.

AGENTS.md
```diff
@@ -31,11 +31,35 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
...
+git push
+```
+
+There is deliberately **no pre-push hook**. A git hook cannot do this job: git resolves which commits to push _before_ running the hook, so fixes a hook commits are left behind and never reach the remote. Doing it in the agent, before the push is issued, is the only way the fixes actually travel with the branch. Humans are not gated at all — this is an agent instruction, not enforcement.
+
+What the fixes are: everything [autofix.ci](https://autofix.ci) would otherwise commit to your PR. It is generated output, not opinion, so it belongs in your commit rather than in a bot commit afterwards. `lint:changed` mirrors those steps in the same order:
```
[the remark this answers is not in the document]
csiyang: Fixed in `ec23a5170`. Real contradiction, and the failure mode you named is the right one — an agent reading the first line stops there and never regenerates GraphQL types.

Now scoped, and it names the exception rather than leaving it to a later paragraph:

> What the fixes are: autofix.ci's Prettier, ESLint and i18next steps — everything it would otherwise commit to your PR **except `codegen`**, which is not covered locally (see below).

`codegen` is excluded deliberately: it only runs where `apollo` is installed globally, which the devcontainer and autofix.ci both do, and a plain host checkout does not. So it stays with CI and GraphQL changes still produce a bot commit.

AGENTS.md
```diff
@@ -31,11 +37,42 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
...
+
+```bash
+# Commit your own work first, so anything dirty afterwards is a generated fix.
+pnpm lint:changed --fix
+
+if [ -n "$(git status --porcelain)" ]; then
+  git add -A && git commit -m "chore: lint fixes"
+fi
+
+git push
```
jianwei1: **Concern:** nothing checks whether `lint:changed --fix` succeeded, so the recipe pushes through a failed lint.

`--fix` exits 1 when ESLint finds errors autofix can't repair. Ran the snippet verbatim against that case:

```
🛑 - lint failed (file and rule named above)
 ELIFECYCLE  Command failed with exit code 1.
[main c50a63a] chore: lint fixes
REACHED git push
```

It commits the partial fixes and pushes anyway. An agent following this literally ships known lint errors and reds the `lint` check — with a `chore: lint fixes` commit on top that implies the opposite.

This is the same class as the unconditional-commit bug CodeRabbit caught, one line up. Your reply on that thread says you verified "both branches — dirty commits, clean skips", which is the two branches of the `git status --porcelain` guard; the path where the fixer itself exits non-zero isn't covered by that. And by your own framing there — the whole design rests on agents following the snippet literally — a snippet that proceeds past its own failure is worth closing.

```bash
pnpm lint:changed --fix || { echo "unfixable errors above — fix them, then re-run"; exit 1; }
```

`set -e` at the top would do it too, though the explicit form reads better as an instruction since agents may run these as separate commands rather than as one script.

---

tools/scripts/lint-changed.sh
```diff
@@ -1,27 +1,126 @@
...
+#   1. Prettier   — formatting, across every changed file of any type.
+#   2. ESLint     — code rules, across changed JS/TS only.
+#   3. i18next    — extracted translations, for changed projects only.
+#
+# Deliberately not mirrored, because none of them can produce a commit:
+# type-check and subgraph-check only report (and subgraph-check needs a Hive
+# token); prisma-generate writes nothing that is tracked. codegen is the one
+# real gap — it can commit, but it `rm -rf`s a project's whole __generated__
+# directory before regenerating from a deprecated global apollo CLI, which is
+# not something a push hook should do to a working tree. It stays with CI.
```
jianwei1: **Concern:** this header contradicts itself twice, and both are regressions of fixes made earlier in this same PR.

**1. The coverage claim.** Line 4-5 says a clean run means "autofix.ci has nothing left to commit". Line 13-16 then says "codegen is the one real gap — it can commit… It stays with CI."

This is the exact defect CodeRabbit raised on these lines, which you fixed in `873de128` ("took your wording verbatim in both places"). `f375ebe2` rewrote this block when the i18next pass landed and lost the scoping. It's also the same defect you fixed in AGENTS.md in `ec23a51` — the doc got scoped to "Prettier, ESLint and i18next … except `codegen`", the script didn't. Both threads are resolved, so nothing is watching for the drift.

The failure mode is the one you named yourself on the AGENTS.md thread: a reader who stops at the header concludes codegen is covered and never regenerates GraphQL types.

**2. Stale hook reference.** Line 16 — "not something a **push hook** should do to a working tree" — against line 41, "There is **no git hook** on this." Left over from before `d82ff43`.

Suggested for line 4-5:

```
# Mirrors the autofix.ci steps this can apply — Prettier, ESLint and i18next —
# in the same order, so a clean run here leaves autofix.ci nothing to commit
# except codegen (see below):
```

and line 16 → "…which is not something a pre-push step should do to a working tree."

---

jianwei1 (commented):
## Re-review at `ec23a51`

**Critical:** none. **Concerns:** 2 (inline). CI green across 21 checks.

### My three earlier concerns — all fixed, all re-verified

| Concern | I verified |
|---|---|
| Sibling worktree failing as "formatting issues" | Sibling → refused up front, both paths named, exit 1. Nested → passes, exit 0 |
| Missing binary reported as a content failure | Removed a stub from `.bin` → `🛑 - i18next-cli missing from …`, exit 1 |
| Codex detection doc drift | Surface removed entirely with the hook |

The worktree guard tests the right thing — whether Node's upward walk from the worktree actually reaches `MAIN_ROOT/node_modules` — rather than approximating it. And `exit 1` instead of `return 1`, with the comment explaining that the caller's "could not find node_modules/.bin" would be actively wrong there, is the kind of detail that makes the next failure legible.

### On the pivot

The premise holds up: generated output belongs in a human commit, not a bot commit on top of it. The throwaway-repo check that a hook's commit gets left behind is the sort of thing most people would assume rather than test, and it's the right thing to have tested.

Also confirmed, since the whole recipe depends on it: `pnpm lint:changed --fix` forwards `--fix` correctly through pnpm 10.33.4, in both `pnpm run` and shorthand form — pnpm does swallow some flags, and this isn't one of them.

And since the new guard actively steers people toward nested worktrees, I checked that they don't get swept into the file list: `git ls-files --others` reports them as directory paths and the `-f` guard drops them. `git add -A` in the recipe warns about the embedded repo but stages nothing.

### The two inline items

Both are small, and both sit just outside what's already been checked — one past the branches your verification covered, one in a block a later commit rewrote after it had been fixed.

Leading AGENTS.md with the push rule rather than burying it at line 34 is a real improvement, separately from any of this.

---

AGENTS.md
csiyang: Fixed in `165595a81`. You are right, and the part I want to acknowledge is the process point: my "verified both branches" covered the two branches of the `git status --porcelain` guard, not the path where the fixer itself exits non-zero. Different axis, and I did not test it.

Reproduced it before fixing. My first attempt was a bad probe — an unused variable, which is a *warning* here, so ESLint exited 0 and nothing fired. Used a real error-level rule instead (`@typescript-eslint/no-floating-promises`):

```
=== the documented recipe, verbatim ===
→ GUARD FIRED: unfixable errors, stopped before push
recipe exit=1

=== and the old, unguarded recipe for comparison ===
→ REACHED git push  (this is what jianwei1 reported)
```

Took your explicit form rather than `set -e`, for the reason you gave — agents run these as separate commands, so a per-command guard survives that:

```bash
pnpm lint:changed --fix || { echo "unfixable errors above — fix them, then re-run"; exit 1; }
```

tools/scripts/lint-changed.sh
csiyang: Fixed in `165595a81`. Confirmed the regression exactly as you described:

```
$ git show 873de128 -- tools/scripts/lint-changed.sh | grep "left to commit"
-# means autofix.ci has nothing left to commit:
+# means autofix.ci has no formatting or linting changes left to commit:

$ git show f375ebe2 -- tools/scripts/lint-changed.sh | grep "left to commit"
-# means autofix.ci has no formatting or linting changes left to commit:
+# clean run here means autofix.ci has nothing left to commit:
```

So I fixed it, then reverted it four commits later while rewriting the block for the i18next pass, having told CodeRabbit I had taken its wording "verbatim in both places". Your point that both threads were resolved so nothing was watching is the real lesson — resolving a thread retires the only check on that line.

Header now reads:

```
# Mirrors the autofix.ci steps this can apply — Prettier, ESLint and i18next —
# in the same order, so a clean run here leaves autofix.ci nothing to commit
# except codegen (see below):
```

and the stale "push hook" phrasing on line 16 is now "pre-push step", left over from `d82ff43` as you spotted.

AGENTS.md
```diff
@@ -31,11 +37,42 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
...
+
+```bash
+# Commit your own work first, so anything dirty afterwards is a generated fix.
+pnpm lint:changed --fix || { echo "unfixable errors above — fix them, then re-run"; exit 1; }
+
+if [ -n "$(git status --porcelain)" ]; then
+  git add -A && git commit -m "chore: lint fixes"
+fi
+
+git push
```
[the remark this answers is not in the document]
csiyang: Fixed in `be0d9fd2a`. Took the guards as proposed, and the finding is stronger than the framing suggests — in a worktree the commit does not merely *risk* failing, it fails every time:

```
$ git commit -m "chore: lint fixes"
.husky/pre-commit: line 2: .husky/_/husky.sh: No such file or directory
COMMIT EXIT=1
```

`.husky/_` is generated by `husky install` during `pnpm install`, and worktrees have no `node_modules`, so it is never there. Worktrees are the normal workflow in this repo, so an agent following the recipe applied the fixes, silently failed to commit them, and pushed without them — the exact bug this PR exists to fix, arriving through its own instructions.

So the guards alone would have turned "silently pushes unlinted" into "always fails in a worktree". Added `--no-verify` to the generated commit as well:

```bash
git add -A || { echo "could not stage generated fixes; aborting push"; exit 1; }
git commit --no-verify -m "chore: lint fixes" || { echo "could not commit generated fixes; aborting push"; exit 1; }
```

Narrow and justified rather than blanket — the pre-commit hook only validates the branch name, already validated at branch creation, and commitlint accepts this exact message (`echo "chore: lint fixes" | commitlint` → exit 0). Both hooks would pass on content; they fail purely on the missing bootstrap. The inline comment says not to carry `--no-verify` over to your own commits.

Verified end to end in a worktree: fixes applied, commit succeeds, recipe reaches push. Raising husky-in-worktrees separately, since it costs a `--no-verify` on every commit made from one.

AGENTS.md
```diff
@@ -31,11 +37,47 @@ This is an **Nx monorepo** (TypeScript). Apps live in `apps/`, GraphQL APIs in `
...
+# Commit your own work first, so anything dirty afterwards is a generated fix.
+pnpm lint:changed --fix || { echo "unfixable errors above — fix them, then re-run"; exit 1; }
+
+if [ -n "$(git status --porcelain)" ]; then
+  git add -A || { echo "could not stage generated fixes; aborting push"; exit 1; }
+  # --no-verify because husky's .husky/_ bootstrap only exists where pnpm install
+  # has run, so this commit fails in a worktree. The hooks would pass on content:
+  # the branch name was validated at creation and commitlint accepts this exact
+  # message. Do not carry --no-verify over to your own commits.
+  git commit --no-verify -m "chore: lint fixes" || { echo "could not commit generated fixes; aborting push"; exit 1; }
```
jianwei1, line 55: **Concern:** the last line tells agents not to use the only thing that works in a worktree.

Your diagnosis is right, and it generalises further than the comment allows: the missing bootstrap breaks **every** commit in a worktree, not just this one. Reproduced independently:

```
=== MAIN ===                  === WORKTREE ===
pre-commit ran                .husky/pre-commit: 2: .: cannot open .husky/_/husky.sh
main commit OK                COMMIT EXIT=1
```

With `--no-verify` in the worktree: `OK`. So the recipe's own first line — "Commit your own work first" — hits the same wall, and "Do not carry `--no-verify` over to your own commits" tells the agent not to use the workaround three lines below it. Meanwhile the sibling-worktree guard added earlier in this PR steers people *into* worktrees.

You've already called this — "raising husky-in-worktrees separately, since it costs a `--no-verify` on every commit made from one" — and that's the right split; this shouldn't grow into a husky fix. The ask is just that the note not be actively misleading until then. One clause:

```
# message. Do not carry --no-verify over to your own commits — though in a
# worktree those hit the same missing bootstrap today; tracked separately.
```

**Separately, line 52 — "the branch name was validated at creation".** Nothing validates at `git worktree add -b`; `validate-branch-name` runs *from this same pre-commit hook*, so it validates at first commit — and if every worktree commit uses `--no-verify`, never locally at all. The real backstop is autofix.ci's `validate-branch-name -t $GITHUB_HEAD_REF` step, which is a stronger justification than the one written. Worth saying instead:

```
# the branch name is checked by autofix.ci on the PR, and commitlint accepts
# this exact message.
```

Two things for the separate issue, since I tested them:

- Guarding the source line clears the bootstrap failure — `[ -f "$(dirname "$0")/_/husky.sh" ] && . "$(dirname "$0")/_/husky.sh"` — worktree commit succeeds, and the hook body still runs where the bootstrap exists. Husky v9 dropped that line entirely; the repo pins `^8.0.3`.
- `npx validate-branch-name` then needs to resolve too. In the nested layout `resolve_bin_dir` mandates, npx should walk up to the main checkout's `.bin` by the same mechanism the guard already relies on — worth confirming rather than assuming.

---

jianwei1 (commented):
## Re-review at `be0d9fd2`

**Critical:** none. **Concern:** 1 (inline), doc-only — no behaviour change. CI green across 22 checks.

### Both my last concerns are fixed

**The recipe's failure check** — now guarded. Worth noting how you got there: your first probe was an unused variable, which is a *warning* here, so ESLint exited 0 and nothing fired; you switched to `@typescript-eslint/no-floating-promises` to get a real error-level failure. Catching that your own test couldn't fail is the part that matters.

**The script header** — rescoped to "…nothing to commit except codegen (see below)", and `push hook` → `pre-push step`. You pulled both `git show` diffs to confirm the fix-then-revert across `873de128` → `f375ebe2`, and drew the general lesson yourself: resolving a thread retires the only check on that line. That's the durable takeaway from this whole PR, more than any individual fix in it.

### On the `--no-verify` addition

Good change, and carefully scoped rather than a shrug — narrowed to one commit with a fixed message, both hooks checked for what they'd actually do, `commitlint` verified against that exact string, and the exception documented inline where someone will read it.

Turning CodeRabbit's "can continue to push" into "fails every time in a worktree" is a sharper finding than the one reported. Worth spelling out what that implies: the guards *alone* would have converted a silent wrong push into a hard stop with no way through, so the `--no-verify` wasn't optional polish on top of them — it was what kept the fix from replacing one bug with another.

The inline comment is the one thing left, and it's a clause.

### Overall

The substance is done. What remains is one clause in a code comment and one word in a justification, neither of which changes behaviour — if you'd rather fold both into the husky-in-worktrees issue and merge this as-is, I have no objection.

Four rounds in, every fix on this PR has come with a reproduction attached. That's why it converged.

---

AGENTS.md
csiyang: Both fixed in `03658db8d`. The second one I had simply asserted without checking, and it was wrong.

Verified before changing it:

```
$ cat .husky/pre-commit
. "$(dirname "$0")/_/husky.sh"
npx validate-branch-name

$ grep -A2 "lint branch name" .github/workflows/autofix.ci.yml
run: pnpm exec validate-branch-name -t $GITHUB_HEAD_REF
```

So validation happens at first commit via the hook, not at `git worktree add -b`, and with `--no-verify` on every worktree commit it never runs locally at all. autofix.ci is the real backstop. Comment now says that instead.

On the misleading note — you are right that it warned against the only thing that works, three lines below using it, and that the recipe's own "commit your own work first" hits the same wall. Took your clause:

```
# this exact message. Do not carry --no-verify over to your own commits — though
# in a worktree those hit the same missing bootstrap today; tracked separately.
```

Thanks for the husky research — the `[ -f ... ] &&` guard on the source line, husky v9 having dropped it while we pin `^8.0.3`, and the point that `npx validate-branch-name` then has to resolve too. That is most of the investigation for the separate issue; I will carry it over rather than rediscover it, and confirm the npx resolution rather than assume it works the way `resolve_bin_dir` does.

jianwei1 (approved):
## Approving at `03658db8`

Both items from the last round are fixed, taken as suggested, and the `diff` confirms the commit touches exactly those two lines and nothing else. CI green across 21 checks; all 11 threads resolved.

Worth noting you checked the claim rather than taking my word for it — pulling `.husky/pre-commit` and the `lint branch name` step out of `autofix.ci.yml` before rewriting the line, and saying plainly that the original had been asserted without checking. Same on the husky research: carrying it into the separate issue rather than rediscovering it, and flagging that you'll confirm the `npx validate-branch-name` resolution rather than assume it behaves the way `resolve_bin_dir` does. Assuming that is exactly how the `--config` resolution question would have bitten.

### Everything raised across the review

| Raised | Outcome |
|---|---|
| Sibling worktree failing as a formatting error | Refused up front with an accurate message |
| Missing binary reported as a content failure | Explicit `-x` guard naming the binary |
| Codex detection doc drift | Surface removed with the hook |
| Recipe pushing through a failed lint | Guarded; reproduced with a real error-level rule |
| Script header self-contradiction ×2 | Rescoped; stale `push hook` phrasing gone |
| `--no-verify` note misleading + wrong justification | Both corrected |

### What holds up

The mechanism is verified, not just read. `--dry-run --ci` genuinely reports drift — `anyFileUpdated` is computed before the `isDryRun` write guard, which was the one place this could have silently no-opped the way the ESLint-only gate did. `--ignore-unknown` is required and safe at both edges, including all-`.prettierignore`d file sets, so `/sync-ce` pushes don't spuriously fail. The 249 bracketed Next.js route paths pass through Prettier intact. i18n grouping is complete at five configs, since `apps/arclight`'s `extract-translations` target is an empty no-op. And `pnpm lint:changed --fix` forwards `--fix` correctly through pnpm 10.33.4, which the whole recipe depends on.

### The pattern worth keeping

Every fix on this PR arrived with a reproduction attached, and twice the reproduction corrected the fix: the unused-variable probe that turned out to be a *warning* and so proved nothing, and CodeRabbit's "can continue to push" that turned out to be "fails every time in a worktree" — which is what made `--no-verify` necessary rather than optional, since the guards alone would have replaced a silent wrong push with a hard wall.

The most durable takeaway is your own, from the header regression: resolving a thread retires the only check on that line. That one had been fixed, reverted four commits later, and confirmed fixed to CodeRabbit in between — with both threads closed, nothing was watching.

Known remaining gaps, both deliberate and both documented: `codegen` stays with CI, and husky-in-worktrees is tracked separately.

---

PR #9499, opened 2026-08-17
fix(NES-1866): delete selected card when a deselected edge is backspaced

Estherqa:
## Description

Fixes [NES-1866](https://linear.app/jesus-film-project/issue/NES-1866/backspace-deletes-a-journey-map-connection-you-already-deselected).

On the journey map, clicking a connection arrow selects it (highlights, shows the delete X). Clicking empty canvas visibly deselects it. But the delete handler kept its **own** copy of the selection and never cleared it on deselect, so pressing **Backspace/Delete** removed the connection the user believed was deselected and orphaned the downstream card, with nothing selected on screen.

## Root cause

In `useDeleteOnKeyPress`, `onSelectionChange` only updated the remembered selection when an edge was present, and did nothing on deselect, so the remembered selection drifted out of sync with what is visible. Every sibling selection tracker (`BaseEdge`, `BaseNode`, `CustomEdge`) already handles the empty/deselect case; this hook was the outlier.

## Fix

Add the missing `else` branch so an empty selection falls back to the currently selected card (`StepBlock`), else clears it. Deselect the arrow, and Backspace now deletes the selected card (the ticket expected behaviour). Deleting a genuinely selected connection is unchanged.

## Testing

- Added a unit test: select an edge, deselect it, press Delete, and the selected card is deleted while `deleteEdge` is not called.
- Manual QA on the preview: open a journey with 2+ connected cards, go to the journey map, click an arrow, click empty canvas, then Backspace deletes the selected card, not the connection.

Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keyboard deletion so that, after deselecting an edge, pressing Delete removes the selected card instead of the previously selected edge.

* **Tests**
  * Added regression coverage for switching selection from an edge to a card.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useDeleteOnKeyPress/useDeleteOnKeyPress.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Slider/JourneyFlow/libs/useDeleteOnKeyPress/useDeleteOnKeyPress.spec.tsx
@@ -114,4 +114,53 @@ describe('useDeleteOnKeyPress', () => {
     })
     await waitFor(async () => expect(deleteBlock).toHaveBeenCalled())
   })
+
+  it('should delete the selected card when a previously selected edge is deselected', async () => {
+    mockUseKeyPress.mockReturnValue(false)
+    const stepBlock = {
+      __typename: 'StepBlock',
+      id: 'step.id'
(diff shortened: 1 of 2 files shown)


Estherqa:
Closed automatically when the branch was renamed to lead with the ticket ID. Replaced by #9500 (same commit, correct branch/title). Please review there.

PR #9501, opened 2026-08-18
docs(context-map): give both remaining sections a bolded Handoff verdict

csiyang:
Docs only. Two `validate-map` warnings against `main` prompted this; one was a real gap in the map, the other was not.

**`apps/journeys/CONTEXT-intake.md` § "Chat / AI assistant"** genuinely lacked an explicit verdict — it had one embedded in `**Status:**` and no `**Handoff:**` line — so it gains one line, `**Handoff:** route to a human.` Every section closes with an explicit agent-able / human / how-to-FAQ / ops verdict, so a reader has one consistent place to look; the `**Status:**` line is untouched and still explains why there is nothing more to say.

**`apps/journeys-admin/CONTEXT-intake.md` § Integrations is deliberately untouched.** Its warning is a bug in the hatsu validator, not a problem with the map: `CONTEXT-MAP-contract.md` says a label line matches "after stripping leading indentation and an optional `- ` list marker", and `extractHandoff` is `/^\s*\*\*Handoff:\*\*/`, which never strips the marker. The bullet form is contract-conforming, so it stays, and the parser is fixed in [JesusFilm/hatsu#6](https://github.com/JesusFilm/hatsu/pull/6). **With that PR this branch is clean; against hatsu `main` before it lands, the one warning is the parser bug and not a regression here.** The two PRs are independent — either order works.

**`CONTEXT-MAP-contract.md`** stops describing what `main` happens to look like — the observation that went stale and made this PR necessary in the first place. Its "known deviation on `main`" note becomes a one-sentence rule: a section with no diagnosis layer yet still closes with a verdict.

§Section anatomy also stops teaching authors the parser's line forms: bold labels are stated as the recommended layout, not syntax, and consumers read bodies as prose — the only thing checked is that a section closes with a `**Handoff:**` verdict, in any line form.

### Verification

- `validate-map` against this branch, both measured: **0 errors, 0 warnings** with [hatsu#6](https://github.com/JesusFilm/hatsu/pull/6)'s parser fix, **0 errors, 1 warning** against hatsu `main` before it — the bullet-form bug described above.
- Prettier clean on both changed files.
- Diff against `main` is 2 files, +10/−14: one added line in journeys, the rest the contract.

Linear: ENG-3706

## Summary by CodeRabbit

* **Documentation**
  * Clarified accepted formatting for section labels, including flush-left, indented, and bulleted forms.
  * Clarified that bold labels may include their associated values.
  * Updated requirements so every section includes a handoff verdict, including sections without a diagnosis layer.
  * Added guidance to escalate out-of-scope Chat or AI assistant issues to a human.

### DIFF
--- a/CONTEXT-MAP-contract.md
+++ b/CONTEXT-MAP-contract.md
@@ -58,16 +58,13 @@ Authors may add further keys freely (additive); consumers must ignore unknown ke
   `— T4 (cache) · T5 (optimistic drift)`, `— T6, expectation mismatch`). Sections with no T-token
   are untyped (how-to / FAQ / flagged entries) — allowed.
 - `type_tags` in frontmatter equals the union of T-tags found in the file's headings.
-- Bodies are built from bold-label lines (`**Signatures:**`, `**Localizing question (reporter):**`,
-  `**Then ask:**`, `**Ready when:**`, `**Status:**`, `**Look first (fixer):**`, `**Handoff:**`, …).
-  A label line matches after stripping leading indentation and an optional `- ` list marker — the
-  flush-left, indented (list-item continuation), and list-bullet (`- **Handoff:** …`) forms are
-  all accepted. The bold may extend past the label over the whole line, label and value
-  together (e.g. `**Localizing question (reporter): does a refresh fix it?**`).
-  The label vocabulary is **open** — authors may introduce new labels; consumers must not assume a
-  fixed set or order — with one exception:
-- **`Handoff:` is reserved.** Every section closes its diagnosis with a `**Handoff:**` verdict
-  (agent-able / human / how-to-FAQ / ops), the one label consumers may key on semantically.
+- Bodies are prose. Bold labels (`**Signatures:**`, `**Localizing question (reporter):**`,
+  `**Then ask:**`, `**Ready when:**`, `**Status:**`, `**Look first (fixer):**`, …) are the
+  recommended layout so a reader knows where to look; the vocabulary is open, and consumers read
+  bodies as prose — they do not parse labels.
+- **One exception: `Handoff:`.** Every section closes with a `**Handoff:**` verdict
+  (agent-able / human / how-to-FAQ / ops). The validator warns when a section lacks one, and
+  recognises it flush-left, indented, or as a `- ` bullet.
   - A section with no diagnosis layer yet still closes with one: route it to a human, and let a
     `**Status:**` line explain why there is nothing more to say.
 
--- a/CONTEXT-MAP-contract.md
+++ b/CONTEXT-MAP-contract.md
@@ -61,17 +61,19 @@ Authors may add further keys freely (additive); consumers must ignore unknown ke
 - Bodies are built from bold-label lines (`**Signatures:**`, `**Localizing question (reporter):**`,
   `**Then ask:**`, `**Ready when:**`, `**Status:**`, `**Look first (fixer):**`, `**Handoff:**`, …).
   A label line matches after stripping leading indentation and an optional `- ` list marker — the
-  flush-left, indented (list-item continuation), and list-bullet (`- **Handoff:** …`) forms all
-  occur on `main` today. The bold may extend past the label over the whole line, label and value
-  together (e.g. `**Localizing question (reporter): does a refresh fix it?**`).
+  flush-left, indented (list-item continuation), and list-bullet (`- **Look first (fixer):** …`)
+  forms all occur on `main` today. The reserved `Handoff:` label below is the exception: consumers
+  key on it semantically, so it is written flush-left or indented, never as a list bullet. The bold
+  may extend past the label over the whole line, label and value together (e.g.
+  `**Localizing question (reporter): does a refresh fix it?**`).
   The label vocabulary is **open** — authors may introduce new labels; consumers must not assume a
   fixed set or order — with one exception:
 - **`Handoff:` is reserved.** Every section closes its diagnosis with a `**Handoff:**` verdict
   (agent-able / human / how-to-FAQ / ops), the one label consumers may key on semantically.
-  - _Known deviation on `main` today: `apps/journeys/CONTEXT-intake.md` § "Chat / AI assistant"
-    embeds its verdict in `**Status:**` with no `**Handoff:**` line — left as-is deliberately
-    (unreleased feature, plan not settled; Siyang, 2026-08-13). The validator reports it as a
-    warning until the section gets a real diagnosis layer._
+  - A section with no diagnosis layer yet still closes with one: `apps/journeys/CONTEXT-intake.md`
+    § "Chat / AI assistant" carries `**Handoff:** route to a human.` while its `**Status:**` line
+    explains why there is nothing more to say. "Too new to diagnose" is a verdict, not an excuse
+    for omitting one.
 
 ## Index ↔ frontmatter sync
 
--- a/CONTEXT-MAP-contract.md
+++ b/CONTEXT-MAP-contract.md
@@ -61,18 +61,15 @@ Authors may add further keys freely (additive); consumers must ignore unknown ke
 - Bodies are built from bold-label lines (`**Signatures:**`, `**Localizing question (reporter):**`,
   `**Then ask:**`, `**Ready when:**`, `**Status:**`, `**Look first (fixer):**`, `**Handoff:**`, …).
   A label line matches after stripping leading indentation and an optional `- ` list marker — the
-  flush-left, indented (list-item continuation), and list-bullet (`- **Look first (fixer):** …`)
-  forms are all accepted. The reserved `Handoff:` label below is the exception: consumers key on it
-  semantically, so it is written flush-left or indented, never as a list bullet. The bold may
-  extend past the label over the whole line, label and value together (e.g.
-  `**Localizing question (reporter): does a refresh fix it?**`).
+  flush-left, indented (list-item continuation), and list-bullet (`- **Handoff:** …`) forms are
+  all accepted. The bold may extend past the label over the whole line, label and value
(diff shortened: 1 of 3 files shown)


CONTEXT-MAP-contract.md
```diff
@@ -58,20 +58,15 @@ Authors may add further keys freely (additive); consumers must ignore unknown ke
...
-    embeds its verdict in `**Status:**` with no `**Handoff:**` line — left as-is deliberately
-    (unreleased feature, plan not settled; Siyang, 2026-08-13). The validator reports it as a
-    warning until the section gets a real diagnosis layer._
+- Bodies are prose. Bold labels (`**Signatures:**`, `**Localizing question (reporter):**`,
+  `**Then ask:**`, `**Ready when:**`, `**Status:**`, `**Look first (fixer):**`, …) are the
+  recommended layout so a reader knows where to look; the vocabulary is open, and consumers read
+  bodies as prose — they do not parse labels.
+- **One exception: `Handoff:`.** Every section closes with a `**Handoff:**` verdict
+  (agent-able / human / how-to-FAQ / ops). The validator warns when a section lacks one, and
+  recognises it flush-left, indented, or as a `- ` bullet.
```
jianwei1: **Concern:** this asserts validator behaviour that the PR description says doesn't exist.

Here: _"The validator warns when a section lacks one, and recognises it flush-left, indented, or as a `- ` bullet."_

The PR body: `extractHandoff` is `/^\s*\*\*Handoff:\*\*/`, which "never strips the marker" — which is precisely why `apps/journeys-admin/CONTEXT-intake.md:217` (`- **Handoff:** OAuth reconnect → …`) still warns. So the contract now states as present fact the one form that is currently broken.

The stated goal of this PR was to stop describing what `main` happens to look like. This swaps a stale description of the files for a false description of the tool — and it's the more expensive kind, because an author who trusts it writes a bullet Handoff and gets an unexplained warning.

Suggest framing it as a requirement on consumers rather than a report on one of them:

> All three forms — flush-left, indented, and `- ` bullet — are contract-conforming; consumers must accept them.

The semantics (agent-able / human / how-to-FAQ / ops is what consumers key on) are what the contract is for; which line forms a given validator currently parses is an implementation detail that will go stale again.

---

CONTEXT-MAP-contract.md
```diff
@@ -58,20 +58,15 @@ Authors may add further keys freely (additive); consumers must ignore unknown ke
...
+  - A section with no diagnosis layer yet still closes with one: route it to a human, and let a
+    `**Status:**` line explain why there is nothing more to say.
```
jianwei1, line 68: **Concern:** deleting the deviation note leaves the remaining warning undocumented.

The old `_Known deviation on main today…_` note was the in-repo record of *why* `validate-map` wasn't clean. Replacing the stale part of it is right — but nothing takes over that job. The PR description acknowledges one warning survives ("expected, not a regression"), and that acknowledgement lives in a PR body nobody reads six weeks from now. Post-merge the repo has a validator reporting 1 warning and no in-repo explanation, which is the exact failure mode the old note existed to prevent.

Worth one line, scoped to the tool rather than to the map, so it deletes itself when fixed:

> _Until the hatsu parser strips the `- ` list marker (ENG-…), `validate-map` warns on the bullet form in `apps/journeys-admin/CONTEXT-intake.md` § Integrations. Expected; remove this note when the fix lands._

Unlike the note it replaces, that one has a defined end date.

---

jianwei1 (commented):
## Review summary

Docs-only, tightly scoped, and the direction is right — the contract should state rules, not snapshot what `main` currently looks like. Two things to consider before merge, both about the replacement text rather than the change itself.

**Concerns (inline, both on `CONTEXT-MAP-contract.md`):**

1. **L66-67 asserts validator behaviour this PR says is broken.** The new text claims the validator "recognises it flush-left, indented, or as a `- ` bullet"; the PR description says `extractHandoff` is `/^\s*\*\*Handoff:\*\*/` and never strips the marker. It swaps a stale description of the files for a false description of the tool.
2. **L68-69 — the deleted deviation note leaves the surviving warning undocumented.** The PR acknowledges one warning persists by design, but only in the PR body. After merge, `validate-map` reports 1 warning with nothing in-repo explaining it.

**Nits (not worth inline threads):**

- _"Every section closes with a `**Handoff:**` verdict"_ is literally violated by an existing section: in `apps/journeys-admin/CONTEXT-intake.md` the § Integrations Handoff sits at `:217` with the `**Growth Spaces:** dormant…` note following at `:218`, so it doesn't close the section. The old wording ("closes **its diagnosis** with") tolerated that; the new wording doesn't. "Every section carries a `**Handoff:**` verdict, closing its diagnosis" would keep the rule true. In a document whose whole purpose is precision, this one matters more than it looks.
- The deleted sentence about bold spanning label and value is still the live form at `apps/journeys-admin/CONTEXT-intake.md:39` and `:196`. Harmless now that consumers read bodies as prose, but that form is now undocumented.
- `apps/journeys/CONTEXT-intake.md` frontmatter still reads `updated: 2026-08-13`. The contract defines `updated` as "date of last substantive edit", and this PR argues the missing verdict was a real gap in the map — substantive by its own framing.

**Verified independently:** all 18 sections across both intake files now carry exactly one `**Handoff:**` (journeys 7/7 including the added one, journeys-admin 11/11), so the tightened rule holds repo-wide as of this branch.

**What's good here:**

- The rule generalises instead of carving out an exception — "a section with no diagnosis layer yet still closes with one: route it to a human, and let `**Status:**` explain why" is something any future thin section can follow, and it's what made the old note deletable.
- Correctly declines to change the journeys-admin bullet. Bending contract-conforming content to fit a buggy parser would be the wrong fix, and the PR says so with the regex to back it.
- The added line matches the neighbouring § Custom domains section's `**Status:**` + `**Handoff:**` shape exactly, blank lines included.

**CI:** green apart from `danger`, which fails only on "Please request a reviewer for this PR" — process, not code.

---

CONTEXT-MAP-contract.md
csiyang: Agreed on the problem, took a different fix: the clause is deleted rather than reworded (570c138eb).

`All three forms — flush-left, indented, and `- ` bullet — are contract-conforming; consumers must accept them` is still a rule about line forms, phrased as a requirement instead of a report. That's the thing this contract just spent three commits removing: a body is prose, and how a label is laid out isn't the contract's business. Rewording keeps a sentence that has to be maintained against a parser; deleting it means there's nothing left to go stale.

So the bullet is now:

> **One exception: `Handoff:`.** Every section carries a `**Handoff:**` verdict (agent-able / human / how-to-FAQ / ops), closing its diagnosis. The validator warns when a section lacks one.

Your semantics point is exactly right and is what survives — agent-able / human / how-to-FAQ / ops is what consumers key on.

Two of the nits went in with it. **"closes with"** was literally false, as you spotted — § Integrations has `**Growth Spaces:**` after the verdict — so it's "carries … closing its diagnosis", which also drops a positional requirement. And the label example list is down from six to two, since a sentence whose point is that consumers don't parse labels doesn't need six of them.

Declined: restoring the bold-spanning-label sentence. It's layout mechanics, and adding one form back invites the rest.

CONTEXT-MAP-contract.md
csiyang: Declining this one, deliberately — and it's the item I went back and forth on most.

The note would be a record of a hatsu parser bug living in core's authoring contract, and it's time-bound. That's the same shape as the `_Known deviation on main today…_` note it would replace, which went stale twice and is what made this PR necessary in the first place. "Remove this note when the fix lands" doesn't change that — the old note had an owner and a date too.

The warning is recorded where it's actionable: **ENG-3742** in the hatsu repo, which now scopes it as a detection fix (`extractHandoff` should strip an optional `- ` marker), plus the PR description here. And the person who runs `validate-map` is a hatsu dev with that ticket, not a map author reading this contract — the contract's audience never sees the warning.

You're right that the PR body isn't a durable record. I'd rather that pressure push the parser fix along than add a fourth time-bound note to a document we've now spent four commits de-timestamping.

Also declined, for completeness: bumping `updated:` on `apps/journeys/CONTEXT-intake.md`. Fair catch on the internal inconsistency, but the section is behind a flag and unreleased — nothing about the diagnosis changed, only where the verdict is written, so the date would overstate it.

And thank you for independently checking all 18 sections carry exactly one verdict. That's the check that made the tightened rule safe to state.

jianwei1 (approved):
## Approving

Both concerns from the previous round are closed.

**The false validator claim is gone** — and deleting the clause was the better fix. My suggestion was to restate it as a requirement on consumers; you were right that this is still a rule about line forms, which is exactly what the contract spent three commits removing. A reworded sentence has to be maintained against a parser forever; a deleted one can't go stale. What survives is the semantics — agent-able / human / how-to-FAQ / ops — which is the part consumers actually key on.

**The undocumented-warning concern I'll drop**, on your reasoning: the note would put a time-bound record of a hatsu parser bug inside core's authoring contract, which is structurally the same thing as the "known deviation on `main`" note that went stale twice and started this PR. With ENG-3742 and now [hatsu#6](https://github.com/JesusFilm/hatsu/pull/6) tracking it, it's recorded where it's actionable and where the person who actually sees the warning will look.

Worth noting what the review process produced here: the PR got *smaller* twice under scrutiny — the reverts in `0308485`, then the deletion in `570c138`. What lands is one added line in a map file plus the contract wording that made it necessary, and the contract no longer contains a single sentence describing what `main` happens to look like or what a particular tool happens to do today. That's the durable version.

Verified at `570c138`: every section carries exactly one `**Handoff:**` verdict (journeys 7/7, journeys-admin 11/11); `apps/journeys-admin/CONTEXT-intake.md` is byte-identical to `main`, so the reverts held; Prettier clean on all three files; CI fully green.

---

PR #9504, opened 2026-08-18
feat(api-media): durable, retriable video-to-Algolia sync (QA-543)

tanflem:


tanflem:

Closing as **already shipped** — QA-543 landed on `main` via #9442 on 2026-08-10, eight days before this PR was opened from the stale `helm/qa-543-durable-video-algolia-sync` branch. #9513 flagged this as a likely duplicate; this confirms it.

**19 of the 22 files are byte-identical to `main`**, including the entire `apis/api-media/src/workers/videoAlgoliaSync/` directory and all of `algoliaVideoUpdate`.

The three that differ — `updateAvailableLanguages.ts`, `updateAvailableLanguages.spec.ts`, `video.spec.ts` — differ because `main` moved *further ahead*, not because this branch holds unique work. Both of the branch's real changes are already there:

- switching the call site from `updateVideoInAlgolia` to `enqueueVideoAlgoliaSync(videoId, videoOnlyScope, logger)` — `main` line 80
- extracting `findContainerParentIds` — `main` line 153

**Merging this would be a regression.** `main` has since replaced `addLanguageToVideo`'s read-then-set with an atomic `$executeRaw` UPDATE, specifically to close a lost-update race:

> Concurrent published uploads for the same video (different languages finishing around the same time) must not lose each other's language. A read-then-set (`findUnique` → `set: [...current, next]`) is a classic lost-update race.

This branch still carries that racy `findUnique` → `set: [...currentLanguages, languageId]` version, so merging it would reintroduce the bug `main` fixed.

Nothing is lost by closing. Follow-up work on this area is tracked in #9515.

PR #9505, opened 2026-08-18
fix(api-media): stop mux-videos processDownloads retry loop on permanently-invalid Mux metadata (VMT-239 follow-up)

tanflem:


tanflem:

Closing as **already fixed on main** — all three commits on this branch are redundant:

| Commit | Status on `main` |
| --- | --- |
| `f6a209445` stop persisting zero size/bitrate renditions | `hasValidRenditionMetadata` is already defined in `apis/api-media/src/lib/downloads/processing.ts` (line 68) and applied at the two call sites (lines 227, 260) |
| `c08d5539f` lint fixes | git dropped it during rebase: *"patch contents already upstream"* |
| `034d2fbcd` stop the `processDownloads` retry loop | the bug it fixes no longer exists — see below |

The last commit is the only one worth explaining. It guarded against a variant whose Mux metadata is still incomplete being re-selected forever *within a single run*, by tracking `attemptedVariantIds`. That was a real bug against the old **un-paginated** `while (hasMore)` loop.

`main` has since replaced that loop with a cursor-paginated one over `videoVariantDownload`:

```ts
nextCursor = zeroMetadataDownloads.at(-1)?.id ?? null
if (zeroMetadataDownloads.length < take || nextCursor == null) hasMore = false
```

With `cursor: { id: nextCursor }, skip: 1` and `orderBy: [{ videoVariantId: 'asc' }, { id: 'asc' }]`, the cursor advances monotonically whether or not the repair succeeded, so an unfixable variant simply falls out of the run. The `attemptedVariantIds` guard has nothing left to protect.

`main`'s selection predicate is also strictly broader than this branch's — `OR: [{ size: null }, { size: 0 }, { bitrate: null }, { bitrate: 0 }]` versus `OR: [{ size: 0 }, { bitrate: 0 }]` — so it catches null-metadata rows this branch would have missed.

No code from this branch is being lost. Re-open if you disagree with the reading of the cursor loop.

PR #9506, opened 2026-08-18
docs(solutions): add investigation notes for the Ama watch-page bug

tanflem:
## Summary by CodeRabbit

* **Documentation**
  * Added a research write-up documenting the Ama language availability issue on the JESUS film Watch page.
  * Recorded reproduction steps, observed behavior, investigated possibilities, related work, conclusions, and unresolved questions.
  * Clarified that the exact cause remains unresolved while preserving historical analysis for future investigation.


tanflem:

Closing — not a pattern we're keeping in this repo.

The analysis itself isn't lost. The confirmed root cause is recorded in #9513, including the specific finding that matters: the bug is in `forge`, not `core`. The watch-page language picker (`apps/web/src/lib/content.ts`) lists any published + HLS language, but the route gate rejects any slug missing from the generated corpus (`packages/watch-url-policy/src/public-watch-language-slugs.ts`), which is built from `Language.bcp47`. Roughly 39 admin-added `Language` rows — Ama (id `20118`, bcp47 `nyi`) among them — have no `bcp47` set, so `classify()` returns `unknown` and `notFound()` fires before any GraphQL lookup runs.

That makes this a whole bug class, not an Ama-specific one: every published dub in one of those ~39 languages 404s identically. The three candidate fixes (backfill `bcp47`, key the corpus generator on `Language.slug`, or filter the picker to match the corpus) are also captured in #9513.

The actual fix belongs in `forge`. Nothing further is needed in `core`.

PR #9507, opened 2026-08-18
feat(api-media): add parentVariantsOnly publish mode and download-size backfill

### LINKED ISSUE #9382: Repair zero-size Downloads and recover missing parent language Variants

tanflem:
Closes #9382

Replaces #9491, closed when its source branch (`phoebe/issue-9382`) was renamed out from under it. Same commits, no code changes.

## Summary by CodeRabbit

- **New Features**
  - Added a parent-variants-only publishing mode to identify and optionally recover missing parent language variants.
  - Publishing results now show missing and successfully recovered language IDs.
  - Added a resumable download-size backfill command with dry-run/apply modes, provider detection, batching, filtering, progress tracking, and audit reports.
  - Improved download-size recovery using provider metadata and HTTP fallbacks.

- **Documentation**
  - Added usage guidance for parent-variant recovery and download-size backfill operations.


tanflem:
Addressed the outstanding review threads in f49f74925.

**Taken as raised**

- **Error detail discarded** — both `catch` blocks in `processCandidate` now log the error with `downloadId` and `provider`, and the Mux asset lookup warns instead of swallowing. This was the substantive one: an operator reading the audit report could not tell a Mux outage from missing rendition metadata, since both surfaced as `errorCode: 'unknown'`.
- **Trim `DOWNLOAD_SIZE_BACKFILL_PROVIDER`** — done, matching every sibling env read.
- **Report-stream `error` listener** — added, but it records the error and fails the run rather than only logging. Logging and continuing would silently drop audit records while still reporting success.
- **Non-zero exit on failures** — done, matching the contract in `recover-parent-variants.ts`. Also exits non-zero when the audit report is incomplete.

**Narrowed**

- **`Number.isSafeInteger` on all three size paths** — applied to `toPositiveSize` only, and as a strict digit-string check rather than `isSafeInteger` alone. Two notes on the original suggestion:
  - `isSafeInteger` does not do what the finding assumes. `Number('1e10')` and `Number('0x1F')` are both safe integers, so they pass. Rejecting malformed headers needs the `/^[0-9]+$/` pattern that `parseMuxFilesize` in `resolveMuxSize.ts` already uses — this brings `httpSize.ts` in line with that.
  - `parseContentRangeTotal` already constrains its input with `/\/(\d+)\s*$/`, and `resolveR2Size` reads a trusted `bigint` from our own database. Neither needed changing.
  - The real justification is not petabyte-scale files but that the repair query only reselects rows still `null` or `<= 0`, so any bogus positive value written here is never revisited.

**Declined**

- **Removing the `catch` around the Mux lookup so failures reach `processCandidate` as `failed`** — this would regress behaviour. `resolveMuxSize` treats a null asset as “no rendition metadata” and falls back to HTTP sizing, which usually still resolves the length. Letting it throw converts a recoverable case into a hard failure, and combined with the new non-zero exit code would make any run containing a deleted Mux asset permanently exit 1. It now logs and still returns null.

Full `api-media` suite passes (861 tests), type-check clean.

mikeallisonJS (approved):

PR #9508, opened 2026-08-18
feat(api-languages): sync WESS native language names into LanguageName

### LINKED ISSUE #9483: Sync WESS native (autonym) language names into LanguageName

tanflem:
Closes #9483

Replaces #9489 — same commits, republished under a compliant branch name (the original `phoebe/issue-9483` fails the branch-naming CI check and I'm not touching it directly after #9491 got closed by an in-place rename).

## Summary by CodeRabbit

* **New Features**
  * WESS language imports now capture and store native language names, including English’s autonym.
  * Import results report separate counts for languages and native names created or updated.
  * Completion messages and progress reporting now include native-name totals.
  * Existing native names are preserved when later import data omits them.

* **Documentation**
  * Updated importer documentation to explain native-name handling, primary-name behavior, blank values, and preservation rules.


tanflem:
Replaced the per-row `prisma.languageName.findUnique` lookup (used to decide whether to keep an existing autonym's primary flag when `NATIVE_LAN_NAME` is absent this run) with an optional `primary` param on `upsertLanguageNameEntry` — `undefined` now just omits the key from the Prisma `update` payload, leaving the existing flag untouched. Same correctness, no extra query per row (this saves roughly 46k needless SELECTs on a ~48k-row import). Verified: 118/118 tests passing, lint clean, typecheck clean.

mikeallisonJS (commented):
Reviewed against #9483. Solid — all 8 in-scope stories are covered, and the Prisma → GraphQL → codegen chain is complete (subgraph, gateway and `graphql-env.d.ts` all carry `nativeNamesImported`). Two things before merge:

**1. English-gloss demotion is skipped when `row.name == null`.** The demotion lives inside `if (row.name != null && row.id !== englishLanguageId)`. If a WESS response ever carries a native name but a blank English label, the pre-existing English row keeps `primary: true` while the new autonym is also written `primary: true` — two primary rows for one language. Narrow, but it's exactly the invariant the README paragraph promises ("the two rows never disagree about which one is primary").

**2. The spec contradicts itself and the code silently resolved it.** #9483 says primary "remains `true` only for languages where no native name exists", but the code passes `primary: undefined` there (leave untouched). The code is right — writing `true` would re-promote English over a previously-stored autonym and break story 5. Worth correcting the issue text rather than the code, so the next reader doesn't "fix" it back.

Smaller notes, all optional:

- `primary?: boolean` is a tri-state (`true` = promote, `false` = demote, `undefined` = leave alone) that needs ~12 lines of comment to be legible. A `'set-true' | 'set-false' | 'leave'` union would carry itself.
- The "absence is a no-op / primary never disagrees" rule is written out four times — README bullet, the `primary?` JSDoc, and both inline comments in `upsertLanguage`. Four copies to keep in sync.
- `const hasNativeName = row.nativeName != null` and then `if (row.nativeName != null)` again — if either drifts, the returned tally diverges from what was actually written.
- New tests cover `normalizeWessLanguageRow` only. The genuinely new behaviour (primary demotion, the id-529 English gloss skip, the `primary ?? true` create semantics) is untested. The spec file has no Prisma mock at all today, so this is a pre-existing gap rather than one you created.

Also: the rollout note says this should run against stage before production. Did that happen? Nothing in the PR records it.

tanflem:
@mikeallisonJS — replying here rather than inline, since this came in as a top-level review body and there's no thread to resolve. Both blocking items are addressed in 6573c7dab.

## 1. Two primary rows — fixed

Confirmed, and the sequence is exactly as you described. The demotion was the `primary: hasNativeName ? false : undefined` argument on the English-gloss upsert, so it only ran inside `if (row.name != null && row.id !== englishLanguageId)`. A row with `NATIVE_LAN_NAME` and no `LAN_NAME` skipped that branch entirely while the autonym insert below still ran with `primary: true`, leaving the stored English row primary alongside it.

**I chose demote-then-insert unconditionally, not skip-the-insert.** `row.name` carries no information about the autonym — it's the English gloss. Gating the autonym write on it would discard real native-name data for a reason unrelated to the autonym, and that write *is* the feature (stories 1, 2, 4). The thing that must not be conditional is the demotion, so that's what moved:

```ts
if (row.nativeName != null) {
  await prisma.languageName.updateMany({
    where: { parentLanguageId: row.id, languageId: { not: row.id }, primary: true },
    data: { primary: false }
  })
  await upsertLanguageNameEntry({ ...,  primary: true })
}
```

The English upsert no longer passes `primary` at all. Two things fall out:

- It also catches a primary row sitting in some **third** language — older data, or a `languageNameUpdate` edit (more on that below). Demoting only the English row would have left that one standing.
- "Absence is a no-op" is now a no-op in the full sense: a run without `NATIVE_LAN_NAME` does no demotion *and* no write, so a stored autonym keeps its flag and English is never promoted back.

## 2. Spec contradiction — amended, code unchanged

Agreed the code was right and the issue was wrong. Amended #9483 rather than the code. The contradiction was between two Implementation Decisions bullets: "English-row primary flip" said English stays `true` "only for languages where no native name exists", while "Sync semantics on absence" said a run without `NATIVE_LAN_NAME` changes nothing — which, for a language that already has a stored autonym, prescribes two primary rows. The spec was also silent on the mirror case (autonym present, `LAN_NAME` absent), which is where the code actually broke.

Resolved as: *"no native name exists"* means **no autonym is stored for the language**, not "this run didn't supply one" — and the demotion belongs to the autonym write, unconditionally. Both bullets rewritten, the invariant now stated outright as its own bullet, and an **Amendment (2026-09-16)** section at the end records what changed and why, so the next reader doesn't revert it. I also corrected the Testing Decisions bullet that ruled out a Prisma-mock seam, since this PR now adds one — flagging that as a second edit beyond the contradiction itself, in case you'd rather it read differently.

## 3. Regression coverage — your fourth note

You were right that the new behaviour was untested. `wess-languages-import.spec.ts` now covers `runWessLanguagesImport` end to end, with `prismaMock.languageName` backed by an in-memory table that honours the real `@@unique([parentLanguageId, languageId])` upsert semantics — so the assertions are on **the rows a run leaves behind**, not on call arguments:

| case | asserts |
| --- | --- |
| autonym, no English gloss *(the bug)* | exactly 1 primary, and it's the autonym; English kept at `primary: false` |
| both names | exactly 1 primary |
| stale primary in a third language | exactly 1 primary, demoted correctly |
| English gloss, no autonym | exactly 1 primary; stored autonym untouched |

Verified red against the unfixed source (2 failed / 23 passed, both `expected [ …(2) ] to have a length of 1 but got 2`) and green after (25/25). Full workspace suite: 15 files, 122 tests, green. Run with:

```bash
npx vitest run --root apis/api-languages --config $PWD/apis/api-languages/vitest.config.mts \
  --coverage.enabled=false src/scripts/wess-languages-import.spec.ts
```

README bullet updated to describe the unconditional demotion.

## 4. Partial unique index — proposing only, deliberately no migration

Worth having as a backstop, but **I have not added a migration** — migration ordering here is fragile and this one needs a decision plus a production audit first. Putting it up as an open question:

```sql
CREATE UNIQUE INDEX CONCURRENTLY "LanguageName_one_primary_per_parent"
  ON "LanguageName" ("parentLanguageId") WHERE "primary";
```

Three things to settle before anyone writes that file:

1. **Audit production first.** CONTEXT.md says prod is the source of truth and seeds everything downstream, so a violation there fails the deploy, not a local run. Read-only check:
   ```sql
   SELECT "parentLanguageId", COUNT(*) AS primary_rows
   FROM "LanguageName" WHERE "primary"
   GROUP BY "parentLanguageId" HAVING COUNT(*) > 1;
   ```
   Non-empty means a backfill has to land before the index does — and given the bug above shipped nowhere yet, anything it finds is pre-existing from another source.
2. **It constrains `languageNameUpdate` too, which isn't ready for it.** `language.ts:298-321` creates `primary: true` rows with no demotion at all: when `nameLanguageId == null` it writes `{ parentLanguageId: id, languageId: id, primary: true }` outright. Under the index that mutation starts throwing P2002 for publishers whose language already has a primary name. It needs the same demote-then-write treatment first, which is its own change.
3. **Prisma can't declare it.** No partial-unique support in `schema.prisma`, so it'd be hand-written SQL in a migration with nothing in the schema to match it — future `migrate dev` runs will see drift. Liveable, but it should be a conscious choice, and there's no precedent for it in this repo's migrations today.

My read: worth doing, but as a follow-up that pairs the index with the `languageNameUpdate` fix, after the audit. Happy to open that issue if you agree.

## 5. Your optional notes — what I left alone

Deliberately not touched, since they're your code and refactors rather than defects:

- **Tri-state `primary?: boolean`** — partly dissolved by the fix. The English gloss no longer passes it, so only `true` (autonym) and omitted remain; the `false` case is gone and the JSDoc shrank. A `'set-true' | 'leave'` union would now be a two-line change if you still want it.
- **Rule written four times** — still true, and the inline comment in `upsertLanguage` grew rather than shrank, because the *why* of the unconditional demotion is the part a future reader is most likely to undo. Consolidating is yours to call.
- **`hasNativeName` vs the repeated `row.nativeName != null`** — left as-is on purpose: swapping the `if` to `hasNativeName` risks losing the null-narrowing that `value: row.nativeName` depends on under `tsc7`, and I didn't want a type-check gamble riding along with a correctness fix.

## 6. Stage run

Can't answer that one — nothing in the branch or the PR records a stage run, and I have no way to verify it from here. Over to you or @phoebe, since rollout says it gates merge.

---

`pnpm lint:changed --fix` clean, `nx affected --target=type-check` green across 21 projects.

tanflem:

## Triage — 2026-09-16

**Verdict: ♻️ Close as duplicate / superseded**

### Where this stands

**Duplicate of #9607.** Both PRs close #9483 and touch the exact same 9 files. #9607 is newer (created 2026-09-16, +509/-28, 41 checks green) and supersedes this one (+357/-30). This PR also carries 1 unresolved CodeRabbit thread (Major, data-integrity) on `wess-languages-import.ts`.

### What needs to happen next

**Close this PR** in favour of #9607. Before closing, check that the CodeRabbit data-integrity point raised here was carried across to #9607 — it was opened against this branch and may not have been re-raised there.

PR #9526, opened 2026-08-19
fix(api-journeys): restore publisher access to template customization fields (QA-563)

csiyang:
Fixes [QA-563](https://linear.app/jesus-film-project/issue/QA-563/error-when-deselecting-featured-template-and-saving).

## Problem

Saving Template Settings (e.g. after deselecting **Featured**) failed with:

> User is not allowed to update journey customization field

for publishers who hold no userTeam/userJourney role on the template. The `journeyUpdate` mutation succeeded, then `journeyCustomizationFieldPublisherUpdate` threw FORBIDDEN — and because the dialog runs its mutations sequentially, the `journeyFeature` mutation never fired, so the featured change was lost as well.

## Root cause

`canManageTemplateField` (introduced in the NestJS→Pothos migration, #9255) mistranslated the legacy CASL rules: it requires publishers to *also* hold a team/journey role on the journey.

In the legacy ACL, `can(Manage, 'Journey', { template: true })` for publishers was declared **after** `cannot(Manage, 'Journey', 'template')`; CASL's last-rule-wins semantics meant that rule re-unlocked the `template` field for publishers on **any** existing template, no membership needed. The membership-scoped publisher field rules only governed a different case — converting your own non-template journey into a template. The migration collapsed both cases into the stricter one, so every publisher who isn't a member of `jfp-team` has been unable to save template settings since.

## Fix

Restore the legacy truth table in `canManageTemplateField`, except deliberately narrowed for publishers to global templates (see second row):

| Case | Result |
| --- | --- |
| publisher + jfp-team template | allow (no membership needed) |
| publisher + other team's local template, no role | deny — deliberately narrower than legacy, whose CASL rule ordering accidentally unlocked every team's templates; publishers only govern the global (jfp-team) template library |
| publisher + non-template + own team/journey role | allow (convert-to-template path) |
| local template + team/journey role | allow (unchanged) |
| anything else | deny |

The function's only caller is `journeyCustomizationFieldPublisherUpdate`, and non-publisher paths are untouched.

## Testing

- **New** `journeyCustomizationFieldPublisherUpdate.mutation.spec.ts` — exercises the real GraphQL mutation with the real ACL (prisma mocked). The publisher-without-membership case reproduced the exact reported error red-first and passes with the fix; guard cases confirm non-publishers, jfp-team managers without the publisher role, and non-template journeys are still denied.
- `journey.acl.spec.ts` — added unit coverage for `canManageTemplateField`'s full truth table (previously untested).
- Full `api-journeys` suite, `type-check`, and `pnpm lint:changed --fix` all pass.

## Also included: AGENTS.md — pre-push ritual gains type-check

The first push of this branch failed CI's `lint-work` type-check even though the documented agent pre-push ritual passed: the ritual only runs `pnpm lint:changed --fix`, and nothing at push time required a type-check. The agent pre-push instructions in `AGENTS.md` now gate `pnpm exec nx affected --target=type-check --base=origin/main` (the same target `lint-work` runs) before `git push`, and document the two traps that let this slip through: each project's `type-check` target runs `tsc7` (stricter than editor tsc), and nx prints the same trailing summary block on success and failure, so the exit code — not the tail of piped output — is the signal. Verified red/green: the command exits 1 on the exact code that broke CI and 0 on the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Publishers can manage fields on JFP templates without additional journey or team roles.
  * Managing fields on local templates and converting journeys into templates now requires appropriate journey or team permissions.
  * Unauthorized users and non-publishers remain restricted from template field updates.

* **Tests**
  * Added comprehensive coverage for publisher, team-member, role-based, conversion, and unauthorized access scenarios.

* **Documentation**
  * Updated development guidance to include type-checking affected projects before pushing changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 ## Before you push
 
-**Asked to push? Run `pnpm lint:changed --fix` first, commit anything it changes, and push that in the same go.** Never push without it.
+**Asked to push? Run `pnpm lint:changed --fix` first, commit anything it changes, type-check the affected projects, and push that in the same go.** Never push without it.
 
 It applies the same fixes [autofix.ci](https://autofix.ci) would otherwise commit to your PR afterwards — formatting, lint fixes, translation extraction. They are generated output, so they belong in your commit, not in a bot commit on top of it. There is no pre-push hook and nothing will stop you: if you skip this, the bot does it for you and the PR grows a `fix: lint issues` commit. Details in [Lint before push](#lint-before-push-agents).
 
@@ -55,6 +55,10 @@ if [ -n "$(git status --porcelain)" ]; then
   git commit --no-verify -m "chore: lint fixes" || { echo "could not commit generated fixes; aborting push"; exit 1; }
 fi
 
+# Type-check what the branch touched — CI's lint-work job runs this exact
+# target, and lint:changed cannot catch type errors (see below).
+pnpm exec nx affected --target=type-check --base=origin/main || { echo "type-check failed — fix the errors above, then re-run"; exit 1; }
+
 git push
 ```
 
@@ -78,7 +82,9 @@ Not covered locally: **`codegen`** is the one autofix.ci step that can still com
 
 It is deliberately not gated, because it only works where `apollo` is already installed globally. The devcontainer installs it in [`post-create-command.sh`](.devcontainer/post-create-command.sh) (`npm i -g nx foreman apollo graphql`) and autofix.ci does the same, which is why it works in both. On a plain host checkout `apollo` is absent, so `npx` tries to fetch it and fails with `EOVERRIDE` from the `next` override in `package.json` — gating it would mean requiring that global install everywhere, which we chose not to do. Work in the devcontainer, or expect autofix.ci to commit the regenerated files.
 
-The remaining steps cannot produce a commit at all: `type-check` and `subgraph-check` only report (and `subgraph-check` needs a Hive token), and `prisma-generate` writes nothing that is tracked.
+**`type-check`** cannot produce a commit (it only reports), but it is still gated in the ritual above because CI's `lint-work` job fails on it and nothing in `lint:changed` covers types. Two traps: each project's `type-check` target runs `tsc7` (the TypeScript 7 preview via `tsconfig.ts7.json`), which is stricter than editor tsc — code your IDE accepts can still fail it. And nx prints the same trailing summary block on success and failure, so never judge the outcome from the tail of piped output — check the exit code.
+
+The remaining steps cannot produce a commit at all: `subgraph-check` only reports (and needs a Hive token), and `prisma-generate` writes nothing that is tracked.
 
 ### Documented Solutions
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,7 +57,8 @@ fi
 
 # Type-check what the branch touched — CI's lint-work job runs this exact
 # target, and lint:changed cannot catch type errors (see below).
-pnpm exec nx affected --target=type-check --base=origin/main || { echo "type-check failed — fix the errors above, then re-run"; exit 1; }
+git fetch origin main --quiet || true
+pnpm exec nx affected --target=type-check --base=origin/main || { echo "type-check failed — either real type errors above, or an environment gap (origin/main unfetched, node_modules not installed in this worktree)"; exit 1; }
 
 git push
 ```
@@ -82,7 +83,7 @@ Not covered locally: **`codegen`** is the one autofix.ci step that can still com
 
 It is deliberately not gated, because it only works where `apollo` is already installed globally. The devcontainer installs it in [`post-create-command.sh`](.devcontainer/post-create-command.sh) (`npm i -g nx foreman apollo graphql`) and autofix.ci does the same, which is why it works in both. On a plain host checkout `apollo` is absent, so `npx` tries to fetch it and fails with `EOVERRIDE` from the `next` override in `package.json` — gating it would mean requiring that global install everywhere, which we chose not to do. Work in the devcontainer, or expect autofix.ci to commit the regenerated files.
 
-**`type-check`** cannot produce a commit (it only reports), but it is still gated in the ritual above because CI's `lint-work` job fails on it and nothing in `lint:changed` covers types. Two traps: each project's `type-check` target runs `tsc7` (the TypeScript 7 preview via `tsconfig.ts7.json`), which is stricter than editor tsc — code your IDE accepts can still fail it. And nx prints the same trailing summary block on success and failure, so never judge the outcome from the tail of piped output — check the exit code.
+**`type-check`** cannot produce a commit (it only reports), but it is still gated in the ritual above because CI's `lint-work` job fails on it and nothing in `lint:changed` covers types. It is the slowest step of the four — expect tens of seconds for a typical branch (a few seconds per affected project, plus nx overhead), so a pause is normal. Two traps: each project's `type-check` target runs `tsc7` (the TypeScript 7 preview via `tsconfig.ts7.json`), which is stricter than editor tsc — code your IDE accepts can still fail it. And nx prints the same trailing summary block on success and failure, so never judge the outcome from the tail of piped output — check the exit code. The command also exits non-zero on environment gaps (`origin/main` not fetched, `node_modules` missing in a fresh worktree) — rule those out before hunting for type errors.
 
 The remaining steps cannot produce a commit at all: `subgraph-check` only reports (and needs a Hive token), and `prisma-generate` writes nothing that is tracked.
 
--- a/apis/api-journeys/src/schema/journey/journey.acl.ts
+++ b/apis/api-journeys/src/schema/journey/journey.acl.ts
@@ -243,11 +243,13 @@ export function canManageTemplateField(
 
   if (user.roles?.includes('publisher') !== true) return false
 
-  // Publishers manage any existing template without needing a team/journey
-  // role: in the legacy CASL, can(Manage, 'Journey', { template: true }) was
-  // declared after cannot(Manage, 'Journey', 'template'), so it re-unlocked
-  // the template field for every template (QA-563).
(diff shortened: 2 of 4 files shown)


apis/api-journeys/src/schema/journeyCustomizationField/journeyCustomizationFieldPublisherUpdate.mutation.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      roles
+    })
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    mockGetUserFromPayload.mockReturnValue(mockUser)
+    prismaMock.$transaction.mockImplementation(async (cb: any) => await cb(tx))
+    prismaMock.journeyCustomizationField.findMany.mockResolvedValue(
+      updatedFields as any
```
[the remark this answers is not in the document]
csiyang: The `mockRoles` part is adopted — it is now typed with the generated `Role` enum (53239c142). The remaining casts are deliberate and follow this repo's established spec conventions at the prisma deep-mock boundary: `blockOrderUpdate.mutation.spec.ts` and `adminJourneys.query.spec.ts` use the same `(cb: any)` transaction mock and `as any` fixture returns. Typing the callback as `Prisma.TransactionClient` would not remove a cast, only relocate it (`tx` is a partial mock, so it becomes `tx as unknown as Prisma.TransactionClient`), and fully typing the journey fixture would mean populating every generated `Journey` field with data irrelevant to the ACL under test. tsc7 type-checks the file clean as of 53239c142.

apis/api-journeys/src/schema/journey/journey.acl.ts
```diff
@@ -241,11 +241,14 @@ export function canManageTemplateField(
...
-  )
-    return true
+  if (user.roles?.includes('publisher') !== true) return false
 
-  return false
+  // Publishers manage any existing template without needing a team/journey
+  // role: in the legacy CASL, can(Manage, 'Journey', { template: true }) was
+  // declared after cannot(Manage, 'Journey', 'template'), so it re-unlocked
+  // the template field for every template (QA-563).
+  if (journey.template === true) return true
```
jianwei1: **Concern:** the restoration is wider than the reported bug needs — worth making the product decision explicit rather than inheriting it.

I verified your root-cause analysis independently against the pre-migration CASL at `39ce7c0` (the base of #9255), and it holds exactly, including for non-jfp teams: `can(Manage, 'Journey', { template: true })` in the publisher block carries no `fields`, so it is relevant to the `template` field, and it is declared after `cannot(Manage, 'Journey', 'template')` — last-matching-rule wins, so a publisher with no membership was allowed on any template. I walked all eight cells of the truth table against this function and every one matches legacy; the change is strictly a widening for publishers on templates where they hold no role, and nothing else moves.

The question is whether legacy is the right target here. QA-563 is about **jfp-team** templates, but `journey.template === true` also grants every publisher write access to *other teams' local templates*, and `journeyReadAccessWhere` already gives publishers `{ template: true }` read across all teams — so this is API-reachable, not just theoretical. The new test `allows any template without team or journey roles (QA-563)` pins `localTemplateNoRoles` → `true`, so it is clearly deliberate.

What gives me pause is that legacy's breadth looks like a byproduct of CASL rule ordering rather than a decision — the same accident you documented. That makes "restore it exactly" and "restore what was intended" two different targets, and the publisher-convert field rules just below it in the legacy file become dead code for templates under the broad reading, which is what you would expect if the breadth was unintentional.

So: is a publisher meant to be able to rewrite another team's private template's customization description? If yes, this is right as-is and the answer is worth a line in the PR description so the next reader does not re-litigate it. If not, scoping this branch to `journey.teamId === 'jfp-team'` fixes QA-563 with no widening at all.

---

AGENTS.md
```diff
@@ -55,6 +55,10 @@ if [ -n "$(git status --porcelain)" ]; then
   git commit --no-verify -m "chore: lint fixes" || { echo "could not commit generated fixes; aborting push"; exit 1; }
 fi
 
+# Type-check what the branch touched — CI's lint-work job runs this exact
+# target, and lint:changed cannot catch type errors (see below).
+pnpm exec nx affected --target=type-check --base=origin/main || { echo "type-check failed — fix the errors above, then re-run"; exit 1; }
```
jianwei1: **Nit:** the failure message will misattribute environment failures.

I confirmed both claims in the surrounding prose — `lint-work`'s Type Check step really does run `nx affected --target=type-check`, and api-journeys' target really is `pnpm tsc7 -p apis/api-journeys/tsconfig.ts7.json` — so the gate is well justified.

Two small things about the command itself, given this text is copy-pasted and run by agents:

1. It also exits non-zero when `origin/main` is unfetched or `node_modules` is missing — both common in exactly the worktrees this document says agents work in (the `--no-verify` note two blocks up exists precisely because worktrees lack the husky bootstrap). The handler then prints "type-check failed — fix the errors above," sending the agent hunting for type errors that do not exist. A `git fetch origin main` ahead of it, or a message naming both causes, would save that detour.
2. The neighbouring lint steps are careful to give timings ("under a second", "5–20s per touched workspace") and this one gives none, though it is by far the slowest of the four. Worth a rough number so agents know whether a long pause is normal.

---

jianwei1 (commented):
## Review summary

**Critical:** none. **Concerns:** 2 (one inline on the ACL, one below on scope). **Nit:** 1 (inline on `AGENTS.md`).

CI is green on `c701ee0` and the fix looks correct. I checked the central claim independently rather than taking the description on trust — fetching the pre-migration CASL at `39ce7c0` (base of #9255) and walking CASL's last-matching-rule semantics. The root-cause analysis holds exactly, and I walked all eight cells of the truth table against the new function: every cell matches legacy, so this is strictly a widening for publishers on templates where they hold no role. Your scope claim also checks out — `canManageTemplateField` has exactly one caller, and `journeyTemplate.mutation.ts` (which actually flips the `template` boolean) goes through `journeyAcl` and is untouched.

### Concern — unrelated process change bundled into a bugfix

The `AGENTS.md` type-check gate is a sensible addition and the reasoning behind it is sound. But it changes the repo-wide ritual every agent contributor follows, and it is riding in a PR whose reviewers are focused on an ACL truth table — the two need different reviewers looking for different things. It would get better scrutiny as its own PR. Not blocking; flagging because process changes that arrive as PR riders tend to go in unread.

### What is done well

- `journeyCustomizationFieldPublisherUpdate.mutation.spec.ts` is a genuine integration test, not a unit test wearing a costume — it drives the real yoga pipeline, so roles resolve through the actual `prisma.userRole.findUnique` lookup in `yoga.ts` and the real ACL runs. That is why it could reproduce the reported error red-first, and it is the right level for an authorization regression.
- The guard cases are well chosen, particularly the non-publisher jfp-team **manager** — the case that most looks like it should pass and must not.
- The inline comment at `journey.acl.ts:246-249` records *why* the legacy semantics are what they are. Without it the next reader would almost certainly "tidy" this straight back into the bug.

### Adjacent observation — out of scope, do not let it gate this PR

While confirming the blast radius I read `journeyTemplate.mutation.ts`, the mutation that flips the `template` boolean. Its permission check is gated behind `isGlobalTemplate`:

```ts
const isGlobalTemplate = journey.team?.id === 'jfp-team'
if (isGlobalTemplate && !journeyAcl(Action.Manage, ...)) throw FORBIDDEN
```

For any journey **not** in `jfp-team` there is no authorization check at all — the resolver only requires `isAuthenticated` or `isAnonymous`. On my reading, any authenticated user who knows a journey id could convert another team's journey into a template. Entirely pre-existing and outside this diff, but it is in the neighbourhood you are already auditing and probably deserves its own issue.

Skipped as noise: the `as any` casts at the prisma deep-mock boundary (already litigated and withdrawn upstream), the missing `extensions.code` assertions on the denial tests, and the `mockResolvedValueOnce` that quietly no-ops `recalculateJourneyCustomizable` on the success path — I checked, and no once-queue leaks between tests.

---

apis/api-journeys/src/schema/journey/journey.acl.ts
csiyang: Agreed — the breadth was legacy's accident, not a decision worth inheriting. Narrowed in 6fe80babf: the publisher branch now requires `teamId === 'jfp-team'`, so publishers govern the global template library only and other teams' local templates stay membership-only (a publisher holding an actual team/journey role still gets in through the membership branches). Tests pin both directions and the PR description documents the decision. One known edge accepted with this: `journeyUpdate`'s separate publisher rule still allows metadata edits on any template, so a publisher saving another team's local template hits FORBIDDEN on the customization mutation — consistent with them not being meant to touch it.

AGENTS.md
csiyang: Both applied in 30559a496: the gate now fetches origin/main first, the failure message names the environment gaps (unfetched base, missing node_modules) alongside real type errors, and the traps paragraph gives the step a rough timing so the pause reads as normal.

jianwei1 (approved):
Approving — both findings from my earlier review are resolved.

**`6fe80ba`** narrows the publisher branch to `teamId === 'jfp-team'`. I re-walked all eight cells of the truth table against the new code: the QA-563 fix is intact (publisher + jfp-team template + no role → allow), the accidental widening is gone (publisher + another team's local template + no role → deny), publishers holding real membership still get in through the membership branch, and every non-publisher cell is unchanged from `main`. The tests were split to pin both directions rather than just the permissive one, and the comment now records that the divergence from legacy is a decision rather than a translation error — which is what stops the next reader from "fixing" it back.

**`30559a4`** applies both halves of the `AGENTS.md` nit: `git fetch origin main` ahead of the gate, a failure message that names the environment gaps alongside real type errors, and a timing estimate so the pause reads as normal. The `|| true` on the fetch is the right call — an offline agent proceeds against a stale base instead of being blocked.

CI is green on `30559a4`. Nice work on the root-cause write-up; the legacy CASL analysis held up exactly when I checked it against `39ce7c0`.

---

PR #9527, opened 2026-08-20
fix(journeys-admin): featured checkbox desync and stale refetch race (QA-564)

csiyang:
Fixes [QA-564](https://linear.app/jesus-film-project/issue/QA-564/featured-checkbox-desyncs-from-saved-state-in-template-settings-dialog) — follow-on from [QA-563](https://linear.app/jesus-film-project/issue/QA-563/error-when-deselecting-featured-template-and-saving) / #9526. Frontend-only; deliberately split from the backend PR per FE/BE separation (no schema or deploy-order coupling — either can ship first).

## Problem

Stage testing of the QA-563 server fix surfaced remaining client bugs in `TemplateSettingsDialog`:

1. Ticking **Featured** → Save showed the success snackbar but sometimes didn't feature the template.
2. Reopening Template Settings showed the Featured box permanently unticked for the rest of the session, even though the journey was correctly featured.

## Root causes

1. **Uncontrolled checkbox.** `MetadataTabPanel` used `defaultChecked={values.featured}`. With Formik's `enableReinitialize`, a journey cache update landing between the user's click and Save silently resets `values.featured` while the box still displays the tick — submit then sees "no change", skips `journeyFeature`, and reports success without saving.
2. **Stale refetch race.** `handleSubmit` ran journeyUpdate → customization update (with an unawaited `refetchQueries: ['GetPublisherTemplate']`) → `journeyFeature`. The refetch could read the journey *before* `journeyFeature` committed and land *after* its response, overwriting `featuredAt` back to `null` in the Apollo cache for the rest of the session.

## Fix

- Checkbox is now controlled: `checked={values.featured}`.
- The save now mirrors `LocalTemplateDetailsDialog` (per review): the three mutations run via `Promise.allSettled` so one failure cannot silently drop the others, and `refetchQueries: ['GetPublisherTemplate']` is replaced with a targeted `cache.modify` patch of `journeyCustomizationDescription` — with no refetch, nothing can race `journeyFeature`, and the ordering invariant ceases to exist rather than being held by statement order.

## Testing

Red/green verified. Regression tests pin: the checkbox following `featuredAt` across rerenders; the reported symptom direction (box snaps back when a reinitialize resets the form mid-edit); and the partial-failure guarantee (a `journeyFeature` rejection still saves the customization description — verified red against the sequential structure). All 34 tests across the `TemplateSettingsItem` tree pass. `lint:changed --fix` and affected type-check clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the featured setting so its checkbox stays synchronized with the journey’s current state.
  * Ensured featured status updates are applied before customization changes, preventing stale settings after refresh.
  * Avoided unnecessary featured-status updates when the value has not changed.

* **Tests**
  * Added regression coverage for checkbox state synchronization and update ordering.

### DIFF
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
@@ -1,6 +1,7 @@
 import { MockedProvider } from '@apollo/client/testing'
 import { fireEvent, render, waitFor, within } from '@testing-library/react'
 import { SnackbarProvider } from 'notistack'
+import { ReactElement } from 'react'
 
 import { JourneyProvider } from '@core/journeys/ui/JourneyProvider'
 import {
@@ -585,5 +586,117 @@ describe('TemplateSettingsDialog', () => {
       </MockedProvider>
     )
     expect(queryByRole('tab', { name: 'Categories' })).not.toBeInTheDocument()
+  })
+
+  // QA-564: the checkbox was uncontrolled (defaultChecked), so its visual
+  // state could diverge from Formik's — a cache-driven reinitialize would
+  // reset values.featured while the box still displayed the user's tick,
+  // making Save skip journeyFeature and report success without saving.
+  it('keeps the featured checkbox in sync with journey.featuredAt', async () => {
+    const tree = (featuredAt: string | null): ReactElement => (
+      <MockedProvider mocks={[]}>
+        <SnackbarProvider>
+          <JourneyProvider
+            value={{
+              journey: { ...defaultJourney, featuredAt },
+              renderMode: 'admin'
+            }}
+          >
+            <TemplateSettingsDialog open onClose={onClose} />
+          </JourneyProvider>
+        </SnackbarProvider>
+      </MockedProvider>
+    )
+
+    const { getByRole, rerender } = render(tree(null))
+    expect(getByRole('checkbox')).not.toBeChecked()
+
+    rerender(tree('2026-08-20T00:00:00.000Z'))
+    await waitFor(() => expect(getByRole('checkbox')).toBeChecked())
+  })
+
+  // QA-564: the customization update triggers an unawaited refetch of
+  // GetPublisherTemplate. If it were issued before journeyFeature, it could
+  // read pre-feature state and land after journeyFeature's response,
+  // overwriting featuredAt in the cache with a stale null.
+  it('applies the featured change before the customization update fires its refetch', async () => {
+    const journeyUpdateResult = vi.fn(() => ({
+      data: {
+        journeyUpdate: {
+          __typename: 'Journey',
+          id: defaultJourney.id,
+          title: defaultJourney.title,
+          description: defaultJourney.description
+        }
+      }
+    }))
+    const customizationResult = vi.fn(() => ({
+      data: { journeyCustomizationFieldPublisherUpdate: [] }
+    }))
+    const featureResult = vi.fn(() => ({
+      data: {
+        journeyFeature: {
+          __typename: 'Journey',
+          id: defaultJourney.id,
+          featuredAt: '2026-08-20T00:00:00.000Z'
+        }
+      }
+    }))
+
+    const { getByRole } = render(
+      <MockedProvider
+        mocks={[
+          {
+            request: {
+              query: JOURNEY_SETTINGS_UPDATE,
+              variables: {
+                id: defaultJourney.id,
+                input: {
+                  title: defaultJourney.title,
+                  description: defaultJourney.description,
+                  strategySlug: null,
+                  tagIds: [],
+                  creatorDescription: null,
+                  languageId: '529'
+                }
+              }
+            },
+            result: journeyUpdateResult
+          },
+          {
+            request: {
+              query: JOURNEY_CUSTOMIZATION_DESCRIPTION_UPDATE,
+              variables: { journeyId: defaultJourney.id, string: '' }
+            },
+            result: customizationResult
+          },
+          {
+            request: {
+              query: JOURNEY_FEATURE_UPDATE,
+              variables: { id: defaultJourney.id, feature: true }
+            },
+            result: featureResult
+          }
+        ]}
+      >
+        <SnackbarProvider>
+          <JourneyProvider
+            value={{ journey: defaultJourney, renderMode: 'admin' }}
+          >
+            <TemplateSettingsDialog open onClose={onClose} />
+          </JourneyProvider>
+        </SnackbarProvider>
+      </MockedProvider>
+    )
+
+    fireEvent.click(getByRole('checkbox'))
+    fireEvent.click(getByRole('button', { name: 'Save' }))
+
+    await waitFor(() => expect(customizationResult).toHaveBeenCalled())
+    expect(featureResult).toHaveBeenCalled()
+    expect(journeyUpdateResult).toHaveBeenCalled()
+    expect(featureResult.mock.invocationCallOrder[0]).toBeLessThan(
+      customizationResult.mock.invocationCallOrder[0]
+    )
   })
 })
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
@@ -615,11 +615,36 @@ describe('TemplateSettingsDialog', () => {
     await waitFor(() => expect(getByRole('checkbox')).toBeChecked())
   })
 
-  // QA-564: the customization update triggers an unawaited refetch of
-  // GetPublisherTemplate. If it were issued before journeyFeature, it could
-  // read pre-feature state and land after journeyFeature's response,
-  // overwriting featuredAt in the cache with a stale null.
-  it('applies the featured change before the customization update fires its refetch', async () => {
+  // QA-564: the reported symptom — user ticks the box, a cache-driven
+  // reinitialize resets Formik's values, and the (previously uncontrolled)
+  // box kept displaying the tick, so Save saw "no change" and skipped
+  // journeyFeature while reporting success. The controlled box must snap
+  // back so display and form state never diverge.
+  it('snaps the checkbox back when a reinitialize resets the form mid-edit', async () => {
+    const tree = (journey: typeof defaultJourney): ReactElement => (
+      <MockedProvider mocks={[]}>
+        <SnackbarProvider>
+          <JourneyProvider value={{ journey, renderMode: 'admin' }}>
+            <TemplateSettingsDialog open onClose={onClose} />
+          </JourneyProvider>
+        </SnackbarProvider>
+      </MockedProvider>
+    )
+
+    const { getByRole, rerender } = render(tree(defaultJourney))
+    fireEvent.click(getByRole('checkbox'))
+    expect(getByRole('checkbox')).toBeChecked()
+
+    // A translation-poll/cache update changes another field while
+    // featuredAt is still null — enableReinitialize resets the form.
+    rerender(tree({ ...defaultJourney, title: 'Updated Title' }))
+    await waitFor(() => expect(getByRole('checkbox')).not.toBeChecked())
+  })
+
+  // QA-564: mutations run via Promise.allSettled so one failure cannot
+  // silently drop the others — a journeyFeature rejection must not lose the
+  // user's customization edit.
+  it('still saves the customization description when journeyFeature fails', async () => {
     const journeyUpdateResult = vi.fn(() => ({
       data: {
         journeyUpdate: {
@@ -633,17 +658,8 @@ describe('TemplateSettingsDialog', () => {
     const customizationResult = vi.fn(() => ({
       data: { journeyCustomizationFieldPublisherUpdate: [] }
     }))
-    const featureResult = vi.fn(() => ({
-      data: {
-        journeyFeature: {
-          __typename: 'Journey',
-          id: defaultJourney.id,
-          featuredAt: '2026-08-20T00:00:00.000Z'
-        }
-      }
-    }))
 
-    const { getByRole } = render(
+    const { getByRole, getByText } = render(
       <MockedProvider
         mocks={[
           {
@@ -675,7 +691,7 @@ describe('TemplateSettingsDialog', () => {
               query: JOURNEY_FEATURE_UPDATE,
               variables: { id: defaultJourney.id, feature: true }
             },
-            result: featureResult
+            error: new Error('user is not allowed to update featured date')
           }
         ]}
       >
@@ -692,11 +708,13 @@ describe('TemplateSettingsDialog', () => {
     fireEvent.click(getByRole('checkbox'))
     fireEvent.click(getByRole('button', { name: 'Save' }))
 
-    await waitFor(() => expect(customizationResult).toHaveBeenCalled())
-    expect(featureResult).toHaveBeenCalled()
-    expect(journeyUpdateResult).toHaveBeenCalled()
-    expect(featureResult.mock.invocationCallOrder[0]).toBeLessThan(
-      customizationResult.mock.invocationCallOrder[0]
+    await waitFor(() =>
+      expect(
+        getByText('Field update failed. Reload the page or try again.')
+      ).toBeInTheDocument()
     )
+    expect(customizationResult).toHaveBeenCalled()
+    expect(journeyUpdateResult).toHaveBeenCalled()
+    expect(onClose).not.toHaveBeenCalled()
   })
 })
--- a/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
+++ b/apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
@@ -1,5 +1,7 @@
+import { InMemoryCache } from '@apollo/client'
 import { MockedProvider } from '@apollo/client/testing'
 import { fireEvent, render, waitFor, within } from '@testing-library/react'
+import { GraphQLError } from 'graphql'
 import { SnackbarProvider } from 'notistack'
 import { ReactElement } from 'react'
 
@@ -691,7 +693,13 @@ describe('TemplateSettingsDialog', () => {
               query: JOURNEY_FEATURE_UPDATE,
               variables: { id: defaultJourney.id, feature: true }
             },
-            error: new Error('user is not allowed to update featured date')
+            // GraphQL error, not a network error — models the real ACL
+            // rejection shape and exercises the generic-message branch.
+            result: {
+              errors: [
+                new GraphQLError('user is not allowed to update featured date')
+              ]
+            }
           }
         ]}
       >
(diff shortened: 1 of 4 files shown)


apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -101,17 +101,21 @@ export function TemplateSettingsDialog({
           }
         }
       })
+      if (Boolean(journey.featuredAt) !== values.featured)
+        await journeyFeature({
+          variables: { id: journey.id, feature: values.featured }
+        })
```
jianwei1: **Concern:** the reorder changes what survives a partial failure.

These are sequential `await`s in one `try`. Before this PR, `journeyFeature` ran last, so a failure there still left title/description **and** the customization description saved. Now a `journeyFeature` rejection short-circuits before `journeyCustomizationDescriptionUpdate` ever fires — the user sees the generic "Something went wrong" and has no way to know their customization edit was silently dropped.

That bites hardest in exactly the QA-563 scenario (deselect-featured erroring on ACL). Since this PR is deliberately shippable ahead of #9526, landing it first would regress deselect-and-save from "featured failed, everything else saved" to "featured failed, customization lost".

The sibling dialog already handles this — `LocalTemplateDetailsDialog.tsx:218` collects the mutations into `tasks` and runs `await Promise.allSettled(tasks)` specifically so one failure doesn't block the others. Worth either adopting that here, or merging #9526 first to close the window.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -101,17 +101,21 @@ export function TemplateSettingsDialog({
...
+      // Must run after journeyFeature: the refetch it triggers is not awaited,
+      // so if it were issued first it could read pre-feature state and land
+      // after journeyFeature's response, overwriting featuredAt in the cache
+      // with a stale null for the rest of the session (QA-564).
       await journeyCustomizationDescriptionUpdate({
         variables: {
           journeyId: journey.id,
           string: values.journeyCustomizationDescription
         },
         refetchQueries: ['GetPublisherTemplate']
```
jianwei1: **Concern:** the race is ordered around rather than removed — and the sibling dialog already has the structural fix.

`LocalTemplateDetailsDialog.tsx:203-213` solves the identical cache-staleness problem for this *same* `journeyCustomizationFieldPublisherUpdate` mutation with a targeted `update(cache) { cache.modify(...) }`, with a comment noting the mutation returns customization fields rather than a Journey, so the Journey entity gets patched directly. That approach needs no refetch at all.

Adopting it here would drop `refetchQueries: ['GetPublisherTemplate']`, remove the ordering constraint entirely (right now it's held only by statement order plus a comment), and save a full `JourneyFields` round-trip per save.

To be clear, the diagnosis is right and the constraint is real: `JOURNEY_FIELDS` does include `featuredAt` (`libs/journeys/ui/src/libs/JourneyProvider/journeyFields.ts:33`) and `GET_PUBLISHER_TEMPLATE` spreads it, so a stale refetch genuinely would overwrite it. And documenting the constraint in code *and* pinning it with a test is the right pair for the approach taken. It's just an invariant that doesn't need to exist. Fine as a follow-up if you'd rather keep this PR tight.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
```diff
@@ -586,4 +587,116 @@ describe('TemplateSettingsDialog', () => {
...
+        </SnackbarProvider>
+      </MockedProvider>
+    )
+
+    const { getByRole, rerender } = render(tree(null))
+    expect(getByRole('checkbox')).not.toBeChecked()
+
+    rerender(tree('2026-08-20T00:00:00.000Z'))
+    await waitFor(() => expect(getByRole('checkbox')).toBeChecked())
+  })
```
jianwei1, line 618: **Concern:** this doesn't reproduce the failure mode QA-564 actually reported.

The bug in the description is: *user ticks the box → a cache-driven reinitialize resets `values.featured` → the box still displays the tick*. This test never clicks anything — it renders with `featuredAt: null`, rerenders with a date, and asserts the box is checked. That's the inverse direction.

It is a legitimate red/green (React sets the checkedness dirty flag in `postMountWrapper`, so `defaultChecked` can't follow a rerender after mount — it does fail without the fix). But a test that clicks the checkbox and *then* rerenders with `featuredAt` still `null`, asserting the box goes back to unchecked, would pin the actual reported symptom: no divergence between what's displayed and what Formik holds.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
```diff
@@ -586,4 +587,116 @@ describe('TemplateSettingsDialog', () => {
...
+
+  // QA-564: the customization update triggers an unawaited refetch of
+  // GetPublisherTemplate. If it were issued before journeyFeature, it could
+  // read pre-feature state and land after journeyFeature's response,
+  // overwriting featuredAt in the cache with a stale null.
+  it('applies the featured change before the customization update fires its refetch', async () => {
```
jianwei1: **Nit:** the name overstates what's asserted — "…fires its refetch", but no `GetPublisherTemplate` query is active in this test, so no refetch actually occurs and nothing about it is checked.

Mutation ordering is a fair proxy, so either rename to match (`applies the featured change before the customization update`), or mount an active `GetPublisherTemplate` query so the refetch itself becomes observable.

(The `invocationCallOrder` assertion itself is sound — under `MockLink` those fire at response delivery rather than request issue, which is the same ordering for a sequential await chain.)

---

jianwei1 (commented):
## Review summary

**Concerns:** 3 (see inline) · **Nits:** 1 · **Critical:** none

Both root causes check out. `JOURNEY_FIELDS` includes `featuredAt` (`journeyFields.ts:33`) and `GET_PUBLISHER_TEMPLATE` spreads it, so a stale refetch really would write `null` over the cache — the race isn't theoretical. And `defaultChecked` genuinely can't follow a reinitialize after mount, so both new tests are valid red/green.

The concerns are about the *shape* of the fix rather than its correctness:

1. Reordering the mutations changes partial-failure semantics — a `journeyFeature` rejection now drops the customization edit, which previously survived. Most relevant if this lands before #9526.
2. The refetch race is ordered around rather than removed. `LocalTemplateDetailsDialog` already solves the same problem for the same mutation with a targeted `cache.modify` and no refetch.
3. The checkbox-sync test asserts the inverse of the reported symptom — valid red/green, but it doesn't pin what QA-564 described.

**Worth acknowledging:**

- Dropping `value={values.featured}` alongside the `defaultChecked` → `checked` switch is the right call, not just tidying. Formik's `getValueForCheckbox` short-circuits to `Boolean(checked)` when the current value is a boolean, so `value` was dead weight — and it was rendering `value="false"` on the input, a latent hazard if `featured` ever became array-shaped.
- Documenting the ordering constraint in code *and* pinning it with a test is the right pair. Comment-only would be fragile; test-only would be unexplained.
- Genuinely tight scope: FE-only, no schema coupling, deploy-order independent, and the description says so explicitly.

**Not raised inline:** `enableReinitialize` still wipes in-flight text edits (title, description, customization) on any journey cache update. This PR makes the featured desync *visible* rather than silent, which is a real improvement, but the broader data-loss hazard is unchanged — out of scope here, possibly worth its own ticket.

CI is green on `c4ae0de` (the three earlier `danger` failures are superseded by the 02:48 run, which passed).

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
csiyang: Adopted in 362b81e3c — handleSubmit now mirrors LocalTemplateDetailsDialog: all three mutations run via `Promise.allSettled`, so a journeyFeature rejection no longer drops the customization update (pinned by a new test, verified red against the sequential structure). The ship-before-#9526 window you flagged is closed by structure rather than merge order.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
csiyang: Adopted in 362b81e3c — `refetchQueries` is replaced with the sibling's targeted `cache.modify` patch, so the refetch (and with it the race and the ordering invariant) no longer exists. Good call that the invariant didn't need to exist.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
csiyang: Added in 362b81e3c — a test now clicks the box, triggers a reinitialize via a journey change with `featuredAt` still null, and asserts the box snaps back: display and Formik state can no longer diverge. Kept the original sync test as the complementary direction.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
csiyang: Superseded in 362b81e3c — with the refetch gone the ordering invariant no longer exists, so that test is deleted rather than renamed; the allSettled partial-failure test replaces it as the structural guarantee.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -91,55 +91,71 @@ export function TemplateSettingsDialog({
...
+          cache.modify({
+            id: cache.identify({ __typename: 'Journey', id: journey.id }),
+            fields: {
+              journeyCustomizationDescription() {
+                return values.journeyCustomizationDescription
+              }
+            }
+          })
+        }
       })
```
jianwei1, line 136: **Concern:** this `update` callback is now the load-bearing half of the fix, and nothing tests it.

It replaced the refetch, and it's what keeps `journey.journeyCustomizationDescription` fresh so reopening the dialog doesn't show a stale value. Delete the callback and all 34 tests still pass — the regression would surface only as "save, reopen Template Settings, see the old description".

`LocalTemplateDetailsDialog` doesn't cover its equivalent either, so there's no in-repo pattern to copy — but this PR is the one that newly depends on the behaviour here, and untested load-bearing cache behaviour is how this class of bug arrived in the first place. Passing a shared `new InMemoryCache()` into `MockedProvider` and asserting the `Journey` entity after save would pin it.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
```diff
@@ -586,4 +587,134 @@ describe('TemplateSettingsDialog', () => {
...
+              variables: { journeyId: defaultJourney.id, string: '' }
+            },
+            result: customizationResult
+          },
+          {
+            request: {
+              query: JOURNEY_FEATURE_UPDATE,
+              variables: { id: defaultJourney.id, feature: true }
+            },
+            error: new Error('user is not allowed to update featured date')
```
jianwei1: **Concern:** this models the ACL rejection as a network error, so the test exercises the opposite branch from the real one.

Under `MockLink` the `error` key calls `observer.error()`, so Apollo wraps it as `ApolloError` with `networkError` set — which is why line 713 asserts `'Field update failed. Reload the page or try again.'`. But the real QA-563 failure this message text is quoting is a *GraphQL* error, and that lands on the other branch (`t('Something went wrong, please reload the page and try again')`, `TemplateSettingsDialog.tsx:156`).

So the mock's wording implies one failure shape while the mechanism is another, and the branch the actual ACL rejection hits has no coverage at all.

`result: { errors: [new GraphQLError('user is not allowed to update featured date')] }` still rejects the mutation (default `errorPolicy: 'none'`), models the real case, and covers the untested branch. The `allSettled` guarantee this test exists to prove holds either way.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
```diff
@@ -91,55 +91,71 @@ export function TemplateSettingsDialog({
...
-      }
+      return
     }
+
+    const networkError =
+      failed.reason instanceof ApolloError && failed.reason.networkError != null
+    enqueueSnackbar(
+      networkError
+        ? t('Field update failed. Reload the page or try again.')
+        : t('Something went wrong, please reload the page and try again'),
```
jianwei1: **Nit:** `results.find(...)` on line 140 picks the failure by array position rather than by severity, so the message can be the less actionable one.

If `journeySettingsUpdate` fails with a GraphQL error *and* `journeyFeature` fails with a network error, `failed` is the first, and the user gets the generic "Something went wrong" despite a genuine network problem.

```ts
const networkError = results.some(
  (result) =>
    result.status === 'rejected' &&
    result.reason instanceof ApolloError &&
    result.reason.networkError != null
)
```

The old sequential code had the same bias (first throw wins), so this isn't a regression — just worth picking up while the block is being rewritten anyway.

---

jianwei1 (commented):
## Review summary — round 2 (`362b81e3c`)

All four findings from the first pass are addressed, and the rewrite is better than what I suggested: wiring the two `LocalTemplateDetailsDialog` patterns together also collapsed the ordering invariant entirely, so the code comment now explains *why there is no constraint* instead of guarding one.

Since `handleSubmit` was rewritten rather than tweaked, I re-reviewed the new structure from scratch. **2 Concerns and 1 Nit, no Critical** — see inline.

**On the concurrency the rewrite introduces**, since that's the crux and worth stating as verified rather than assumed: `JourneyUpdateInput` has no `featuredAt`, `JOURNEY_SETTINGS_UPDATE` doesn't select it, and `journeyUpdate.mutation.ts:74-94` writes only the columns present in `input` via Prisma — so `journeyUpdate` and `journeyFeature` touch disjoint columns in both the Apollo cache and Postgres. No lost update. Building `tasks` synchronously before any `await` also means the `Boolean(journey.featuredAt) !== values.featured` guard can't be raced by a mid-submit reinitialize. The parallelism is sound.

**New findings:**

1. **Concern** — the `cache.modify` patch that replaced the refetch has no test coverage. Delete the `update` callback and all 34 tests still pass.
2. **Concern** — the partial-failure test mocks the ACL rejection as a network error, so it exercises the opposite branch from the real QA-563 failure, leaving that branch uncovered.
3. **Nit** — `results.find(rejected)` classifies the error message by array position, so a network failure can be masked by an earlier GraphQL one.

**Worth acknowledging:**

- The three replacement tests are well chosen. `snaps the checkbox back when a reinitialize resets the form mid-edit` is exactly the reported symptom and is red under `defaultChecked`; keeping the original sync test alongside it covers the opposite direction.
- Deleting the ordering test rather than renaming it was the right call — the invariant no longer exists to guard.
- The rewrite incidentally picked up a `t()` wrapper on `'Something went wrong, please reload the page and try again'`, fixing a pre-existing i18n miss. Both keys already exist in `libs/locales/en/apps-journeys-admin.json` (lines 586 and 918), so there's no extraction gap and nothing for autofix.ci to add.
- The thread replies were specific about what changed and where, which made this pass quick.

Concern 1 is the only one I'd want resolved before merge — untested load-bearing cache behaviour is how this class of bug got here in the first place. 2 and 3 are fine as follow-ups.

CI at time of review: `test (1/3)`, `build`, `danger`, CodeQL and schema checks green; `test (2/3)`, `test (3/3)`, `lint-work` and the preview deploy still running.

---

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
csiyang: Pinned in 0986e23ee — a shared `InMemoryCache` seeded with a stale `journeyCustomizationDescription` now asserts the Journey entity is patched on save, verified red with the callback removed. Agreed this was the load-bearing half; thanks for catching that the sibling doesn't cover its own either — worth a follow-up there.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.spec.tsx
csiyang: Fixed in 0986e23ee — the mock now rejects with `result: { errors: [new GraphQLError(...)] }`, matching the real ACL failure shape, and the assertion moved to the generic-message branch, which picks up its first coverage. The allSettled guarantee holds as you noted.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
csiyang: Adopted verbatim in 0986e23ee — the network-error check now scans all settled results via `.some()`, so a genuine network problem isn't masked by an earlier GraphQL error.

apps/journeys-admin/src/components/Editor/Toolbar/Items/TemplateSettingsItem/TemplateSettingsDialog/TemplateSettingsDialog.tsx
csiyang: Update: the sibling's equivalent is now covered in this PR too (6c3b39474) rather than deferred — same seeded-cache pattern, verified red with its callback stubbed. Both dialogs' cache patches are pinned.

jianwei1 (approved):
## Approving ✅

All three round-2 findings are fixed, and I verified each rather than taking the commit messages at their word:

- **`cache.modify` coverage** — the new test would genuinely be red without the `update` callback: absent it, `journeyCustomizationFieldPublisherUpdate` writes nothing to the Journey entity, so `'stale cache value'` survives.
- **Error shape** — `result: { errors: [new GraphQLError(...)] }` rejects under the default `errorPolicy: 'none'`, so it lands on the generic-message branch, and both other mutations are still asserted to complete. The `allSettled` guarantee the test exists to prove still holds.
- **`.some()`** — `results.every(fulfilled)` is exactly equivalent to the old `find(rejected) == null`, and TS narrows `.reason` correctly through the `status === 'rejected'` guard.

One thing I checked rather than assumed, since it would have been an invisible failure: `cache.modify` only invokes a modifier for fields **already present** on the entity, so a test that seeds the field could pass while the patch silently no-ops in production. It doesn't — `journeyFields.ts:114` selects `journeyCustomizationDescription` and `GET_PUBLISHER_TEMPLATE` spreads `JourneyFields`, so the field is present on the real cached entity. The seeding mirrors production.

Pinning `LocalTemplateDetailsDialog`'s identical `update(cache)` in `6c3b394` was good unprompted follow-through — that's exactly where the pattern would drift first.

CI green across all 21 checks on `6c3b394`.

---

**One nit before merge — the description is stale.** It still describes a 3-file change confined to `TemplateSettingsDialog`, but the PR is now 4 files and edits a spec under `JourneyDetails/LocalTemplateDetailsDialog`, outside the `TemplateSettingsItem` tree. The "Fix" section doesn't mention the `.some()` error-message preference or either cache test, and "All 34 tests across the `TemplateSettingsItem` tree pass" is out of date. Worth a refresh since the body is the merge record — not blocking the approval.

**Non-blocking observation for later:** with `allSettled`, partial success is now a reachable state (title saved, featured didn't) and the single generic snackbar can't say which part failed. That ambiguity predates this PR and matches the sibling dialog, so it's consistent rather than a regression — but `results` carries the rejection reasons if it's ever worth naming the failed operation.

Nice iteration on this one. The ordering invariant is gone rather than documented, partial failure is structural rather than incidental, and both halves of the fix are pinned by red-verified tests.

---

Issue #9515, opened 2026-08-19
Rebuild Video.availableLanguages as one cycle-safe, recursively-correct source of truth


tanflem:
## Problem Statement

`Video.availableLanguages` is supposed to be the union of a video's own published-variant languages plus every language available anywhere in its published descendant tree (a collection/series/featureFilm can nest videos to arbitrary depth via a self-referential `children` relation). In practice it is maintained by hand-rolled, incremental application code scattered across many mutation call sites, and that code is provably wrong in more than one way:

- Removing a video from a container never re-checks whether the language it provided should disappear from that container — by the time reconciliation fires, the relation is already severed, so the old parent is structurally unreachable and the stale language sticks forever.
- The one cascade that does exist stops after a single hop up the tree, so a three-level-deep container (e.g. `featureFilm` → `series` → `video`) never receives an update at its outermost level even on the already-shipped add path.
- At least two more call sites bypass the shared helper entirely with their own hand-rolled array math and their own, different definition of "parent," and one wired-up worker job overwrites the field from a video's own variants only, silently discarding every child-derived language on any container it touches.
- The underlying `children`/`parents` relation is an unguarded graph (not a tree) with no cycle prevention anywhere — nothing stops a video from being made its own ancestor today.
- None of this is meaningfully tested: the one existing test for the cascade only ever wires up a single level of parents, so it could not have caught the depth bug even though it technically exists; the parent-variant creation/cleanup functions have placeholder tests that assert only `typeof fn === 'function'`.

The user-facing symptom is real: catalog consumers (the public Arclight API, the watch app, journeys-admin) can show a collection as missing a language its own children actually have, or keep showing a language nothing provides anymore. This is the confirmed root cause blocking two in-flight PRs (#9384, #9386) and the umbrella VMT-318 fix (Wonder Series language-sync bug, tracked under issue #9381).

## Solution

Replace the scattered, incremental, hand-mutated maintenance of `availableLanguages` with a single correct calculation, used everywhere the field is ever written, plus a batch verifier that can prove (and self-heal) the invariant across the whole catalog on demand. Concretely:

- One recompute function becomes the only way this field is ever set. It is implemented as an application-level, cycle-safe upward walk (not a raw SQL recursive query) so it stays testable through this repo's existing mocked-Prisma test seam without introducing a new integration-test tier.
- Every mutation, worker, and script that can affect a video's derived language set is routed through this one function, cascading all the way to the root, not stopping at one hop.
[...]

PR #9528, opened 2026-08-20
fix(api-media): sync videoChildren worker as true replace, not additive

### LINKED ISSUE #9522: availableLanguages: videoChildren worker — additive sync to true replace
## Parent

## What to build

The worker that syncs the `children` relation from the `childIds` scalar array is changed from additive-only to a true replace, so a relation edge removed from `childIds` by any write path can't persist forever as a stale connection. This is independent of the recompute engine — it's a data-integrity hardening of the relation the cascade walks, not a change to the cascade itself.

## Acceptance criteria

- [ ] A video whose `childIds` shrinks (a child id removed) has that child correctly disconnected from the `children` relation after the worker runs, not just new children connected.
- [ ] A video whose `childIds` grows continues to get the new child correctly connected (no regression).
- [ ] Running the worker twice in a row against unchanged data is a no-op (idempotent).

## Blocked by

None (can start immediately) — independent of the recompute engine work.

tanflem:
Closes #9522

## Summary

Sub-issue of #9515. Hardens the `videoChildren` worker (`apis/api-media/src/workers/videoChildren/service/service.ts`) that syncs the `Video.children` relation from the `childIds` scalar array, so it's a true replace instead of additive-only.

## Changes

- `children: { connect: [...] }` → `children: { set: [...] }` — each run now replaces the relation to match `childIds` exactly. A childId removed from the array is disconnected, not left as a stale edge.
- The candidate query is widened from `childIds: { isEmpty: false }` to also include videos where `children: { some: {} }`. Without this, a video whose `childIds` shrinks all the way to empty would be skipped entirely by the old filter, leaving its previously connected children stuck forever — the same latent-drift bug the issue targets, at the boundary case.
- `set` is naturally idempotent: running the worker twice against unchanged data recomputes and reissues the same target list, so the resulting relation state is unchanged.

## Testing

- Added test coverage for: a child removed from `childIds` gets disconnected (not just new children connected), `childIds` shrinking all the way to empty disconnects all stale children, and running the worker twice against unchanged data is idempotent — alongside the existing grow-case coverage.
- `bash tools/scripts/lint-changed.sh` — pass
- `pnpm tsc7 -p apis/api-media/tsconfig.ts7.json` — clean
- `vitest run` against apis/api-media's full suite — 78 files, 782 tests, all passing (includes the 4 `videoChildren/service` tests)

Note: `pnpm exec nx affected` itself is broken in this worktree (`TypeError: WorkspaceContext is not a constructor` on every subcommand, including `nx show projects` and `nx reset`) — unrelated to this change, so verification was done via the direct commands above instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Video child relationships now stay synchronized with the latest configured list.
  * Removed child relationships are automatically disconnected.
  * All child relationships are cleared when no children remain.
  * Repeated synchronization produces consistent results without duplicating changes.

### DIFF
--- a/apis/api-media/src/workers/videoChildren/service/service.spec.ts
+++ b/apis/api-media/src/workers/videoChildren/service/service.spec.ts
@@ -72,6 +72,12 @@ describe('videoChildren/service', () => {
           { id: 'video1', childIds: [] } as unknown as Video
         ])
       await service()
+      expect(prismaMock.video.findMany).toHaveBeenNthCalledWith(2, {
+        select: { id: true, childIds: true },
+        where: {
+          OR: [{ childIds: { isEmpty: false } }, { children: { some: {} } }]
+        }
+      })
       expect(prismaMock.video.update).toHaveBeenCalledWith({
         where: { id: 'video1' },
         data: {
--- a/apis/api-media/src/workers/videoChildren/service/service.spec.ts
+++ b/apis/api-media/src/workers/videoChildren/service/service.spec.ts
@@ -22,7 +22,7 @@ describe('videoChildren/service', () => {
         where: { id: 'video1' },
         data: {
           children: {
-            connect: [{ id: 'video2' }, { id: 'video3' }]
+            set: [{ id: 'video2' }, { id: 'video3' }]
           }
         }
       })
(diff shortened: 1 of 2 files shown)


tanflem:
@coderabbitai review

tanflem:
## Review feedback addressed

**Fixed:** 1 thread(s)
- CodeRabbit: strengthened the "disconnect all children when childIds becomes empty" test to assert the second `findMany` call includes the `{ children: { some: {} } }` OR predicate, so a regression that dropped that branch would actually fail the test instead of passing on an unconditional mock.

**Challenged:** 0 thread(s)
**Skipped:** 0 thread(s)

Kneesal (approved):

Issue #9517, opened 2026-08-19
availableLanguages: cycle-safe, cascade-to-fixpoint recompute engine


tanflem:
## Parent

Sub-issue of #9515 (Rebuild Video.availableLanguages as one cycle-safe, recursively-correct source of truth).

## What to build

The shared `availableLanguages` calculation becomes a single, cycle-safe, full-recompute walker that cascades upward to the root of a container hierarchy instead of stopping after one hop. It is implemented as an application-level (Prisma-backed, level-batched) walker, not a raw SQL recursive query — this repo's api-media tests mock Prisma entirely with no real-database integration tier, so a raw-SQL recursion's correctness (in particular the cycle guard) would not be verifiable by any existing test seam.

Every mutation, worker, and script that already calls through the shared helper (variant create/update/delete, the reconciliation pipeline, the Mux upload worker) inherits correct multi-level cascade behavior automatically — their call signatures don't change, only the correctness of what's behind them.

Incremental array mutation (appending/filtering a language in place) is retired in favor of always recomputing a video's value from its own current published variants unioned with its live children's current values — this is what makes removal symmetric with addition, with no separate "did we forget to remove" logic needed.

## Acceptance criteria

- [ ] A fixture tree at least three levels deep (leaf video → mid-level container → outer container): changing the leaf's published variant updates the outermost container's `availableLanguages`, not just the immediate parent.
- [ ] A fixture where a container's children relation includes itself (directly or indirectly): the walker terminates (throws or returns cleanly) instead of hanging.
- [ ] A fixture with a container whose two children both provide an overlapping language: removing the child that provided it uniquely drops the language; removing the other (still provided by the sibling) does not.
- [ ] The cascade stops walking upward once a level's recomputed value doesn't change (no unnecessary writes to unaffected ancestors).
- [ ] All existing tests for callers that already go through the shared helper (variant create/update/delete, reconciliation, upload worker) still pass unmodified.
- [ ] Incremental add/remove array-mutation helpers are removed or no longer used as a write path; every write goes through full recomputation.

## Blocked by

None (can start immediately) — this is the prefactor everything else in #9515 builds on.

tanflem:
🔒 Phoebe holds a lease on this issue — runtime `23d237d8-ffd4-4f4e-960e-7682f69cf890`, branch `phoebe/bot-0-issue-9517`.

[...]

PR #9531, opened 2026-08-20
fix(api-media): parent-variant create/cleanup use shared recompute + lookup

### LINKED ISSUE #9519: availableLanguages: parent-variant create/cleanup use the shared recompute + lookup
## Parent

## What to build

Parent-variant creation and cleanup (`createEmptyParentVariant`, `checkAndRemoveEmptyParentVariant`) stop hand-rolling their own array math and their own definition of "parent." Today they locate parents via the `childIds` scalar array with no container-label restriction — a second, non-identical definition of "parent" that can disagree with the relation-based, label-restricted lookup the rest of the cascade uses. Both are changed to use the same shared parent lookup and to write through the recompute engine from #9517 instead of inline array mutation.

## Acceptance criteria

- [ ] `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant` locate parents using the same relation-based, container-label-restricted lookup used elsewhere in the cascade, not the `childIds` scalar array.
- [ ] Both write `availableLanguages` through the shared recompute engine, not hand-rolled array mutation.
- [ ] The current placeholder tests for parent-variant creation/cleanup (which assert only that the functions exist) are replaced with real behavioral tests: creating an empty parent variant results in the correct `availableLanguages`; cleaning one up results in the correct `availableLanguages` after removal.
- [ ] All existing callers of these two functions (variant create/update, video publish-flip, `videoPublishChildren`) continue to behave correctly.

## Blocked by

- #9517 (needs the recompute engine as the write path)

tanflem:
Closes #9519

## Summary

Sub-issue of #9515. `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant` (and their callers `handleParentVariantCreation`/`handleParentVariantCleanup`) had their own, non-identical definition of "parent": a `childIds` scalar-array lookup with no container-label restriction, separate from the relation-based, label-restricted `findContainerParentIds` used elsewhere in the availableLanguages cascade. They also mutated `availableLanguages` inline instead of writing through the shared recompute engine from #9517.

## Changes

- `handleParentVariantCreation`/`handleParentVariantCleanup` now locate parents via `findContainerParentIds` instead of `childIds: { has: videoId }`.
- `checkAndRemoveEmptyParentVariant` now enumerates a parent's children via the `children` relation instead of the `childIds` scalar array (the same reasoning applies: `childIds` is a hand-maintained sort-order field, not the source of truth for container membership).
- `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant` now call `updateVideoAvailableLanguages` after mutating a variant, instead of hand-rolling the union/filter of the `availableLanguages` array.
- Replaced the placeholder "functions exist" tests with real behavioral tests that exercise the actual `findContainerParentIds` and `updateVideoAvailableLanguages` implementations (only the unrelated cascade primitives - `addLanguageToVideo`, `removeLanguageFromVideoIfUnused`, `updateParentCollectionLanguages` - stay mocked), asserting the resulting `availableLanguages` value after creating and after cleaning up an empty parent variant.

## Dependency note

Blocked by #9517 (the recompute engine). That work is done (PR #9530) but not yet merged to `main`, so this branch is built on top of `phoebe/bot-0-issue-9517` and this PR's diff currently includes #9530's commit as well. Once #9530 merges, rebasing this branch onto `main` will drop the duplicate commit and shrink the diff to just the two files below.

## Files changed

- `apis/api-media/src/schema/videoVariant/videoVariant.ts`
- `apis/api-media/src/schema/videoVariant/videoVariant.spec.ts`

## Verification

- `pnpm exec nx affected -t lint,type-check,test --base=origin/main` failed to run in this environment (nx's native binary is broken here — pre-existing, unrelated to this change: `nx run api-media:lint` fails the same way with zero changes). Fell back to running the underlying tools directly against `api-media`:
  - `pnpm lint:changed --fix` — passed clean
  - `tsc --noEmit` on both `tsconfig.app.json` and `tsconfig.spec.json` — clean
  - `vitest run` — 78 test files / 788 tests passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved language propagation across nested parent videos.
  - Prevented stale language values when variants are added or removed.
  - Added safeguards for circular parent relationships and unchanged language sets.
  - Improved parent-video variant lookup and cleanup behavior.

- **Tests**
  - Added coverage for cascading updates, language retention and removal, parent lookup, variant creation, and cleanup.


tanflem:
## Review feedback addressed

**Fixed:** 1 thread(s)
**Challenged:** 0 thread(s)
**Skipped:** 0 thread(s) (outdated / already addressed)

### Details

- CodeRabbit flagged that `checkAndRemoveEmptyParentVariant` only recomputed the immediate parent's `availableLanguages` after deleting an empty placeholder variant, with nothing cascading that removal further up to grandparents. Confirmed valid and, on closer look, broader than the reported scope: `video.ts`'s publish-status toggle path calls `handleParentVariantCleanup`/`handleParentVariantCreation` without any subsequent cascade call of its own (unlike the `videoVariant.ts` mutations, which already call `updateParentCollectionLanguages` afterward), so a grandparent's `availableLanguages` could go stale indefinitely once the immediate parent's value changed via that path.
  - Fix: both `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant` now call `updateParentCollectionLanguages(parentVideoId)` after recomputing the immediate parent, so the cascade-to-fixpoint walk continues upward regardless of whether the caller cascades itself.
  - Added assertions to the existing behavioral tests confirming the cascade call fires with the correct parent id.

### Verification

- `pnpm exec nx affected -t lint,type-check --base=origin/main` — still fails to run in this environment (pre-existing nx native-binary breakage, unrelated; confirmed it fails the same way on unmodified `main`).
- `pnpm lint:changed --fix` — passed clean.
- `tsc --noEmit` on both `tsconfig.app.json` and `tsconfig.spec.json` for api-media — clean.
- `vitest run` (api-media) — 787/787 tests passed across 77 files; 1 unrelated pre-existing failure (`video-variant-reconciliation-migration.spec.ts`, missing migration.sql — confirmed present on the branch before this fix too, unaffected by this change).

tanflem:
Mergetron attempted an idle merge-conflict fix (merge `origin/main` into this branch) for PR #9531 but could not resolve it cleanly. The branch was left unchanged (`git merge --abort`). A human should resolve the conflicts manually.

mikeallisonJS (commented):
Reviewed against #9519. The lookup change (AC1–AC3) is right — `findContainerParentIds` replacing the `childIds: { has }` scan, the `children` relation replacing `parentVideo.childIds`, and real assertions replacing the "functions exist" placeholders. Three things concern me:

**1. Both `prisma.$transaction` wrappers are gone.** The old code wrapped `videoVariant.create` + `video.update` (and `delete` + `update`) atomically. Now the variant write and the recompute are separate statements, and if `updateVideoAvailableLanguages` throws, the caller's `.catch` only logs — leaving a created or deleted variant with stale `availableLanguages`. The commit body doesn't mention dropping this, so I want to check it was deliberate.

**2. Concurrent writes to the same parent can clobber each other.** `handleParentVariantCreation` maps `parentIds` through `Promise.all`, and `videoPublishChildren` calls it concurrently for many `(videoId, languageId)` pairs that share one parent. `updateVideoAvailableLanguages` is a read-then-`set:`, and the comment in `updateAvailableLanguages.ts` warns about precisely this: "a plain recompute … does not have the same serialization guarantee … two independent read/write pairs can still interleave and clobber each other's write." The write-path change AC2 asks for is the pattern that file flags as unsafe on a concurrent path.

**3. adf825dbe is out of scope for this issue.** Adding `updateParentCollectionLanguages(parentVideoId)` to both handlers cascades to the hierarchy root — that's #9517/#9515 territory. #9519's ACs are scoped to *lookup* and *write path*. It also means every caller now gets `enqueueVideoAlgoliaSync` + `videoCacheReset` per parent (the old inline `tx.video.update` did not), multiplied by the cascade depth. If the cascade belongs here, it'd be clearer as its own PR against the right issue.

Also, AC4 ("all existing callers continue to behave correctly") has no test or verification behind it, and both live callers — `video.ts:916/919` and `videoPublishChildren.mutation.ts:237` — inherit the new side effects above.

Standards:

- `videoVariant.ts:213` and `:252` are both edited by 33730a675 and both still `console.error`, in silent-swallow handlers on a cascade path. `apis/AGENTS.md` → Logging says Pino everywhere, and the file already imports `logger`. Cheap to fix while you're in there.
- The `updateVideoAvailableLanguages(parentVideoId)` + `updateParentCollectionLanguages(parentVideoId)` pair, under a near-verbatim 8-line comment, is now pasted into both `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant`. "Recompute then cascade" is one idea that every future call site has to remember to spell as two calls — worth a `recomputeAndCascade(videoId)`.
- Both new behavioural tests queue six positional `mockResolvedValueOnce` stubs whose meaning lives only in comments; any reordering inside the implementation silently rebinds them.

tanflem:
## Addressed: transactional atomicity + the read-then-set race

Both defects from the review are fixed in 7b0a63c. The shared-helper deduplication this PR introduced (`findContainerParentIds`, the shared recompute engine) is kept as-is — only the transaction boundary and the concurrency guarantee changed.

### 1. Restored the transaction boundary

The refactor dropped both `$transaction` wrappers, so a parent variant write and its `availableLanguages` recompute were no longer atomic: if the recompute threw, the variant row stayed committed and the parent's `availableLanguages` was left permanently stale, with nothing to retry it.

Rather than reinstating the two hand-rolled `$transaction` blocks that were on `main` (which is what forced the duplicated union/filter logic in the first place), the boundary now lives in the shared engine as a new seam:

```ts
export async function withAvailableLanguagesRecompute<T>(
  videoId: string,
  write: (tx: Prisma.TransactionClient) => Promise<T>,
  options?: { skipCache?: boolean; skipAlgolia?: boolean }
): Promise<T>
```

The caller's write and the recompute run in one transaction and commit or roll back together. `createEmptyParentVariant` and `checkAndRemoveEmptyParentVariant` each pass their single write through it, so the atomicity is restored without either of them going back to hand-rolling the language array.

One related change fell out of this: the cache reset and Algolia enqueue are now deferred until **after** commit. Both re-read the video from the database, so firing them inside the open transaction could publish pre-commit state.

### 2. The read-then-set race on shared parents

The recompute reads the parent's current source data and writes back a computed value, so two concurrent variant writes against the same parent could each read the pre-write state and the second clobber the first — exactly the lost-update race the existing comment on `addLanguageToVideo` warns about.

`addLanguageToVideo`'s own fix (a single self-referential atomic `UPDATE`) doesn't transfer here: the recomputed value is a union over *other* rows (the video's published variants and its published children's stored values), not a mutation of the column in terms of itself. So this takes the other option the comment leaves open — the recompute now holds the video's row lock across the read and the write:

```ts
await tx.$executeRaw`SELECT id FROM "Video" WHERE id = ${videoId} FOR UPDATE`
```

Concurrent recomputes of the same video serialize; the loser blocks, then re-reads the winner's committed state instead of overwriting it. This applies to *every* recompute path, not just parent-variant create/cleanup — `updateVideoAvailableLanguages` and the upward cascade go through it too.

The lock is taken **first**, before the caller's write. A variant insert or delete makes Postgres take a `FOR KEY SHARE` lock on the referenced video row; upgrading that to `FOR UPDATE` afterwards would let two concurrent callers deadlock, each holding the shared lock and waiting on the other. Locking up front gives every transaction the same acquisition order.

I also updated the `addLanguageToVideo` comment, since its claim that "a plain recompute does not have the same serialization guarantee" is no longer true — it now explains why that primitive is still worth keeping (no transaction, no lock wait on the hot path).

Two smaller consequences:

- The cascade's "did this level actually change?" comparison now reads the previous value **under the same row lock** as the recompute, instead of in a separate unlocked query beforehand. It could otherwise be decided against a value a concurrent writer had already replaced, silently stopping the cascade early.
- Checked every `updateVideoAvailableLanguages` call site for transaction nesting (`videoPublishChildren.mutation.ts:505`, `video.ts:1106`) — both run after their surrounding transaction has committed, so there's no nested-transaction or self-deadlock path.

### Tests

9 new/changed assertions, all verified red against this PR's head (`ed041f04f`) and green after the fix.

`videoVariant.spec.ts` — the requested rollback guard. A small fake models Postgres commit/rollback over the prisma mock: a write issued while a `$transaction` callback is in flight is staged and only becomes durable when that callback resolves; a write issued outside one autocommits immediately. That lets the tests assert what actually *survives*, rather than merely that `$transaction` was called.

- `rolls the parent variant create back when the recompute throws` — the create is attempted, nothing is durable. On the pre-fix code this reports `[{ operation: 'videoVariant.create', inTransaction: false }]`, i.e. the committed orphan row.
- `rolls the parent variant delete back when the recompute throws` — same, for cleanup.
- `commits the parent variant create and the recompute together` — both writes durable, both `inTransaction: true`. Red pre-fix because both were autocommitted separately.
- `locks the parent row before reading it, so concurrent recomputes cannot lose an update` — asserts the `FOR UPDATE` statement, its bound video id, and that it precedes the variant write.

`updateAvailableLanguages.spec.ts` — at the engine seam:

- `locks the video row before reading it, and writes while the lock is held` — asserts lock-before-read and write-under-lock via invocation ordering.
- `runs the caller write and the recompute in one transaction, then syncs`
- `propagates a failed recompute and skips the post-commit sync` — nothing is published to cache or Algolia for a change that never committed.

**Seam limitation, stated plainly:** these run against the prisma mock, so they verify the *shape* of the fix (one transaction, lock before read, correct ordering) rather than executing real concurrent transactions against Postgres. That matches how the existing sibling guard for this race is written (`addLanguageToVideo`'s tests assert on the SQL shape). A true interleaving test would need a real database in the unit suite.

### Verification

- `npx vitest run --root apis/api-media --config apis/api-media/vitest.config.mts --coverage=false` — **87 files / 893 tests passed**
- `pnpm lint:changed --fix` — clean
- `pnpm exec nx affected --target=type-check --base=origin/main` — passed

> Note on the suggested command: `apis/api-media` has no `jest.config.ts` — this workspace is on Vitest (`apis/api-media/vitest.config.mts`), per `docs/agents/testing.md`. I used the Vitest equivalent above, and did not use `nx test`.

### One thing I did not change

`removeLanguageFromVideoIfUnused` has the same read-then-set shape inside its transaction, with no row lock. It's pre-existing on `main` and untouched by this PR, so I left it alone rather than widening the diff — flagging it as a likely follow-up.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 24 checks green, no unresolved threads, closes #9519. +1004/-145 across 5 files, rewriting parent-variant create/cleanup to use the shared recompute and lookup.

### What needs to happen next

Run it through `/code-review` before Nisal. Largest of the `availableLanguages` cluster by net change — worth confirming the shared-recompute contract matches what #9530 and #9533 assume.

PR #9532, opened 2026-08-20
fix(api-media): back availableLanguages with NOT NULL default and GIN index (#9523)

### LINKED ISSUE #9523: availableLanguages: NOT NULL default + GIN index migration
## Parent

## What to build

A migration gives `Video.availableLanguages` a database-level `NOT NULL DEFAULT '{}'`, matching the Prisma client's existing non-null type guarantee (today the column has neither, and only one of two relevant call sites defensively guards against a possible `NULL`). A GIN index is added on the same column, backing the existing `hasSome`/`isEmpty` filter usage — most notably Arclight's paginated public catalog listing, which is filtering this column today with no index at all.

## Acceptance criteria

- [ ] Migration applies cleanly against the current schema.
- [ ] The column is `NOT NULL` with a `'{}'` default after migration; no existing row violates the new constraint.
- [ ] A GIN index exists on `availableLanguages`.
- [ ] `EXPLAIN` on Arclight's `/v2/media_components` `hasSome` filter query shows the index being used instead of a sequential scan.
- [ ] The defensive `COALESCE` guard against a possibly-null value (in the atomic append helper, if still present after #9517) can be safely removed, since the constraint now guarantees non-null at the database level.

## Blocked by

None (can start immediately) — independent schema/performance change.

tanflem:
Closes #9523.

## Summary

`Video.availableLanguages` is typed as a non-null `String[]` by the Prisma client, but the column had no constraint behind it — so NULL rows existed. That makes array filters behave surprisingly: `NOT (x = ANY(NULL))` is NULL rather than true, so a NULL row silently fails predicates callers expect it to match.

This backfills the NULLs, adds a `'{}'` default and the `NOT NULL` constraint so schema and client agree, and adds a GIN index — the existing `hasSome`/`isEmpty` array filters (`videos`/`videosCount`, and Arclight's direct-Prisma `/v2/media_components` listing) were doing unindexed array scans.

```sql
UPDATE "Video" SET "availableLanguages" = '{}' WHERE "availableLanguages" IS NULL;
ALTER TABLE "Video" ALTER COLUMN "availableLanguages" SET DEFAULT '{}';
ALTER TABLE "Video" ALTER COLUMN "availableLanguages" SET NOT NULL;
CREATE INDEX "Video_availableLanguages_idx" ON "Video" USING GIN ("availableLanguages");
```

## Provenance

Harvested from #9516, which was closed in favour of the nine-way decomposition of #9515. This is the one slice with no dependency on the recompute engine, so it can land immediately; the harvest map for the rest is recorded on #9516.

## Why this is safe to land first

Making the column `NOT NULL` with a default is backward-compatible with the current code, which guards every array operation with `COALESCE`. Those guards stay correct against a non-null column. Landing this **first** is what later allows #9530 to drop them — doing it in the other order would leave `array_append`/`ANY` operating on NULL and silently no-oping.

## Changes

- `migration.sql` — backfill, default, `NOT NULL`, GIN index
- `schema.prisma` — `@default([])` and `@@index([availableLanguages], type: Gin)`
- `pothos-types.ts` — regenerated; the only semantic change across the entire datamodel is `Video.availableLanguages.hasDefaultValue` flipping `false` → `true` (verified by parsing both versions and diffing the models)

## Validation

`pnpm lint:changed --fix` clean. Migration timestamp `20260819120000` sorts after the current head migration `20260721154000`.

Note this is a schema migration on a large table — the `CREATE INDEX` is not `CONCURRENTLY`, so it takes a brief lock on `Video`. Worth a human eye on deploy timing.


tanflem:

Closing as a duplicate of #9534 — both PRs declare `Closes #9523`.

**Why this one and not #9534:** #9534 is a strict superset. Both add the `NOT NULL DEFAULT '{}'` + GIN index migration, but #9534 also edits `apis/api-media/src/schema/video/lib/updateAvailableLanguages.ts` to drop the defensive `COALESCE` guard — which is acceptance criterion 5 of #9523. This PR is schema-only.

**Why it matters that only one merges:** the two carry different migration directories —

- this PR: `libs/prisma/media/db/migrations/20260819120000_available_languages_not_null_gin_index/`
- #9534: `libs/prisma/media/db/migrations/20260821060000_video_available_languages_not_null_gin_index/`

so they would not conflict textually and both could merge, at which point the second migration attempts to create an index that already exists.

Work continues in #9534.

PR #9536, opened 2026-08-21
refactor(api-media): dedupe Algolia client factories

### LINKED ISSUE #9434: Dedupe the two Algolia client factories in api-media
api-media has two independent Algolia client/config factories:

- `apis/api-media/src/lib/algolia/algoliaClient.ts` — `getRequiredEnv` / `getAlgoliaConfig` / `getAlgoliaClient`, covering `ALGOLIA_INDEX_VIDEOS` and `ALGOLIA_INDEX_VIDEO_VARIANTS`. Used by ~12 call sites across `videoAlgolia.ts`, `algoliaVideoUpdate`, `algoliaVideoVariantUpdate`, and the `algolia-helpers` scripts.
- `apis/api-media/src/lib/languages/updateLanguageInAlgolia.ts` — a near-identical private `getRequiredEnv` plus `getAlgoliaLanguagesClient`, covering `ALGOLIA_INDEX_LANGUAGES`.

The duplication predates #9343.

## Why it was not fixed in #9343

The obvious merge — adding `languagesIndex: getRequiredEnv('ALGOLIA_INDEX_LANGUAGES')` to `getAlgoliaConfig()` — was implemented and then deliberately backed out. Because `getAlgoliaClient()` calls `getAlgoliaConfig()`, which reads *every* index var eagerly, that change makes all ~12 video/video-variant call sites hard-fail wherever `ALGOLIA_INDEX_LANGUAGES` is unset. Terraform (`apis/api-media/infrastructure/locals.tf`) provisions it for deployed environments, but a partial local `.env` or a CI job that only provisions video secrets would start breaking **video** search sync as a side effect of a **languages** change. Not a trade worth making inside a bug fix.

## Suggested approach

Make config resolution lazy or per-index so callers only require the vars they actually use — e.g. `getAlgoliaIndex('languages')`, or split `getAlgoliaConfig()` into per-index getters sharing one `getRequiredEnv`. Then collapse the languages factory into the shared module.

Worth confirming as part of this: api-media writes the languages index with `ALGOLIA_API_KEY`, while api-languages uses a separate `ALGOLIA_API_KEY_LANGUAGES` for the same index. The two-key arrangement should be intentional and documented, or unified.

## Acceptance

- One `getRequiredEnv` and one client factory in api-media.
- Requiring a languages env var cannot break video indexing, and vice versa.
- Existing Algolia specs pass unchanged.

tanflem:
Closes #9434

## Summary

`apis/api-media` had two independent Algolia client/config factories: the shared `algoliaClient.ts` (used by ~12 video/video-variant call sites) and a near-identical private one in `updateLanguageInAlgolia.ts`. The obvious merge was tried and reverted in #9343 because `getAlgoliaConfig()` read every index env var eagerly — adding `languagesIndex` would have made every video call site hard-fail if `ALGOLIA_INDEX_LANGUAGES` was unset.

This PR does the merge the issue suggested: make index resolution **lazy per-index** instead of eager.

## What changed

- `algoliaClient.ts`: `AlgoliaConfig` now exposes `videosIndex`, `videoVariantsIndex`, and `languagesIndex` as **getters**. `appId`/`apiKey` stay eager (every caller needs both). Each index's `ALGOLIA_INDEX_*` env var is only required at the point a caller actually destructures that property — so requiring `ALGOLIA_INDEX_LANGUAGES` can no longer break video/variant callers, and vice versa. No call-site changes needed for the existing ~12 video/variant sites; they already do `const { videosIndex } = getAlgoliaConfig()`.
- `updateLanguageInAlgolia.ts`: dropped its local `getRequiredEnv` + `getAlgoliaLanguagesClient`, now calls the shared `getAlgoliaClient()` / `getAlgoliaConfig().languagesIndex`, matching the pattern every video call site already uses.

## The two-key question

The issue also asked to confirm whether api-media writing the languages index with `ALGOLIA_API_KEY` while api-languages uses a separate `ALGOLIA_API_KEY_LANGUAGES` for the same index is intentional. Both are separately Terraform-provisioned per service (`apis/api-media/infrastructure/locals.tf` vs `apis/api-languages/infrastructure/locals.tf`), which reads as an intentional least-privilege credential split across services rather than accidental drift. Unifying it would be a cross-service infra change and is out of scope for this dedupe.

## Testing

- `pnpm exec nx affected -t lint,type-check,test --base=origin/main` — passes (api-media: 78 files / 779 tests, lint 0 errors, type-check clean).
- `pnpm lint:changed --fix` — clean, no changes.
- Existing Algolia specs (`algoliaVideoUpdate`, `algoliaVideoVariantUpdate`, `videoAlgolia`, `updateLanguageInAlgolia`) pass unchanged — they either fully mock the module or only set/touch the index they care about, which is exactly what the lazy-getter change was designed to keep working.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a dedicated language search index.
  * Language updates and full reindexing now use shared search configuration.

* **Bug Fixes**
  * Search configuration errors are now reported when the affected index is used, while core credentials remain validated immediately.
  * Blank or whitespace-only search configuration values are now rejected with clear validation errors.

* **Tests**
  * Added coverage for search configuration loading, validation, and client creation.

### DIFF
--- a/apis/api-media/src/lib/algolia/algoliaClient.ts
+++ b/apis/api-media/src/lib/algolia/algoliaClient.ts
@@ -5,6 +5,7 @@ export type AlgoliaConfig = {
   apiKey: string
   videosIndex: string
   videoVariantsIndex: string
+  languagesIndex: string
 }
 
 function getRequiredEnv(name: string): string {
@@ -21,13 +22,28 @@ function getRequiredEnv(name: string): string {
  * NOTE: This is intentionally a function (not a module-level constant) so
  * unit tests that import GraphQL schema modules don't immediately throw
  * when Algolia env vars are not present.
+ *
+ * The index names are exposed as getters rather than eagerly-evaluated
+ * properties: appId/apiKey are shared by every caller so those still resolve
+ * up front, but each index's env var is only required at the point a caller
+ * actually reads that index. This lets video call sites destructure
+ * `videosIndex`/`videoVariantsIndex` without ever requiring
+ * ALGOLIA_INDEX_LANGUAGES to be set, and vice versa for language call sites -
+ * requiring one index cannot break callers of another.
  */
 export function getAlgoliaConfig(): AlgoliaConfig {
   return {
     appId: getRequiredEnv('ALGOLIA_APPLICATION_ID'),
     apiKey: getRequiredEnv('ALGOLIA_API_KEY'),
-    videosIndex: getRequiredEnv('ALGOLIA_INDEX_VIDEOS'),
-    videoVariantsIndex: getRequiredEnv('ALGOLIA_INDEX_VIDEO_VARIANTS')
+    get videosIndex(): string {
+      return getRequiredEnv('ALGOLIA_INDEX_VIDEOS')
+    },
+    get videoVariantsIndex(): string {
+      return getRequiredEnv('ALGOLIA_INDEX_VIDEO_VARIANTS')
+    },
+    get languagesIndex(): string {
+      return getRequiredEnv('ALGOLIA_INDEX_LANGUAGES')
+    }
   }
 }
 
--- a/apis/api-media/src/lib/languages/updateLanguageInAlgolia.ts
+++ b/apis/api-media/src/lib/languages/updateLanguageInAlgolia.ts
@@ -1,4 +1,3 @@
-import { type SearchClient, algoliasearch } from 'algoliasearch'
 import type { Logger } from 'pino'
 
 import {
@@ -6,29 +5,7 @@ import {
   prisma as languagesPrisma
 } from '@core/prisma/languages/client'
 
-function getRequiredEnv(name: string): string {
-  const value = process.env[name]
-  if (value == null || value.trim() === '') {
-    throw new Error(`Missing required env var: ${name}`)
-  }
-  return value
-}
-
-interface AlgoliaLanguagesClient {
-  client: SearchClient
-  languagesIndex: string
-}
-
-// Deliberately local rather than reusing lib/algolia/algoliaClient, whose
-// getAlgoliaConfig also requires the video index env vars. Coupling the
-// languages sync to those would break video indexing wherever they are unset.
-function getAlgoliaLanguagesClient(): AlgoliaLanguagesClient {
-  const appId = getRequiredEnv('ALGOLIA_APPLICATION_ID')
-  const apiKey = getRequiredEnv('ALGOLIA_API_KEY')
-  const languagesIndex = getRequiredEnv('ALGOLIA_INDEX_LANGUAGES')
-
-  return { client: algoliasearch(appId, apiKey), languagesIndex }
-}
+import { getAlgoliaClient, getAlgoliaConfig } from '../algolia/algoliaClient'
 
 const languageAlgoliaSelect = {
   id: true,
@@ -122,7 +99,8 @@ export async function updateLanguageInAlgoliaFromMedia(
   logger?: Logger
 ): Promise<void> {
   try {
-    const { client, languagesIndex } = getAlgoliaLanguagesClient()
+    const client = getAlgoliaClient()
+    const { languagesIndex } = getAlgoliaConfig()
     const language = await languagesPrisma.language.findUnique({
       where: { id: languageId },
       select: languageAlgoliaSelect
@@ -160,7 +138,8 @@ interface ReindexLanguagesResult {
 export async function reindexLanguagesWithVideosInAlgolia(
   logger?: Logger
 ): Promise<ReindexLanguagesResult> {
-  const { client, languagesIndex } = getAlgoliaLanguagesClient()
+  const client = getAlgoliaClient()
+  const { languagesIndex } = getAlgoliaConfig()
 
   let count = 0
   let cursor: string | undefined


mikeallisonJS (commented):
Reviewed against #9434. Good dedupe — the lazy-getter approach solves the real problem (#9343) without coupling the language sync to the video index env vars. Two things I'd fix before merge, one of them a silent behaviour change:

**1. Whitespace validation regressed on the languages path.** The deleted local helper rejected whitespace:

```ts
if (value == null || value.trim() === '') {
```

The surviving one in `algoliaClient.ts:12` checks only `value === ''`. After this merge, `ALGOLIA_INDEX_LANGUAGES=" "` is accepted and the language sync writes to a whitespace-named index instead of failing fast. One-character fix, and nothing in #9434 asked for the looser check.

**2. The lazy contract is invisible to the type system.** `AlgoliaConfig` declares `videosIndex: string` etc. as plain properties, but reading one can throw. That means `{ ...getAlgoliaConfig() }`, `JSON.stringify(config)`, structured-clone, or any logger that enumerates the object evaluates **all three getters** — reintroducing exactly the failure #9434 says must not be possible ("requiring a languages env var cannot break video indexing, and vice versa"). No current call site does this (I checked `videoAlgolia.ts`, both `algoliaVideo*Update.ts`, and both `scripts/algolia-helpers/*`), so it's latent, not broken — but it's held safe by convention rather than by the type. The shape #9434 itself suggested, `getAlgoliaIndex('languages')`, has no such trapdoor. Failing that, the laziness is documented in the *function's* JSDoc, which readers of the type never see — worth moving a line onto `AlgoliaConfig`.

Two smaller notes:

- #9434's other ask — "the two-key arrangement should be intentional and documented, or unified" — is argued in the PR body but nothing landed in the repo: no comment beside `apiKey` in `algoliaClient.ts`, nothing in either `infrastructure/locals.tf`. Since this PR closes the issue, the reasoning disappears with it.
- Nothing asserts the acceptance criterion itself. `updateLanguageInAlgolia.spec.ts` happens to cover it incidentally (it sets no video index vars, so eager fields would fail it), but a test that explicitly unsets `ALGOLIA_INDEX_LANGUAGES` and still indexes a video would state the intent — and would catch a future refactor that spreads the config.
- Minor: both languages call sites now do `getAlgoliaClient()` + `getAlgoliaConfig()`, and `getAlgoliaClient` calls `getAlgoliaConfig` internally, so the config is built twice per call.

tanflem:
@mikeallisonJS — replying here rather than in-thread since this was a top-level review body, so there's nothing for GitHub to mark resolved.

**Point 1 (whitespace validation regressed) — fixed in b98ee9aac.**

You're right, and it was exactly the one-character difference you identified. `getRequiredEnv` in `apis/api-media/src/lib/algolia/algoliaClient.ts` now matches what the deleted local helper did:

```ts
if (value == null || value.trim() === '') {
  throw new Error(`Missing required environment variable: ${name}`)
}
```

The check lives in the shared reader rather than in a reinstated local helper, so it covers every Algolia env var — the video indexes and the credentials as well as `ALGOLIA_INDEX_LANGUAGES`. Before the dedupe only the languages path had the trim check, so this is strictly a widening, not a behaviour change for video callers: `ALGOLIA_APPLICATION_ID=" "` was previously accepted by `algoliaClient.ts` and is now rejected too.

Behaviour is otherwise identical to the deleted helper: values are returned verbatim, not trimmed, so a padded-but-non-blank value (`" my-index "`) still passes through unchanged as it always did on both paths. Only the *blank* case is rejected. The error message is the shared reader's existing wording (`Missing required environment variable: …`) rather than the local helper's (`Missing required env var: …`).

**Test:** new `apis/api-media/src/lib/algolia/algoliaClient.spec.ts` covers unset, empty-string and whitespace-only for all five env vars (forcing the lazy getters to evaluate), plus blank credentials through `getAlgoliaClient()`. Against the pre-fix code 7 of the 19 cases fail — the 5 whitespace-only ones and 2 blank-credential ones — and the unset/empty cases pass, which pins the regression to exactly the lost `.trim()`.

```
npx vitest run --root apis/api-media \
  --config $PWD/apis/api-media/vitest.config.mts --coverage=false \
  src/lib/algolia/algoliaClient.spec.ts

 ✓ src/lib/algolia/algoliaClient.spec.ts (19 tests)
```

The surrounding suites (`src/lib/algolia`, `src/lib/languages`, `src/schema/video/videoAlgolia`, 69 tests) still pass, and `nx affected --target=type-check` is clean.

**Your other points are not addressed here** — this change was deliberately scoped to the regression only, so the rest are still open for you to call:

2. Lazy contract invisible in `AlgoliaConfig` — agreed it's latent, not broken; would need either the `getAlgoliaIndex('languages')` shape or at minimum a note on the type.
3. The two-key rationale living only in the PR body.
4. No test asserting the acceptance criterion directly (video indexing with `ALGOLIA_INDEX_LANGUAGES` unset).
5. `getAlgoliaClient()` + `getAlgoliaConfig()` building the config twice per call at the languages call sites.

Happy to take any of those in this PR or a follow-up — say which.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 25 checks green, no unresolved review threads, closes #9434. Small refactor: +111/-29 in 3 files, deduping the Algolia client factories.

### What needs to happen next

Ready for review — no action needed from you.

PR #1793, opened 2023-08-07
fix: add hover only on devices which support hover

storyworks:
# Description

### <samp>🤖 Generated by Copilot at feb5fbb</samp>

This pull request improves the button styles for light and dark themes, by disabling hover effects on touch devices using media queries. The changes affect the `colors.ts` and `components.ts` files in the `shared/ui` library.

https://3.basecamp.com/3105655/buckets/33281894/todos/6321144545

# How should this PR be QA Tested?

Please describe the QA tests that you ran to verify your changes. Please also note any relevant details for your test configuration.

- [ ] Test A
- [ ] Test B

# Walkthrough

### <samp>🤖 Generated by Copilot at feb5fbb</samp>

*  Add media queries to button style overrides to disable hover effect on touch devices ([link](https://github.com/JesusFilm/core/pull/1793/files?diff=unified&w=0#diff-842c7ab4ea57dac211490b47a02ce51683bd3f18c4b45b25a94a04c73a94c6a8L57-R60), [link](https://github.com/JesusFilm/core/pull/1793/files?diff=unified&w=0#diff-842c7ab4ea57dac211490b47a02ce51683bd3f18c4b45b25a94a04c73a94c6a8L75-R81), [link](https://github.com/JesusFilm/core/pull/1793/files?diff=unified&w=0#diff-842c7ab4ea57dac211490b47a02ce51683bd3f18c4b45b25a94a04c73a94c6a8L137-R144), [link](https://github.com/JesusFilm/core/pull/1793/files?diff=unified&w=0#diff-842c7ab4ea57dac211490b47a02ce51683bd3f18c4b45b25a94a04c73a94c6a8L155-R165), [link](https://github.com/JesusFilm/core/pull/1793/files?diff=unified&w=0#diff-7020025fb350dd12ddf8d1cc648fdcbe5e78bcbc325b9491cadc79c56898018eL120-R123))

PR #9542, opened 2026-08-24
fix(journeys): guard hover on touch devices for polls, multiselect and nav (NES-1894)

csiyang:
Fixes poll options appearing pre-selected on touch devices. Reported by Riah on a marketing journey going live around October — the first time that department has used NextSteps as a marketing piece, pushed via ads to a large audience.

## The bug

Touch browsers keep `:hover` on whatever the finger last touched and synthesise mouse events at those coordinates. A card advance is a client-side transition, so whatever ends up under that point paints as hovered — and on a poll option that reads as "already selected".

Nothing is actually selected: `selectedId` starts `null` and clears when the block goes inactive, so no answer is recorded and the card doesn't advance. Reproduced on a phone. Not a regression — these styles date to #7262 / #7286 (Jul–Aug 2025).

## Why it happened

[#1793](https://github.com/JesusFilm/core/pull/1793) guarded button hover behind `@media (hover: hover)` in the base theme in Aug 2023. Poll options were written two years later, define their own `&:hover` at higher specificity, and never got the guard — re-opening exactly the bug #1793 closed.

## The fix

A `hoverOnly()` helper in `libs/shared/ui`, so the guard is the path of least resistance rather than something each new component has to remember:

```ts
hoverOnly(styles) // → { '@media (hover: hover)': { '&:hover': styles } }
```

Routed through it: poll options (list **and** image/grid variants — separate files, each with its own hover block), multiselect options, and the Conductor prev/next chevrons.

`getPollOptionBorderStyles` now returns base borders only, with state colours moved to `getPollOptionBorderColors` so callers apply selectors themselves. That leaves one `@media (hover: hover)` block per style object — otherwise the second spread would silently clobber the first and drop the hover border.

## Multiselect was the worst case

Its hover background (`0.8` alpha) is a shade off its genuine selected background (`0.9`), and unlike a poll, tapping doesn't advance the card — so nothing cleared the phantom highlight; it persisted through the whole selection. The checkbox icon stayed honest throughout.

## Scope

Limited to the card-navigation path. Deliberately excluded: footer buttons, share icons, AI chat (out of scope until release), and gallery/admin surfaces. The card `Button` was checked and is fine — its own `&:hover` only applies during editor inline-editing, so the viewer uses the theme's guarded rule.

Poll options' missing `&.selected` rule is left alone — that was removed intentionally.

## Known limitation

`@media (hover: hover)` resolves **true** on hybrid devices (iPad with a trackpad, touchscreen laptops), so a tap there can still leave a stale highlight. Accepted as rare for a mobile ad audience; the alternatives (a JS touch-neutraliser, or redesigning hover so it can't resemble selection) cost more than the residual risk.

## Testing

- Typecheck clean on `libs/journeys/ui` and `apps/journeys`.
- 97 tests pass — 35 across the poll/multiselect components, 62 in Conductor including NavigationButton.
- Verification is a manual check on a real phone: Playwright emulation would verify the guard, not the platform sticky-hover behaviour it guards against.

> [!WARNING]
> ESLint could not be run locally — the worktree shares a `node_modules` predating this branch's ESLint 10 upgrade, so `eslint-plugin-import-x` is missing. Relying on CI for lint.

Closes NES-1894

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented hover highlights from lingering after taps on touch devices.
  - Improved radio and multiselect option states across pointer and touch interactions.
  - Refined border colors for default, hover, active, and disabled states in light and dark themes.

- **New Features**
  - Added consistent hover behavior for navigation buttons and selectable options.
  - Improved visual feedback for radio question choices and add-option controls.

- **Tests**
  - Updated coverage for option border styles and state-specific colors.

### DIFF
--- a/apps/journeys/src/components/Conductor/NavigationButton/NavigationButton.tsx
+++ b/apps/journeys/src/components/Conductor/NavigationButton/NavigationButton.tsx
@@ -10,6 +10,7 @@ import type { TreeBlock } from '@core/journeys/ui/block'
 import { useBlocks } from '@core/journeys/ui/block'
 import { useJourney } from '@core/journeys/ui/JourneyProvider'
 import { useStepNavigationEvents } from '@core/journeys/ui/useStepNavigationEvents'
+import { hoverOnly } from '@core/shared/ui/hoverOnly'
 import ChevronLeftIcon from '@core/shared/ui/icons/ChevronLeft'
 import ChevronRightIcon from '@core/shared/ui/icons/ChevronRight'
 
@@ -127,17 +128,18 @@ export function NavigationButton({
           size="small"
           onClick={() => handleNavigation(variant)}
           disableRipple
-          sx={{
+          sx={(theme) => ({
             pointerEvents: 'all',
             mx: { xs: 2, lg: 8 },
             p: 2,
-            color: (theme) => theme.palette.common.white,
-            backgroundColor: (theme) => `${theme.palette.grey[700]}33`,
-            '&:hover': {
-              color: (theme) => theme.palette.common.white,
-              backgroundColor: (theme) => `${theme.palette.grey[700]}4d`
-            }
-          }}
+            color: theme.palette.common.white,
+            backgroundColor: `${theme.palette.grey[700]}33`,
+            // Pointer devices only, so the chevron does not stay lit after a tap
+            ...hoverOnly({
+              color: theme.palette.common.white,
+              backgroundColor: `${theme.palette.grey[700]}4d`
+            })
+          })}
         >
           {alignment === 'left' ? <ChevronLeftIcon /> : <ChevronRightIcon />}
         </IconButton>
--- a/libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
+++ b/libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
@@ -15,36 +15,23 @@ import {
   getPollOptionBorderStyles
 } from '../utils/getPollOptionBorderStyles'
 
-const StyledListRadioQuestion = styled(Box)<BoxProps>(({ theme }) => {
-  const borderColors = getPollOptionBorderColors(theme)
-  const optionBorderStyles = {
-    ...getPollOptionBorderStyles(theme),
-    ...hoverOnly({ borderColor: borderColors.hover }),
-    '&:active': { borderColor: borderColors.active },
-    '&.disabled': { borderColor: borderColors.disabled }
-  }
-
-  return {
-    marginBottom: theme.spacing(4),
-    '& .MuiButtonGroup-root': {
-      boxShadow: 'none',
-      gap: theme.spacing(2),
-      '& .MuiButtonGroup-grouped': {
-        border: 'none',
-        borderBottom: 'none',
-        borderRight: 'none',
-        borderRadius: '12px',
-        margin: '0 !important',
-        '&:not(:last-of-type)': {
-          borderBottom: 'none'
-        },
-        '& .MuiButtonGroup-firstButton': optionBorderStyles,
-        '& .MuiButtonGroup-middleButton': optionBorderStyles,
-        '& .MuiButtonGroup-lastButton': optionBorderStyles
+const StyledListRadioQuestion = styled(Box)<BoxProps>(({ theme }) => ({
+  marginBottom: theme.spacing(4),
+  '& .MuiButtonGroup-root': {
+    boxShadow: 'none',
+    gap: theme.spacing(2),
+    '& .MuiButtonGroup-grouped': {
+      border: 'none',
+      borderBottom: 'none',
+      borderRight: 'none',
+      borderRadius: '12px',
+      margin: '0 !important',
+      '&:not(:last-of-type)': {
+        borderBottom: 'none'
       }
     }
   }
-})
+}))
 
 const adminPrimaryColor = adminTheme.palette
   .primary as SimplePaletteColorOptions
@@ -88,8 +75,7 @@ export function ListVariant({
                   borderBottomRightRadius: 8,
                   ...getPollOptionBorderStyles(theme, { important: true }),
                   ...hoverOnly({ borderColor: borderColors.hover }),
-                  '&:active': { borderColor: borderColors.active },
-                  '&.disabled': { borderColor: borderColors.disabled }
+                  '&:active': { borderColor: borderColors.active }
                 }
               }}
             >
--- a/libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
+++ b/libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
@@ -5,38 +5,46 @@ import Typography from '@mui/material/Typography'
 import { useTranslation } from 'next-i18next/pages'
 import { ReactElement } from 'react'
 
+import { hoverOnly } from '@core/shared/ui/hoverOnly'
 import AddSquare4Icon from '@core/shared/ui/icons/AddSquare4'
 import { adminTheme } from '@core/shared/ui/themes/journeysAdmin/theme'
 
 import { StyledListRadioOption } from '../../RadioOption/ListVariant'
-import { getPollOptionBorderStyles } from '../utils/getPollOptionBorderStyles'
+import {
+  getPollOptionBorderColors,
+  getPollOptionBorderStyles
+} from '../utils/getPollOptionBorderStyles'
 
-const StyledListRadioQuestion = styled(Box)<BoxProps>(({ theme }) => ({
-  marginBottom: theme.spacing(4),
-  '& .MuiButtonGroup-root': {
-    boxShadow: 'none',
-    gap: theme.spacing(2),
-    '& .MuiButtonGroup-grouped': {
-      border: 'none',
-      borderBottom: 'none',
-      borderRight: 'none',
-      borderRadius: '12px',
-      margin: '0 !important',
-      '&:not(:last-of-type)': {
-        borderBottom: 'none'
-      },
-      '& .MuiButtonGroup-firstButton': {
-        ...getPollOptionBorderStyles(theme)
-      },
-      '& .MuiButtonGroup-middleButton': {
-        ...getPollOptionBorderStyles(theme)
-      },
-      '& .MuiButtonGroup-lastButton': {
-        ...getPollOptionBorderStyles(theme)
+const StyledListRadioQuestion = styled(Box)<BoxProps>(({ theme }) => {
+  const borderColors = getPollOptionBorderColors(theme)
+  const optionBorderStyles = {
+    ...getPollOptionBorderStyles(theme),
+    ...hoverOnly({ borderColor: borderColors.hover }),
+    '&:active': { borderColor: borderColors.active },
+    '&.disabled': { borderColor: borderColors.disabled }
+  }
+
+  return {
+    marginBottom: theme.spacing(4),
+    '& .MuiButtonGroup-root': {
+      boxShadow: 'none',
+      gap: theme.spacing(2),
+      '& .MuiButtonGroup-grouped': {
+        border: 'none',
+        borderBottom: 'none',
+        borderRight: 'none',
+        borderRadius: '12px',
+        margin: '0 !important',
+        '&:not(:last-of-type)': {
+          borderBottom: 'none'
+        },
+        '& .MuiButtonGroup-firstButton': optionBorderStyles,
+        '& .MuiButtonGroup-middleButton': optionBorderStyles,
+        '& .MuiButtonGroup-lastButton': optionBorderStyles
       }
     }
   }
-}))
+})
 
 const adminPrimaryColor = adminTheme.palette
   .primary as SimplePaletteColorOptions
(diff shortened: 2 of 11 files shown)


apps/journeys/src/components/Conductor/NavigationButton/NavigationButton.tsx
```diff
@@ -127,17 +128,18 @@ export function NavigationButton({
...
-            }
-          }}
+            color: theme.palette.common.white,
+            backgroundColor: `${theme.palette.grey[700]}33`,
+            // Pointer devices only, so the chevron does not stay lit after a tap
+            ...hoverOnly({
+              color: theme.palette.common.white,
+              backgroundColor: `${theme.palette.grey[700]}4d`
+            })
+          })}
```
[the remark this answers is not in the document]
csiyang: Not fixing this here — it's out of scope for this PR.

The finding itself is accurate: this `IconButton` has no `aria-label`, and in fact no `aria-*` attribute at all. But it's pre-existing rather than introduced here — this PR only changed that button's `sx` prop, to route its hover styles through the new `hoverOnly()` helper. No accessibility attribute was added or removed by this diff.

This PR is scoped to the stale-hover-on-touch-devices bug (NES-1894), so we're leaving the label alone to keep the change reviewable against that one concern.

libs/shared/ui/src/libs/hoverOnly/hoverOnly.ts
```diff
@@ -0,0 +1,30 @@
...
+ * client-side transition, whatever ends up under that point paints as hovered —
+ * on a poll option that reads as "already selected". Guarding hover this way is
+ * how the base theme has handled buttons since #1793; components that override
+ * the theme with their own `&:hover` must use this helper or they reintroduce
+ * the bug.
+ *
+ * Hybrid devices (an iPad with a trackpad, a touchscreen laptop) report
+ * `hover: hover`, so a tap there can still leave a stale highlight. Accepted.
+ */
+export function hoverOnly(styles: CSSObject): HoverOnlyStyles {
```
jianwei1, line 33: **Concern:** the return shape makes this helper harder to use than it needs to be.

`hoverOnly()` returns an object whose only key is `'@media (hover: hover)'`. Two spreads into the same style object therefore clobber each other silently — no type error, no lint error, the second one just wins and the first block's styles vanish.

That footgun is what forced the `getPollOptionBorderStyles` → `getPollOptionBorderColors` split in this PR: the util had to stop returning selector keys so there'd only ever be one media block per style object. Most of the diff size, plus the test rewrite, is working around this function's API rather than fixing the bug.

Worth considering the inverted nesting:

```ts
export function hoverOnly(styles: CSSObject) {
  return { '&:hover': { '@media (hover: hover)': styles } }
}
```

Emotion supports media queries nested inside a selector, so the emitted CSS is identical. The difference is the collision key: `&:hover` is a key developers already treat as unique in a style object, so a duplicate is visible at a glance instead of silently eating a media block. I believe that would have let `getPollOptionBorderStyles` keep its selector keys and stay untouched — worth sanity-checking against the preview build before acting on it.

Not blocking, but this helper is explicitly meant to be reused by future components, so its shape is worth settling now rather than after four more call sites exist.

Separately: the clobbering hazard is documented in `getPollOptionBorderColors`'s docstring, but not here. This docstring is where the next caller will look — worth a line either way, whichever shape you land on.

---

libs/shared/ui/src/libs/hoverOnly/hoverOnly.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { hoverOnly } from '.'
+
+/**
+ * hoverOnly guards hover styles behind `@media (hover: hover)` so that sticky
+ * hover on touch devices - which paints the last-touched element as hovered
+ * after a card transition - cannot make poll and multiselect options read as
+ * already-selected (NES-1894).
+ *
+ * These tests pin the wrapper shape: if the media query is ever simplified
+ * away, the bug returns silently.
```
jianwei1: **Concern:** this comment claims more protection than the tests actually give.

These specs pin `hoverOnly`'s return shape, which guards the helper. They don't touch any of the four places the fix actually lives — `StyledListRadioOption`, `StyledGridRadioOption`, `StyledListMultiselectOption`, `NavigationButton`. Both realistic regressions pass every test in this PR:

- a component going back to a bare `&:hover` (exactly how the bug got reintroduced after #1793), and
- a second `@media (hover: hover)` spread clobbering the guarded block — the failure mode the `getPollOptionBorderStyles` split exists to prevent.

So "if the media query is ever simplified away, the bug returns silently" is true of the helper, but the silent return is at the call sites, and nothing here would catch it.

I'm not asking for a new dependency: there's no `@emotion/jest` in the repo and no precedent anywhere in `libs/` or `apps/` for asserting emitted CSS, so closing this properly is a bigger call than this PR should make. Suggest just narrowing the comment to what the tests cover, so the next person doesn't read it as call-site coverage and skip adding a guard.

---

libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
```diff
@@ -5,38 +5,46 @@ import Typography from '@mui/material/Typography'
...
-      },
-      '& .MuiButtonGroup-lastButton': {
-        ...getPollOptionBorderStyles(theme)
+const StyledListRadioQuestion = styled(Box)<BoxProps>(({ theme }) => {
+  const borderColors = getPollOptionBorderColors(theme)
+  const optionBorderStyles = {
+    ...getPollOptionBorderStyles(theme),
+    ...hoverOnly({ borderColor: borderColors.hover }),
+    '&:active': { borderColor: borderColors.active },
+    '&.disabled': { borderColor: borderColors.disabled }
```
jianwei1: **Nit:** `&.disabled` is now dead in two different ways.

This keeps `'&.disabled'`, while `RadioOption/ListVariant.tsx` dropped the one it used to pick up from the old full-object spread and now styles only `'&.Mui-disabled'`.

Neither component ever applies a bare `.disabled` class — both set `className={selected ? 'selected' : ''}`, and MUI emits `Mui-disabled` — so this selector never matched before the PR either. The change isn't a regression; it just leaves the same dead rule present in one file and absent in the other, which reads like an intentional distinction when it isn't.

Cheapest fix is dropping it here too (and the matching `borderColors.disabled` lookup), so `Mui-disabled` is the only disabled selector in the poll option code.

---

libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
```diff
@@ -5,38 +5,46 @@ import Typography from '@mui/material/Typography'
...
+        borderBottom: 'none',
+        borderRight: 'none',
+        borderRadius: '12px',
+        margin: '0 !important',
+        '&:not(:last-of-type)': {
+          borderBottom: 'none'
+        },
+        '& .MuiButtonGroup-firstButton': optionBorderStyles,
+        '& .MuiButtonGroup-middleButton': optionBorderStyles,
+        '& .MuiButtonGroup-lastButton': optionBorderStyles
```
jianwei1: **Nit:** I think this whole block is unreachable, which would mean the refactor above is dressing up dead CSS.

These are nested inside `'& .MuiButtonGroup-grouped'`, so the emitted selector is a *descendant* one — roughly `.MuiButtonGroup-grouped .MuiButtonGroup-firstButton`. MUI puts `MuiButtonGroup-grouped` and `MuiButtonGroup-firstButton` on the same node, not one inside the other, so nothing should ever match.

If that's right, the actual poll option borders come entirely from `StyledListRadioOption`, and the new `optionBorderStyles` const, the `hoverOnly()` call, and the two extra state selectors here have no effect either way. Pre-existing, not introduced here — but the PR is adding structure to it, so it's a good moment to check.

Quick way to confirm: inspect a list poll option on the preview deploy and see whether any rule from this block appears in the computed styles. If it doesn't, deleting the `firstButton`/`middleButton`/`lastButton` entries is simpler than maintaining them, and it takes the `getPollOptionBorderColors` call in this file with it.

---

jianwei1 (commented):
## Review summary

**Critical:** none — nothing here blocks merge.
**Concerns:** 2 (see inline)
**Nits:** 2 (see inline)

The diagnosis is right and the link back to #1793 is exactly the point: this is a re-opened bug, not a new one, and extracting a shared helper is the correct structural answer rather than a fourth one-off guard. CI is green across all 32 checks.

I specifically checked the thing a refactor like this most easily breaks — the `hoverOnly()` spread stays *before* `'&.selected'` in both `MultiselectOption` and `GridVariant`, so with equal specificity the source order still lets selected win over hover. That holds. Also nice: the old util's inconsistent `!important` spacing (`' !important'` for default/hover vs `'!important'` for active/disabled) is quietly normalised, and the remaining unguarded `&:hover` blocks in the `editableLabel` branches are correctly left alone — those are admin inline-editing, not the viewer path. The scope section matches what the code actually does, and the hybrid-device limitation is stated honestly rather than papered over.

The one item I'd genuinely like a reply on is the first inline comment. `hoverOnly()` returning a top-level `@media` key means two spreads silently clobber, and that constraint is what forced the `getPollOptionBorderStyles` split — so most of this diff is working around the helper's shape rather than the bug. Since the helper is explicitly meant to be the path of least resistance for future components, it's worth settling before more call sites depend on it. Not a blocker either way.

Skipping the ESLint warning in the description — `lint` and `lint-work` both passed in CI, so that gap closed itself.

The `aria-label` thread is already resolved and I agree with your reasoning there; not reopening it.

---

apps/journeys/src/components/Conductor/NavigationButton/NavigationButton.tsx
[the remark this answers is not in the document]
csiyang: No follow-up issue, thanks — we're not tracking this one separately for now.

libs/shared/ui/src/libs/hoverOnly/hoverOnly.ts
csiyang: Keeping the current shape, but taking your second point — both hazards are now documented in `hoverOnly`'s own docstring rather than only in `getPollOptionBorderColors`'.

On the main claim, I don't think inverting would have spared `getPollOptionBorderStyles`. The old util already returned a top-level `'&:hover'`, and `RadioOption/ListVariant` already wrote its own `'&:hover'` alongside it — those two already collided, which is exactly why the pre-existing code had to re-spread `borderStyles['&:hover']` inside its own block. Inverting moves the collision from the `@media` key onto the `&:hover` key; it doesn't remove it, so the split would still have been needed.

That said, I think there's a stronger argument for your proposal than the one you made. With the current shape, a `hoverOnly()` spread and a hand-written `&:hover` **don't** collide — different keys, both apply, and the hand-written one stays unguarded. That is precisely how the bug came back after #1793, and it's silent. Inverted, they'd collide and styles would visibly vanish.

What tipped it for me is the other direction: with `@media` on the outside, everything passed to `hoverOnly()` is guaranteed guarded. Inverted, `'&:hover': { '@media (hover: hover)': {...}, color: 'red' }` mixes guarded and unguarded in one block — a new way to half-fix it. So I'd rather make the hazard loud in the docs than trade it for a different one.

Happy to revisit if a second call site makes the collision bite in practice.

libs/shared/ui/src/libs/hoverOnly/hoverOnly.spec.ts
csiyang: Agreed — that was overclaiming, and it was my wording. Narrowed in 6d801ab8a.

The docstring now says explicitly that these tests pin the helper's return shape only, that they won't catch a component reverting to a bare `&:hover` or a guarded block being clobbered by a second spread, and that catching either needs an assertion on emitted CSS which this repo has no precedent for. That way the next person reads it as helper coverage rather than call-site coverage.

libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
csiyang: Verified and dropped in 6d801ab8a.

Confirmed nothing applies a bare `disabled` class — `RadioOption/ListVariant` and `MultiselectOption` both set `className={selected ? 'selected' : ''}`, `RadioOption/GridVariant` sets `Mui-disabled` explicitly, and MUI emits `Mui-disabled` for the rest. So the selector never matched, before or after.

Removed it and the `borderColors.disabled` lookup from the Add Option button. `Mui-disabled` is now the only disabled selector in the poll option code.

libs/journeys/ui/src/components/RadioQuestion/ListVariant/ListVariant.tsx
csiyang: You're right, and this was the most useful catch in the review. Deleted in 6d801ab8a.

Confirmed from MUI's `ButtonGroup` source rather than the preview: `classes.grouped` is passed to each child via `ButtonGroupContext`, and `getButtonPositionClassName(index)` — which returns `classes.firstButton` / `middleButton` / `lastButton` — is passed to the *same* child via `ButtonGroupButtonContext`. Both classes land on one node. MUI's own styles target them as `& .firstButton` from the root, never nested under `.grouped`. Our nesting emitted a descendant selector, so it could never match.

Deleting the three entries reverts that `styled()` to its exact pre-PR form, so the only remaining change in this file is the Add Option button. It also removes the `getPollOptionBorderColors` call from the `styled()` block.

Worth flagging what this means for the PR: I'd presented those selectors as live behaviour that needed preserving, and the "preserve exactly" decision was made on that basis. It was preserving nothing. Poll option borders come from `StyledListRadioOption`, as they already did.

jianwei1 (approved):
## Approved

All four review items are addressed — three by change, one by an argument that beat mine. CI is green across all 31 checks.

**On the `hoverOnly()` shape:** you're right and I was wrong. The old util already returned a top-level `'&:hover'` and `RadioOption/ListVariant` already wrote its own alongside it — the pre-existing `...borderStyles['&:hover']` re-spread only exists because those two already collided. Inverting relocates that collision rather than removing it, so the split would still have been needed. You also made the better version of my own argument before deciding against it: under the current shape a `hoverOnly()` spread and a hand-written `&:hover` *don't* collide, so the hand-written one stays live and unguarded — which is exactly how this regressed after #1793. Trading that for the mixed guarded/unguarded hazard is a fair call, and both are now documented where the next caller will actually look. Good as-is.

**Verified on the rest:**

- Spec docstring now scopes its claim to the helper's return shape, with the call-site gap stated plainly.
- `&.disabled` dropped, and `borderColors.disabled` is still consumed by `'&.Mui-disabled'` in `RadioOption/ListVariant` — nothing stranded.
- The `MuiButtonGroup` deletion checks out. `StyledListRadioQuestion` is now byte-identical to its pre-PR form on `main`, and the removal is behaviour-neutral on independent grounds: that block applied the border properties without `!important` while `StyledListRadioOption` applies the same three with it, so it could never have won even had the selector matched.

Also worth saying: you flagged unprompted that you'd presented those selectors as live behaviour needing preservation, and that the "preserve exactly" decision rested on it. Volunteering that a stated rationale was wrong is the useful thing to do, and it's why this PR ended up smaller than it started.

The diagnosis, the link back to #1793, and choosing a shared helper over a fourth one-off guard are all right. Nice fix.

---

PR #9544, opened 2026-08-25
chore(deps): upgrade Apollo Client 3→4 (#9376)

mikeallisonJS:
Delivers the **`@apollo/client` half of the GraphQL core family** from the upgrade roadmap in #9376. Twelfth roadmap family to land, after ESLint (#9514), MUI (#9445), Vitest (#9484), Nx (#9415) and the rest.

## `graphql` 16 → 17 is blocked, not skipped

The roadmap states the family as *"`graphql` (16 → 17) and `@apollo/client` (3 → 4, major client rewrite)"*. Only the client half is possible today — the federation stack caps `graphql` at 16:

| Package | Latest | `graphql` peer |
| --- | --- | --- |
| `@apollo/subgraph` | 2.14.4 | `^16.5.0` |
| `@apollo/server` | 5.5.1 | `^16.11.0` |
| `@graphql-hive/gateway` | 2.11.2 | `^15.9.0 \|\| ^16.9.0` |
| `@apollo/client` | 4.2.12 | `^16.0.0 \|\| ^17.0.0` ← the only one ready |

So `graphql` stays on 16 and the 16→17 bump waits on those three moving. Worth folding back into the issue.

One roadmap line resolves as a non-finding: *"chase the stray transitive `graphql@4.8.0` resolution surfaced in the lockfile"*. The only `4.8.0` in the lockfile is `@octokit/graphql@4.8.0` — an unrelated package, not a stray `graphql` resolution.

## What moved

| Package | From | To |
| --- | --- | --- |
| `@apollo/client` | 3.13.6 | 4.2.12 |
| `@apollo/client-integration-nextjs` | 0.12 | 0.14 (peers `@apollo/client@^4`) |
| `rxjs` | — | 7.8.2 (new required peer; v4 replaced zen-observable) |

Removed: `@adobe/apollo-link-mutation-queue`, `apollo-link-debounce`, `zen-observable-ts` — see below.

## Codemod, then the parts it can't do

`@apollo/client-codemod-migrate-3-to-4` handled the mechanical sweep across ~1,100 files: entry points (`@apollo/client/react`, `@apollo/client/testing/react`, `@apollo/client/link/*`), types moved into namespaces (`ApolloLink.Result`, `useQuery.Options`, `MockLink.MockedResponse`), link creators → classes, and the `ApolloClient` constructor reshuffle. Apollo's own docs say it gets you ~90% of the way; the remainder was diagnosed case by case.

### Error handling — `ApolloError` is gone

GraphQL errors now arrive as `CombinedGraphQLErrors`; everything else is passed through unwrapped. The interesting part was `error.networkError`, which had no obvious replacement. `LinkError.is()` is the exact one: Apollo registers errors in `registerLinkError` from a `catchError` on the raw link observable, *before* `CombinedGraphQLErrors` is constructed downstream — so it means precisely "AC3 would have set `networkError`". Verified against the v4 source rather than inferred, which is why the snackbar branches keep their original meaning instead of collapsing into "any error".

### Removed hook options

`useQuery` / `useLazyQuery` lost `onCompleted` and `onError` outright. Each site now reads the result instead. Two shared hooks keep their public contract by mirroring the callbacks internally (`useJourneyAnalyticsQuery`, `useJourneyAiTranslateSubscription`) so their callers were untouched.

`useLazyQuery` also lost hook-level `variables`, and executing one outside an event handler now throws. Lazy queries that were fired from an effect (AccessDialog, LocalDetails) became `useQuery` + `skipToken` — which is what Apollo's guide recommends, and drops a render-cycle round trip.

### Three behaviour changes that only surface at runtime

Found by running the suites, not by reading the diff:

- **Subscriptions deliver errors through `next`, then complete.** `onComplete` now fires after a failure, where v3 terminated. Callers read `onComplete` as "translation succeeded" and were switching teams on error. Gated once in the shared subscription hook rather than at six call sites.
- **Mutate promises reject even when `onError` is supplied.** v3 resolved. Three `await mutate({ onError })` call sites were leaking unhandled rejections.
- **`MockLink` gained a random 20–50 ms delay.** The suite was written against v3's immediate responses; pinned back to `{ delay: 0 }` in the ten test setups, as the migration guide suggests.

## Two links had to be vendored

Both are incompatible with v4 in ways no version bump fixes, so they are ported in-repo under `apps/journeys-admin/src/libs/apolloClient/` with behaviour unchanged:

- **`@adobe/apollo-link-mutation-queue`** returns `zen-observable` instances. v4's `QueryManager` calls `.pipe()` on link results, which zen-observable does not have.
- **`apollo-link-debounce`** ships an ES5 build that subclasses `ApolloLink` via a `_super.call(this)` shim. v4's `ApolloLink` is a real class, so it throws *"Class constructor ApolloLink cannot be invoked without 'new'"* the moment the link is constructed.

Both are now covered by specs (there were none before — the debounce link was only exercised indirectly through editor specs). Each spec was checked against a deliberately broken implementation to confirm it fails.

`zen-observable-ts` went with them; nothing imported it once the Adobe link was gone.

## Typing fallout

- `ApolloCache<T>` lost its generic — dropped at 10 sites.
- `ApolloCache.extract()` now returns `unknown`; cast where the result is written into page props. In `useBlockDeleteCommand` this had been masking a real gap — the editor's step blocks carry no `x`/`y`, but `blockRestore`'s response type requires them, so the optimistic response now reads the coordinates back off the cache.
- v4 derives a result's data states from the options handed to a hook. Wrapper hooks that passed a wide `useQuery.Options` were leaving `returnPartialData` unresolved, which put `partial` in the union and handed every caller `DeepPartial` data. Their `TStates` are now narrowed, so consumers keep complete data.
- Server-side `client.query` types `data` as optional under the default error policy. Call sites with inferred generics state `errorPolicy: 'none'` (which is what they already relied on); the rest narrow explicitly.

Sharper variable typing in arclight surfaced two genuine latent bugs: a `select` variable the query never declared, and four `c.req.param()` values passed as required `ID!` while typed `string | undefined`.

## Verification

- **Type-check: 35/35 projects clean.**
- **Tests green** — journeys-admin 3,768, journeys-ui 1,474, api-journeys 1,114, api-media 779, resources 606, videos-admin 466, plus the rest. **Zero unhandled rejections**, which `nx test` fails on even when `vitest run` passes.
- `pnpm install --frozen-lockfile` verified.


csiyang (approved):
Scoped my review to the hand-written, high-risk parts — the ~1,100-file codemod sweep is verified by green CI (35/35 type-check, all suites, zero unhandled rejections), so I read the code that a codemod can't get right. All of it held up:

- **Vendored `MutationQueueLink`** is a faithful rxjs port — queue/skip/cancel semantics match the Adobe original — and both vendored links now have specs where upstream had none.
- **The subscription error-then-complete gate** is the standout fix: v4 fires `onComplete` after a failure and callers read it as "translation succeeded"; the `erroredRef` gate in `useJourneyAiTranslateSubscription` closes that once for six call sites, and the `skip`-when-no-variables pattern handles v4's required-variables typing correctly.
- **`useBlockDeleteCommand`**: the `extract()`-returns-`unknown` change forced reading `x`/`y` off the cache for *all* restored blocks, fixing a real pre-existing gap where only the step (not children/siblings) got coordinates in the optimistic response.
- **No leftovers**: the only remaining references to the removed packages are provenance comments in the vendored files and AGENTS.md — no surviving imports of `zen-observable`, `apollo-link-debounce`, or the Adobe link.

The `LinkError.is()` ≡ `networkError` equivalence I'm taking on the description's word (verified against v4 source per the PR), with the dependent snackbar branches covered by tests — the receipts elsewhere in this PR earn that trust.

The blocked `graphql` 16→17 half with peer-range evidence is worth folding back into #9376 as noted. For a change this size the residual risk is behavioral edge cases only production traffic will surface — worth watching client error rates after deploy, but no reason to hold this.

PR #9547, opened 2026-08-27
feat(api-media): add youtube Service value and ShortLink sourceRef

### LINKED ISSUE #9545: ShortLink: add a youtube Service value and a generic sourceRef column
## Why

The youtube-studio Wheat app (JesusFilm/youtube-studio) is adopting Core ShortLinks for the CTA links it stamps into YouTube video descriptions, replacing the third-party shorteners the channel used historically. Two things are missing for that class of links to be attributable and centrally manageable:

1. A `Service` value that identifies them. Today Wheat can only send `apiMedia`, which the `nxstp.is` domain row (registered for `apiJourneys` only) rejects, and which would misattribute the links anyway.
2. A record of which YouTube channel a link was created for, so that if a channel is compromised or a link misused, every link minted for that channel can be found and repointed or deleted in one query.

## Changes

- Add `youtube` to the `Service` enum in `libs/prisma/media/db/schema.prisma` (Prisma migration plus regenerated schema). Note this is the first value that names a product surface rather than an api-* service; that is deliberate, and worth a sentence in the migration or PR description. The redirect resolver does not branch on `service`, so this is provenance and enumeration only.
- Append `youtube` to the `services` array of the `nxstp.is` `ShortLinkDomain` row (data change, additive; api-journeys' own minting is untouched).
- Add a nullable `sourceRef` string column to `ShortLink`, exposed on the GraphQL type. Convention, documented in the field description so every future writer sees it: a single namespaced `<scheme>:<value>` string, one value per link. First registered scheme: `youtube-channel:<channelId>`. If a link ever needs two attributions, that is the moment for a real column, not a delimiter grammar inside the string.

Wheat will send `service: youtube` and always write `sourceRef` on its mints. The bad-actor query is then `service = youtube AND sourceRef = 'youtube-channel:UC...'`, with bulk repoint via the existing `shortLinkUpdate`.

## Consumer

Blocks the Wheat link picker work in JesusFilm/youtube-studio (see JesusFilm/youtube-studio#635 for the design revision this comes from).

mikeallisonJS:
Closes #9545

Blocks the Wheat link picker work in JesusFilm/youtube-studio (see JesusFilm/youtube-studio#635).

## What this does

The youtube-studio (Wheat) app is adopting Core ShortLinks for the CTA links it stamps into YouTube video descriptions. Three changes, matching the issue:

**`Service.youtube`** — the first enum value that names a **product surface rather than an api-\* service**. That is deliberate: Wheat mints links directly and needs to be attributable as itself instead of borrowing `apiMedia`, which the `nxstp.is` domain row rejected anyway and which would have misattributed the links. Nothing branches on `service` at redirect time, so this is provenance and enumeration only.

**`nxstp.is` grant** — `youtube` appended to the row's `services` array, additive; api-journeys' own minting is untouched.

**`ShortLink.sourceRef`** — nullable string, exposed on the GraphQL type and accepted by `shortLinkCreate`. A single namespaced `<scheme>:<value>` attribution, one value per link. First registered scheme: `youtube-channel:<channelId>`. The convention lives in the field description so every future writer sees it; the reasoning for keeping it one unvalidated string (a second attribution is the moment for a real column, not a delimiter grammar) sits in a code comment rather than in the published supergraph.

## Two things worth a reviewer's attention

**The motivating query has no API surface.** The issue's payoff is finding every link for a compromised channel "in one query", but `shortLinks` still filters only by `hostname` — so today that query is direct SQL against the new index, not something Wheat or an admin UI can issue. The Changes section didn't ask for a filter so I didn't add one, but the Why section reads like an API promise. Flagging it as a decision, not an oversight.

**`@@index([service, sourceRef])` is not in the three bullets.** I added it because it backs the enumeration query the column exists for. Easy to drop if you'd rather keep the diff to the letter of the issue.

Also note `sourceRef` is create-only — `shortLinkUpdate` cannot correct or backfill it. That matches the spec, which only mandates writing on mints, and is defensible if provenance is immutable, but it is a one-way door.

## Migrations

Two files, not one: Postgres refuses to use a new enum value in the transaction that added it, and Prisma wraps each migration in a transaction.

The data migration skips rows with an empty `services` array — empty already means "all services allowed", so a plain append there would have **narrowed** the domain to youtube-only and broken api-journeys minting. Verified against seeded rows rather than assumed: `{apiJourneys}` → `{apiJourneys,youtube}`, other domains untouched, re-running is a no-op. It is also a no-op in any environment with no `nxstp.is` row, so an environment seeded later will not pick up the grant.

## Testing

- api-media: **792/792 pass**, including a new test for a youtube mint carrying `sourceRef`, and one asserting a domain not registered for `youtube` is rejected by the create validator. The `sourceRef` test was written first and failed with `Cannot query field "sourceRef" on type "ShortLink"` before the Pothos wiring.
- Full migration history replays clean on an empty database.
- `pnpm lint:changed --fix` passes with no files changed.
- Pre-existing, **not** from this branch: type-check fails on `api-media`/`arclight`/`short-links`/etc. and `short-links` + `journeys-admin` tests fail — all identically on clean `main` (in-progress Apollo Client 4 migration). None of the 1005 pre-existing type errors touch a file this branch changes, and journeys-admin's failing-spec set on this branch is a strict subset of `main`'s.

## Summary by CodeRabbit

* **New Features**
  * Added YouTube as a supported short-link service.
  * Short links can now include an optional source attribution reference, including YouTube channel references.
  * Attribution details are preserved when creating and retrieving short links.

* **Bug Fixes**
  * Added validation to prevent YouTube short links from being created for unsupported domains.


csiyang (approved):
Reviewed the code at `3e46839` rather than running the suite locally; Nx Cloud reports all 24 CI tasks green.

Approving — the findings below are worth folding in, but none of them block.

## The `nxstp.is` grant lands only in prod

The data migration matches `WHERE "hostname" = 'nxstp.is'`, but that row is created by `apis/api-media/db/seeds/shortLinkDomain.ts:12` from `$JOURNEYS_SHORTLINK_DOMAIN` — a per-environment value (it appears only in `apis/api-media/infrastructure/locals.tf:28`, never in the repo). So:

- In any environment where that hostname differs, the `UPDATE` matches nothing and a `youtube` mint fails the create validator.
- The seed hardcodes `services: [Service.apiJourneys]`, so an environment seeded *after* this migration gets a row with no `youtube` grant. You flag this in the description; the fix is one line.

Not a blocker: nothing in Core consumes `Service.youtube` yet, and `shortLinkDomainUpdate` (`shortLinkDomain.ts:185`) can grant the service at runtime in any environment without a deploy. The failure mode is loud rather than silent — a clear ZodError, not a wrong result.

Two suggestions:

- Add `Service.youtube` to the seed's `services` array. Because the upsert uses `update: {}` it only affects rows that don't yet exist, so there's no risk to prod, and it closes the fresh-environment case permanently.
- Worth confirming whether stage's `JOURNEYS_SHORTLINK_DOMAIN` is `nxstp.is` — I couldn't determine that from the repo. If it isn't, the Wheat integration work will need the grant mutation run against stage first.

## `sourceRef` is readable without auth

`shortLinkByPath` (`shortLink.ts:184`) is the only ShortLink query without `.withAuth` — it's the public redirect resolver, so every field on the type is readable by anyone holding a short link. `youtube-channel:UC...` is public data, so this is fine today, but the field is deliberately unvalidated and open-ended. Consider adding "values must be publicly disclosable" to `SOURCE_REF_DESCRIPTION` so the next scheme registrant doesn't park a user id or internal record id there.

## The index's leading column doesn't match the likely query

`@@index([service, sourceRef])` — the scheme prefix already namespaces the value, so "every link for this channel" is naturally `WHERE "sourceRef" = 'youtube-channel:UC...'` with no `service` predicate, which can't use this index. And nothing couples the two: `shortLinkCreate` accepts `service: apiJourneys` alongside `sourceRef: 'youtube-channel:UC...'`, so a lookup that *does* filter on `service = 'youtube'` can silently miss rows. `@@index([sourceRef])` serves both shapes.

## Minor

- No `CONCURRENTLY` note on the `CREATE INDEX`. `20260324194149_.../migration.sql:2` sets the precedent of at least commenting it; if `ShortLink` is large this blocks writes for the deploy's duration.
- The new test at `shortLink.spec.ts:566` largely duplicates `:825` — both mock `shortLinkDomain.findFirst → null` and assert the same ZodError. Its only added coverage is the `hasEvery: ['youtube']` assertion on the mock call, which checks the shape of a query rather than a behavior. It doesn't demonstrate what the name claims (that an `apiJourneys`-only domain rejects a youtube mint), because the mock can't model the filter. Fine to keep — just noting it's thinner than it reads.

## On your two flags

- **No API surface for the motivating query** — agreed that's a decision, not an oversight. But the index exists specifically for that query and Wheat is the named consumer, so adding `sourceRef: t.arg.string()` to the `shortLinks` connection (`shortLink.ts:133`) is a handful of lines and closes the loop. I'd lean toward doing it now rather than shipping an index nothing can reach.
- **Create-only `sourceRef`** — defensible. Immutable provenance is a reasonable stance and matches the spec.

## What's good

The two-migration split for the enum is correct and correctly explained. The empty-`services` guard is a real trap avoided — a plain `array_append` there would have narrowed the domain and broken api-journeys minting. Generated files are regenerated consistently across all five consumers, and putting the convention in the field description rather than a regex is the right call for a registry that grows by documentation.


---

PR #9548, opened 2026-08-29
feat(api-journeys): add journeysByVideo publisher query

### LINKED ISSUE #9546: api-journeys: publisher query for journeys that use a given media video
## Why

The youtube-studio Wheat app is adding a link picker for the CTA link in YouTube video descriptions. One of the options is a NextSteps journey, and the natural journey to offer is one that already contains the video being published. Journey blocks reference `videoId` internally, but no gateway query answers "which journeys embed this video", so today the operator has to know and paste the journey URL by hand.

## Proposal

A publisher-gated query, roughly:

```graphql
journeysByVideo(videoId: ID!, languageId: ID): [Journey!]!
```

returning at least id, title, slug (or public URL), and status, resolved by looking through VideoBlocks. Auth `isPublisher`, the same gate as the rest of the journeys admin surface.

Semantic decisions that belong to the journeys owners, not the consumer:

- Match on the variant or the root video, and whether `languageId` narrows or is ignored.
- Whether unpublished and template journeys appear in the results.
- Whether team visibility scopes the results for the calling publisher.

## Consumer

JesusFilm/youtube-studio will add a "journeys using this video" list to the Wheat link picker once this exists. The picker ships first with a paste-a-URL flow, so this is an enhancement, not a blocker for them.

mikeallisonJS:
Closes #9546

## What

Adds a publisher-gated query answering "which journeys embed this media video":

```graphql
journeysByVideo(videoId: ID!, languageId: ID): [Journey!]!
```

The youtube-studio Wheat link picker will use it to offer a NextSteps journey as the CTA target for a video being published, instead of asking the operator to know and paste the journey URL.

## Semantic decisions

The issue explicitly left four calls to the journeys owners. Each is decided here and documented in the GraphQL field description, so the consumer reads them off the schema:

| Decision | Choice | Why |
| --- | --- | --- |
| Root video vs. variant | Root `videoId`, on VideoBlocks whose `source` is internal **or unset** | A media video id only ever appears on internal blocks; YouTube/Cloudflare/Mux hold a provider id in the same column |
| `languageId` | Narrows to blocks playing that variant; omitted matches any variant | The picker may or may not care which dub is embedded |
| Unpublished / templates | Excluded | The caller needs a link a viewer can actually open |
| Team visibility | Not scoped to the publisher's teams | Finding a journey they don't own is the point, and `isPublisher` already gates the field |

## Ordering — most viewed first

Requested on top of the issue. `orderBy: { journeyVisitors: { _count: 'desc' } }`, with `publishedAt desc, id asc` as tiebreakers, so it stays a single statement rather than a count-then-sort round trip.

`JourneyVisitor` is upserted once per unique visitor per journey, so this ranks by lifetime unique visitors. Worth knowing: that structurally favours older journeys over recently popular ones. A recency window would need Plausible or a filtered `Event` count, both of which cost a second query and lose the in-database sort — happy to switch if the picker wants recency.

## Two nullable-column traps this had to dodge

Both were caught in review and then **confirmed against a real Postgres**, not just reasoned about:

- **`Journey.template` is nullable.** `template: { not: true }` compiles to `NOT (template = true)`, which SQL's three-valued logic evaluates to NULL for a legacy row — silently dropping it. Uses `OR: [{ template: false }, { template: null }]`.
- **`Block.source` is nullable with no database default.** VideoBlocks written before the column existed carry a media video id and no source. Filtering `source: 'internal'` alone dropped them: querying video `7_0-nfs0803` returned **0 journeys** before the fix and **3** after. The seed database has 21 such blocks against 29 internal ones, so this was not a hypothetical.

## Response cache

`Query.journeysByVideo` is added to `ttlPerSchemaCoordinate` with TTL `0`. `useResponseCache` runs with `session: () => null` and a default TTL of Infinity, and returns *before* execution on a hit — so scope auth never runs, and one publisher's result would otherwise be served to any caller repeating the same query and variables.

## Verification

- 5 new unit tests; full `api-journeys` suite passes (1119 tests).
- `nx affected --target=type-check --base=origin/main` green across 21 projects.
- Ordering proven against the local database by temporarily boosting two journeys' visitor counts and confirming they moved to positions 1 and 2 (test rows cleaned up afterwards).
- Schemas regenerated through the full chain: subgraph, gateway compose, frontend codegen.

## Not included

No `take`/limit argument — the issue pins the signature and the candidate set is bounded by "journeys embedding one video". Easy to add if a popular video makes the list unwieldy in the picker.

## Summary by CodeRabbit

* **New Features**
  * Added a publisher-only `journeysByVideo` query to find published journeys containing a specified video.
  * Supports optional filtering by language variant.
  * Results are ranked by unique visitors, with recent publications prioritized.
  * Excludes template journeys and videos hosted through external providers.


csiyang (approved):
Verified every load-bearing claim against the code:

- **Both nullable-column traps are real and correctly dodged.** `Block.videoId`, `Block.source`, and `Journey.template` are all nullable in the Prisma schema, so `source: 'internal'` alone and `template: { not: true }` would both silently drop legacy rows exactly as described. The spec asserting the `where` clause exactly — with a comment marking the ORs as load-bearing — is the strongest possible guard against someone "simplifying" them later.
- **The response-cache entry follows the established pattern.** `useResponseCache` runs with `session: () => null`, the sibling auth-gated queries are already TTL-0 with the same cross-user rationale, and the new entry is even in alphabetical order.
- **Auth is properly tested.** `isPublisher` is a real authenticated scope in the builder, and both rejection tests also assert `findMany` was never called — proving auth runs before the database, not just that an error string comes back.
- **The shared-column claim holds**: one `videoId` column serves all `VideoBlockSource` values, so provider ids can never match a media video id under the internal/null filter.

Deciding the issue's four delegated semantics explicitly and documenting them in the schema description — where the consumer will actually read them — is exactly right, as are the honest caveats (lifetime-visitor ordering favouring older journeys, no `take` yet). Nothing to flag.

PR #9573, opened 2026-09-02
feat(resources): remove header tab buttons on resources page

lumberman:
## What this does

Two changes to the [/resources](https://www.jesusfilm.org/resources) page.

### 1. Remove the header tab buttons

Removes the header tab-button section from the page. The section showed the Resources, Journeys, Watch, and Metaverse tabs on desktop, and one dropdown button on mobile. The other pages that use the shared `Header` keep the section.

- Adds a `hideTabButtons` prop to `Header`. When true, the `Header` does not render the `BottomAppBar` or its spacer.
- Adds a `hideHeaderTabButtons` prop to `PageWrapper` that passes through to `Header`.
- Sets the prop on the `PageWrapper` in `ResourcesView`.

Why a new prop: the existing `hideBottomAppBar` prop does not remove the bar. It hides the bar until the user scrolls, and `VideoHero` depends on that behavior. The new prop removes the bar fully and does not change `VideoHero`.

### 2. Align the section cards with the headings

The cards sat 32 px right of the section headings: 24 px from a `Container` wrapper, plus 8 px from the card hover padding. The `slidesOffsetBefore={-32}` compensation in `ContentCarousel` sets a margin through the `sx` prop, but the Swiper stylesheet can win the cascade with `margin-left: auto`. The margin then computes to 0 and the compensation fails.

The fix removes the `Container` wrapper and the `slidesOffsetBefore` value in `ResourceSection`. A `Box` with `ml: -2` now offsets the card padding, so the card image aligns with the heading at all breakpoints. `ContentCarousel` is a shared library component, so the fix stays local to `ResourceSection`.

## Tests

- New tests in `Header.spec.tsx`: the bar shows when the flags are on, and the bar is absent when `hideTabButtons` is true.
- `nx run resources:test`: all tests pass.
- `nx affected --target=type-check` and `pnpm lint:changed --fix`: pass, no fixes.
- Manual check in Chrome on localhost: the headings, the search bar, and the card images share one left edge.

## Note for the reviewer

The tab-section removal applies to the /resources page only. Tell us if the section must go from more pages — the same prop covers that.

## Summary by CodeRabbit

* **User Experience**
  * Removed header tab buttons from the Resources page on desktop and mobile.
  * Adjusted resource card carousel alignment to better match section headings.
* **Bug Fixes**
  * Ensured the header navigation bar appears correctly when relevant feature flags are enabled, unless explicitly hidden.
* **Tests**
  * Added coverage for showing and hiding the header navigation bar under supported conditions.

### DIFF
--- a/apps/resources/src/components/Header/Header.spec.tsx
+++ b/apps/resources/src/components/Header/Header.spec.tsx
@@ -3,6 +3,8 @@ import useScrollTrigger from '@mui/material/useScrollTrigger'
 import { fireEvent, render, screen } from '@testing-library/react'
 import type { Mock } from 'vitest'
 
+import { FlagsProvider } from '@core/shared/ui/FlagsProvider'
+
 import { Header } from './Header'
 
 vi.mock('@mui/material/useScrollTrigger', async () => ({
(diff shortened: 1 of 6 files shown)


lumberman:
Second commit on this branch: the section cards now align with the section headings on /resources. The Container padding and the card hover padding pushed the cards 32 px right, and the slidesOffsetBefore compensation failed because the Swiper stylesheet can override the sx margin. A local Box offset in ResourceSection replaces both.

tanflem (approved):

PR #9575, opened 2026-09-03
ci: right-size 5 runners

mikeallisonJS:
Right-sizes the five selected CI jobs from the 30-day utilization review. The three test recommendations share one matrix job, so one `runs-on` update applies to all three shards.

## Apply for reliability

- **autofix.ci / lint-work**, 2 to 4 vCPU, medium confidence. About 501 monthly runs showed 94.4% peak memory and full CPU utilization. Estimated monthly cost changes from $8.14 to $9.77 (+$1.63), while projected runtime improves about 40%. Median runtime moves from 167s to 99.2s and p95 from 421.8s to 267.6s. [Median sample](https://app.blacksmith.sh/JesusFilm/runs/33550237688/jobs/99997596229?tab=metrics), [p95 sample](https://app.blacksmith.sh/JesusFilm/runs/31037111323/jobs/92411838623?tab=metrics).

## Test before applying

- **Main / build**, 4 to 8 vCPU, medium confidence. About 500 monthly runs saturated all four cores. Estimated monthly cost changes from $8.25 to $10.66 (+$2.41), with projected runtime improving about 35%. Median runtime moves from 80s to 51.3s and p95 from 229.3s to 155.2s. [Median sample](https://app.blacksmith.sh/JesusFilm/runs/31017710993/jobs/92346030648?tab=metrics), [p95 sample](https://app.blacksmith.sh/JesusFilm/runs/33209225083/jobs/98978030953?tab=metrics).
- **Main / test (1/3)**, 4 to 8 vCPU, medium confidence. About 499 monthly runs saturated all four cores. Estimated monthly cost changes from $9.68 to $12.99 (+$3.31), with projected runtime improving about 33%. Median runtime moves from 76s to 50.6s and p95 from 306.6s to 214.4s. [Median sample](https://app.blacksmith.sh/JesusFilm/runs/31066280421/jobs/92504562396?tab=metrics), [p95 sample](https://app.blacksmith.sh/JesusFilm/runs/32167622008/jobs/95810803665?tab=metrics).
[...]

PR #9576, opened 2026-09-03
ci: use Blacksmith node cache in main workflow

mikeallisonJS:
Switches the Main workflow's dependency-installing jobs to Blacksmith's setup-node action. This avoids concurrent cold-key jobs racing to save the same pnpm cache while preserving the existing Node version and cache inputs.

```yaml
- uses: useblacksmith/setup-node@v5
  with:
    node-version: 22
    cache: 'pnpm'
```

Applied to `build`, `graphql-schema-check`, and all `test` shards. `pnpm lint:changed --fix` and the affected type-check were attempted, but this bare checkout has no `node_modules`, so `node_modules/.bin` and `nx` were unavailable.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
         uses: pnpm/action-setup@v4
         with:
           run_install: false
-      - uses: actions/setup-node@v5
+      - uses: useblacksmith/setup-node@v5
         with:
           node-version: 22
           cache: 'pnpm'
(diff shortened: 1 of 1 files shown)


csiyang (approved):
Verified the swap:

- **Complete within scope** — those three are the only `setup-node` uses in `main.yml` (`prisma-migration-check` doesn't install Node), so nothing was missed.
- **Established pattern** — `useblacksmith/setup-node@v5` is already on `main` in the ECS frontend deploy workflows at the same version tag, and all affected jobs run on Blacksmith runners, which the action requires.
- **Proven by this PR's own CI** — `build`, `graphql-schema-check`, and all three test shards ran green with the new action, so cache restore/save works.
- **No conflict with #9575** — different lines of `main.yml`; they merge cleanly in either order.

One observation, not an ask: ten other workflows still use `actions/setup-node` — most notably `autofix.ci.yml`'s `lint-work`, which #9575 just resized and would benefit from the same cache treatment. Worth a follow-up if the Codesmith review doesn't get there on its own.

PR #9577, opened 2026-09-03
chore(deps): remove unused Mermaid CLI

mikeallisonJS:
Removes the unused root Mermaid CLI dependency and its Puppeteer build allowlist entry so routine installs no longer download Chrome for this package. Repository searches found no Mermaid CLI invocation in code, workflows, or scripts.

```yaml
onlyBuiltDependencies:
  - protobufjs
  - sharp
```

The lockfile was regenerated with `pnpm remove @mermaid-js/mermaid-cli --save-dev --lockfile-only`. Required lint and affected type-check commands could not run because this fresh environment has no `node_modules` (`node_modules/.bin` and `nx` were unavailable).

## Summary by CodeRabbit

* **Chores**
  * Removed the Mermaid CLI development dependency.
  * Disabled build scripts for the Puppeteer package during installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -239,7 +239,6 @@
     "@eslint/js": "^10.0.1",
     "@graphql-hive/cli": "^0.48.0",
     "@graphql-tools/executor-http": "^1.0.9",
-    "@mermaid-js/mermaid-cli": "^11.0.0",
     "@nx/esbuild": "23.1.0",
     "@nx/eslint": "23.1.0",
     "@nx/eslint-plugin": "23.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -614,9 +614,6 @@ importers:
       '@graphql-tools/executor-http':
         specifier: ^1.0.9
         version: 1.3.3(@types/node@22.18.8)(graphql@16.10.0)
-      '@mermaid-js/mermaid-cli':
-        specifier: ^11.0.0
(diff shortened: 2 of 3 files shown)


csiyang (approved):
Traced why the dependency existed rather than just confirming it's unreferenced: mermaid-cli was `prisma-erd-generator`'s renderer (the docs troubleshooting page still describes `mmdc` failures from ERD generation), and that generator is already completely gone from the tree — not in `package.json`, zero lockfile entries, no `generator erd` block in any Prisma schema. So the CLI was a genuine orphan left behind by an earlier cleanup, and the two remaining `mmdc` mentions are historical troubleshooting notes, not live tooling.

The subtler half checks out too: dropping `puppeteer` from `onlyBuiltDependencies` is safe because full `puppeteer` left the dependency tree along with mermaid-cli — only `puppeteer-core@13/@20` remain as transitives, and `-core` downloads no Chrome and has no build script to allowlist. The lockfile's six `+` lines are re-resolution formatting noise (deprecation notes, an `optional: true`), not version churn. CI's green install-plus-full-suite run is empirical proof nothing needed the removed pieces.

One follow-up observation, non-blocking: five `DISABLE_ERD: true` env entries linger across `main.yml` (×3), `autofix.ci.yml`, and `api-deploy-worker.yml` — guards for the ERD generator that no longer exists. Same class of dead config as this PR removes; a follow-up sweep would finish the job.

PR #9600, opened 2026-09-16
fix(api-media): keep draft video uploads and manual draft switches in draft

tanflem:
Draft audio uploads published themselves once Mux finished, and switching a Variant back to Draft in videos-admin did nothing. Reported by Holly Gooden for the children of `7_KnowGodBP`.

## What was wrong

**Draft uploads published themselves.** The `processVideoUploads` worker called `createOrUpdateVideoVariant()` without `published`, so it fell back to the parameter's `true` default (introduced in 1f8550a2f). That intent was stored on the `VideoVariantReconciliation` row, and the sweep then published a Variant the uploader had explicitly left as Draft.

**Switching back to Draft silently did nothing.** Since 48cb34348d, `videoVariantUpdate` wrote `currentVariant.published` — the value the row already had — whenever a caller set `published`. The editor got a success toast and no change.

## The fix

- The worker reads `published` off the durable `VideoVariantUpload` row it already loads. Jobs queued before this deploys carry no publish intent in their payload, so reading the row covers those too.
- `videoVariantUpdate` writes an explicit unpublish immediately. Publishing stays with reconciliation, which only publishes once the media is ready — that deferral was the point of 48cb34348d, so this does not restore the pre-48cb34348d line.

## Testing

Regression tests at both seams; each was confirmed to fail before its fix. Verified end to end against a local stack: a Draft upload stayed Draft through Mux and the sweep, and a manual switch to Draft applied immediately and survived the sweep.

## Follow-up

Existing affected Variants (the `7_KnowGodBP` children, and any Draft upload processed since 2026-06-23) are already published and will not self-correct; they need a one-off correction. #9601 builds on this branch with the publish-side feedback gaps Holly also ran into.


tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, no unresolved review threads. +191/-6 across 4 files, keeping draft video uploads and manual draft switches in draft. The `danger` failure is **stale** — it ran on 2026-09-16 before the assignee and reviewer were set, and the only rules `danger` can fail on are PR title, assignee and reviewer, all of which now pass.

### What needs to happen next

Ready for review. Re-run the `danger` job (or push any commit) to turn the check green. Review alongside #9599 and #9601.

PR #9601, opened 2026-09-16
feat: publish variants as soon as media is ready, and show publishes in progress

tanflem:
Follow-up to #9600. While testing that fix, two more gaps turned up in the upload/publish flow — both predate it.

**Stacked on #9600** — base this branch's merge on that one; the diff here is only the follow-up work.

## Publishing looked like it did nothing

Since 48cb34348d, publishing waits for the reconciliation sweep, so the row stayed Draft for up to 15 minutes with a "Variant updated" toast and nothing explaining the delay. That is a large part of how this reads to an editor as "the toggle doesn't work".

- `videoVariantUpdate` now publishes immediately when the Variant has playable media — the same condition the sweep applies (`videoVariantHasUsableMedia`, extracted so both paths share it). A Variant still waiting on Mux continues to publish via reconciliation, so a dead Variant never reaches the watch site.
- `VideoVariant.reconciliation` is exposed (publisher only) so the UI can see a publish that is still in progress.
- The dialog now says the publish is waiting on processing rather than reporting a change the editor cannot see.

## A finished upload stayed invisible

An upload leaves the audio page's in-progress list once the worker creates its Variant. That stopped the page's polling, and nothing refetched the Variants, so a newly uploaded audio language only appeared after a full page reload.

- The page refetches Variants when an upload completes.
- A Variant whose reconciliation is working towards published shows a `Publishing...` chip and polls until it settles, instead of a stale Draft chip.

## Testing

New tests at each seam, each confirmed to fail without its change: api-media publish timing, the chip's pending state, the refetch when an upload finishes, and the dialog's message. Full api-media and videos-admin audio suites pass.


tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, closes nothing explicitly. +488/-22 across 13 files — publishes variants as soon as media is ready and surfaces in-progress publishes. Touches publish timing, which is user-visible.

### What needs to happen next

Run it through `/code-review` before Nisal. The `danger` failure is **stale** (predates the reviewer assignment) and clears on the next run. Note this overlaps #9600 and #9599, which also change publish/draft semantics — review the three as a set.

PR #9602, opened 2026-09-16
fix(videos-admin): open videos from the library tab

tanflem:
Every row click in the Library tab 404s.

## What was wrong

The video list opened `${pathname}/${id}`, i.e. it appended the Video id to whatever page it was rendered on. From `/videos` that gives `/videos/<id>`, which exists. From the Library tab it gives `/videos/library/<id>`, which is not a route — so the editor gets a 404 instead of the Video. Broken since the Library tab was added in 0655784ef (#9361).

Found while testing #9600 against a local stack.

## The fix

Push the Video's own route instead of deriving it from the current path.

## Testing

`VideoList.spec.tsx` mocked the component under test away (`vi.mock('./VideoList')`), so the row click had no coverage at all and every other case in the file was `it.skip`. It now renders the real list and asserts the navigation target from both `/videos` and `/videos/library`, plus that a locked row still does not navigate. The library case fails against the old behaviour.

Independent of #9600 and #9601 — no overlapping files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated video list navigation so unlocked videos consistently open their own detail pages from both the main videos view and the library.
  * Locked videos remain non-navigable.
* **Tests**
  * Expanded coverage with integration tests for unlocked and locked video interactions.


tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, no unresolved review threads. +82/-45 in 2 files, fixing videos not opening from the library tab. The `danger` failure is **stale** — it predates the assignee and reviewer being set.

### What needs to happen next

Ready for review. Re-run `danger` to clear the red check. No code action needed.

PR #9603, opened 2026-09-16
fix(api-media): availableLanguages not dropping stale language on child removal (#9518)

### LINKED ISSUE #9518: availableLanguages: removing a child correctly drops a stale language
## Parent

## What to build

Removing a video from a container's children correctly drops a language it uniquely provided from that container's `availableLanguages`. Today, removing a child fires reconciliation for its variants, but by the time that reconciliation runs the parent-child relation is already severed, so the old parent is structurally unreachable through the normal path and the stale language sticks forever.

`videoUpdate`'s child-removal handling is changed to explicitly recompute the old parent (using the engine from #9517) after the relation is severed, and cascade that recompute upward to the root.

## Acceptance criteria

- [ ] A container with a single child that uniquely provides language X: removing that child from the container's `childIds` results in the container's `availableLanguages` no longer including X.
- [ ] A container with two children, one providing language X uniquely and the other not: removing the child that does not provide X leaves the container's `availableLanguages` unchanged.
- [ ] The recompute triggered by removal cascades to the root, not just the immediate parent (a grandparent's `availableLanguages` also updates when appropriate).
- [ ] Adding a child to `childIds` in the same mutation continues to work as before (no regression on the already-correct add path).

## Blocked by

- #9517 (needs the cascade engine to route the recompute through)

tanflem:
Closes #9518

## Summary

`videoUpdate`'s `childIds` handling previously relied on reconciliation requests to keep a container's `availableLanguages` in sync, but `reconcileParentVariants` only ever **adds** a language, never removes one. Removing the last child providing a given language left that language permanently stale on the parent, because by the time anything else looked for the old parent, the relation was already severed.

This fixes it by explicitly recomputing the container's own `availableLanguages` from its current source data (own published variants ∪ live children's current values) right after the children relation is updated, and cascading that recompute upward through the container hierarchy when it actually changes something.

## Dependency note

This issue was blocked by #9517 (the cascade-to-root recompute engine), which has an open, reviewed PR (#9530) that hasn't merged to `main` yet. Since #9518's acceptance criteria explicitly require cascading past the immediate parent, I adopted the same cascade implementation from that branch here (`apis/api-media/src/schema/video/lib/updateAvailableLanguages.ts` is otherwise untouched on `main` since #9530 branched, so this is a clean, faithful port — cycle guard, fixpoint-stop, and all).

**Whichever of #9530 and this PR merges first will make the other's diff to `updateAvailableLanguages.ts`/`updateAvailableLanguages.spec.ts` a no-op on rebase** — the content is identical.

## Key decisions

- `videoUpdate`: after the `children: { set: [...] }` update, recompute the container's own value via `updateVideoAvailableLanguages(video.id, { skipCache: true, skipAlgolia: true })` — cache/Algolia sync are skipped here since the resolver already does both for this video unconditionally later in the same request. If the recompute changed the stored value, cascade via `updateParentCollectionLanguages(video.id)` (which does sync cache/Algolia for each parent it touches, since nothing else in the resolver covers them).
- `updateParentCollectionLanguages` upgrades from a single-hop parent update to a cycle-safe walker that cascades to the root, stopping early once a level's recomputed value doesn't change.
- `calculateAvailableLanguages`/`updateVideoAvailableLanguages` now also return the previously-stored value alongside the recomputed one, so callers can tell whether a recompute actually changed anything.
- `addLanguageToVideo` (the atomic hot-path primitive for a video's own value) is unchanged — kept for its lost-update-race immunity on concurrent uploads.

## Testing

- `updateAvailableLanguages.spec.ts`: cascade-to-root across 3+ levels, cycle termination, fixpoint-stop, diamond/shared-ancestor correctness, per-branch failure isolation, and removal-symmetry (`drops a language when the child that uniquely provided it is removed` / `keeps a language still provided by the sibling that remains`) against a fixture graph.
- `video.spec.ts`: new tests for `videoUpdate` — recomputes and cascades when childIds changes the value, does **not** cascade when the recompute is unchanged, does **not** recompute at all when `childIds` isn't part of the update (add path unaffected).

## Verification

`pnpm exec nx affected -t lint,type-check,test --base=origin/main` fails in this sandbox with `WorkspaceContext is not a constructor` — reproduces identically on `origin/main` with no args, a pre-existing environment issue (also noted in PR #9516/#9530's verification notes). Ran the underlying tools directly instead:
- `pnpm lint:changed --fix` — clean (0 errors; 2 pre-existing, unrelated `no-empty` warnings in `video.ts`).
- `pnpm tsc7 -p apis/api-media/tsconfig.ts7.json` — clean.
- `vitest run` for `apis/api-media` — 891/891 real tests passing. One pre-existing, unrelated failing suite (`video-variant-reconciliation-migration.spec.ts`, a `process.cwd()`-relative fixture path) — confirmed identical on `origin/main` via `git stash`.


tanflem:
⚠️ Mergetron hit an unexpected error working this unit (attempt 1/3, `git`): Command failed: git push origin phoebe/bot-0-issue-9518. It stays in the queue and will retry.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 27 checks green, no unresolved threads, closes #9518. +601/-28 fixing `availableLanguages` failing to drop a stale language on child removal.

### What needs to happen next

Run it through `/code-review` before Nisal. The `danger` failure showing on this PR is **stale** — it ran before the assignee and reviewer were set, and clears on the next run.

PR #9604, opened 2026-09-16
fix: demote superseded variant upload attempts on the Audio Languages tab (#9561)

### LINKED ISSUE #9561: Demote superseded Variant Upload attempts on the Audio Languages tab
## Problem Statement

A publisher opens the Audio Languages tab for a Video and sees a list of "Upload not complete" rows for languages that are already published and streaming. Each row says the upload cannot be resumed and instructs them to *"Add this audio language again."*

Following that instruction is actively harmful: the language already works, and re-adding it means re-uploading a multi-gigabyte master for no reason. If that re-upload also fails, it leaves behind another permanent row, so the list grows over time.

There is no way to clear the rows. The publisher cannot tell, from the page, which rows represent work still outstanding and which are dead history from an attempt that was already retried successfully. They are left either doing redundant work or ignoring the warnings entirely — and ignoring them means a genuinely failed upload gets ignored too.

This was reported on Augustine, where six rows had been displayed since July for three languages (Farsi, French, Turkish) that were all live and streaming. Four of the six were repeat attempts at the same French file.

## Solution

Distinguish a Variant Upload attempt that a later successful attempt has made irrelevant from one that still needs the publisher's attention.

An attempt is **superseded** when a later attempt for the same Video, Audio Language and Edition reached `variantCreated`. Superseded attempts are demoted, not hidden: they collapse into a "Previous attempts" disclosure that keeps the status, filename, timestamp and any error readable on expand, but carries no call to action, because there is nothing to do.

Every attempt that is *not* superseded renders exactly as it does today — full visibility, full actionability. A publisher re-uploading an Audio Language that already has a published Variant still sees that failure in full, because no later attempt has succeeded.

Attempt history is retained. Nothing is deleted, and the record of repeated failures stays available, since that pattern is itself the signal that an upload path is unreliable.

## User Stories

1. As a publisher, I want rows for Audio Languages that are already live to stop telling me to add the language again, so that I don't re-upload a multi-gigabyte file for work that is already done.
2. As a publisher, I want the Audio Languages tab to show only genuinely outstanding upload work by default, so that I can trust the warnings I see.
3. As a publisher, I want a failed re-upload of an Audio Language that already has a published Variant to remain fully visible, so that a replacement dub that failed is never silently swallowed.
4. As a publisher, I want superseded attempts kept rather than deleted, so that I can still see that a language took several tries.
5. As a publisher, I want superseded attempts collapsed behind a single summary line, so that a language that took five attempts occupies one row rather than five.
6. As a publisher, I want the collapsed summary to say how many previous attempts there were, so that I can tell at a glance whether a language was troublesome.
7. As a publisher, I want to expand the previous attempts and see each one's status, filename, timestamp and error, so that I can diagnose a recurring upload problem.
8. As a publisher, I want superseded attempts to carry no "Add again" button, so that I cannot start redundant work by reflex.
9. As a publisher, I want superseded attempts visually distinguished from outstanding ones, so that severity reads correctly at a glance.
10. As a publisher, I want an attempt superseded only by a later success for the *same* Edition, so that a completed `base` upload does not hide an outstanding Burned In upload for the same language.
11. As a publisher, I want an attempt superseded only by a later success for the *same* Audio Language, so that one language's success never masks another's failure.
12. As a publisher, I want supersession judged by attempt ordering rather than by whether a Variant currently exists, so that re-uploads over an existing Variant behave correctly.
13. As a publisher, I want an attempt that failed *after* the most recent success to stay outstanding, so that the newest failure is always the one I am asked to act on.
14. As a publisher, I want a video whose uploads all succeeded to show no incomplete rows at all, so that a healthy Video looks healthy.
15. As a publisher, I want a video with no upload history to render normally, so that the change is invisible where it does not apply.
16. As a publisher, I want the page to stop polling every few seconds once no outstanding attempts remain, so that a Video with stale history is not refetched indefinitely.
17. As a publisher, I want polling to continue while a genuinely outstanding attempt exists, so that in-progress uploads still update live.
18. As a publisher, I want an in-flight resume to keep working while superseded attempts are collapsed nearby, so that the existing resume flow is unaffected.
19. As a publisher, I want the stale-Mux retry path to keep working, so that the existing recovery routes are unaffected.
20. As a publisher, I want the upload debug details copy action to remain available on superseded attempts, so that I can still gather diagnostics for a support request.
21. As a publisher who has just completed an upload, I want the attempt I retried to move into history immediately on refresh, so that the list reflects reality without a page reload.
22. As a publisher, I want the message on a genuinely unresumable attempt to describe what to do accurately, so that I am not told to "add again" when what I actually want is to start a fresh upload for that language.
23. As a publisher, I want a Video where every incomplete attempt is superseded to show a quiet history affordance rather than a wall of warnings, so that the page communicates "nothing to do here".
24. As a support engineer, I want the retained history queryable, so that I can investigate upload reliability without restoring deleted rows.
25. As an engineer, I want the supersession rule to consult only Variant Upload records, so that its behavior does not change when a Variant is unpublished or removed.
26. As an engineer, I want the rule expressed once, so that outstanding-versus-superseded cannot drift between the rendered list and the polling condition.
27. As a publisher, I want an attempt at `failed` status that a later success superseded to be treated as history too, so that terminal failures do not linger as if actionable.
28. As a publisher, I want ordering within the collapsed history to be newest-first, so that the most recent attempt is the one I read first.

## Implementation Decisions

**Where the rule lives.** Supersession is computed in the videos-admin Audio Languages layout, client-side. No schema change, no new API field, no codegen or subgraph check. The rule is scoped to the one surface that needs it today.

**The query must be widened.** The Variant Upload query currently filters server-side to incomplete statuses only, so successful attempts are never fetched and supersession cannot be determined from the data currently in scope. The query gains `variantCreated` in its requested statuses. Volume is bounded — one Video, existing limit of 100 — so no pagination change is needed.

**The predicate.** An attempt is superseded when, from a prototype validated against production data:

```
superseded(u)  ⟺  ∃ s :  s.videoId    = u.videoId
                     ∧  s.languageId = u.languageId
                     ∧  s.edition    = u.edition
                     ∧  s.status     = 'variantCreated'
                     ∧  s.createdAt  > u.createdAt
```

Edition is part of the key, not just language. The Variant table is not consulted at all — this is deliberate, and is what makes a failed re-upload over an existing Variant stay visible.

**Three-way partition.** The fetched list divides into: successful attempts (used only as supersession evidence, never rendered as incomplete), superseded attempts (rendered as collapsed history), and outstanding attempts (rendered exactly as today). The partition is derived once and both the rendered list and the polling condition read from it.

**Polling.** The polling condition must key on outstanding attempts only. Today it keys on the length of the incomplete list, which is why Videos with stale rows have been polling every three seconds indefinitely. Widening the query without changing this would make that worse, not better, by adding successful attempts to the same list.

**The upload's `version` field is not usable** as a discriminator. It is client-supplied on start rather than auto-incremented; all five attempts at the same French file carried `version: 1`.

**Component contract.** The incomplete-uploads component receives the outstanding attempts and the superseded attempts as distinct inputs rather than deriving the split itself, keeping the rule in one place.

**Copy.** The message for a genuinely unresumable outstanding attempt should describe starting a fresh upload for that language rather than "adding the language again", which misdescribes the action when the language already exists. Superseded attempts get a neutral historical label and no action.

**No deletion.** Nothing in this change removes records. Retention is the policy; demotion is the mechanism.

## Testing Decisions

**What makes a good test here.** Tests assert what a publisher sees and what the page does — which rows render, what each says, which carry an action, whether polling starts. They do not reach for the supersession helper directly or assert on its shape; it stays unexported so no test can couple to it. Fixtures vary the upload list and assert on rendered output, which keeps the tests honest if the rule is later moved server-side.

**Seam.** One, and it already exists: the Audio Languages layout spec. It renders the client layout with `useQuery` mocked, which is the highest available seam and the same one every existing upload behavior is tested through. No new seam is introduced.

**Prior art.** The existing suite already covers this surface closely — there are tests for querying and rendering server-filtered incomplete uploads, for lifecycle-specific labels, messages and actions, for routing unresumable uploads to add-again without calling resume, for the stale Mux threshold and retry, and for an active resumed upload disappearing from the incomplete list. New tests follow those directly.

**Cases to cover.** An attempt with a later success is demoted; an attempt with no later success stays outstanding; an attempt whose only later success is in a different Edition stays outstanding; same for a different Audio Language; a failure *after* the most recent success stays outstanding; a language with several superseded attempts collapses to one summary reporting the count; expanding the summary reveals each attempt's detail; superseded attempts render no action control; polling starts when an outstanding attempt exists and does not when only superseded ones remain; the existing resume and stale-retry flows are unchanged.

**Required update to an existing test.** The test asserting the query variables for server-filtered incomplete uploads will need its expected status list updated, since the query now also requests `variantCreated`. This is an intended change, not a break.

## Out of Scope

- Exposing supersession as a field from the media API. Deferred until a second consumer exists; the rule is client-side for now.
- A discard or delete mutation for Variant Upload records. None exists today, which is a real gap, but this spec makes it unnecessary for the reported problem rather than closing it.
- Removing orphaned objects from R2. Deleting an upload record never removed bytes; incomplete multipart uploads accumulating in the bucket need a bucket lifecycle rule and are unrelated to this display change.
- Investigating why multi-gigabyte browser uploads fail. Six abandoned transfers of 7.6–8.6 GB files, one needing five attempts across four days, is a real reliability question and deserves its own issue.
- The stranded Variant Reconciliation rows found during the same investigation. Unrelated defect, separate fix.
- The production data cleanup, which has already been carried out.

## Further Notes

The reported instance has already been resolved in production: seven stale Variant Upload records across two Videos, together with six associated R2 asset records, were removed under a guarded transaction, with a restorable CSV snapshot retained. Catalog-wide there were only seven such rows across two Videos, and a check for incomplete attempts with no corresponding live Variant returned zero — no upload work had actually been lost anywhere.

This spec exists because deletion fixes the symptom once. Without the display rule, the next abandoned upload recreates the problem, and there is no way for a publisher to clear it.

The supersession predicate was validated against the production rows before they were deleted: it identifies exactly the seven that were removed and fires on zero surviving records.

Worth noting for whoever picks this up: the polling behavior means affected Videos have been refetching every three seconds continuously since July. Fixing the partition fixes that as a side effect, but only if the polling condition is switched to read outstanding attempts rather than the full incomplete list.

tanflem:
Closes #9561

## Summary

The Audio Languages tab was showing "Upload not complete" / "Add this audio language again" rows for languages that are already live and streaming, whenever a later re-upload attempt for the same video+language+edition had succeeded. Following the "add again" instruction re-uploads a multi-gigabyte master for no reason.

This change demotes those attempts into a collapsed "Previous attempts" history instead of hiding or deleting them, while leaving any attempt that is *not* superseded (including a failed retry over an existing Variant) fully visible and actionable, exactly as before.

## Key decisions

- **Query widened, client-side only.** `GET_VIDEO_VARIANT_UPLOADS` now also requests `variantCreated` rows, purely as supersession evidence — those rows themselves are never rendered. No schema change, no codegen.
- **Rule expressed once.** A single, unexported `partitionVideoVariantUploads()` helper in `layout.tsx` splits fetched uploads into `outstandingUploads` / `supersededUploads`. Both the rendered list and the polling condition read from this one partition, so they can't drift apart.
- **Predicate matches the spec exactly:** an attempt is superseded when a later attempt for the same `videoId` + `languageId` + `edition` reached `variantCreated` (compared by `createdAt`). The Variant table is never consulted, so a failed re-upload over an existing published Variant stays visible.
- **Component contract.** `IncompleteVideoVariantUploadItems` now takes `outstandingUploads` and `supersededUploads` as distinct props rather than deriving the split itself.
- **Collapsed history.** Superseded attempts group by language+edition into one "N previous attempts" disclosure (MUI `Collapse`, `unmountOnExit`), newest-first, with status/filename/timestamp/error visible on expand and no action button. The copy-debug-details action is preserved.
- **Polling fixed.** The 3s poll now keys on `outstandingUploads.length` instead of the full incomplete list, so a Video with only stale/superseded history stops polling indefinitely (this was the reported bug: affected videos had been polling every 3s since July).
- **Copy fix.** The message for a genuinely unresumable outstanding attempt now says to start a new upload for the language, instead of "Add this audio language again" (which misdescribes the action once the language already exists).

## Testing

Added coverage in `layout.spec.tsx` for: demotion by a later same-language/edition success, staying outstanding when the only later success is a different edition, staying outstanding when it's a different language, staying outstanding when the failure happened *after* the most recent success, collapsing multiple superseded attempts into one summary with the count and revealing each on expand, and polling starting/stopping based on outstanding vs. superseded attempts. Updated the existing query-variables assertion to include `variantCreated` in the requested statuses.

## Test plan

- [x] `pnpm lint:changed --fix` — clean
- [x] `nx affected --target=type-check` — clean (see note below on the workaround used)
- [x] Full `videos-admin` vitest suite — 474 passed, 10 skipped, 0 failed
- [ ] Manual QA on preview: open a Video's Audio Languages tab with repeated upload attempts for a language, confirm superseded ones collapse and outstanding ones behave unchanged

Note: `nx` itself fails to start in this worktree with `TypeError: WorkspaceContext is not a constructor` (a native-addon load issue unrelated to this change). Type-check was run with `NX_SKIP_NATIVE_FILE_CACHE=true`, and tests were run directly via `vitest` using the project's own `vitest.config.mts`. Both are logged in the verification artifact.


tanflem:
Mergetron update for #9561: 1 new commit(s) pushed to this branch.

tanflem:
Mergetron attempted an idle CI fix for PR #9604 but could not resolve the failing checks. The branch was left unchanged. A human should investigate the CI failures.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 26 checks green, closes #9561. +588/-16 across 3 files. Supersedes #9569, which duplicates it and is currently failing lint.

### What needs to happen next

Run it through `/code-review` before Nisal, including the Major CodeRabbit point raised on #9569's copy of `audio/layout.tsx`.

PR #9607, opened 2026-09-16
feat(api-languages): sync WESS native (autonym) language names into LanguageName (#9483)

### LINKED ISSUE #9483: Sync WESS native (autonym) language names into LanguageName

tanflem:
Closes #9483

## Summary

Extends the WESS languages import (the "Run WESS import" flow) to read `NATIVE_LAN_NAME` and store it as a language's own autonym `LanguageName` row, which becomes that language's Primary Name once known.

- `WessLanguageRow` gains `nativeName`, parsed via the existing `readStringField` alias-list convention (`nativeName`, `NativeName`, `NATIVE_LAN_NAME`) — blank/whitespace normalizes to `null` for free.
- `upsertLanguage` writes the autonym row (`parentLanguageId: id, languageId: id`) whenever `nativeName != null`, unconditionally demoting every other `LanguageName` row for that language first — the demotion never depends on whether an English label was present in the same run (this is the case the spec's 2026-09-16 amendment called out as the actual shipped defect on #9508).
- The English-label write checks for an existing stored autonym before deciding whether a brand-new row defaults to primary, and never writes `primary` on update — preserving "at most one primary row per language" across all four `LAN_NAME`/`NATIVE_LAN_NAME` presence/absence combinations, on both fresh and previously-imported languages.
- English's own WESS row (id `529`) flows through the same autonym branch as any other language (one row, not a special case); the English-label branch is skipped only for that id to avoid a duplicate write to the same composite key.
- `runWessLanguagesImport` and `WessImportResult` gain `nativeNamesImported`, surfaced in the "Run WESS import" button's success message.
- README (`apis/api-languages/src/scripts/README.md`) documents `NATIVE_LAN_NAME` alongside the fields it already documents.

## Testing

- `wess-language-parsers` normalization gains cases for `NATIVE_LAN_NAME` mapping, alias fallbacks, and blank/whitespace-only handling.
- Per the testing amendment on the issue, `wess-languages-import.spec.ts` now also covers `runWessLanguagesImport` against a Prisma mock backed by an in-memory `languageName` table honouring the real `@@unique([parentLanguageId, languageId])` upsert semantics — asserting on the rows a run leaves behind for each presence/absence combination (fresh language, previously-imported language, English's own row, blank native name, multi-row aggregation).

## Test plan

- [x] `pnpm lint:changed --fix` — clean
- [x] `pnpm tsc7 -p apis/api-languages/tsconfig.ts7.json` — clean
- [x] `vitest run` for `api-languages` — 128 passed
- [x] Sanity check of `apps/videos-admin`'s `WessImportButton.spec.tsx` — 7 passed
- [ ] Run the "Run WESS import" button against a stage environment before this reaches production (per the spec's Rollout section — needs a human/publisher)

**Note:** `pnpm exec nx affected`/`nx run` fail in this environment with `TypeError: WorkspaceContext is not a constructor` for every invocation (reproduced even on `nx --version`'s project graph call) — an environment issue unrelated to this change. Verified via the underlying tools directly instead (lint-changed.sh, tsc7, vitest), and regenerated `schema.graphql`/`api-gateway/schema.graphql` by running the exact commands those nx targets invoke.


tanflem:
Mergetron update for #9483: 1 new commit(s) pushed to this branch.

tanflem:
Mergetron attempted an idle CI fix for PR #9607 but could not resolve the failing checks. The branch was left unchanged. A human should investigate the CI failures.

tanflem:

## Triage — 2026-09-16

### Where this stands

Mergeable, 41 checks green, closes #9483. +509/-28. This supersedes #9508, which touches the identical 9 files and should be closed.

### What needs to happen next

Run it through `/code-review` before Nisal — and carry over the Major data-integrity point CodeRabbit raised on #9508's copy of `wess-languages-import.ts`, which has not been re-raised here.

tanflem:
## Carried over from #9508: primary-name transition is not atomic

#9508 is the older duplicate of this PR (same 9 files, same issue #9483) and is headed for closure in favour of this one. Before it goes, one finding raised there needs to move across, because it was never re-raised here — this PR currently has no unresolved review threads, so nothing is tracking it.

**Origin:** CodeRabbit, `Data Integrity & Integration / Major`, on `wess-languages-import.ts` in #9508.

**The finding:** the demotion and the promotion are two separate writes. In this branch that is the `languageName.updateMany({ data: { primary: false } })` that clears every other name for the language, followed by the `upsertLanguageNameEntry` that writes the autonym with `primaryOnCreate/primaryOnUpdate: true`. If the second write fails, the first has already committed — and the language is left with **no primary name at all**, which is worse than the two-primaries state this code exists to prevent.

**Still applies here.** I checked: there is no `$transaction` anywhere in this branch's copy of the script, and the two writes are plain sequential `await`s.

**Suggested fix:** wrap the demotion and the autonym upsert in a single `prisma.$transaction`, plus a failure-path test that forces the autonym write to fail and asserts the demotion rolled back.

### Note on the other #9508 finding — no action needed

#9508 also carried a correctness fix (`6573c7dab`) for a double-`primary: true` bug: there, the demotion rode along on the English-gloss upsert guarded by `row.name != null`, so a WESS row with `NATIVE_LAN_NAME` but no `LAN_NAME` promoted the autonym without demoting the English row.

**This PR does not have that bug.** The `updateMany` here sits outside the `if (row.name != null && row.id !== englishLanguageId)` block, with a comment saying it must not depend on whether an English label was written. This branch is an independent reimplementation rather than a descendant of #9508, and it got that part right on its own — so that fix does not need porting. Only the atomicity point above does.

PR #9608, opened 2026-09-16
Mergetron: fix Language.name[0] call sites broken by native-name primary flip (#9511)

### LINKED ISSUE #9511: Fix Language.name[0] call sites broken by native-name primary flip
## Problem

#9483 makes a language's native (autonym) name its Primary Name once WESS supplies one — the previously-always-English `LanguageName` row for that language flips to `primary: false`, and the autonym row becomes `primary: true`. `Language.name` is ordered `primary: desc`, so any consumer that reads `name[0]` with no `languageId`/`primary` argument will silently start receiving the native-script autonym instead of English, for any language where one now exists.

Most consumers in the codebase were already written defensively for this two-row model (explicit `languageId: "529"`, or `.find(n => n.primary)` / `.find(n => !n.primary)` splits) and are unaffected. A small, enumerable set is not:

1. `apps/player/src/components/PlaylistList/PlaylistList.tsx:64` — `item.videoVariant.language?.name?.[0]?.value`, embeddable public player's playlist language label. Backed by `apps/player/src/lib/queries/getPlaylist.ts:22-27` (`language { id name { value } }`, no args).
2. `apps/watch/src/components/DialogDownload/DialogDownload.tsx:308` — `{language.name[0].value}`, watch site's Download dialog language chip. Backed by `videoContentFields.ts:49-52` (`name { value primary }`, no args).
3. `apps/resources/src/components/DownloadDialog/DownloadDialog.tsx:195` — same pattern, resources app's Download dialog.
4. `apps/journeys-admin/.../VisitorInfo/VisitorJourneysList/utils/transformToJourney/transformToJourney.ts:26` — `event.language?.name?.[0]?.value`, admin visitor-session event subtitle. Backed by `GET_VISITOR_EVENTS`'s `language { id name { value } }` in `VisitorJourneysList.tsx:26-28`.
5. `apps/journeys-admin/.../VisitorInfo/VisitorJourneysList/utils/data.ts:334` — duplicate mapping site for the same feature as #4.

## Solution

Update each of the 5 call sites to select the English/local name explicitly instead of relying on array position — matching the pattern already used everywhere else in this codebase for the same field, e.g. `VideoControls.tsx`'s `.find(({primary}) => !primary)?.value ?? name[0]?.value`. Either fix at the query level (pin `languageId: "529"` or add a `primary` argument, per what each existing query already supports) or at the read site (`.find` instead of `[0]`), whichever matches the existing GraphQL query shape at each location.

## Scope

- Fix all 5 sites listed above.
- No new GraphQL fields or schema changes expected — the backend (#9483) already exposes what's needed; this is a client-side read-pattern fix.
- Verify each fixed surface still shows the English/local label as before, and that array position is no longer load-bearing.

## Out of scope

- Any other `Language.name` consumer not listed above — the wider consumer sweep already confirmed the rest are unaffected.
- The #9483 backend change itself.

## Dependency

Blocked by #9483 — the primary-flag flip in that change is what makes these 5 sites unsafe. Do not start until #9483 has merged and its native-name sync has actually run (otherwise there's nothing to reproduce the bug against).

## Further Notes

Full consumer-impact sweep (which sites are safe vs. affected, and why) was produced during the grilling session for #9483 — see that issue's history for the complete list of ~immune and already-defensive call sites, not repeated here.

tanflem:
Closes #9511

## Summary

#9483 made a language's native (autonym) name its Primary Name once WESS supplies one. `Language.name` is ordered `primary: desc`, so any consumer reading `name[0]` with no `languageId`/`primary` argument silently starts receiving the native-script autonym instead of English wherever that flip has happened.

This fixes 3 of the 5 sites named in the issue:

1. `apps/player/src/components/PlaylistList/PlaylistList.tsx` — embeddable public player's playlist language label
2. `apps/resources/src/components/DownloadDialog/DownloadDialog.tsx` — resources app's Download dialog language chip
3. `apps/journeys-admin/.../VisitorJourneysList/utils/transformToJourney/transformToJourney.ts` and the duplicate mapping in `utils/data.ts` — admin visitor-session event subtitle

The other 2 sites named in the issue (`apps/watch/src/components/DialogDownload/DialogDownload.tsx` and its `videoContentFields.ts` query) no longer exist — the `watch` app was removed in #9490 before this issue could be worked.

## Approach

Each site now reads `.find(({ primary }) => !primary)?.value ?? name[0]?.value`, matching the existing pattern used elsewhere in the codebase (e.g. `VideoControls.tsx`) rather than introducing a new helper. Where the GraphQL query didn't already select `primary`, it was added to the query (`getPlaylist.ts`, `GET_VISITOR_EVENTS`) so the field is available client-side, rather than only patching the read site.

Since this sandbox has no global `apollo` install, the apollo-generated `GetVisitorEvents.ts` type was updated by hand to match what `nx codegen journeys-admin` would produce (documented gap in `AGENTS.md`).

## Test plan

- [x] Added a regression test per fixed surface asserting the local/English name is shown even when the autonym is `primary: true` and appears first in the `name` array (`PlaylistList.spec.tsx`, `DownloadDialog.spec.tsx`, `transformToJourney.spec.ts`)
- [x] Updated existing mock fixtures across both apps to include the now-required `primary` field
- [x] `pnpm exec nx affected -t lint,type-check,test --base=origin/main` passes (see note below)

Note: in this sandbox, `nx`'s native binding needs `NX_SKIP_NATIVE_FILE_CACHE=true` (the `/tmp` mount here is `noexec`, which breaks nx's native-module file cache), and ESLint on the largest projects needs a raised `--max-old-space-size` or it OOMs on the default heap. Both are environment quirks unrelated to this change — the gate is green with those two env vars set.

## Summary by CodeRabbit

* **New Features**
  * WESS language imports now recognize native language names and report how many were added or updated.
  * Native autonyms are stored as primary names, while other names remain available as translations.
  * Language displays in journeys, playlists, and downloads now prefer the local name over the native autonym when available.
* **Documentation**
  * Updated import documentation to explain native-name handling and preservation behavior.


tanflem:
Mergetron update for #9511: 1 new commit(s) pushed to this branch.

tanflem:

## Triage — 2026-09-16

**Verdict: 🔧 Fixes needed before review**

### Where this stands

**Title violates commitlint, plus missing metadata.** `danger` reports three fails:
- PR title does not match commitlint convention — `Mergetron: fix Language.name[0] ...` has no conventional-commit type prefix.
- No assignee.
- No requested reviewer.

Also 2 unresolved CodeRabbit threads (both Minor / quick win) on `wess-languages-import.ts` and the scripts `README.md`.

### What needs to happen next

1. Rename the PR to something like `fix(api-languages): repair Language.name[0] call sites after native-name primary flip (#9511)`.
2. Add an assignee and request `@Kneesal`.
3. Clear the 2 quick-win CodeRabbit threads.

PR #9616, opened 2026-09-23
feat: develop core on a host checkout without the dev container

mikeallisonJS:
## Summary

Lets `core` be developed on a host checkout (WSL2 / Linux / macOS) without the dev container, while keeping the dev container working and sharing the same backing services and data between the two.

- **Backing services split out**: `db`, `redis`, `serverless-redis-http`, `maildev`, `clickhouse`, `plausible` move to a root `docker-compose.services.yml` that `.devcontainer/docker-compose.yml` pulls in via `include:`. It pins the `core_devcontainer` project name, so host and dev container use the same containers and volumes (existing databases are preserved — volume names are unchanged).
- **`localhost` everywhere**: all services share one network namespace anchored on `db` (`network_mode: service:db`), and the dev container `app` joins it, so every service is `localhost` inside and outside the container. Prisma migrate targets, redis fallbacks and docs updated. Legacy names (`redis`, `maildev`, `plausible`, …) are aliased to `127.0.0.1` inside the namespace as a shim for configs not yet migrated.
- **Host bootstrap**: `tools/scripts/setup-local.sh` (`pnpm setup:local [--base]`) — prereq checks, `compose up --wait`, `test-user` role, `pnpm install`, plausible seed (idempotent). `post-create-command.sh` reuses it (`--no-services`), no more hardcoded `/workspaces/core`.
- **No global installs on host**: `foreman` is a devDependency, `Procfile` uses `pnpm exec nx`, `pnpm dev:api` / `services:up` / `services:up:base` / `services:stop` scripts. `apollo` stays global-only (`AGENTS.md` explains why).
- **Hygiene**: `* text=auto eol=lf` in `.gitattributes` (fixes ~8k phantom-modified files on Windows clones), `.nvmrc` + `engines.node` (22), devcontainer scripts use `corepack enable` (honours `packageManager` instead of a stale pnpm pin), `NODE VARIANT` aligned with CI.
- **Docs**: new `02-getting-started/05-local-setup-without-dev-container.mdx` (per-OS native deps, WSL2 guidance, host env vars incl. `TZ=UTC` for tests); `.vscode/extensions.json` mirrors the devcontainer list.

## ⚠️ Merge prerequisite: Doppler `dev` configs

The `dev` configs for **api-journeys, api-media, api-users, api-languages, api-analytics, arclight, journeys, journeys-admin** still reference `redis` / `db` / `maildev` / `plausible` / `clickhouse` as hostnames. Inside the dev container the alias shim keeps them working, but on a host checkout they fail (`ENOTFOUND redis`). A Doppler admin needs to change those hosts to `localhost` when this lands. The `libs/prisma/*` `.env`s are fetched from the same projects.

## Verified

**Dev container** (rebuilt on this branch): post-create completes, Plausible on :8000, existing Postgres data intact, `pnpm -v` = 10.33.4, `nf start` works.

**WSL2 host** (fresh clone, same containers): `pnpm setup:local` → `fetch-secrets` (20 projects) → Prisma `migrate status` up to date for journeys/users/media/languages over `localhost` → `pnpm dev:api` (gateway composes on :4000, APIs on 4002–4008, 0 connection errors with `localhost` envs) → `journeys-admin:serve` responds → `api-users:test` (78) and `journeys-admin:test` (3,777, with `TZ=UTC`) pass.

## Notes for reviewers

- The devcontainer CLI reads `.devcontainer/*` config from the folder Cursor was opened on, not the workspace volume — with the clone-into-volume flow both must be on this branch to test.
- `services:stop` deliberately uses `stop`, not `down`: the project is shared with the dev container, and `--remove-orphans` would delete the `app` container.
- A "Rebuild Container" does not always recreate backing services whose config changed (clickhouse needed a `--force-recreate`).

## Summary by CodeRabbit

* **New Features**
  * Added local development setup scripts and commands for running services, APIs, and analytics with or without a dev container.
  * Added shared Docker services for PostgreSQL, Redis, MailDev, ClickHouse, and Plausible.
  * Updated development support to Node.js 22.
  * Added recommended editor extensions.

* **Documentation**
  * Added a guide for setting up the project without the dev container.
  * Clarified local database access and Doppler login instructions.

* **Bug Fixes**
  * Improved local PostgreSQL and Redis connectivity by using localhost-based defaults.


csiyang (commented):
The architecture here is right — `network_mode: service:db` is what makes "localhost everywhere" true in *both* modes, which is the precondition for pointing the Doppler dev configs at one set of hostnames. Verified the migration is data-safe (project name pinned, volume names unchanged), the alias shim is correctly scoped (devcontainer keeps working with un-migrated Doppler values; only the new host mode waits on the admin), and the verification story — devcontainer rebuild plus a fresh WSL2 clone with suites green — is as thorough as this kind of PR gets.

Four asks before approval, three of which are the open CodeRabbit threads (all real, all cheap):

1. **Bind the published ports to loopback** (`127.0.0.1:5432:5432` etc. on `db`). Technically pre-existing exposure — the old devcontainer compose also published unbound — but this PR is exactly the moment: it moves the file, and its purpose is putting these services on roaming host laptops. Unauthenticated Redis plus committed postgres credentials on every interface is a real shared-network exposure. Six tokens, breaks nothing: publishing is host-side only, and namespace-internal traffic (`app` → db etc.) doesn't touch it.
2. **The `[ "$i" -eq 30 ]` off-by-one** in `setup-local.sh` — success on try 30 reports failure. Faithfully copied from the old `post-create-command.sh`, but it's a one-line fix in a brand-new file (CodeRabbit's re-probe suggestion is right).
3. **`--single-transaction` on the plausible seed** — the "already seeded" guard checks the `users` table, so a partial seed (users committed, later insert fails) permanently skips re-seeding. One flag makes the guard sound.
4. **Docs note on the namespace coupling**: if the `db` container is ever recreated (image bump, config change), every joined container — including the devcontainer `app` — is left holding a dead network namespace and needs restarting. The PR body hints at this ("clickhouse needed `--force-recreate`") but the new getting-started page doesn't warn about it; two lines there would save the first person who hits it a confusing hour.

Not concerns, for the record: the Doppler merge-gate sequencing is sound (CI and deployed configs don't use these hostnames; only the new host mode is affected until the admin updates them), `foreman` resolves via `node_modules/.bin` for `pnpm dev:api` while the devcontainer keeps its global copy, and `* text=auto eol=lf` is safe given Prettier already enforces LF. Happy to approve once the above land.

PR #9617, opened 2026-09-23
docs(infrastructure): ADR for Plausible EBS snapshot retention

mikeallisonJS:
## Summary

Adds `infrastructure/docs/adr/0001-plausible-ebs-snapshot-retention.md`, the first ADR in the infrastructure context. It records the decision to bound EBS snapshots of the Plausible ClickHouse volumes (7 daily + 4 weekly in prod, 3 daily in stage) with an AWS Data Lifecycle Manager policy, delete the existing 1,462+ snapshot backlog after a verified restore, and retire the in-chart schedule once DLM is producing snapshots.

Snapshot storage was $522 in August 2026 and rising ~$25/month, 12% of the account's spend.

## Context correction against the chart

The draft described the schedule as living in `volume-snapshot.yaml` with no retention. The chart actually has:

- `templates/snapscheduler.yaml`: a SnapScheduler `SnapshotSchedule` (daily, `retention.maxCount: 5`, no `claimSelector`).
- `templates/snapshot-class.yaml`: `VolumeSnapshotClass` with `deletionPolicy: Retain`.
- `templates/volume-snapshot.yaml`: two one-shot `VolumeSnapshot`s.

So pruning past five removes the Kubernetes object but keeps the EBS snapshot, which is the actual mechanism behind the unbounded backlog. The ADR's context section now says this, and the alternatives list gains "flip `deletionPolicy` to `Delete`" as the cheaper in-cluster option, rejected for the same visibility reason the draft already gave for in-cluster retention.

## Follow-ups noted in the ADR

- `apis/api-analytics/infrastructure/snapscheduler.yaml` is a copy of the same schedule applied to the `default` namespace by that service's deploy scripts.
- Audit other PVCs in both clusters for the same pattern.


mikeallisonJS:
Closing: the ADR itself will not be merged as a PR; the implementing infrastructure change is coming as a separate PR.

PR #9618, opened 2026-09-23
feat(infrastructure): bound Plausible ClickHouse EBS snapshots with a DLM policy

mikeallisonJS:
## Summary

Implements the Plausible EBS snapshot retention ADR. Snapshot storage (`USE2-EBS:SnapshotUsage`) was $522 in August 2026, rising ~$25/month, 12% of the account's spend, from 1,462+ snapshots of the two 500 GiB `clickhouse-gp3` volumes.

**Why the backlog is unbounded:** the Plausible chart's SnapScheduler schedule (`templates/snapscheduler.yaml`) prunes `VolumeSnapshot` objects past five, but the `VolumeSnapshotClass` (`templates/snapshot-class.yaml`) has `deletionPolicy: Retain`, so every EBS snapshot behind a pruned object is kept forever.

**What this adds**

- `infrastructure/modules/aws/ebs-snapshot-policy`: an AWS Data Lifecycle Manager policy plus its IAM role (`AWSDataLifecycleManagerServiceRole`) for one set of PVC-backed volumes on one EKS cluster.
- `module.eks` gains a `cluster_name` output.
- Prod: `daily` at 01:00 UTC keeping 7, `weekly` Sunday 02:00 UTC keeping 4.
- Stage: `daily` at 01:00 UTC keeping 3.

**How volumes are targeted.** Both environments share one AWS account and the same PVC names, so a policy keyed on `kubernetes.io/created-for/pvc/name` alone cannot give prod and stage different retention. The module finds the volumes for a cluster by that PVC-name tag *and* attachment to instances tagged `eks:cluster-name`, adds a `SnapshotPolicy = jfp-dlm-plausible-clickhouse-<env>` tag to them with `aws_ec2_tag`, and points DLM at that tag. Snapshots copy the volume tags and get `Name`/`Env`/`SnapshotPolicy` of their own, so they are attributable in Cost Explorer once cost-allocation tags are activated.

**Guard rails.** Plan fails with a clear message if no running nodes are found for the cluster, or if no matching volume is attached (for example ClickHouse is down). Failing loudly is deliberate: applying in that state would remove the tag and DLM would quietly stop snapshotting.

## Verified

- `tofu fmt -recursive` clean, `tofu validate` passes (OpenTofu 1.11.8, the Atlantis version). The only warnings are pre-existing Datadog deprecations.
- Not verified: an actual plan. No AWS credentials on this machine, so the Atlantis plan on this PR is the first real check that the `eks:cluster-name` instance tag and the PVC-name volume tags resolve as expected. Please read the plan for exactly one or two `aws_ec2_tag` resources per environment before applying.

## Deliberately not in this PR (per the ADR's sequencing)

1. Removing the SnapScheduler schedule from the Plausible chart. That comes once DLM is confirmed producing snapshots, so there is never a gap.
2. Deleting the ~1,460 snapshot backlog. One-off operational step after a verified restore from a recent snapshot; not code.
3. `apis/api-analytics/infrastructure/snapscheduler.yaml` is a second copy of the same schedule, applied to the `default` namespace by that service's deploy scripts. Should go with item 1.

## Noticed along the way

The ArgoCD applications for Plausible point at `infrastructure/kube/plausible`, but the chart directory is `infrastructure/kube/plausible-analytics`. Not touched here.

## Summary by CodeRabbit

* **New Features**
  * Added automatic backups for Plausible ClickHouse data in production: daily snapshots are retained for 7 days, with weekly snapshots retained for 4 weeks.
  * Added daily backups for Plausible ClickHouse data in staging, retaining the 3 most recent snapshots.


csiyang (dismissed):
Verified the whole causal chain in-repo: `deletionPolicy: Retain` is exactly as claimed in `snapshot-class.yaml`, SnapScheduler's `maxCount: 5` daily schedule is there, the second copy in `apis/api-analytics/infrastructure/` exists, and `aws_eks_cluster.this` backs the new output. The module is textbook against `infrastructure/AGENTS.md` — reusable, fully variable-driven with validation blocks that match real DLM limits (≤4 schedules, retain ≤1000, six-field `cron(...)` form; both cron expressions are correct), every resource tagged.

The design decisions are right: tag-and-target is the only way to split prod from stage given one account and identical PVC names; the `for_each` over volume IDs resolves at plan time (data-source-only dependencies), so Atlantis can plan it; and failing the plan loudly when the workload is down — rather than silently untagging — is exactly the correct trade. The ADR sequencing (SnapScheduler stays until DLM is proven; backlog deletion waits on a verified restore) is disciplined.

Two observations, neither blocking:

1. **Silent-lapse drift is the residual risk this design carries.** The preconditions cover *plan time*, not *drift over time*: if a volume is ever recreated out-of-band (ClickHouse reinstall, PVC recreation), the replacement has no `SnapshotPolicy` tag until the next infra plan — and DLM won't error, it will just snapshot nothing. Suggest a follow-up: a snapshot-age alarm (or a scheduled plan) so a lapse is noticed in days rather than at restore time.
2. The precondition means *unrelated* infra PRs will fail their Atlantis plan if ClickHouse happens to be mid-restart at that moment. Deliberate and right — just worth remembering when someone hits a confusing plan failure on an unrelated change.

Echoing the PR's own instruction because it's the one thing that matters at apply time: **the Atlantis plan on this PR is the first real test of the `eks:cluster-name` / PVC-tag resolution — confirm it shows exactly one or two `aws_ec2_tag` resources per environment before applying.**

PR #9621, opened 2026-09-23
fix: update Core Next.js to 16.3.6

tataihono:
## Summary

Update Next.js and its companion tooling to 16.3.6 for [GHSA-vcvr-r3jv-pc5j](https://github.com/vercel/next.js/security/advisories/GHSA-vcvr-r3jv-pc5j) (CVE-2026-94545). The maintainer advisory lists `>=16.2.0 <16.3.6` as affected and 16.3.6 as patched. This deliberately crosses from 16.2 to 16.3; 16.2.12 remains within the advisory's affected range, so it is not a patched backport for this issue. The change aligns `@next/bundle-analyzer`, `@next/eslint-plugin-next`, `@next/third-parties`, `eslint-config-next`, and the Next.js override, then refreshes the lockfile.

## Verification

A frozen dependency install, `pnpm lint:changed --fix`, and `pnpm exec nx affected --target=type-check --base=origin/main` passed locally across 35 affected projects after generating Prisma clients. CI, review, and the journeys-admin preview are merge gates.

## Monitoring

After merge, confirm the affected Next.js applications deploy from `main` and their main routes start normally. The advisory describes a conditional Node.js `ImageResponse` exploit; this PR does not assert whether the applications met those conditions.

## Summary by CodeRabbit

* **Chores**
  * Updated Next.js-related packages to newer versions.


csiyang (commented):
The bump itself checks out — the four specifier changes plus the `overrides` pin are consistent, and the lockfile's top-level changes are confined to the Next family (the rest of the churn is in-range floats from the refresh, covered by CI). Three notes, all body-text or one-liners, none touching the shipped code:

1. **Name the security advisory in the body.** It says "patched security release" but not which vulnerability. That ID is what lets a reviewer verify now that 16.3.6 actually contains the fix, and what answers "were we exposed to X and when did we close it?" later — without it, this PR can't be audited as a security fix.

2. **Note the deliberate minor-line jump.** The diff moves 16.2 → 16.3, but the body's framing reads like a patch-line bump. The registry shows a 16.2.12 backport line exists, so taking 16.3.6 looks like a runway choice (be on the newest minor so future advisories land as patch bumps, absorbed now in a calm window rather than later under pressure) — which is sound, but worth one sentence in the body so the next reader doesn't re-run the "why not 16.2.12?" analysis, and so approvers know they're signing off a minor (feature/behavioral changes alongside the fix) rather than a patch.

3. **`@next/third-parties` was left at `^16.2.3`** (resolves 16.2.5, now peering against `next@16.3.6`) while `@next/bundle-analyzer` and `@next/eslint-plugin-next` moved to `^16.3.6`. Looks like an oversight rather than a choice — one line to bump it, or one line to say it's deliberate.

One merge gate rather than a note: the `journeys-admin` deploy preview — the biggest Next consumer — was still pending when I checked; worth confirming it's green before merge.

csiyang (approved):

